PR #188, opened 2015-01-27
Implement key in cpp

### LINKED ISSUE #187: C++ interface does not implement getting the 'key' of a message
Line 122 in src-cpp/rdkafkacpp_int.h:
```
 const std::string *key () const { return NULL; /* FIXME */ }
```
So when consuming via the C++ interface, we are not able to get the key of the message.
I'll propose a fix via a pull request in a short while.

kodekarl:
This should fix #187 (Message::key() not implemented).
The example program will print the key if the message has one.


kodekarl:
Sorry, need to check something...

PR #189, opened 2015-01-27
Implement Message::key() in C++ interface

### LINKED ISSUE #187: C++ interface does not implement getting the 'key' of a message

kodekarl:
This should fix #187 by adding a key_ member to the message implementation and returning that as key (or NULL if they key is empty/unset).
The C++ example prints the key if present, like the C example does. None of the examples actually produce keys (so this code was tested with a temporary example code).

### DIFF
--- a/examples/rdkafka_example.cpp
+++ b/examples/rdkafka_example.cpp
@@ -440,6 +440,9 @@ int main (int argc, char **argv) {
         case RdKafka::ERR_NO_ERROR:
 	  /* Real message */
 	  std::cerr << "Read msg at offset " << msg->offset() << std::endl;
+	  if (msg->key()) {
+		  std::cerr << "Key: " << *msg->key() << std::endl;
+	  }
           printf("%.*s\n",
                  static_cast<int>(msg->len()),
                  static_cast<const char *>(msg->payload()));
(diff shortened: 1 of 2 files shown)


kodekarl:
My implementation sort of expects that keyed messages are the norm, and so "pays" a std::string for all messages. If keyed messages are rare, the string can be created on-demand instead (on first call to key()), at the expense of one extra heap allocation for those that do use keyed messages.
I also assume that Message should keep ownership of the returned object; if we may assume that the caller gets ownership, we can make it simpler (for Message, not the caller).

edenhill:
That's a good point.
It is hard to say if keyed messages are the norm or not, or if the consumer applications care about the key, so I'm inclined not to create it on Message creation, but on demand instead.

Returning a heaped key that needs application-side deletion makes the API more complex, so what about a middle route where the key is created on demand in key() but the Message object retains ownership of it and deletes the maybe-created key on Message deletion?

kodekarl:
Yes, that's what I meant. I'll update the code later today.

kodekarl:
I'm a bit new to github. I read that if I claim this PR fixes #187 (with those words), housekeeping for the owner becomes a little easier.

edenhill:
Looks good, thanks!

Issue #193, opened 2015-02-02
Brokers leak because of refcount mismatch


thelamb:
To reproduce:

`valgrind --leak-check=full ./rdkafka_example -b first-server,second-server -t your-topic -P`

(Wait about 30 seconds before quitting)

Output:
```
==29251== 1,460 (1,424 direct, 36 indirect) bytes in 1 blocks are definitely lost in loss record 10 of 10
==29251==    at 0x4C29DB4: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==29251==    by 0x40AA1A: rd_kafka_broker_add (rdkafka_broker.c:4112)
==29251==    by 0x415E9B: rd_kafka_brokers_add (rdkafka_broker.c:4291)
==29251==    by 0x402C8D: main (rdkafka_example.c:500)
```

There are 2 causes that I could identify:
### rd_kafka_brokers_add 
  `rd_kafka_brokers_add` calls `rd_kafka_broker_find`, which increments the broker's refcount if found. `rd_kafka_brokers_add` does not call `rd_kafka_broker_destroy` on the return value of `rd_kafka_broker_find`

Note:
`rd_kafka_broker_find` does not have a comment that the caller should call `rd_kafka_broker_destroy` on the return like `rd_kafka_broker_find_by_nodeid` has.

### rd_kafka_metadata_refresh_cb
`rd_kafka_metadata_refresh_cb` calls `rd_kafka_broker_any`, which increments the broker's recount if found. `rd_kafka_metadata_refresh_cb` does not call `rd_kafka_broker_destroy` on the return value.

 
I have a proposed fix ready and will submit a pull request

edenhill:
Now this is what a bug report is supposed to look like! Great work!

PR #194, opened 2015-02-02
Fix mismatches in broker_keep and broker_destroy that cause brokers to l...

thelamb:
The fix on brokers_add is not too sexy, but it's not enough to only call broker_destroy on line 4389 (after cnt++) because this will leak when `rkb->rkb_source != RD_KAFKA_CONFIGURED` (#193)

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -1147,6 +1147,8 @@ static void rd_kafka_metadata_refresh_cb (rd_kafka_t *rk, void *arg) {
         else
                 rd_kafka_broker_metadata_req(rkb, 1 /* all topics */, NULL,
                                              NULL, "periodic refresh");
+
+		rd_kafka_broker_destroy(rkb);
 }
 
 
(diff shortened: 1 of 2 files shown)


thelamb:
Does GitHub not like `git commit --amend` when submitting a change to the pull request? I don't see your previous comments any more. 
I'm used to the workflow of one commit per topic/branch and rebasing instead of git merge. Let me know if I should just make a second commit on the same branch in the future if there are changes needed to the pull request.

Anyway your comment made sense. I fixed it by destroying only after the original if/else. Doing 
```c++
if( rkb ) {
    if( rkb->rkb_source == RD_KAFKA_CONFIGURED )
        cnt++;
    rd_kafka_broker_destroy(rkb);
else if( rd_kafka_broker_add( ... ) )
    cnt++;
```
as you proposed doesn't call `rd_kafka_broker_add` in case `rkb->rkb_source != RD_KAFKA_CONFIGURED`. So I believe the current solution is the cleanest.

edenhill:
It is okay with multiple commits in the same pull request if they are fixing stuff in the pull request itself, like now. But it doesn't really matter too much unless things get too cluttered up.

Thanks!

PR #200, opened 2015-02-16
Allow message payload to be NULL.

secretmike:
NULL message payloads are used with compaction topics introduced in 0.8.1
which use the NULL payload to indicate the corresponding key should be deleted.

(see [Log Compaction](https://cwiki.apache.org/confluence/display/KAFKA/Log+Compaction#LogCompaction-Theanatomyofalog) for more details)

The [Kafka protocol guide](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets) also notes that the Value (payload) may be NULL. 

This commit makes these changes:

- prevents NULL payloads from being copied when `RD_KAFKA_MSG_F_COPY` is set.
- prevents payload bytes being copied into `rkbuf` when NULL.
- sets `rkm_len` to `-1` when payload is NULL (just like key).
- Allows NULL payloads to be received in a fetch.

C is not my strong point so I'm sure there are things missed. If you can point me in the right direction I'd be happy make any changes, or if you'd rather clean things yourself that works for me too :)

Thanks!

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -2127,6 +2127,10 @@ static int rd_kafka_broker_produce_toppar (rd_kafka_broker_t *rkb,
 
 
 		/* Value(payload) length */
+		if (rkm->rkm_payload == NULL) {
+			/* Set length to -1 to indicate a NULL value. */
+			rkm->rkm_len = -1;
+		}
 		msghdr->part4.Value_len = htonl(rkm->rkm_len);
 		msghdr->part3.Crc =
 			crc32(msghdr->part3.Crc,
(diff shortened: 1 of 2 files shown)


edenhill:
Thanks for the PR, this is a good fix!

edenhill:
Null message support is now in master.
Thanks for your contribution!

secretmike:
Thank you!

edenhill:
Can you verify this works as intended for you?

secretmike:
I'll be trying it out, but probably not until next week.

PR #207, opened 2015-02-26
Switch to use zlib crc32 function

tburmeister:
Hi Magnus,

I have been using librdkafka for a project at work - it has made it really easy to add Kafka support to our existing real time data applications and the performance has been excellent. In looking for bottlenecks, however, my manager and I discovered that our application was spending a more time executing rd_crc32_update than just about anything else. I did some testing and discovered that the zlib crc32 function is roughly 4 times faster than rd_crc32_update, which makes sense since I believe the version of zlib I am using stores static tables allowing for 4 byte at-a-time CRC calculations as opposed to just 1 byte. zlib is already a build requirement, so using the zlib crc32 function seems like an easy way to get a slight performance boost (depending on the zlib version being used.)

Best,
Taylor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,13 +20,14 @@ CONFIGURATION.md: src/rdkafka.h examples
 		mv CONFIGURATION.md.tmp CONFIGURATION.md; \
 	rm -f CONFIGURATION.md.tmp)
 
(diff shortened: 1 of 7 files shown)


edenhill:
This is great, thanks!

PR #213, opened 2015-03-11
rdkafka_example: make consumer handle multiple partitions

vincentbernat:
When no partition is specified by the user, let consume all the
partitions for the provided topic. The code was stolen from kafkacat.

### DIFF
--- a/examples/rdkafka_example.c
+++ b/examples/rdkafka_example.c
@@ -55,6 +55,13 @@ static 	enum {
 	OUTPUT_RAW,
 } output = OUTPUT_HEXDUMP;
 
+/* Partition's at EOF state array */
+int *part_eof = NULL;
+/* Number of partitions that has reached EOF */
+int part_eof_cnt = 0;
+/* Threshold level (partitions at EOF) before exiting */
+int part_eof_thres = 0;
+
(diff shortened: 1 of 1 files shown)


edenhill:
Thanks for this addition vincent!

I would like to keep rdkafka_example.c as simple as possible as not to scare off potentional users,
so I'd like to keep the current consume-single-partition-without-queues code.
What I'll do is break out this multi-partition patch into its own code block.

chengchengpei:
Now, the consumer can consume all partitions if partition is not specified?

edenhill:
@chengchengpei This PR shows the current way to do this, but with the next librdkafka release there will support for proper balanced consumer groups where you subscribe to a set of topics and the partition assignor assigns each client those topic's partitions according to an assignment strategy.

chengchengpei:
@edenhill When implement of consumer group will be released? This feature is important to our prod.

edenhill:
In about a month.

vincentbernat:
@edenhill But you also need Kafka 0.9, right? (notably for proper balanced consumer groups)

edenhill:
@vincentbernat That's right!

PR #219, opened 2015-03-19
Allow key to be used in pointer and length style.

rthalley:


rthalley:
reworking this, forgot one thing :)

PR #221, opened 2015-03-20
Wip cpp metadata

eugpermar:
Currently, the test is only able to connect to localhost. It should be able to read the test config file and act in consequence.

As always, any comments will be appreciated


edenhill:
Thanks @eugpermar , merged in #434

eugpermar:
Glad to help!

PR #234, opened 2015-04-07
Add a built in consistent hashing partitioner

NZKoz:
For our particular use case mapping keyed messages on to a consistent partition is important, it's probably important for anyone who wants to use log compaction.  I'm not sure what you'd want in terms of tests so I've done nothing. :trollface: 

Uses rd's crc32 to hash keys ensuring that the same key will map to the same partition.

Similar in spirit to the Scala code's default partitioner:

```scala
def partition(key: Any, numPartitions: Int): Int = {
  Utils.abs(key.hashCode) % numPartitions
}
```

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -541,6 +541,16 @@ int32_t rd_kafka_msg_partitioner_random (const rd_kafka_topic_t *rkt,
 					 int32_t partition_cnt,
 					 void *opaque, void *msg_opaque);
 
+/**
+ * Consistent partitioner
+ * Uses consistent hashing to map identical keys onto identical partitions.
+ *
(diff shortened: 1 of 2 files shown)


edenhill:
Thanks, looks good.
One question; is it compatible with the Scala partitioner?
I.e., same key gives same partition?

DEvil0000:
Not compatible with Scala partitioner. Scalas default hashing algorithm is plattform dependent.
```Scala
def hashCode(): Int

Calculate a hash code value for the object.

The default hashing algorithm is platform dependent.
...
returns
    the hash code value for this object. 
```
Or MurmurHash is being used.

So Scalas hashing results will not be the same.

edenhill:
Okay, fair enough, no point in trying to be compatible then.
Thanks

PR #235, opened 2015-04-09
use topic level group_id for rkbuf sizing

alfred-landrum:
rd_kafka_toppar_offsetcommit_request writes topic-level rkt_conf.group_id into the request, but uses the global-level rk_conf.group_id for buffer sizing.


edenhill:
Ouch, good catch!

PR #236, opened 2015-04-09
improve build rpm

toidi:
Build rpm with

```
$ make rpm
```

command.

Process looks like:
1. `Makefile` get current `librdkafka` version.
2. Create tar archive from git.
3. Build SRPM with mock.
4. Build RPM with mock from SRPM.


edenhill:
Thanks!

PR #240, opened 2015-04-13
Put rdkafka_example debugging contexts string in right place.

rthalley:
When looking at help for rdkafka_example (the C version) it prints the list of possible debugging contexts after the -A option, instead of as a second line of text for the -d option.


edenhill:
Thanks!

PR #243, opened 2015-04-15
Ignore supplied key or payload length if the pointer is null

ept:
Minor API usability tweak. I was producing messages with a NULL payload, but a non-zero payload length. This causes an invalid produce request to be sent to the Kafka broker, and the broker responds by logging an exception (java.nio.BufferUnderflowException) and closing the connection.

Test case: https://gist.github.com/ept/a7d6edc2e46f896b03a6

This is really my mistake for passing invalid parameters to rd_kafka_produce. However, it might be nicer if the length parameters are ignored if NULL data is passed in, rather than creating an invalid produce request. This simple patch forces the length to be zero in that case.


edenhill:
Thanks!

Issue #245, opened 2015-04-15
Does any C++ API invoke rd_kafka_consume_callback


michema:
Hi, Magnus:

The current C++ API consume invokes rd_kafka_consume(). Is there any plan to integrate rd_kafka_consume_callback() for C++ API?

Thanks,

Michelle

edenhill:
The plan is to support all relevant C APIs in the C++ API, and this is an ongoing effort.
Requesting specific APIs, like you do here, speeds that up :)
So is this a request for consume_callback in C++?

michema:
Yes, because it has a better performance.

kodekarl:
I made a draft of this (in the context of #168) some time ago. I'll rebase it, fix it up a little, and let you have a look. Hang on.

edenhill:
Now fixed, by kodekarl, in PR #250

Issue #168, opened 2014-11-26
there is no batch consume interface for C++ 


itismewxg:
NULL

edenhill:
Please try out the new consume_callback interface and reopen this issue if it doesn't solve your problem.

itismewxg:
Thanks! gonna try~

PR #246, opened 2015-04-15
Added Consumer::consume_callback as front-end of rd_kafka_consume_callback

kodekarl:
Implements #168 / #245. Lets the example program use it when called with -e option.

This is one way of doing it, kept reasonably close to the C function. Another would be to define a callback base class (like e.g. EventCb) and use a instance of one of those as argument to consume_callback. If you'd rather see that instead, I can draft it tomorrow or the next day?

### DIFF
--- a/examples/rdkafka_example.cpp
+++ b/examples/rdkafka_example.cpp
@@ -118,6 +118,34 @@ class MyHashPartitionerCb : public RdKafka::PartitionerCb {
   }
 };
 
+void msg_consume(RdKafka::Message* message, void* opaque) {
+  switch (message->err()) {
+    case RdKafka::ERR__TIMED_OUT:
+      break;
+
+    case RdKafka::ERR_NO_ERROR:
+      /* Real message */
+      std::cerr << "Read msg at offset " << message->offset() << std::endl;
+      if (message->key()) {
+        std::cerr << "Key: " << *message->key() << std::endl;
+      }
+      printf("%.*s\n",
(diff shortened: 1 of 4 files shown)


edenhill:
Looks good, just one comment on rdkafka_example.cpp to make -e work.

kodekarl:
Sorry about that misunderstanding; it was really mainly to have _some_ code that called the function ;)
Perhaps I should look into how to write tests ...
Anyway, I'll put the example back together. Would it make sense to have an option to use the consume_callback function instead of the regular consume? Like, for comparing/measuring?

edenhill:
Yeah, that's a good idea,  add an option to toggle between standard consume and callback.

kodekarl:
The toggle will be in a separate PR.

kodekarl:
Please hold this PR a little. I really want to explore that idea I presented in comment 1 (using an interface class), possibly _instead of_ the function-interface it has now.

edenhill:
Yes, class based callbacks are preferable.

kodekarl:
I'm happy with it now. If you are as well, shouldn't I make a new - clean - PR with only the final result in it? If you're not quite happy, let me know...

edenhill:
Looks good! Just a minor nitpick on poll().
You dont need to, but a clean PR would be very much appreciated

kodekarl:
I really like clean PRs (using Gerrit at work and it's getting to me), so I'll make one.

PR #250, opened 2015-04-17
Added Consumer::consume_callback as front-end of rd_kafka_consume_callback

kodekarl:
* New base class ConsumeCb allows for easy definition of consume_callback classes.
* New function Consumer::consume_callback consumes as many messages as possible, calls ConsumeCb::consume_cb for each message.
* Example C++ code re-organized a little; function msg_consume extracted from loop, prints on cout instead of cerr.

Implements #168, implements #245.

### DIFF
--- a/examples/rdkafka_example.cpp
+++ b/examples/rdkafka_example.cpp
@@ -51,6 +51,7 @@
 
 
 static bool run = true;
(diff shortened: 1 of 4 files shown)


edenhill:
Awesome, thanks!

PR #255, opened 2015-04-21
mklove update

Whissi:
Updates mklove in librdkafka to https://github.com/edenhill/mklove/commit/9bd461ae4d

* Fixes colored output
* Adds `--disable-debug-symbols` option
* Disables *-Werror* per default (can be re-enabled using `--enable-werror`)


edenhill:
Travis is whining about the Makefile, see here:
https://travis-ci.org/edenhill/librdkafka/builds/59434859

Whissi:
Yup, saw that. I created a test branch to see what's happening. I am unable to reproduce it yet on my local Ubuntu system which should be equal to travis...

Whissi:
Mhh... 
```
make[1]: *** No rule to make target `../src/librdkafka.a', needed by `rdkafka_example'.  Stop.
```
https://travis-ci.org/Whissi/librdkafka/builds/59442310

Weird, when I don't call `mkl_check_terminal_color_support` Travis seems to pass (https://travis-ci.org/Whissi/librdkafka/jobs/59445700) but I don't know what's the problem. I cannot reproduce :/

edenhill:
Maybe Travis itself is to blame, look at this:
http://www.traviscistatus.com/

Lets hold off until things settle over there.

Whissi:
Well, the origin error message seems to be
```
/bin/sh: 2: Syntax error: "(" unexpected (expecting "fi")
make: *** [mklove-check] Error 2
```
Now I think I found the problem, see https://travis-ci.org/Whissi/librdkafka/builds/59454065#L190

The reset code is weird. When I use the same value, I am getting the same problem. But why is `tput` returning such a value...

Whissi:
OK, we are having two problems:

1. `make` will fail when using `make -n`: https://travis-ci.org/Whissi/librdkafka/jobs/59480813 - that cost me hours. Happening without my changes...

2. Color codes containing chars like ";" or "(" are causing problems. Sounds like an escaping problem, but I don't know how to fix that at the moment. Maybe we shouldn't pass color codes from configure to the Makefile (workaround), see: https://travis-ci.org/Whissi/librdkafka/builds/59484093 (not sure why the Makefile doesn't use colors).

Let me think about the problems...

Whissi:
I finally fixed the problem. As quoted in the PR for mklove, the problem was the quoted color code:
```bash
A_RESET=        "^[(B^[[m"
```
Due to the quote, chars like "(" or ";" got a special meaning. We could have added escaping, but chances are high that we would miss some chars. The good news is: We don't need quotes at all, so we don't need to add escaping ;)

Whissi:
It has passed but please don't merge yet. Seems like MacOS don't like `/bin/echo -e`... it is showing "-e": https://travis-ci.org/edenhill/librdkafka/jobs/60394884#L114

Fix should be easy.

edenhill:
Reminds me of this game: http://en.wikipedia.org/wiki/15_puzzle

Whissi:
I had two options: Rewrite *configure.host* to define `ECHOCMD` for each platform (and export to *Makefile*) or just use the recommended `printf` command instead. I chose `printf`.

Let's see what Travis will say.

If it passes, please have a special look at https://github.com/edenhill/librdkafka/commit/687cd352637ecb199785c9586d6e3c26ef460538#diff-df77a2872d82ce3fe58a21d565c31a97L105
This should fix
```bash
make[1]: Entering directory '/home/thomas/repositories/librdkafka/src'
Checking librdkafka integrity
RET=true ; \
for f in librdkafka.so.1 librdkafka.a ; do \
        printf "%-30s " $f ; \
        if [ -f "$f" ]; then \
                printf "OK\n" \
        else \
                printf "MISSING\n" \
                RET=false \
        fi \
done \
$($RET))
/bin/sh: -c: line 10: syntax error near unexpected token `)'
/bin/sh: -c: line 10: `$($RET))'
../mklove/Makefile.base:94: recipe for target 'file-check' failed

```

Whissi:
Woohoo, finally passed. Working colored output on Linux and MacOS. If the fix for `$RET` is correct you are free to merge.

If this will be merged I'll backport the changes to *mklove*.

Whissi:
Last commit is for https://travis-ci.org/edenhill/librdkafka/jobs/59878718#L109 I just found.

Whissi:
The build error is a Travis problem, the same commit passed in my Travis account. Please trigger a rebuild...

edenhill:
Thanks Thomas, great work!

Issue #259, opened 2015-04-23
make tests: test 14 fails for me


DEvil0000:
I was seeing test#14 fail in differet ways
```
############## 0014-reconsume-191.test ################
### Running test 0014-reconsume-191.test in bare mode ###
Using topic "rdkafkatest_0013"
Topic rdkafkatest_0013, testid 9078722994771939079
Created    kafka instance rdkafka#producer-0
### Test failed at 0014-reconsume-191.c:51:dr_cb(): ###
Message delivery failed: Broker: Invalid message
 
### Test random seed was -108662397 ###
###
### Test 0014-reconsume-191.test in bare mode FAILED! ###
###
```
```
############## 0014-reconsume-191.test ################
### Running test 0014-reconsume-191.test in bare mode ###
Using topic "rdkafkatest_0013"
Topic rdkafkatest_0013, testid 9078722994771939079
Created    kafka instance rdkafka#producer-0
### Test failed at 0014-reconsume-191.c:51:dr_cb(): ###
Message delivery failed: Broker: Invalid message
 
### Test random seed was -108662397 ###
###
### Test 0014-reconsume-191.test in bare mode FAILED! ###
###
```
and one time it worked I think

DEvil0000:
I think it is a multi threading related issue - helgrind:
```
############## 0014-reconsume-191.test ################
### Running test 0014-reconsume-191.test in helgrind mode ###
==2841== Helgrind, a thread error detector
==2841== Copyright (C) 2007-2011, and GNU GPL'd, by OpenWorks LLP et al.
==2841== Using Valgrind-3.7.0 and LibVEX; rerun with -h for copyright info
==2841== Command: ./0014-reconsume-191.test
==2841==
Using topic "rdkafkatest_110e5fdc4f2dfb83_0013"
Topic rdkafkatest_110e5fdc4f2dfb83_0013, testid 6935047690210883586
Created    kafka instance rdkafka#producer-0
Destroying kafka instance rdkafka#producer-0
Created    kafka instance rdkafka#consumer-1
Iteration #0: Consuming from partition 0
Iteration #0: consumed 10000 messages
==2841== ---Thread-Announcement------------------------------------------
==2841==
==2841== Thread #1 is the program's root thread
==2841==
==2841== ---Thread-Announcement------------------------------------------
==2841==
==2841== Thread #7 was created
==2841==    at 0x6349FC8: clone (clone.S:111)
==2841==
==2841== ----------------------------------------------------------------
[...]

Issue #258, opened 2015-04-23
make tests: test 12 fails for me


DEvil0000:
1. I changed default broker config to create topics with 1 partition what was causing test#12 to fail on produce to partition#2
-> I changed it back to 2 and recreated the topic
2. test#12 fails on reconsume with offset -5000
(debug=topic)
```
############## 0012-produce_consume.test ################
### Running test 0012-produce_consume.test in bare mode ###
Using topic "rdkafkatest_0012"
Topic rdkafkatest_0012, testid 152358064405547136
Created    kafka instance rdkafka#producer-0
%7|1429791778.444|TOPIC|rdkafka#producer-0| New local topic: rdkafkatest_0012
%7|1429791778.446|STATE|rdkafka#producer-0| Topic rdkafkatest_0012 changed state unknown -> exists
%7|1429791778.446|PARTCNT|rdkafka#producer-0| Topic rdkafkatest_0012 partition count changed from 0 to 2
%7|1429791778.446|BRKDELGT|rdkafka#producer-0| Broker ubuntu:9092/0 is now leader for topic rdkafkatest_0012 [1] with 0 messages (0 bytes) queued
%7|1429791778.446|BRKDELGT|rdkafka#producer-0| Broker ubuntu:9092/0 is now leader for topic rdkafkatest_0012 [0] with 0 messages (0 bytes) queued
%7|1429791778.446|PARTCNT|rdkafka#producer-0| Partitioning 0 unassigned messages in topic rdkafkatest_0012 to 2 partitions
%7|1429791778.446|UAS|rdkafka#producer-0| 0/0 messages were partitioned in topic rdkafkatest_0012
%7|1429791778.447|PARTCNT|rdkafka#producer-0| No change in partition count for topic rdkafkatest_0012
%7|1429791778.447|TOPICUPD|rdkafka#producer-0| No leader change for topic rdkafkatest_0012 [1] with leader 0
%7|1429791778.447|TOPICUPD|rdkafka#producer-0| No leader change for topic rdkafkatest_0012 [0] with leader 0
Start produce to partition 0
Produced 5000 messages to partition 0, waiting for deliveries
Start produce to partition 1
Produced 5000 messages to partition 1, waiting for deliveries
Destroying kafka instance rdkafka#producer-0
Created    kafka instance rdkafka#consumer-1
%7|1429791779.525|TOPIC|rdkafka#consumer-1| New local topic: rdkafkatest_0012
Consuming 5000 messages from partition 0
[...]

PR #260, opened 2015-04-24
fixed some race conditions & changed one lock to recursive

DEvil0000:
usage of the recursive lock was just a quick fix, should be improved in the future.
#258 #259
but still is is not fixing all my problems


DEvil0000:
I am not sure if this is still a good thing

PR #269, opened 2015-05-07
issue in broker address caching

yazgoo:
When a broker goes down, it seems that rdkafka should keep its address in cache for a little while before retrying.
From what I understand, rkb->rkb_t_rsal_last should contain the last time the cache was updated.
But it is not set.
So, we always have
````C
 rkb->rkb_t_rsal_last + rkb->rkb_rk->rk_conf.broker_addr_ttl == 0 + 300 000 /*ms*/
````
Which is always less than
````C
time(NULL)
````
Which means rkb->rkb_rsal is always set to NULL, and rd_getaddrinfo() is always called.
Do you agree?

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -573,7 +573,7 @@ static int rd_kafka_broker_resolve (rd_kafka_broker_t *rkb) {
                                              rkb->rkb_nodename, errstr);
 			return -1;
 		}
-	    rkb->rkb_t_rsal_last = time(NULL);
+		rkb->rkb_t_rsal_last = time(NULL);
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


edenhill:
You are right, that is a bug.
But I'm not really sure we want to provide second-level name resolution caching, the libc resolver already does that, and it is aware of record TTL (e.g., correct caching time), so I think the current buggy behaviour might actually be correct.
What should be done is to remove the t_rsal_last checks altogether.

fsaintjacques:
IMHO, you do _not_ want to cache DNS resolution on the client side (instead of kernel or libnss), this leads to potential headaches for devops/developpers unaware of the behaviour. For example, this makes it really hard to debug with `/etc/hosts` overwrite.

edenhill:
Will disable address caching instead.

PR #270, opened 2015-05-07
Updated README: Added php-rdkafka

arnaud-lb:
php-rdkafka is a thin PHP binding for librdkafka. Although it uses classes, it attempts to not deviate from librdkafka's API.


edenhill:
Cool! Thanks for letting me know of your bindings.
Out of curiosity, why did you choose to conceive new bindings rather than using existing phpkafka?

arnaud-lb:
At the time I started, I believe EVODelavega/phpkafka didn't exist yet, and salebab/phpkafka looked more like a proof of concept than a production ready binding (e.g. the broker list is a global variable shared by all instances, and consume(), produce() always start with a rd_kafka_new()). Also, there was a few personal forks with different fixes in each of them.

I though it would be simpler to just start from scratch. Also, I preferred to keep the API as close as possible to librdkafka's, and to build a layer on top of it in PHP if felt necessary.

edenhill:
I think you made the right choice.
Good luck going forward and let me know if you need anything.

PR #271, opened 2015-05-07
Adding support for passing optional opaque data if exists for deliver…

redsand:
…y reporting.

Sometimes temporal data is needed for this, not a data value set in cb.


redsand:
My fault, I was looking for the rkmessage->_private opaque void ptr.

PR #273, opened 2015-05-10
fix rd_kafka_q_concat0 deadlock with self

alfred-landrum:


edenhill:
Thanks!

edenhill:
Did you have a specific case where this happened? I'm surprised I havent seen this before.

alfred-landrum:
It was happening during a a call to rd_kafka_consume_start_queue.

In my code using librdkafka, I've got a 'start' call that creates a new queue, then calls consume_start_queue for ~64 partitions, and a 'stop' call that calls 'rd_kafka_consume_stop', then rd_kafka_queue_destroy. I was using these calls for flow control, so if enough messages were received after 'start', 'stop' would get called until the higher levels are ready to process more messages. This issue would show up eventually after several start, consume, stop sequences.

edenhill:
Okay, thanks for the details.

Are you really sure you need to `_stop()` the consumers?
If you dont call any of the `.._consume..()` functions, fetched messages will queue up internally in librdkafka in the `fetchq`. When the fetchq reaches `queued.max.messages.kbytes` librdkafka will stop fetching messages from the broker until there are less than `queued.max.messages.kbytes` in the queue again (because your application consumed some messages).

alfred-landrum:
You're welcome - yeah, this was just the beginning implementation my consumer control code. I plan to have something like a 'pause' and 'resume' that just control if the consume calls are actually made.

Issue #298, opened 2015-06-02
rd_kafka_msg_new() and error handling when RD_KAFKA_MSG_F_FREE is set


rthalley:
rd_kafka_msg_new() doesn't say what the contract is with respect to who is responsible for freeing the payload on errors.  Indeed, it is inconsistent.

If a failure occurs before the 'rkm' is created, then errno is set, -1 returned, and payload memory is NOT freed.

If the partitioner fails, then the rkm is destroyed (causing the payload to be freed if RD_KAFKA_MSG_F_FREE is set), errno is set, and -1 is returned.

This behavior makes it impossible for the caller to know whether or not it needs to free the payload itself, as sometimes it does and sometimes it doesn't.

I propose that the contract be amended such that the payload is *always* freed on failure if RD_KAFKA_MSG_F_FREE is set.  If you agree, I'll send a patch.

edenhill:
Good finding, thats definately a bug!

I think it might be more useful not to free the provided pointer on immediate errors though since it allows the application to take alternative actions (retry, queue, whatever..).
Otherwise such an application needs to duplicate the data in case of immediate error.

rthalley:
I can do it that way too.  Should be a simple matter of just clearing the free flag if there's an error.  Will send a patch.

skandragon:
I mentioned to @rthalley offline that it would be good to check the delivery report path as well, as it gets the data.  It's not clear from the docs if it is freed or not, or I missed that part of the docs!

PR #300, opened 2015-06-03
rd_kafka_msg_new() should never free the payload on failure if

rthalley:
RD_KAFKA_MSG_F_FREE was specified.  (Issue #298)


edenhill:
Looks good, just clarify what an error is in the public API (rdkafka.h).
Also, update the corresponding comment in src-cpp/rdkafkacpp.h

Thanks

rthalley:
Comments updated as requested.

edenhill:
Awesome, thanks!

Issue #296, opened 2015-06-01
consumer partition seemingly stuck when offset request fails


alfred-landrum:
With a broker running 0.8.2, and with an aggressive data retention policy that's causing log files to roll often, some partitions on my consumer get 'stuck' such that no further messages are delivered.

In the kafka broker log, I'll see something like this:

```
...
[2015-05-30 00:30:26,149] INFO Rolled new log segment for 'receiver-db1fc9b7-85b2-4016-8eda-66a356cb2ae8-35' in 1 ms. (kafka.log.Log)
[2015-05-30 00:30:26,152] WARN [KafkaApi-0] Error while responding to offset request (kafka.server.KafkaApis)
java.lang.ArrayIndexOutOfBoundsException: 7
        at kafka.server.KafkaApis.fetchOffsetsBefore(KafkaApis.scala:469)
        at kafka.server.KafkaApis.fetchOffsets(KafkaApis.scala:449)
        at kafka.server.KafkaApis$$anonfun$17.apply(KafkaApis.scala:411)
        at kafka.server.KafkaApis$$anonfun$17.apply(KafkaApis.scala:402)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.immutable.Map$Map1.foreach(Map.scala:109)
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
        at scala.collection.AbstractTraversable.map(Traversable.scala:105)
        at kafka.server.KafkaApis.handleOffsetRequest(KafkaApis.scala:402)
        at kafka.server.KafkaApis.handle(KafkaApis.scala:61)
        at kafka.server.KafkaRequestHandler.run(KafkaRequestHandler.scala:59)
        at java.lang.Thread.run(Thread.java:745)
...
```

When the above happens, my consumer seemingly doesn't read any further data from the partition that just rolled. My current guess is that there's a kafka broker bug here related to the segment roll code racing with the offset api. I'd like to figure out why the driver stops pulling in further data from the partition.

In my setup, I've got auto.offset.reset set to 'smallest'. I'm using the rd_kafka_consume_batch_queue api, with 64 partitions per topic, feeding a single queue.  I'm not using the rdkafka offset store code; I'm tracking offsets outside of rdkafka and calling rd_kafka_consume_start_queue at program start.

[...]

PR #301, opened 2015-06-03
dont update rktp_fetch_state on consumer lag request fail

alfred-landrum:
Addresses #296


edenhill:
Not sure why the last commit fails on travis.
Have a look at both the clang and gcc error printouts to figure it out.

Apart from that the fix looks good

alfred-landrum:
I just messed up the diff/apply when transferring from our internal snapshot of librdkafka, updated & retrying.

edenhill:
Thanks Alfred!

PR #325, opened 2015-07-06
Make configure script work under FreeBSD

mre:
When executing the configure file, I get an error on FreeBSD 10.1:

    > ./configure
    -bash: ./configure: /bin/bash: bad interpreter: No such file or directory

The reason being that bash is at /usr/local/bin/bash but the configure script uses /bin/bash which is not available by default. A portable way is to use env to get the correct path.
See http://stackoverflow.com/questions/5159711/what-is-the-proper-way-to-make-a-bash-script-portable-between-linux-and-freebsd
Tested on Mac Yosemite, Ubuntu Trusty and FreeBSD 10.1.

### DIFF
--- a/configure
+++ b/configure
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 
 MKL_CONFIGURE_ARGS="$0 $*"


edenhill:
Good stuff!
I'll need to close this PR though since it should really be filed on  https://github.com/edenhill/mklove and then backported to librdkafka (and kafkacat), but I'll apply it manually on mklove and do the backporting.


It'll also need some testing on Solaris, let's see if I can dig up an old vbox image.

mre:
Just tested on Solaris 10.1 and it works over there.
Thanks for backporting. :+1: 
This can be closed I guess.

edenhill:
Thanks!

PR #332, opened 2015-07-15
Win32

yazgoo:
This version supports building for MINGW librdkafka.so (the C++ version has undefined references)
Just do a:

````shell
$ CXX=x86_64-w64-mingw32-g++ CC=x86_64-w64-mingw32-gcc CFLAGS="-DLIBRDKAFKA_EXPORTS -DInterlockedAdd=_InterlockedAdd -DMINGW_VER -D_WIN32_WINNT=0x0760 -I/path/to/zlib/include/ -w" LDFLAGS="-L/path/to/zlib/dll/" LIBS="-lwsock32 -lws2_32" ./configure
````

I use this version with fuse_kafka for windows with no problem so far.
I disabled mkl_require socket if CC contains mingw:

````C
if (echo $CC | grep -v mingw >/dev/null); then mkl_require socket; fi
````


edenhill:
Thanks @yazgoo.
Can you check the AppVeyor build failures?
https://ci.appveyor.com/project/edenhill/librdkafka/build/1.0.16

PR #334, opened 2015-07-17
rd_kafka_q_serve: ensure rd_kafka_q_destroy() is called.

mutability:
On FreeBSD (and perhaps elsewhere) pthread_cond_init / pthread_mutex_init
do heap allocations that are only freed when the corresponding _destroy
functions are called.

rd_kafka_q_init() calls the init functions, so we must make sure that we
call rd_kafka_q_destroy(), even if the underlying queue is stack-allocated.

This is a slow but steady leak - something like 6MB/hour.


edenhill:
Thanks, good find!

mutability:
What needs to be done so this can be merged?

edenhill:
Real sorry for loosing track of this PR. Will merge now

mutability:
Thanks!

PR #357, opened 2015-08-20
Make rwlocks prefer writer

frew:
The particular issue that inspired this change was that when consuming at a high volume metadata updates would get stuck indefinitely behind consumers that weren't making any progress. Since rdkafka doesn't use rwlocks recursively anywhere and other than that the restriction there's pretty minimal overhead, just converted all rwlocks to prefer writer.

First time contributing to rdkafka, so just let me know if you'd like me to pull out this stanza to a common library or add any particular testing.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -1257,7 +1257,11 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *conf,
 
 	rd_kafka_keep(rk); /* application refcnt */
 
-	pthread_rwlock_init(&rk->rk_lock, NULL);
+	pthread_rwlockattr_t rwattr;
+	pthread_rwlockattr_init(&rwattr);
+	pthread_rwlockattr_setkind_np(&rwattr, PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP);
+	pthread_rwlock_init(&rk->rk_lock, &rwattr);
+	pthread_rwlockattr_destroy(&rwattr);
 
 	rd_kafka_q_init(&rk->rk_rep);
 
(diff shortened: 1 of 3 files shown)


frew:
Oh, hrm. I guess the non-write-starving version is Linux specific. I could make it a config check?

edenhill:
Thanks Fred, this looks good but as you say needs some porting modifications and I think its best if that work is done on the dev branch where all the thread primitives have been abstracted.

What do you mean by consumers that dont make any progress: they're not receiving any messages from the broker for that topic+partition, or the application's code is stuck processing the consumed messages?

In tinycthread.h change the rwlock_init() function/define to take an additional argument (`int pref_wr`)
and do a check for `#ifdef PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP` on the non-win32 version when setting up the `attr` field.
You should probably change it into a function (on the non-win32 side at least) while you're at it.
Also make sure all variables are declared at the beginning of the function scope.

frew:
Hi Magnus,

Really sorry for the delay getting back to you. Unfortunately, we ended up going with another language for implementing the server I observed this issue in and it doesn't look like I'll have a chance to dig into the circumstances that cause it to happen/iterate on the pull request. I'm going to go ahead and close it, but obviously feel free to use the code as a starting point if anybody else runs into the same issue.

-Fred

edenhill:
Thanks Fred.

Out of curiousity, was this the main issue pushing you to an implementation in another language?

frew:
Pretty much just the standard C/C++ considerations in an environment that doesn't have a lot of C++ infrastructure (legwork to select a viable subset of the language + common set of libraries, developer education, likelihood of hard-to-diagnose bugs, etc.) - nothing kafka/librdkafka related.

PR #376, opened 2015-09-19
add build instruction for osx users

tattsun:
The former build instruction is confusing OSX users because it will fail on OSX now.
(OSX users should use clang instead of gcc)

How about adding an build instruction for OSX users?


edenhill:
Why do you need to specify clang as compiler? Configure should be able to figure that out by itself.
What happens if you run it without arguments?

tattsun:
You're right. My environment was strange.

Homebrew was overridden it.

PR #377, opened 2015-09-21
Add Consumer::OffsetTail().

bwelling:
This can be used to create a negative offset, like the
RD_KAFKA_OFFSET_TAIL() macro in the C binding.

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -624,6 +624,13 @@ class Consumer : public virtual Handle {
   virtual int consume_callback (Queue *queue, int timeout_ms,
                                 RdKafka::ConsumeCb *consume_cb,
                                 void *opaque) = 0;
+
+  /**
+   * Converts an offset into the logical offset from the tail of a topic.
+   *
+   * 'offset' is the (positive) number of items from the end.
+   */
+  static int64_t OffsetTail(int64_t offset);
 };
 
 
--- a/src-cpp/ConsumerImpl.cpp
+++ b/src-cpp/ConsumerImpl.cpp
@@ -71,6 +71,9 @@ RdKafka::Consumer *RdKafka::Consumer::create (RdKafka::Conf *conf,
   return rkc;
 }
 
+int64_t RdKafka::Consumer::OffsetTail (int64_t offset) {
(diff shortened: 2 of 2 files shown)


src-cpp/rdkafkacpp.h
```diff
@@ -548,6 +549,15 @@ class Consumer : public virtual Handle {
                                 int timeout_ms,
                                 ConsumeCb *consume_cb,
                                 void *opaque) = 0;
+
```
edenhill: Can you make this a private method outside rdkafkacpp.h so that it only contains virtual methods?

src-cpp/rdkafkacpp.h
```diff
@@ -388,6 +388,7 @@ class Topic {
   static const int64_t OFFSET_END       = -1; /* Consume from end */
   static const int64_t OFFSET_STORED    = -1000; /* Consume from offset
                                                   * retrived from offset store*/
+  static const int64_t OFFSET_TAIL_BASE = -2000; /* internal: do not use */
```
edenhill: See comment below. Remove this and use the proper TAIL macro from rdkafka.h in the private method.

bwelling:
I'm not completely sure what you mean.  It's currently a static method, so while the function body itself could be moved to a source file, it can't be made virtual unless you want it to require a Consumer object that's unused.

edenhill:
Yeah, just move the implementation out of the public header file, just keep the declaration there.
Thanks

src-cpp/rdkafkacpp.h
```diff
@@ -548,6 +548,13 @@ class Consumer : public virtual Handle {
                                 int timeout_ms,
                                 ConsumeCb *consume_cb,
                                 void *opaque) = 0;
+
+  /**
+   * Converts an offset into an offset from the tail of a topic.
```
edenhill: should probably say "into the logical offset from the tail of a topic".
Otherwise people might think that the returned value is an absolute offset.

bwelling:
Any chance this could be pulled?  All of the comments have been addressed.

edenhill:
Sure, can you rebase your patch on latest master? There is a merge conflict here.

bwelling:
I think that should fix it.  Thanks!

edenhill:
Awesome, thanks!

PR #384, opened 2015-10-15
C++ Queue Wrapping

macisamuele:
The implemetation of the Queue tries to respect the 1:1 matching with the underlining C functions. The goal of this wrapping is to simplify the consuming of multiple topic+partitions in a single thread.
May be it will be required the update of the wiki page related to it ([wiki page](https://github.com/edenhill/librdkafka/wiki/Consuming-from-multiple-topics-partitions-from-a-single-thread))
Issue: https://github.com/edenhill/librdkafka/issues/383


edenhill:
Seems like Queues in C++ is the flavour of the month, here is another pull request:
https://github.com/edenhill/librdkafka/pull/387

PR #388, opened 2015-10-19
Limit producer queue size to a specified number of payload bytes.

rthalley:
Add queue.buffering.max.bytes to producer configuration, which sets a maximum on the number of payload bytes that can be queued on the producer.

Limiting by total size is helpful as the count and time-baesd limits do not give good control over how much space will be used for buffering, which means you're likely to either overbuffer and run out of memory, or underbuffer and lose messages.


skandragon:
bump.  This is a good thing for us as we need to limit memory.

edenhill:
Fixed in f7a7a1d

Issue #398, opened 2015-10-30
C++ Topic::name() constness


incessantier:
In the C++ API, a "const Topic*" pointer is passed around (e.g. partitioner callbacks), but Topic::name() is not a const method.  In order to get the topic name in a callback, you need to cast away the constness of the pointer.  Does Topic::name() really make internal mods, or could it be const?

edenhill:
You are right, it should be const. same for partition_available().
Care to file a PR?

incessantier:
Not at all, I'll get it going in the next couple days.

PR #404, opened 2015-11-02
Issue #398 Made name() and partition_available const()

incessantier:
Updated Topic::name() and Topic::partition_available() to be const methods.  Updated Handle::name() to be const.


edenhill:
Great, thanks!

PR #467, opened 2015-12-22
Avoiding conditional directives that break statements.

RomeroMalaquias:


edenhill:
This patch is not tested and breaks the logic.
Additionally I'm not really sure why you see the need for these types of patches, I see you are filing quite a few similar ones for other projects as well.

PR #475, opened 2015-12-27
Support certificate chains for TLS client authentication

hnousiainen:
Use SSL_CTX_use_certificate_chain_file() instead of SSL_CTX_use_certificate_file() to allow the use of complete certificate chains for TLS client authentication.


edenhill:
Thanks!

PR #476, opened 2015-12-27
Add pkg-config files to -devel RPM spec

hnousiainen:
Pkgconfig files were missing from -devel package RPM spec, failing 'make rpm' with unpackaged files.


edenhill:
:+1:

PR #478, opened 2015-12-28
Win x64 build

ah-:
Clean up the build configurations a bit and add a msvc x64 build.

Passes tests and seems to work, but there's a bunch of warnings due to int being 32 bit.
Needs some work around size_t/int usage to make it compile cleanly.


edenhill:
Superb!

PR #483, opened 2015-12-30
Compile cleanly / sort out sizeof(size_t) != sizeof(int)

ah-:
I'm trying to get librdkafka to compile 100% clean with msvc, clang and gcc.

This got longer than expected, so I would appreciate if you had a look.
It's not finished yet, but the remaining warnings are all about rather tricky issues.

Mostly replacing ints with size_t where the int was actually just assigned to the return value of something like strlen or where the value is passed to something like malloc.

Watch out for cases where I might have overlooked things that are now size_ts might actually be negative. I think I got them all now, but best to triple-check.

How would you prefer to continue with this? I think part of this PR should be fairly uncontroversial, another part needs checking by you, and then there's also a bunch of remaining warnings left.

Tests pass on win32/64 and with clang on osx, but haven't run them with valgrind yet.

### DIFF
--- a/src/rdcrc32.h
+++ b/src/rdcrc32.h
@@ -98,7 +98,8 @@ static __inline RD_UNUSED
 rd_crc32_t rd_crc32_update(rd_crc32_t crc, const unsigned char *data, size_t data_len)
 {
 #if WITH_ZLIB
-        return crc32(crc, data, data_len);
+        rd_assert(data_len <= UINT_MAX);
+        return crc32(crc, data, (uInt) data_len);
 #else
     unsigned int tbl_idx;
 
--- a/src/rdkafka_buf.c
+++ b/src/rdkafka_buf.c
@@ -77,8 +77,8 @@ void rd_kafka_buf_alloc_recvbuf (rd_kafka_buf_t *rkbuf, size_t size) {
 /**
  * Rewind write offset pointer and iovec poiner to a previous stored value.
  */
-void rd_kafka_buf_rewind (rd_kafka_buf_t *rkbuf, int iovindex, int new_of,
-	int new_of_init) {
+void rd_kafka_buf_rewind (rd_kafka_buf_t *rkbuf, int iovindex, size_t new_of,
+	size_t new_of_init) {
 	rkbuf->rkbuf_msg.msg_iovlen = iovindex;
 	rkbuf->rkbuf_wof = new_of;
 	rkbuf->rkbuf_wof_init = new_of_init;
@@ -303,15 +303,15 @@ void rd_kafka_bufq_purge (rd_kafka_broker_t *rkb,
 
 
 
-int rd_kafka_buf_write_Message (rd_kafka_buf_t *rkbuf,
-				int64_t Offset, int8_t MagicByte,
-				int8_t Attributes,
-				const rd_kafkap_bytes_t *key,
-				const void *payload, int len,
-				int *outlenp) {
+size_t rd_kafka_buf_write_Message (rd_kafka_buf_t *rkbuf,
+				   int64_t Offset, int8_t MagicByte,
+				   int8_t Attributes,
+				   const rd_kafkap_bytes_t *key,
+				   const void *payload, int32_t len,
+				   int *outlenp) {
 	int32_t MessageSize;
-	int begin_of;
-	int of_Crc;
+	size_t begin_of;
+	size_t of_Crc;
 
 	/*
 	 * MessageSet's per-Message header.
@@ -478,7 +478,8 @@ void rd_kafka_buf_callback (rd_kafka_broker_t *rkb, rd_kafka_resp_err_t err,
  */
 rd_kafkap_bytes_t *rd_kafkap_bytes_from_buf (const rd_kafka_buf_t *rkbuf) {
         rd_kafka_assert(NULL, rkbuf->rkbuf_msg.msg_iovlen == 1);
-        return rd_kafkap_bytes_new(rkbuf->rkbuf_wbuf, rkbuf->rkbuf_wof);
+        rd_kafka_assert(NULL, rkbuf->rkbuf_wof < INT32_MAX);
+        return rd_kafkap_bytes_new(rkbuf->rkbuf_wbuf, (int32_t) rkbuf->rkbuf_wof);
 }
 
 
--- a/src/rdkafka_buf.h
+++ b/src/rdkafka_buf.h
@@ -39,11 +39,11 @@ typedef struct rd_kafka_broker_s rd_kafka_broker_t;
 /* Advance/allocate used space in marshall buffer.
  * Point PTR to available space of size LEN on success. */
 #define _MSH_ALLOC(rkbuf,PTR,LEN)  do {					\
-                int __LEN = (LEN);                                      \
+                size_t __LEN = (LEN);                                   \
                 if (msh_of + __LEN >= msh_size)                         \
                         rd_kafka_buf_parse_fail(rkbuf,			\
 						"Not enough room in marshall buffer: " \
-						"%i+%i > %i",		\
+						"%"PRIdsz"+%"PRIdsz" > %"PRIdsz, \
 						msh_of, __LEN, msh_size); \
                 (PTR) = (void *)(msh_buf+msh_of);                       \
                 msh_of += __LEN;                                        \
@@ -286,8 +286,8 @@ typedef struct rd_kafka_bufq_s {
 void rd_kafka_buf_destroy_final (rd_kafka_buf_t *rkbuf);
 void rd_kafka_buf_auxbuf_add (rd_kafka_buf_t *rkbuf, void *auxbuf);
 void rd_kafka_buf_alloc_recvbuf (rd_kafka_buf_t *kbuf, size_t size);
-void rd_kafka_buf_rewind(rd_kafka_buf_t *rkbuf, int iovindex, int new_of,
-	int new_of_init);
+void rd_kafka_buf_rewind(rd_kafka_buf_t *rkbuf, int iovindex, size_t new_of,
+	size_t new_of_init);
 struct iovec *rd_kafka_buf_iov_next (rd_kafka_buf_t *rkbuf);
 void rd_kafka_buf_push0 (rd_kafka_buf_t *rkbuf, const void *buf, size_t len,
 			int allow_crc_calc);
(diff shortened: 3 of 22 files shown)


edenhill:
+1 for effort!

I think it would make more sense to squash all these into one single commit, it is all the same thing and should be an atomic commit. It also makes reviewing simpler.

These changes needs to be verified across the board, including old CentOS:es, FreeBSD, Solaris 10, Solaris 11, various OSX versions, Cygwin.

edenhill:
Easiest way to squash the commits is probably to do:
`git rebase -i master` then mark all but the first one with `s`

src/rdkafka_buf.c
```diff
@@ -358,7 +358,7 @@ int rd_kafka_buf_write_Message (rd_kafka_buf_t *rkbuf,
 		rd_kafka_buf_push(rkbuf, payload, len);
 
 	/* Finalize Crc */
-	rd_kafka_buf_update_u32(rkbuf, of_Crc,
+	rd_kafka_buf_update_u32(rkbuf, (uint32_t) of_Crc,
```
edenhill: The offset is size_t, why cast to uint32_t?

src/rdkafka_buf.h
```diff
@@ -39,11 +39,11 @@ typedef struct rd_kafka_broker_s rd_kafka_broker_t;
 /* Advance/allocate used space in marshall buffer.
  * Point PTR to available space of size LEN on success. */
 #define _MSH_ALLOC(rkbuf,PTR,LEN)  do {					\
-                int __LEN = (LEN);                                      \
+                size_t __LEN = (LEN);                                      \
                 if (msh_of + __LEN >= msh_size)                         \
                         rd_kafka_buf_parse_fail(rkbuf,			\
 						"Not enough room in marshall buffer: " \
-						"%i+%i > %i",		\
+						"%zu+%zu > %zu",		\
```
edenhill: From what I understand the only portable formattering for (s)size_t is `..%"PRIdsz"`

src/rdkafka_cgrp.c
```diff
@@ -1146,7 +1146,7 @@ static void rd_kafka_cgrp_terminate0 (rd_kafka_cgrp_t *rkcg, rd_kafka_op_t *rko)
 
         rd_kafka_dbg(rkcg->rkcg_rk, CGRP, "CGRPTERM",
                      "Terminating group \"%.*s\" in state %s "
-                     "with %d partition(s)",
+                     "with %zu partition(s)",
```
edenhill: %"PRIdsz"

src/rdkafka_offset.c
```diff
@@ -173,7 +173,7 @@ static int64_t rd_kafka_offset_file_read (rd_kafka_toppar_t *rktp) {
 	char buf[22];
 	char *end;
 	int64_t offset;
-	int r;
+	size_t r;
```
edenhill: the fread() below should probably be changed to:
if ((r = fread(buf, 1, ...)) < 1) {  rather than == -1

edenhill: The `if ((r = fread(...)) == -1)` check can probably be removed completely. Just do  `r = fread(..);` and let the sub-sequent `r == 0` handle it (as is)

src/rdkafka_range_assignor.c
```diff
@@ -89,7 +89,7 @@ rd_kafka_range_assignor_assign_cb (rd_kafka_t *rk,
 
                  rd_kafka_dbg(rk, CGRP, "ASSIGN",
                               "range: Topic %s with %d partition(s) and "
-                              "%d subscribing member(s)",
+                              "%zu subscribing member(s)",
```
edenhill: %"PRIdsz

src/rdkafka_topic.c
```diff
@@ -784,7 +784,7 @@ void rd_kafka_topic_partitions_remove (rd_kafka_itopic_t *rkt) {
 	rd_kafka_toppar_t *rktp;
         shptr_rd_kafka_toppar_t *s_rktp;
         shptr_rd_kafka_itopic_t *s_rkt;
-	int i;
+	size_t i;
```
edenhill: Doesn't make sense to have this as explicit size_t, it just an iterator so keep it as int.

src/rdlist.c
```diff
@@ -31,27 +31,27 @@
 
 
 void rd_list_dump (const char *what, const rd_list_t *rl) {
-        int i;
-        printf("%s: (rd_list_t*)%p cnt %d, size %d, elems %p:\n",
+        size_t i;
```
edenhill: same here, int

src/rdlist.c
```diff
@@ -31,27 +31,27 @@
...
+        printf("%s: (rd_list_t*)%p cnt %zu, size %zu, elems %zu:\n",
```
edenhill: %"PRIdsz..

src/rdlist.c
```diff
@@ -82,7 +82,7 @@ static void rd_list_remove0 (rd_list_t *rl, int idx) {
 
 void *rd_list_remove (rd_list_t *rl, void *match_elem) {
         void *elem;
-        int i;
+        size_t i;
```
edenhill: keep as int

src/rdlist.c
```diff
@@ -98,7 +98,7 @@ void *rd_list_remove (rd_list_t *rl, void *match_elem) {
 void *rd_list_remove_cmp (rd_list_t *rl, void *match_elem,
                           int (*cmp) (void *_a, void *_b)) {
         void *elem;
-        int i;
+        size_t i;
```
edenhill: keep as int

src/rdlist.c
```diff
@@ -127,7 +127,7 @@ void rd_list_destroy (rd_list_t *rl, void (*free_cb) (void *)) {
 		free_cb = rl->rl_free_cb;
 
         if (rl->rl_elems) {
-                int i;
+                size_t i;
```
edenhill: keep as int

src/rdlist.c
```diff
@@ -149,7 +149,7 @@ void *rd_list_elem (const rd_list_t *rl, int idx) {
 
 void *rd_list_find (const rd_list_t *rl, const void *match,
                     int (*cmp) (const void *, const void *)) {
-        int i;
+        size_t i;
```
edenhill: keep as int

src/rdlist.h
```diff
@@ -36,8 +36,8 @@
  */
 
 typedef struct rd_list_s {
-        int    rl_size;
-        int    rl_cnt;
+        size_t rl_size;
+        size_t rl_cnt;
         void **rl_elems;
```
edenhill: Not sure this is warranted?

src/rdlist.h
```diff
@@ -125,7 +125,7 @@ void *rd_list_elem (const rd_list_t *rl, int idx);
 /**
  * Returns the number of elements in list.
  */
-static __inline RD_UNUSED int rd_list_cnt (const rd_list_t *rl) {
+static __inline RD_UNUSED size_t rd_list_cnt (const rd_list_t *rl) {
```
edenhill: return int, anything else is just messy imho

ah-:
Cool, thanks. This is exactly the early feedback I was looking for, will fix and rebase once I'm done.

src/rdkafka_offset.c
ah-: Yeah, noticed that as well. It's unsigned anyways, so the -1 check doesn't make too much sense.

ah-:
Updated, rebased and squashed.

There's two warnings left:

``` rdrand.c(40): warning C4267: '=' : conversion from 'size_t' to 'int', possible loss of data```
is due to rd_jitter only returning 32bit ints, while it tries to shuffle a potentially 64bit long list.

```rdkafka_broker.c(1082): warning C4244: '=' : conversion from 'SOCKET' to 'int', possible loss of data```
not quite sure how to fix that cleanly.

File: src/rdcrc32.h [hunk 15]
```diff
@@ -70,7 +70,7 @@ rd_crc32_t rd_crc32_reflect(rd_crc32_t data, size_t data_len);
  *****************************************************************************/
 static __inline rd_crc32_t rd_crc32_init(void)
 {
-#if WITH_ZLIB
+#if WITH_ZLIB && !defined(_MSC_VER)
```
ah-: Due to zlib crc32 only running on uInt, which is 32 bit.

win32/librdkafka.vcxproj
```diff
@@ -92,7 +92,7 @@
     <ClCompile>
       <PrecompiledHeader>NotUsing</PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
-      <PreprocessorDefinitions>WIN32;_DEBUG;_WINDOWS;_USRDLL;LIBRDKAFKA_EXPORTS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>WIN32;_DEBUG;_WINDOWS;_USRDLL;LIBRDKAFKA_EXPORTS;_CRT_SECURE_NO_WARNINGS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
ah-: To disable warnings about strncpy being unsafe. It's used properly and strncpy_s isn't available on all platforms.

win32/librdkafka.vcxproj
```diff
@@ -225,7 +225,9 @@
     <ClCompile Include="..\src\rdlist.c" />
     <ClCompile Include="..\src\rdlog.c" />
     <ClCompile Include="..\src\rdrand.c" />
-    <ClCompile Include="..\src\snappy.c" />
+    <ClCompile Include="..\src\snappy.c">
+      <ExcludedFromBuild>true</ExcludedFromBuild>
```
ah-: Since it's unused on win32 anyways and causes a couple of warnings.

src/rdcrc32.h
edenhill: I dont really understand what you are trying to achieve here.
Are you trying to disable the use of zlib on win32? If so you should remove WITH_ZLIB from src/win32_config.h, but that seems questionable.

edenhill:
re rdrand.c:40 - just cast as necessary.

re SOCKET to int - I'll fix that after the merge.

src/rdcrc32.h
ah-: Make it work with all possible inputs. It currently silently gives you the wrong crc32 if you give it a buffer that's >2 GB.

But I suppose that's not going to happen, I'll change it to an assert < 2GB and use zlib.

ah-:
Rebased onto master and updated to just cast for rdrand.c and crc32.

edenhill:
Big thanks, @ah-, great work!

PR #498, opened 2016-01-07
Make SASL support configurable

coolfrood:
This PR makes SASL support configurable with `--enable-sasl` and `--disable-sasl` flags.


edenhill:
Looks good, thanks!

PR #499, opened 2016-01-08
Fix offset commit bug

shinya1020:
If a file is opened in `a+` mode, output is always appended to the end of file.
Therefore, in function `rd_kafka_offset_file_commit` ( https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_offset.c#L260-L342 ), `fseek` is ignored and offset file is not updated.


edenhill:
The problem is that "r+" does not create the file if it does not exist, so this will not work for new offset files.

edenhill:
Oh, nevermind, the file is already created from the open_cb.

edenhill:
Thanks!

PR #501, opened 2016-01-08
Fix consumer_callback does not pass opaque to consume_cb

shinya1020:
`consumer_callback` does not pass opaque to `consume_cb`.

### DIFF
--- a/src-cpp/ConsumerImpl.cpp
+++ b/src-cpp/ConsumerImpl.cpp
@@ -155,7 +155,7 @@ int RdKafka::ConsumerImpl::consume_callback (RdKafka::Topic* topic,
                                              RdKafka::ConsumeCb *consume_cb,
                                              void *opaque) {
   RdKafka::TopicImpl *topicimpl = static_cast<RdKafka::TopicImpl *>(topic);
-  ConsumerImplCallback context(topic, consume_cb, NULL);
+  ConsumerImplCallback context(topic, consume_cb, opaque);
   return rd_kafka_consume_callback(topicimpl->rkt_, partition, timeout_ms,
                                    &ConsumerImplCallback::consume_cb_trampoline, &context);
 }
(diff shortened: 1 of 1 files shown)


edenhill:
Thanks!

kparadkar:
@edenhill I am seeing this issue on V0.11.3 in C library (it used to work in 0.9.1)
in rd_kafka_consume0 latest code passes rd_kafka_poll_cb with RD_KAFKA_Q_CB_RETURN parameter returning immediately and do not update opaque - 

case RD_KAFKA_OP_FETCH:
                if (!rk->rk_conf.consume_cb ||
                    cb_type == RD_KAFKA_Q_CB_RETURN ||
                    cb_type == RD_KAFKA_Q_CB_FORCE_RETURN)
                        return RD_KAFKA_OP_RES_PASS; /* Dont handle here */
                   else {
                        struct consume_ctx ctx = {
                                .consume_cb = rk->rk_conf.consume_cb,
                                .opaque = rk->rk_conf.opaque };

                        return rd_kafka_consume_cb(rk, rkq, rko, cb_type, &ctx);
                }
                break;

Was this change intentional? Should I change my test ?

edenhill:
@kparadkar What API are you calling to consume messages and where do you expect to see the opaque?

kparadkar:
@edenhill I use rd_kafka_consumer_poll to consume messages. consume_cb updates the opaque in test code and asserts if opaque is not updated. This was test code which was added on older release of librdkafka and test used to work then which is broken now.

edenhill:
The consume callback does not work / is not supported with consumer_poll()

kparadkar:
rdkafka.h says - 
/**
 * @brief \b Consumer: Set consume callback for use with rd_kafka_consumer_poll()
 *
 */
RD_EXPORT
void rd_kafka_conf_set_consume_cb (rd_kafka_conf_t *conf,
                                   void (*consume_cb) (rd_kafka_message_t *
                                                       rkmessage,
                                                       void *opaque));

Is it documented somwhere that consume_cb is not supported with rd_kafka_consumer_poll ?

PR #504, opened 2016-01-09
Don't busy loop in rd_kafka_timers_run

ah-:
Idle consumers on Win used ~3-4% CPU without doing anything. Turns out they spent their time in a loop in rd_kafka_timers_run, because rd_clock only updates every now and then.

So make it only loop once if called with RD_POLL_NOWAIT, even if rd_clock doesn't change.

This brings idle CPU usage down to practically nothing. The same issue affects osx, but there the effect was smaller, I suspect linux is similar.


edenhill:
Looks good, thanks!

PR #506, opened 2016-01-11
Fix snappy on win32

ah-:
rd_ctz had _BitScanReverse mixed up with _BitScanForward, after that it just worked.

Also puts in a couple of casts to make it not complain during compilation, because snappy is quite optimized.


edenhill:
Big thanks!

PR #514, opened 2016-01-22
changes for Visual Studio 2010 build.

klonikar:
Needed to move C declarations to top of block and in struct initializations, named inializations had to be changed to old C style struct init.


edenhill:
Thanks for your work @klonikar.

I'm a bit hesitant at accepting these changes since they are quite extensive and remove a number of language constructs that make sure the code remains maintainable.

You should really be upgrading to a newer compiler version and if you are stuck on VS2010 for other parts of your project you can still compile librdkafka on a newer VS version and use that in your project.

I will be looking into providing pre-built win32&64 DLLs for librdkafka (both as downloadable installer as well as NuGet), but that is a couple of months out.

klonikar:
We are stuck at the moment on the old VS compiler. Also, I am not sure the binaries (DLLs) built with newer compilers are acceptable as they might require their own visual studio runtime DLLs.

I thought the changes were quite simple: Moving the declarations to the top of the block and initializing all elements of structs instead of initializing some of the struct elements by name.

As you can see, moving declarations to the top will not cause much of maintainability issues since its a simple one. This is the change in most of the files.

The change involving initializing struct fields by name is only in a couple of files (rdkafka_conf.c for the rd_kafka_properties array is the major one).

Had the C code been in C++, both the above would not have been issues. The problem is due the C compiler not supporting C99.

PR #522, opened 2016-01-28
Stop exporting internal symbols from shared libraries

takano-akio:
After the change to rewrite the lds-gen script in Python (in 0fdac6aa), internal symbols are exported in the shared libraries. This branch undoes the change.

The extra exported symbols somehow prevented my application from working, but I haven't tracked down how this happened.


edenhill:
Thanks!

PR #527, opened 2016-02-03
Make the build a bit faster

thedrow:


edenhill:
This fails on osx:
https://travis-ci.org/edenhill/librdkafka/jobs/106722515

thedrow:
Can the tests run in parallel?

edenhill:
The tests are run in parallel already.

What is the underlying problem with the current build speeds on Travis?
It is not a problem I've really encountered.

thedrow:
No problem. I just pick low hanging fruits if I see them.

thedrow:
Done. I'll file an issue about ccache not being available on OSX.

thedrow:
See https://github.com/travis-ci/travis-ci/issues/5567 for relevant travis issue.

thedrow:
@edenhill Done.

DEvil0000:
how about using the real core number instead of -j2?
something like this:
cpuCount=`lscpu -p | grep -v ^# | wc -l`
make -j$cpuCount

thedrow:
The real core number is not available in Travis due to a bug in their virtualization layer.
Each worker has two cores.

edenhill:
Thanks @thedrow!

PR #541, opened 2016-02-11
Include stdarg.h to avoid implicit use of va_start and va_arg in rdkafka_request.c

ewencp:


ewencp:
This is included elsewhere so I think it shouldn't be a cross platform compatibility issue, although there are different sets of preprocessor `#define`s before it in different files so I can't be sure.

edenhill:
Looks good, thanks!

PR #548, opened 2016-02-17
Change default partitioner to consistent-random (closes #546)

### LINKED ISSUE #546: Change default partitioner to consistent?
More of an open question, but wondering what you think about changing the default partitioner to _consistent_ by default. I'll admit the default _random_ partitioner caught me a bit off guard: After reading the Kafka docs, I figured messages with the same key would be assigned to the same topic, which is the default behavior in the JVM implementation.

Now setting the partitioner is not difficult in C, so I understand this isn't an issue with `librdkafka` per se, but I've been using the library through [HaskaKafka](https://github.com/cosbynator/haskakafka), which doesn't have a direct mechanism for this AFAICT. I have contacted the maintainer there about exposing this ([HaskaKafka Issue #17](https://github.com/cosbynator/haskakafka/issues/17))

sczizzo:
Set default partitioner to _consistent_ (rather than _random_) to behave more like Java/Scala implementations. Closes Issue #546

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -1153,7 +1153,6 @@ int rd_kafka_topic_partition_available(const rd_kafka_topic_t *rkt,
 /**
  * @brief Random partitioner.
  *
- * This is the default partitioner.
  * Will try not to return unavailable partitions.
  *
  * @returns a random partition between 0 and \p partition_cnt - 1.
@@ -1179,6 +1178,21 @@ int32_t rd_kafka_msg_partitioner_consistent (const rd_kafka_topic_t *rkt,
 					 int32_t partition_cnt,
 					 void *opaque, void *msg_opaque);
 
+/**
+ * Consistent-Random partitioner.
+ *
+ * This is the default partitioner.
+ * Uses consistent hashing to map identical keys onto identical partitions, and
+ * messages without keys will be assigned via the random partitioner.
+ *
+ * @returns a \"random\" partition between 0 and partition_cnt - 1 based on
+ *          the CRC value of the key (if provided)
+ */
+RD_EXPORT
+int32_t rd_kafka_msg_partitioner_consistent_random (const rd_kafka_topic_t *rkt,
+           const void *key, size_t keylen,
+           int32_t partition_cnt,
+           void *opaque, void *msg_opaque);
 
 
 /**@}*/
--- a/src/rdkafka_msg.c
+++ b/src/rdkafka_msg.c
@@ -340,6 +340,27 @@ int32_t rd_kafka_msg_partitioner_consistent (const rd_kafka_topic_t *rkt,
     return rd_crc32(key, keylen) % partition_cnt;
 }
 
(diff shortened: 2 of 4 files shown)


sczizzo:
Of course as soon as I open this I realize I neglected to update the comment above...

edenhill:
Existing applications that rely on the default partitioner to perform random distribution of key-less messages (e.g.. for load balancing) are in for a surprise.

edenhill:
Can you rebase the patch on latest master so it can be merged?

arnaud-lb:
@sczizzo: The default partitioner in the Java client is not strictly equivalent to rd_kafka_msg_partitioner_consistent.

In the Java client, the [default partitioning strategy](https://apache.googlesource.com/kafka/+/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java) works like this:

> * If a partition is specified in the record, use it
> * If no partition is specified but a key is present choose a partition based on a hash of the key
> * If no partition or key is present choose a partition in a round-robin fashion

rd_kafka_msg_partitioner_consistent will always assign messages to the same partition when a key is not provided.

edenhill:
Thanks for the info @arnaud-lb, I think we should mimic what the Java client does, either by modifying consistent's current functionality to fall back on the random partitioner in case there is no key, or by providing an additional partitioner that does that (to avoid changing the existing consistent partitioner behaviour)

sczizzo:
Rebased.

Totally agree with all this. A new default partitioner that mimics Java would be ideal. The _consistent_ partitioner is definitely going to upset users who don't key messages; I turned this on one of our testing environments yesterday, and surprised quite a few consumers.

sczizzo:
Thought I might take a crack at this proposed partitioner:
```
int32_t rd_kafka_msg_partitioner_consistent_keyed (const rd_kafka_topic_t *rkt,
                                             const void *key, size_t keylen,
                                             int32_t partition_cnt,
                                             void *rkt_opaque,
                                             void *msg_opaque) {
    if (keylen == 0)
      return rd_kafka_msg_partitioner_random(rkt,
                                             key,
                                             keylen,
                                             partition_cnt,
                                             rkt_opaque,
                                             msg_opaque);
    else
      return rd_kafka_msg_partitioner_consistent(rkt,
                                                 key,
                                                 keylen,
                                                 partition_cnt,
                                                 rkt_opaque,
                                                 msg_opaque);
}
```

edenhill:
@sczizzo Looks good, apart from the name, "_keyed" doesnt really make it distinct from the existing partitioners. Any other suggestions? E.g.:
 semi_consistent
 consistent_random / random_consistent
 consistent_roundrobin
 ...?

sczizzo:
Agreed. I guess _consistent_random_ and _random_consistent_ make most sense to me

edenhill:
Great, can you update the PR accordingly?

edenhill:
Great!
One last thing to do is add a dummy call to this new function in tests/0006-symbols.c
I see the consistent one is missng too so please add that while you are at it.

Thanks

edenhill:
Thanks!

PR #562, opened 2016-02-24
Client only connect to broker which locally known topic's partition locate in

JindongChen:
Client only connect to broker which locally known topic's partition locate in when topic.metadata.refresh.sparse is true.

This is to decreasing TCP link numbers of Kafka brokers if you have thousands of kafka client.


edenhill:
That's a neat fix but unfortunately I think there will be problems in a two-broker scenario if the partition leader broker goes down, librdkafka will not know about the new leader broker broker since there is no way to query metadata.

What I plan on doing is wait to connect to a broker until there is a request waiting to be sent, but that is post 0.9.1

PR #572, opened 2016-03-04
Represent feature listing as table

bruth:
Signed-off-by: Byron Ruth <b@devel.io>


edenhill:
Thanks, this looks good on github (markdown) but is very hard to read in text format (the README file is included in the Debian package, for example).

bruth:
Ah good call.

Issue #482, opened 2015-12-29
Busy loops when fetch.message.max.bytes is smaller than received message


edenhill:
From https://github.com/edenhill/kafkacat/issues/58:
> When kafkacat tries to fetch a message bigger than librdkafka fetch.message.max.bytes, kafkacat > loops forever and consumes all bandwidth and cpu.
> Instead it should throttle / print a message / return?

librdkafka should detect this and post an error op + back off.

edenhill:
See #320

Issue #320, opened 2015-06-29
Indicate fetch.message.max.bytes too small


edenhill:
Some way to properly indicate that fetch.message.max.bytes might be too small.
See https://github.com/edenhill/kafkacat/issues/41

PR #597, opened 2016-04-04
Dynamically increase the max fetch size when needed

takano-akio:
This patch allows the max fetch size (currently fixed at `fetch.msg.max.bytes`) to be increased dynamically when a larger message has to be consumed.

Motivation: I have an application where a client have to seek around in a topic and consume a small amount of data after each seek. I'd like to set `fetch.msg.max.bytes` to a small value to minimize bandwidth usage. However, currently this has a danger that a consumer may become permanently stuck when there is a message whose size exceeds this value. I'd like a way to use a small transfer sizes while avoiding the danger.

Regarding the implementation, it's possible that I have made mistakes because I'm not very familiar with the code. In particular I'm not sure that I correctly implemented the check to identify a case where the next message is too large to consume.

Related issues: #320 and #482.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3001,6 +3001,16 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
                                 rd_atomic64_add(&rktp->rktp_c.msgs,
 						rd_kafka_q_len(&tmp_opq));
 				rd_kafka_q_concat(&rktp->rktp_fetchq, &tmp_opq);
+                        } else {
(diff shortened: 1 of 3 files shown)


takano-akio:
Any comments on this? If this implementation is unsound and/or unacceptable, is there an alternative approach that may be taken?

asayers:
I'd also like to have this functionality. @edenhill, do you have any thoughts on this patch?

edenhill:
Thanks for your great patch @takano-akio.
It will be included in the next release.

PR #605, opened 2016-04-09
added virtual destructors to all classes that have a virtual function

danny794:
This ensures that the objects are properly cleaned up and also solves compilation issue when flag -Werror=non-virtual-dtor is present.


edenhill:
Thanks @danny794 , looks good!

PR #613, opened 2016-04-15
Fix a potential signed integer overflow that gcc 5.1 complains about

rthalley:
gcc 5.1 complains 
 rdlog.c:39:6: error: assuming signed overflow does not occur when
  assuming that (X + c) >= X is always true [-Werror=strict-overflow]


xkrt:
Please merge this, I cant build librdkafka in fresh Ubuntu Xenial 16.04 LTS

edenhill:
Thanks @rthalley !

Issue #539, opened 2016-02-10
KafkaConsumer: support per-partition queue consumption


FDePourcq:
Hi, question:
In a multithreaded application that listens to 1 kafka topic I have the choice between 
1 consumer for each thread versus 1 consumer for all threads.

But is the first scenario, 1 consumer for each thread, even possible? I see a lot of static keywords in the code. (And that global opaque value per singleton topic doesn't motivate either. The amount of connections this would open in the background could make it entirely foolish.)

edenhill:
Do you want each app thread to consume a single (set of) partition(s)?
Or do you want to consume all partitions from a single thread?

You are right in that each consumer instance will create 1 + N threads where N is the number of brokers.

FDePourcq:
each partition would be assigned to just one consumer, there are enough partitions... so each app thread would be the sole consumer of the partitions assigned to it

edenhill:
Okay, are you planning on using the high-level consumer or the simple consumer?
The former puts all messages in the same queue so that isn't perfect for your use-case, but the latter allows consumption per-partition and thus per-thread.

FDePourcq:
If I understand it well ... the high-level kafkaconsumer (both c++ kafkaconsumer as C-kafkacat-kafkaconsumer) have a singleton queue ... which would make it impossible to have one for each thread ... but the app could internally still organize itself to distribute the messages after arrival among the threads.

And alternatively I could go to the old style simple consumer, 1 per thread, but then I would have to do the rebalancing and offset-management from scratch. (and the whole offset-commit-to-broker-story would be entirely gone too)

To me it looks easiest to have 1 high-level-kafkaconsumer and distribute the messages behind it to the different threads. (Hoping and suspecting that it would be no performance-bottleneck.)

FDePourcq:
( the code i focus on is based on the highlevel kafkaconsumer-code in kafkacat.c )

edenhill:
KafkaConsumer, right, but not kafkacat - kafkacat simply enumerates the available partitions and consumes all of them , but there is no consumer group balancing involved (yet!).

If you need rebalancing you really dont want to do that yourself, then it is better to use KafkaConsumer as-is and dispatch messages to your worker threads from a single consumer->consume() thread precisely as you say.

Having said that it would not be a huge change to allow per-partition consumption in the high-level consumer, or more precisely, partition queue message routing.

edenhill:
Something like:
````
[...]

PR #628, opened 2016-04-23
issue #539 - support redirect of partition queue

dwieland:
#539

### DIFF
--- a/src-cpp/KafkaConsumerImpl.cpp
+++ b/src-cpp/KafkaConsumerImpl.cpp
@@ -119,6 +119,17 @@ RdKafka::Message *RdKafka::KafkaConsumerImpl::consume (int timeout_ms) {
 }
 
 
+RdKafka::Message *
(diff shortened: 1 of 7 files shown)


edenhill:
Some remaining things:
 * You will need to add the symbol to tests/0006-symbol.c  (with all arguments being 0 or NULL)
 * Add a C++ interface (optional)
 * Add a test case (optional)

dwieland:
I will take a look at the tests later on.

src-cpp/KafkaConsumerImpl.cpp
```diff
@@ -172,6 +183,24 @@ RdKafka::KafkaConsumerImpl::assign (const std::vector<TopicPartition*> &partitio
 
 
 RdKafka::ErrorCode
+RdKafka::KafkaConsumerImpl::redirectPartitionQueue (const TopicPartition *partition,
```
edenhill: Lets use the same name as in the C code as to make documentation references agnostic. (e.g. setPartitionQueue)

edenhill:
work is being done in #953

PR #643, opened 2016-05-04
test-suite-fix: tests that expect more than one partition now create topic ahead-of-time

janmejay:
These tests were failing because they were trying to publish to partitions that didn't exist (topics were being created with 1 partition (default)).


janmejay:
Not sure about this, but appveyor run seems to have failed due to an unrelated problem.

edenhill:
Thanks!

Issue #394, opened 2015-10-27
migrated to unknown broker


lea4n:
there are so many logs like this，even the data is still input to kafka.
%5|1445931284.285|TOPICBRK|test_kafka_client_#producer-0| Topic test_traffic_in [8] migrated to unknown testker 1: requesting metadata update
%5|1445931284.285|TOPICBRK|test_kafka_client_#producer-0| Topic test_traffic_in [2] migrated to unknown testker 2: requesting metadata update
%5|1445931284.285|TOPICBRK|test_kafka_client_#producer-0| Topic test_traffic_in [5] migrated to unknown testker 1: requesting metadata update
%5|1445931284.285|TOPICBRK|test_kafka_client_#producer-0| Topic test_traffic_in [4] migrated to unknown testker 0: requesting metadata update
%5|1445931284.285|TOPICBRK|test_kafka_client_#producer-0| Topic test_traffic_in [7] migrated to unknown testker 0: requesting metadata update
%5|1445931284.285|TOPICBRK|test_kafka_client_#producer-0| Topic test_traffic_in [1] migrated to unknown testker 1: requesting metadata update

There are 3 broker in my cluster, only one topic with six partitions

edenhill:
You dont need to worry about that, the condition should not be logged.

oyld:
Consumer has the same logs, I thought some error occured...

edenhill:
These log messages can be ignored..

edenhill:
Fixed on master

hensansi:
I have the same behavior on the 0.8 branch, can you fix it there too? 

```
172.17.0.1 - - [21/Jan/2016:10:15:24 +0000] "PUT /enqueue/product.new HTTP/1.1" 201 47 "-" "-"
%5|1453371341.704|TOPICBRK|rdkafka#producer-121| Topic product.new [0] migrated to unknown broker 0: requesting metadata update
172.17.0.1 - - [21/Jan/2016:10:15:24 +0000] "PUT /enqueue/product.new HTTP/1.1" 201 47 "-" "-"
%5|1453371341.836|TOPICBRK|rdkafka#producer-122| Topic product.new [0] migrated to unknown broker 0: requesting metadata update
172.17.0.1 - - [21/Jan/2016:10:15:41 +0000] "PUT /enqueue/product.new HTTP/1.1" 201 47 "-" "-"
```

edenhill:
Done

lea4n:
These log messages can be ignored..

PR #652, opened 2016-05-11
Silence 'migrated to unknown broker' log (issue #394)

hzhxxx:
what time to support Consumer group and Consumer offset?


edenhill:
You should be using master branch for 0.8 brokers too, the 0.8 branch shouldnt be used.

PR #683, opened 2016-06-02
Metadata Updates

incessantier:
1.  Adds ability to cache the latest metadata.  Cached metadata is saved to the broker struct and protected with the broker's mutex.
2.  Adds ability to retrieve the cached metadata by passing a < 0 timeout to the metadata call.
3.  Fixes a reference message request "leak":
   -  Metadata requests were not actually timing out correctly (the rkbuf timeout wasn't set to the configuration value).
   -  Metadata requests were always set to retry.  So even when shutting down the broker the message queue was populated and processing.  Set to not retry, relying instead on the periodic request of the metadata, allowing a clean shutdown.


edenhill:
The metadata layer has been rewritten.
Will make sure to add a get-cached-metadata variant.

Thanks for this though!

PR #694, opened 2016-06-16
Add rd_kafka_query_watermark_offsets call to rd_kafka_topic_leader_query

JohanAR:
Query watermark offsets uses rd_kafka_toppar_get2 to create a topic which
in turn calls rd_kafka_topic_new0, which unlike rd_kafka_topic_new does
not call rd_kafka_topic_leader_query.

This had the effect that rd_kafka_toppar_leader would return NULL until
rd_kafka_topic_leader_query was called by rd_kafka_topic_scan_tmr_cb,
which is run every 1000ms by a timer. Because of this query watermark
offsets would occasionally take 1000ms longer to finish, or return
WAIT_COORD error if timeout was shorter than that.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -1911,6 +1911,9 @@ rd_kafka_query_watermark_offsets (rd_kafka_t *rk, const char *topic,
 		return RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION;
 	rktp = rd_kafka_toppar_s2i(s_rktp);
 
+	/* Query for the topic leader (async) */
+	rd_kafka_topic_leader_query(rk, rktp->rktp_rkt);
+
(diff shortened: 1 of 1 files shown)


src/rdkafka.c
```diff
@@ -1911,6 +1911,9 @@ rd_kafka_query_watermark_offsets (rd_kafka_t *rk, const char *topic,
 		return RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION;
 	rktp = rd_kafka_toppar_s2i(s_rktp);
 
+	/* Query for the topic leader (async) */
```
edenhill: We only need to do this if there is no leader found, so maybe move this down a couple of lines below the failing rd_kafka_toppar_leader() check?

PR #698, opened 2016-06-20
Add a export api: rd_kafka_brokers_are_all_down

jerryyyq:
The api can help the caller to judge whether refuse client datas


edenhill:
Thanks for your effort but I belive this should be handled more generally by events that tells the application when brokers, topics and partitions are available or not. This saves checking these two atomics for each produce() call.

PR #726, opened 2016-07-08
Correct reference to enable.auto.commit in topic configuration doc fo…

ottomata:
…r auto.commit.enable


edenhill:
That file is automatically generated, you need to edit rdkafka_conf.c and run make

Issue #740, opened 2016-07-27
Lack of RdKafka::KafkaConsumer::consume_callback


lambdaknight:
On RdKafka::Consumer, there is the handy consume_callback method, but it is nowhere to be found on RdKafka::KafkaConsumer. Is there a technical reason for this or just an oversight?

cedarwu:
Same question.
I want to use high level RdKafka::KafkaConsumer in C++, but why no consume_callback in KafkaConsumer.

edenhill:
fixed on master

PR #741, opened 2016-07-27
Add support for setting "consume_cb" in the C++ wrapper of librdkafka

lambdaknight:


lambdaknight:
I opened up issue #740 and found I had a little bit of time, so I banged out this patch really quick. It worked when I tested it with the `rdkafka_consumer_example_cpp` example (which I appropriately modified), but I'm not super familiar with the library yet, so I don't know what I may be missing. I don't have a lot of extra time to devote to it, so if my patch is complete, yay. But if not, feel free to use whatever parts of my patch are useful.

edenhill:
Thanks!
(sorry for the loooong delay)

PR #742, opened 2016-07-27
Only store initial offset when fetcher was first started

yunjing:
There is a race condition for offset store. `rd_kafka_cgrp_partitions_fetch_start` should not overwrite the stored offset except when it was first started.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -1028,9 +1028,6 @@ rd_kafka_cgrp_partitions_fetch_start0 (rd_kafka_cgrp_t *rkcg,
                         shptr_rd_kafka_toppar_t *s_rktp = rktpar->_private;
                         rd_kafka_toppar_t *rktp = rd_kafka_toppar_s2i(s_rktp);
 
-			if (!RD_KAFKA_OFFSET_IS_LOGICAL(rktpar->offset))
-				rd_kafka_offset_store0(rktp, rktpar->offset, 1);
-
 			if (!rktp->rktp_assigned) {
 				rktp->rktp_assigned = 1;
 				rkcg->rkcg_assigned_cnt++;


edenhill:
Hi @yunjing,

can you elaborate on this fix?
You move offset_store0() down below the async op_fetch_start() call, but rktpar->offset doesnt change  during that time so why do you see the need for the move?
Also op_fetch_start() or op_seek() are asynchronous so  nothing will really have happened by the time they return

yunjing:
Hi @edenhill,

The problem I saw was that `rd_kafka_cgrp_partitions_fetch_start` may be invoked after the fetcher has started (as you commented in the `else` branch). As a result, without the fix, there is a possibility that the following scenario could happen:
1. A topic has three messages in the broker (index 0, 1, 2)
2. A consumer fetched all three from that topic
3. The consumer stores the current offset 3
4. The stored offset is overridden with 0 asynchronously

Because it only affects the stored offset, if there are new incoming messages to this topic, the consumer continues with the right offset and eventually stored the correct, current offset. However, if you kill the consumer and restart before any new messages arrived, it will incorrectly re-consume the three messages again.

meox:
any update on this issue?

yunjing:
@meox would love to have this reviewed.

edenhill:
Huhm, I dont think it should store the offset at all in that situation.
Offsets should only be stored (for future commits) when a message is delivered to the application (if enable.auto.offset.store=true) or explicitly by the application itself.
Doing it here, and not even checking the conf property, might very well commit offsets prematurely.

yunjing:
I agree with the assessment. Before we come up with a permanent solution, do you think this PR is a strict improvement of the existing logic?

edenhill:
I dont think we want to keep the store0() at all.

yunjing:
Okay, I see. I am not that familiar with the code, which is why I moved the two lines. So removing them won't have other side effects?

edenhill:
I did some quick analysis of the code yesterday and couldnt find a reason for why that store would be there, so I'm sure it should be removed.

yunjing:
OK. I updated the PR.

edenhill:
Thanks for locating this bug!

PR #769, opened 2016-08-29
Fix tiny WIN32 problems building 32-bit debug and release projects.

PKRoma:
Thanks for a great product.

In building the WIN32 debug and release versions of all projects, a few inconsistencies were found (library paths and libraries missing, VS2015 toolset being referenced, etc.). These have all been fixed and now all 32-bit projects build. The resultant changes to the .sln and .vcxproj files are in the pull request.

Also, the README.win32 has been updated to state that Microsoft Visual Studio 2013 should be used, and that the latest v1.0.2 version of OpenSSL should be used, because the actual latest version v1.1.0 has a changed API and this breaks compilation.

### DIFF
--- a/README.win32
+++ b/README.win32
@@ -1,11 +1,11 @@
 
-Native win32 build instructions using Microsoft Visual Studio (MSVC).
+Native win32 build instructions using Microsoft Visual Studio 2013 (MSVC).
 
 Requirements:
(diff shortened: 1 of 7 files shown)


edenhill:
Thanks a bunch!

PR #809, opened 2016-09-28
Add missing return in rd_kafka_curr_msgs_get if RD_KAFKA_PRODUCER.

PKRoma:
There is a pattern in the rd_kafka_curr_msgs_* functions in rdkafka_int.h to avoid referencing the rk_curr_msgs structure if the application is not a RD_KAFKA_PRODUCER.

However, for the rd_kafka_curr_msgs_get function, while there exists a check for RD_KAFKA_PRODUCER and return values are set to 0, there is no subsequent return statement and the function falls through to manipulating the rk_curr_msgs structure even when the application is not a RD_KAFKA_PRODUCER, causing exceptions on trying to lock the structure on WIN32.

This patch adds the missing return statement.


edenhill:
Thank you for finding and fixing this issue!

PR #810, opened 2016-09-29
Change the order of parameter explanation in intro for rd_kafka_produce fn

rustycodemonkey:
I just changed the order of your explanation in the introduction file so that it reflect the actual function.
Sorry for being pedantic.


edenhill:
Thanks :)

PR #812, opened 2016-09-29
slove consumer hang when it close

yanzhiqiang:
1：I  just changed the rdkafka close broker thread so that  broker thread can be closed.
2：I encounted one problem consumer close will be hanged because rd_kafka_broker_terminating(rb) not true。

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -569,9 +569,9 @@ static void rd_kafka_destroy_internal (rd_kafka_t *rk) {
                 if (rk->rk_conf.term_sig)
 			pthread_kill(rkb->rkb_thread, rk->rk_conf.term_sig);
 #endif
-
-                rd_kafka_broker_destroy(rkb);
-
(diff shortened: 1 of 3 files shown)


src/rdkafka.c
```diff
@@ -569,9 +569,9 @@ static void rd_kafka_destroy_internal (rd_kafka_t *rk) {
                 if (rk->rk_conf.term_sig)
 			pthread_kill(rkb->rkb_thread, rk->rk_conf.term_sig);
 #endif
-
-                rd_kafka_broker_destroy(rkb);
-
+		while(rd_kafka_broker_terminating(rkb)) {
+                	rd_kafka_broker_destroy(rkb);
```
edenhill: It is hanging because there are outstanding refcounts to the broker, which means something references it (typically a partition with messages on it).
Forcefully decreasing the refcounts like this will eventually cause a use-after-free crash of the objects that reference the broker.

PR #824, opened 2016-10-07
Remove logo setting from Doxygen so recent versions won't complain about it being missing.

ewencp:


ewencp:
@edenhill Not sure if this is the ideal solution or not. Apache can be picky about the use of project logos so I didn't go with adding the file in this case and instead just removed it so we'll be logo-free for now.

PR #836, opened 2016-10-13
rd_kafka_new: delete conf only on success

julien-lecomte:
This patch allows the user to be sure the rd_kafka_conf_t* is only deleted on success when calling rd_kafka_new.

It works by first duplicating the user conf if supplied, or creating a new conf. This means that this duplicate conf may be deleted from the same places in rd_kafka_new as before.

At the end of rd_kafka_new: delete the user supplied conf to keep current behavior.

If we were to change behavior and have user create and delete conf, it's as easy as just deleting the last lines of function rd_kafka_new.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -1097,28 +1097,30 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *conf,
 			  char *errstr, size_t errstr_size) {
 	rd_kafka_t *rk;
 	static rd_atomic32_t rkid;
+        rd_kafka_conf_t *use_conf;             
 #ifndef _MSC_VER
         sigset_t newset, oldset;
 #endif
 
 	call_once(&rd_kafka_global_init_once, rd_kafka_global_init);
 
-
-	if (!conf)
-		conf = rd_kafka_conf_new();
+        /* Use a copy of conf if supplied, otherwise generate a new one so we can always
+         * delete it upon return without additional conditionals
+         */
+        use_conf = (conf ? rd_kafka_conf_dup(conf) : rd_kafka_conf_new());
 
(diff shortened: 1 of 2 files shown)


src/rdkafka.c
```diff
@@ -1103,22 +1103,23 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *conf,
 
 	call_once(&rd_kafka_global_init_once, rd_kafka_global_init);
 
-
-	if (!conf)
-		conf = rd_kafka_conf_new();
+        /* Use a copy of conf if supplied, otherwise generate a new one so we can always
+         * delete it upon return without additional conditionals
+         */
+        rd_kafka_conf_t* use_conf = (conf ? rd_kafka_conf_dup(conf) : rd_kafka_conf_new());
```
edenhill: nitpick: pointer on variable side, not type side: `rd_kafka_conf_t *use_conf = ..`

Compare:
`my_type_t* a, b;` <-- looks like a and b have same type attr
to
`my_type_t *a, b;` <-- more evident that they do not

edenhill: Inline (not at start of scope) variable declarations are not allowed (for portability's sake), put it with the rest of the variables at the top.

src/rdkafka.h
```diff
@@ -1346,7 +1346,7 @@ int32_t rd_kafka_msg_partitioner_consistent_random (const rd_kafka_topic_t *rkt,
  *
  * \p conf is an optional struct created with `rd_kafka_conf_new()` that will
  * be used instead of the default configuration.
- * The \p conf object is freed by this function and must not be used or
+ * The \p conf object is freed by this function on success must not be used or
```
edenhill: "on success **and** must not.."

edenhill:
Thanks!

PR #850, opened 2016-10-21
fix crash in rd_kafka_new() when conf is NULL

wukezhan:
when we use the latest code of https://github.com/edenhill/librdkafka and https://pecl.php.net/package/rdkafka, the php core dumped, so i fixed it.

please review if it's ok, thanks:)

the php code:
```php
$rk = new RdKafka\Producer();
$rk->setLogLevel(LOG_DEBUG);
```

the core dump backtrace:
```
Program received signal SIGSEGV, Segmentation fault.
rd_kafka_new (type=RD_KAFKA_CONSUMER, conf=0x0, errstr=0x7fffffffa9a0 "\316\006", errstr_size=512) at rdkafka.c:1113
1113	        if (!conf->socket_cb) {
Missing separate debuginfos, use: debuginfo-install audit-libs-2.3.7-5.el6.x86_64 cyrus-sasl-lib-2.1.23-15.el6_6.2.x86_64 freetype-2.3.11-15.el6_6.1.x86_64 glibc-2.12-1.149.el6_6.9.x86_64 keyutils-libs-1.4-5.el6.x86_64 krb5-libs-1.10.3-37.el6_6.x86_64 libcom_err-1.41.12-21.el6.x86_64 libcurl-7.19.7-46.el6.x86_64 libgcc-4.4.7-11.el6.x86_64 libgomp-4.4.7-11.el6.x86_64 libidn-1.18-2.el6.x86_64 libjpeg-turbo-1.2.1-3.el6_5.x86_64 libselinux-2.0.94-5.8.el6.x86_64 libssh2-1.4.2-1.el6_6.1.x86_64 libxml2-2.7.6-17.el6_6.1.x86_64 libyaml-0.1.6-1.el6.x86_64 nspr-4.10.8-1.el6_6.x86_64 nss-3.19.1-3.el6_6.x86_64 nss-softokn-freebl-3.14.3-22.el6_6.x86_64 nss-util-3.19.1-1.el6_6.x86_64 openldap-2.4.39-8.el6.x86_64 openssl-1.0.1e-42.el6.x86_64 pam-1.1.1-20.el6.x86_64 postgresql-libs-8.4.20-3.el6_6.x86_64 sqlite-3.6.20-1.el6.x86_64 zlib-1.2.3-29.el6.x86_64
(gdb) backtrace
[...]

PR #851, opened 2016-10-21
FIX SIGSEGV if rd_kafka_new called with NULL conf

eugpermar:


eugpermar:
Nobody can use this feature at this moment, so another valid option could be to only allow to call `rd_kafka_new` with a valid conf, never a NULL one.

eugpermar:
I didn't see the provided solution in PR #850 minutes ago. I think that one is better since it has no sense to check if default config has these cb. I close this one.

PR #852, opened 2016-10-21
Fix for async offset commit issue on AIX

asharma339:
While running tests on IBM-AIX, I experienced a crash while running 0030-offset_commit for  manual async commit. What's happening is that, it is evaluating both expressions and repq being NULL for async case, wreaks havoc when NULL is being passed to rd_kafka_q_keep().

### DIFF
--- a/src/rdkafka_offset.c
+++ b/src/rdkafka_offset.c
@@ -413,6 +413,7 @@ rd_kafka_commit (rd_kafka_t *rk,
         rd_kafka_cgrp_t *rkcg;
 	rd_kafka_resp_err_t err;
 	rd_kafka_q_t *repq = NULL;
+	rd_kafka_replyq_t rq = RD_KAFKA_NO_REPLYQ;
 
         if (!(rkcg = rd_kafka_cgrp_get(rk)))
                 return RD_KAFKA_RESP_ERR__UNKNOWN_GROUP;
@@ -420,15 +421,14 @@ rd_kafka_commit (rd_kafka_t *rk,
         if (!async)
                 repq = rd_kafka_q_new(rk);
 
-        err = rd_kafka_commit0(rk, offsets, NULL,
-			       !async ? RD_KAFKA_REPLYQ(repq, 0) :
-			       RD_KAFKA_NO_REPLYQ, NULL, NULL);
+        if (!async) 
+		rq = RD_KAFKA_REPLYQ(repq, 0);
+ 
+        err = rd_kafka_commit0(rk, offsets, NULL, rq, NULL, NULL);
 
-        if (!async) {
+        if (!err && !async) {
 		err = rd_kafka_q_wait_result(repq, RD_POLL_INFINITE);
-                rd_kafka_q_destroy(repq);
-        } else {
-                err = RD_KAFKA_RESP_ERR_NO_ERROR;
+		rd_kafka_q_destroy(repq);
         }
 
 	return err;
--- a/src/rdkafka_queue.h
+++ b/src/rdkafka_queue.h
@@ -380,7 +380,11 @@ uint64_t rd_kafka_q_size (rd_kafka_q_t *rkq) {
 #endif
 
 /* Construct temporary on-stack replyq for indicating no replyq. */
-#define RD_KAFKA_NO_REPLYQ  (rd_kafka_replyq_t){NULL, 0}
+#if ENABLE_DEVEL
+#define RD_KAFKA_NO_REPLYQ (rd_kafka_replyq_t){NULL, 0, rd_strdup(__FUNCTION__)}
+#else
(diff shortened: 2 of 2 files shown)


src/rdkafka_offset.c
```diff
@@ -420,9 +420,17 @@ rd_kafka_commit (rd_kafka_t *rk,
         if (!async)
                 repq = rd_kafka_q_new(rk);
 
-        err = rd_kafka_commit0(rk, offsets, NULL,
-			       !async ? RD_KAFKA_REPLYQ(repq, 0) :
-			       RD_KAFKA_NO_REPLYQ, NULL, NULL);
+	/* Avoid conditional expression for following as that results in 0030-offset_commit
+	 * test to fail on AIX machine.
+	 */
+	rd_kafka_replyq_t rq;
```
edenhill: Put declarations at beginning of scope (line 416)

src/rdkafka_offset.c
```diff
@@ -420,9 +420,17 @@ rd_kafka_commit (rd_kafka_t *rk,
...
+        if (!async) {
```
edenhill: remove braces for single-line if/else-statements

src/rdkafka_offset.c
```diff
@@ -420,9 +420,17 @@ rd_kafka_commit (rd_kafka_t *rk,
...
+		rq = RD_KAFKA_REPLYQ(repq, 0);
+        }
+        else {
+		rq = RD_KAFKA_NO_REPLYQ;
```
edenhill: You could initialize rq to this on declaration

src/rdkafka_offset.c
asharma339: Done

src/rdkafka_offset.c
asharma339: Done

asharma339:
It is correct that lot of places use ternary operator in the code, but this place is unique for couple of reasons. Most of the places have scalars as the second and third operands for the ternary operator. This place uses a macro and object construction with member passed in curly brackets. In my opinion, this combination is resulting in arguments to object constructor being evaluated and resulting in a call to rd_kafka_q_keep(), which I verified by print statements. I checked xlc documentation, their claim is that either second or third operand is evaluated based on first operand to ternary operator. But silent on when macros are involved.  In my testing this is the only place failing. My suggestion is to keep an eye on any such issues if I come across on IBM and also follow up on research on this unique issue for IBM.  Until then, we should keep this simple fix.

src/rdkafka_offset.c
```diff
@@ -413,16 +413,22 @@ rd_kafka_commit (rd_kafka_t *rk,
...
         if (!(rkcg = rd_kafka_cgrp_get(rk)))
                 return RD_KAFKA_RESP_ERR__UNKNOWN_GROUP;
 
         if (!async)
                 repq = rd_kafka_q_new(rk);
 
-        err = rd_kafka_commit0(rk, offsets, NULL,
-			       !async ? RD_KAFKA_REPLYQ(repq, 0) :
-			       RD_KAFKA_NO_REPLYQ, NULL, NULL);
+	/* Avoid conditional expression for following as that results in 0030-offset_commit
```
edenhill: Remove this comment, the commit message is enough, dont need reasoning in the code.

src/rdkafka_offset.c
```diff
@@ -413,16 +413,22 @@ rd_kafka_commit (rd_kafka_t *rk,
...
+	 * test to fail on AIX machine.
+	 */
+
+        if (!async) 
+		rq = RD_KAFKA_REPLYQ(repq, 0);
+ 
+        err = rd_kafka_commit0(rk, offsets, NULL, rq, NULL, NULL);
 
         if (!async) {
```
edenhill: I think this if .. else should be changed to:
    if (!err && !async) { 
      err = rd_kafka_q_wait..
     ..
    }

asharma339:
Done.

src/rdkafka_offset.c
```diff
@@ -413,18 +413,20 @@ rd_kafka_commit (rd_kafka_t *rk,
...
+        if (!async) 
+		rq = RD_KAFKA_REPLYQ(repq, 0);
+ 
+        err = rd_kafka_commit0(rk, offsets, NULL, rq, NULL, NULL);
 
-        if (!async) {
+        if (!err && !async) {
 		err = rd_kafka_q_wait_result(repq, RD_POLL_INFINITE);
                 rd_kafka_q_destroy(repq);
         } else {
```
edenhill: remove this else and the error assignment, err is already set from rd_kafka_commit0() return value when doing non-async

asharma339:
Done

edenhill:
Thanks!

Issue #837, opened 2016-10-13
librdkafka on IBM machine with native compiler


abhisharma:
# Description

I tried to build and test 32 bit librdkafka on IBM machine using native compiler. Here is what I observed: 

Librdkafka-0.9.1
- I can build but rd_atomic32_get() (i.e. __sync_add_and_fetch()) operation crashes on running tests.
- I replaced  __sync_add_and_fetch() operations with __fetch_and_add() operations, but I cannot compile code because long long is not permitted on __fetch_and_add()  and I cannot use __fetch_and_addlp() either with 32 bit build. Then I just typecasted/typedefed int64\* to int32\* which builds but tests crash on atomic_\* operations.

Librdkafka-master
- I can build but rd_atomic32_get() (i.e. __sync_add_and_fetch()) operation crashes on running tests.
- If I undef HAVE_ATOMICS_32 and HAVE_ATOMICS_64, so that atomic operations use mutex, then it builds fine but tests hang which looks like following:

386258.376|METADATA|0001_multiobj#producer-4| 10.122.140.84:9092/bootstrap:   Topic bib_rnd64b000005215_0001 partition 1 Leader 0
%7|1476386258.376|METADATA|0001_multiobj#producer-4| 10.122.140.84:9092/bootstrap:   Topic bib_rnd64b000005215_0001 partition 3 Leader 2
%7|1476386258.376|METADATA|0001_multiobj#producer-4| 10.122.140.84:9092/bootstrap:   Topic bib_rnd64b000005215_0001 partition 0 Leader 2
%7|1476386258.376|METADATA|0001_multiobj#producer-4| 10.122.140.84:9092/bootstrap: Requested topic bib_rnd64b000005215_0001 seen in metadata
%7|1476386259.297|PRODUCE|0001_multiobj#producer-4| mrplnjdmrsmr02:9092/2: produce messageset with 1 messages (64 bytes)
%7|1476386259.299|MSGSET|0001_multiobj#producer-4| mrplnjdmrsmr02:9092/2: MessageSet with 1 message(s) delivered
**%7|1476386259.299|REQERR|0001_multiobj#producer-4| mrplnjdmrsmr02:9092/2: ProduceRequest failed: Broker: Unknown topic or partition: explicit actions 0x4
[...]

PR #857, opened 2016-10-24
Fix for AIX-- exclude endian.h on AIX

asharma339:
A few more changes for AIX:
1. malloc/calloc return NULL for 0 size memory request.
2. As described in issue  #837, need to use mutexes for atomic ops.

### DIFF
--- a/src/snappy_compat.h
+++ b/src/snappy_compat.h
@@ -19,7 +19,7 @@
 #endif
 
 
-#elif !defined(__WIN32__) && !defined(_MSC_VER) && !defined(sun)
+#elif !defined(__WIN32__) && !defined(_MSC_VER) && !defined(sun) && !defined(_AIX)
 #  include <endian.h>
 #endif
 


src/rd.h
```diff
@@ -79,13 +79,17 @@
 */
 static RD_INLINE RD_UNUSED void *rd_calloc(size_t num, size_t sz) {
 	void *p = calloc(num, sz);
+#ifndef _AIX  // On AIX, calloc(0) is NULL
```
edenhill: Have you observed a case where it allocs zero bytes?
If so, this will break that code.

src/rdatomic.h
```diff
@@ -2,6 +2,11 @@
 
 #include "tinycthread.h"
 
+#if defined(_AIX)
```
edenhill: This is what ./configure is for

src/rd.h
asharma339: This assertion fails when running tests and I printed to check the size which was 0. 
 I didn't  witness any crash after adding this change.  As far as free() is concerned, it should not cause any issue because on IBM free(NULL) is a No-Op.

asharma339:
Cleaned up. Now the change is just for excluding endian.h

edenhill:
Thanks.
Can you rebase the PR so that the fixes-after-review are combined with the original commit?
Something like `git rebase -i origin/master` then set the fixes-on-fixes as 'fixup'

asharma339:
Done

PR #858, opened 2016-10-25
Fix timestamp extraction for compressed messages

yunjing:
Currently timestamp extraction fails for compressed messages. It would report `-1` instead of the actual timestamp because the timestamps for compressed messages are stored at the block level. This PR passes down the header for the block to each of the uncompressed messages.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3471,6 +3471,17 @@ static char *rd_kafka_snappy_java_decompress (rd_kafka_broker_t *rkb,
 #endif
 
 
+/**
+ * Message header
+ */
+struct msg_hdr {
+	int64_t Offset;
+	int32_t MessageSize;
+	uint32_t Crc;
+	int8_t  MagicByte; /* MsgVersion */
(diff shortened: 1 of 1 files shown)


yunjing:
@edenhill any suggestion on this? We are trying to open source a project in a week, but this bug is blocking us.

edenhill:
Hey, I'll look at this during the weekend and get back to you.

yunjing:
Thanks @edenhill. Really appreciated!

qix:
Hi, we've just pushed an open source project for which we require these changes in order to run in production: https://github.com/smyte/kafka_store

For now we're recommending that users install our fork, which we'll revert as soon as this makes it into a release.

edenhill:
Thanks!

edenhill:
Can you verify the fix on your end, thanks?

yunjing:
Verified that it worked by compiling kafkacat with librdkafka master and consuming from several topics in production.

We have not yet incorporated the latest librdkafka in our production services.

PR #859, opened 2016-10-25
Fix relative offset calculation

yunjing:
`rd_kafka_q_fix_offsets` purges messages by comparing a relative offset to the absolute min offset. It only works when the absolute min offset is 0.

### DIFF
--- a/src/rdkafka_queue.c
+++ b/src/rdkafka_queue.c
@@ -766,6 +766,7 @@ void rd_kafka_q_fix_offsets (rd_kafka_q_t *rkq, int64_t min_offset,
 		if (unlikely(rko->rko_type != RD_KAFKA_OP_FETCH))
 			continue;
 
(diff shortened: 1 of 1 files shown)


src/rdkafka_queue.c
```diff
@@ -766,7 +767,7 @@ void rd_kafka_q_fix_offsets (rd_kafka_q_t *rkq, int64_t min_offset,
 		if (unlikely(rko->rko_type != RD_KAFKA_OP_FETCH))
 			continue;
 
-		if (rko->rko_u.fetch.rkm.rkm_offset < min_offset &&
```
edenhill: I think a cleaner fix is to just move the rkm_offset assignment above the if statement, this:
`rko->rko_u.fetch.rkm.rkm_offset += base_offset;`

yunjing:
Fixed.

edenhill:
Thanks for locating this issue!

PR #861, opened 2016-10-25
changes after checking with cppcheck 

joseAndresGomezTovar:
A tool for static C/C++ code analysis: more info here
(http://cppcheck.sourceforge.net/)
---

Summary: Prefer prefix ++/-- operators for non-primitive types.
Message: Prefix ++/-- operators should be preferred for non-primitive
types. Pre-increment/decrement can be more efficient than
post-increment/decrement. Post-increment/decrement usually involves
keeping a copy of the previous value around and adds a little extra
code.
---

Summary: Variable 'topic_' is assigned in constructor body. Consider
performing initialization in initialization list.
Message: When an object of a class is created, the constructors of all
member variables are called consecutively in the order the variables are
declared, even if you don't explicitly write them to the initialization
list. You could avoid assigning 'topic_' a value by passing the value to
the constructor in the initialization list.

First included by: src-cpp\ConfImpl.cpp
---

Summary: %zd in format string (no. 2) requires 'ssize_t' but the
argument type is 'size_t {aka unsigned long long}'.
Message: %zd in format string (no. 2) requires 'ssize_t' but the
argument type is 'size_t {aka unsigned long long}'.
---

Summary: Either the condition 'if(rko_orig)' is redundant or there is
possible null pointer dereference: rko_orig.
Message: Either the condition 'if(rko_orig)' is redundant or there is
possible null pointer dereference: rko_orig.
---

there are advise/errors but I don´t know if it is an error or not... I
am noob with kafka

1)
file rdkafka_broker.c

Summary: Dead pointer usage. Pointer 'msg' is dead if it has been
assigned '&msg2' at line 1760.
Message: Dead pointer usage. Pointer 'msg' is dead if it has been
assigned '&msg2' at line 1760.

2)
file
rdkafka_cgrp.c

Summary: Conversion of string literal "RD_KAFKA_OP_OFFSET_COMMIT" to
bool always evaluates to true.
Message: Conversion of string literal "RD_KAFKA_OP_OFFSET_COMMIT" to
bool always evaluates to true.

3)
file
rdrand.c

Summary: Obsolete function 'alloca' called. In C99 and later it is
recommended to use a variable length array instead.
Message: The obsolete function 'alloca' is called. In C99 and later it
is recommended to use a variable length array or a dynamically allocated
array instead. The function 'alloca' is dangerous for many reasons
(http://stackoverflow.com/questions/1018853/why-is-alloca-not-considered-good-practice
and http://linux.die.net/man/3/alloca).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ gmon.out
 *.rpm
 staging-docs
 tmp
+win32/librdkafka.VC.db
--- a/README.win32
+++ b/README.win32
@@ -24,3 +24,11 @@ Artifacts:
   - remaining tools (rdkafka_performance, etc)
   - SASL support (no official Cyrus libsasl2 DLLs available)
   - LZ4 support
+
+Configuration:
+ - You can disable/enable SSL using the definition WITH_SSL=0
+   0: disable
+   1: enable (by default)
+ - You can disable/enable ZLIB using the definition WITH_ZLIB=0
+   0: disable
+   1: enable (by default)
--- a/examples/kafkatest_verifiable_client.cpp
+++ b/examples/kafkatest_verifiable_client.cpp
@@ -77,8 +77,7 @@ public:
                 minOffset(-1), maxOffset(0) {
     printf("Created assignment\n");
   }
-  Assignment(const Assignment &a) {
-    topic = a.topic;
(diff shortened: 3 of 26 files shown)


edenhill:
I havent had time to look at this yet.
What is the reason you close it?

PR #863, opened 2016-10-25
fix func rd_kafka_new declare use_conf but not use cause conf->socket…

wadee:
fix func rd_kafka_new declare use_conf but not use cause conf->socket_cb segmentfault


edenhill:
Thanks!

PR #865, opened 2016-10-26
Replacement for builtins for sun and ibm

asharma339:
Since  __builtin_clz, __builtin_ctz, __builtin_ctzll are not available for  AIX and Sun, provided the replacements

### DIFF
--- a/src/snappy.c
+++ b/src/snappy.c
@@ -126,7 +126,58 @@ static inline bool is_little_endian(void)
 	return false;
 }
 
-#ifndef _MSC_VER
+#if defined(__xlc__) // xlc compiler on AIX
+#define rd_clz(n)   __cntlz4(n)
+#define rd_ctz(n)   __cnttz4(n)
+#define rd_ctz64(n) __cnttz8(n)
+
+#elif defined(__SUNPRO_C) // Solaris Studio compiler on sun  
+/*
+ * Source for following definitions is Hacker’s Delight, Second Edition by Henry S. Warren
+ * http://www.hackersdelight.org/permissions.htm
+ */
+u32 rd_clz(u32 x) {
+   u32 n;
+
+   if (x == 0) return(32);
+   n = 1;
+   if ((x >> 16) == 0) {n = n +16; x = x <<16;}
+   if ((x >> 24) == 0) {n = n + 8; x = x << 8;}
+   if ((x >> 28) == 0) {n = n + 4; x = x << 4;}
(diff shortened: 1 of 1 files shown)


src/snappy.c
```diff
@@ -126,7 +126,55 @@ static inline bool is_little_endian(void)
 	return false;
 }
 
-#ifndef _MSC_VER
+#if defined(_AIX)
+#define rd_clz(n)   __cntlz4(n)
+#define rd_ctz(n)   __cnttz4(n)
+#define rd_ctz64(n) __cnttz8(n)
+
+#elif defined(sun)
```
edenhill: I believe sun is defined by gcc too and gcc should be providing proper builtins for these things.

src/snappy.c
```diff
@@ -126,7 +126,55 @@ static inline bool is_little_endian(void)
 	return false;
 }
 
-#ifndef _MSC_VER
+#if defined(_AIX)
```
edenhill: This should *probably* be done in configure but we might get away with it here.
I think there's also an important point to separate toolchain from platform, for example the gcc toolchain will most likely provide proper builtins on all platforms.

asharma339:
Done. Used toolchain based checks.

src/snappy.c
```diff
@@ -126,7 +126,55 @@ static inline bool is_little_endian(void)
 	return false;
 }
 
-#ifndef _MSC_VER
+#if defined(__xlc__)
```
edenhill: Please add a comment what this is (.e.g, "xlc compiler on AIX")

src/snappy.c
```diff
@@ -126,7 +126,55 @@ static inline bool is_little_endian(void)
...
+#define rd_clz(n)   __cntlz4(n)
+#define rd_ctz(n)   __cnttz4(n)
+#define rd_ctz64(n) __cnttz8(n)
+
+#elif defined(__SUNPRO_C)
+// Source for following defintions is Hacker’s Delight, Second Edition by Henry S. Warren
```
edenhill: typo: definitions

What's the license for these?

src/snappy.c
asharma339: Done

src/snappy.c
asharma339: Fixed typo. I have added the link to use content which says : "You are free to use, copy, and distribute any of the code on this web site, whether modified by you or not. You need not give attribution. "

asharma339:
Done

edenhill:
Thank you!

PR #869, opened 2016-10-27
Use compiler specific atomic ops for sun studio compiler

asharma339:
This is to make atomic functions work  using native sun solaris compiler. I could not find any good way to do it thru config.


edenhill:
Looks great, thanks!

PR #870, opened 2016-10-28
fix bug rd_kafka_timer_t microsecond interval overflow.

lvxv:
./rdkafka_performance -P -b xxxx:xx -t test_topic -i 3600000
% Configuration property "statistics.interval.ms" value -694967 is outside allowed range 0..86400000

statistics.interval.ms 86400000
int rtmr_interval max :  2147483647(2^31)

86400000 * 1000 > 2147483647(2^31)

### DIFF
--- a/examples/rdkafka_performance.c
+++ b/examples/rdkafka_performance.c
@@ -48,7 +48,7 @@
 
 static int run = 1;
 static int forever = 1;
-static int dispintvl = 1000;
+static rd_ts_t dispintvl = 1000;
 static int do_seq = 0;
 static int exit_after = 0;
 static int exit_eof = 0;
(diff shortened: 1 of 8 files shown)


examples/rdkafka_performance.c
```diff
@@ -1000,7 +1000,7 @@ int main (int argc, char **argv) {
         if (!stats_intvlstr) {
                 /* if no user-desired stats, adjust stats interval
                  * to the display interval. */
-                snprintf(tmp, sizeof(tmp), "%i", dispintvl / 1000);
+                snprintf(tmp, sizeof(tmp), "%ld", dispintvl / 1000);
```
edenhill: Use `"%"PRId64` here to be portable

edenhill:
Thanks!

PR #880, opened 2016-10-31
add snappy function header to be compatible with google snappy

hustlijian:
add snappy function header to be compatible with google snappy


edenhill:
Thank you!

PR #903, opened 2016-11-14
adds check of defined LIBKAFKARD_STATICLIB to enable static library b…

skarlsson:
I currently build libkafkard out of source with cmake but it would be convenient to make it possible to build  static libs. This pull request only enables that - it does not change anything for existing builds.


skarlsson:
oops typo should be LIBRDKAFKA_STATICLIB closing and fixing..

PR #905, opened 2016-11-14
adds check of defined LIBRDKAFKA_STATICLIB to enable static library

skarlsson:
I currently build librdkafka out of source with cmake but it would be convenient to make it possible to build static libs. This pull request only enables that - it does not change anything for existing builds.

### DIFF
--- a/README.win32
+++ b/README.win32
@@ -24,3 +24,7 @@ Artifacts:
   - remaining tools (rdkafka_performance, etc)
   - SASL support (no official Cyrus libsasl2 DLLs available)
   - LZ4 support
+
+If you build librdkafka with an external tool (ie CMake) you can get rid of the 
+__declspec(dllexport) / __declspec(dllimport) decorations by adding a define
+-DLIBRDKAFKA_STATICLIB to your CFLAGS
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -54,10 +54,14 @@
 
 #ifdef _MSC_VER
 #undef RD_EXPORT
+#ifdef LIBRDKAFKA_STATICLIB
+#define RD_EXPORT
+#else
 #ifdef LIBRDKAFKACPP_EXPORTS
 #define RD_EXPORT __declspec(dllexport)
 #else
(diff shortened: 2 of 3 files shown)


edenhill:
Looks good, can you add a small snippet to README.win32 on when (and how) to use LIBRDKAFKA_STATICLIB?

skarlsson:
sure, but as I said I needed to build outside of librdkafka source tree so i ended up just globbing all files except rdkafka_sasl.c


this is what i did in my CmakeLists.txt

...

file(GLOB librdkafka_files
  ${CMAKE_CURRENT_SOURCE_DIR}/../librdkafka/src/*.c
  ${CMAKE_CURRENT_SOURCE_DIR}/../librdkafka/src/*.h
  )

file(GLOB librdkafka_cpp_files
  ${CMAKE_CURRENT_SOURCE_DIR}/../librdkafka/src-cpp/*.cpp
  ${CMAKE_CURRENT_SOURCE_DIR}/../librdkafka/src-cpp/*.h
  )

list(REMOVE_ITEM librdkafka_files ${CMAKE_CURRENT_SOURCE_DIR}/../librdkafka/src/rdkafka_sasl.c)

add_definitions(-DLIBRDKAFKA_STATICLIB)
add_library(librdkafka STATIC ${librdkafka_files})
add_library(librdkafkacpp STATIC ${librdkafka_cpp_files})

Do you want this alternative build explained in the README.win32?

And BTW I also noticed that that openssl 1.1.0+ changed the lib names

edenhill:
There's ongoing work by @olibre to add proper cmake support to librdkafka, as an alternative to VS and mklove: https://github.com/olibre/cmake-for-rdkafka/issues/1

I dont think you need to describe the cmake process, just mention the define and why it is needed.

And librdkafka only supports OpenSSL 1.0 at this point.

skarlsson:
Sure, I'll add a comment. 

I did not test running anything against a ssl enabled kafka broker but everything compiles with 1.1.0c anv visual studio 2015 update 2 (and runs fine against a normal kafka)

edenhill:
Thans Svante!

PR #962, opened 2016-12-20
Fix undefined references to static const ints (961)

bbfgelman1:
This patch fixes an issue when compiling code with librdkafka
with all optimizations disabled. The library declares a number
of static constant integer values in headers, but not in
compiled files. This causes any code that passes these constants
directly to a function accepting a const reference to result in
undefined reference errors when linking. The error is fixed by
also declaring the variables in their corresponding .cpp files.

This issue affects GCC when compiling with optimizations disabled.

### DIFF
--- a/src-cpp/TopicImpl.cpp
+++ b/src-cpp/TopicImpl.cpp
@@ -33,6 +33,15 @@
 
 #include "rdkafkacpp_int.h"
 
+const int32_t RdKafka::Topic::PARTITION_UA = RD_KAFKA_PARTITION_UA;
+
+const int64_t RdKafka::Topic::OFFSET_BEGINNING = RD_KAFKA_OFFSET_BEGINNING;
+
+const int64_t RdKafka::Topic::OFFSET_END = RD_KAFKA_OFFSET_END;
+
+const int64_t RdKafka::Topic::OFFSET_STORED = RD_KAFKA_OFFSET_STORED;
+
+const int64_t RdKafka::Topic::OFFSET_INVALID = RD_KAFKA_OFFSET_INVALID;
 
 RdKafka::Topic::~Topic () {
 
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -1044,13 +1044,13 @@ class RD_EXPORT Topic {
    * The unassigned partition is used by the producer API for messages
    * that should be partitioned using the configured or default partitioner.
    */
-  static const int32_t PARTITION_UA = -1;
+  static const int32_t PARTITION_UA;
 
   /** @brief Special offsets */
-  static const int64_t OFFSET_BEGINNING = -2; /**< Consume from beginning */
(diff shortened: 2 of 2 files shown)


bbfgelman1:
This fixes the issue https://github.com/edenhill/librdkafka/issues/961

src-cpp/ProducerImpl.cpp
```diff
@@ -33,6 +33,16 @@
 
 #include "rdkafkacpp_int.h"
 
+const int RdKafka::Producer::RK_MSG_FREE;
+
+const int RdKafka::Producer::RK_MSG_COPY;
+
+const int RdKafka::Producer::RK_MSG_BLOCK;
+
+#ifndef MSG_COPY
```
edenhill: I think these always needs to be declared since the librdkafka and app compile defines may differ.
To be safe #undef MSG_COPY and MSG_FREE first here (but not in rdkafkacpp.h)

edenhill: Right.
Seeing how those two are legacy ones (to be backwards compatible with old appS), do you think we could get away with not declaring them here at all? It is not like those apps will suddenly break because the new constants are now fixed.

bbfgelman1:
I changed the constants in `RdKafka::Producer` from using `static const int` to using an anonymous `enum`. They can be referred to in the same way as before (`RdKafka::Producer::RK_MSG_FREE`), but the names now behave like literal values rather than static member variables. 

I did not apply the same change to the constants in `RdKafka::Topic` so that their explicit types are preserved.

edenhill:
Fails on VS 2013:
https://ci.appveyor.com/project/edenhill/librdkafka/build/0.9.3-pre-54-ucuvrpyw/job/o9k2domaj75thde1#L189

bbfgelman1:
C++17 will introduce `inline` variables to address exactly this problem. Until that is widely supported, please consider the solution implemented in this PR.

src-cpp/TopicImpl.cpp
```diff
@@ -33,15 +33,15 @@
 
 #include "rdkafkacpp_int.h"
 
-const int32_t RdKafka::Topic::PARTITION_UA;
+const int32_t RdKafka::Topic::PARTITION_UA = -1;
```
edenhill: With these moved from the cpp header file can we use the C defines from rdkafka.h for these?

src-cpp/ProducerImpl.cpp
```diff
@@ -33,7 +33,6 @@
 
 #include "rdkafkacpp_int.h"
 
-
```
edenhill: remove whitespace diff

bbfgelman1:
Would you like me to squash all the commits in this PR?

edenhill:
Yes please!

edenhill:
Thanks alot!

bbfgelman1:
Thanks for merging!

PR #974, opened 2016-12-29
fix: use out of scope variable

HeChuanXUPT:
rd_kafka_broker_send may use the address of msg2, out of  scope


HeChuanXUPT:
https://github.com/edenhill/librdkafka/pull/861
has mention it too

edenhill:
Thanks!

PR #979, opened 2017-01-04
validate topic name when add a new one

HeChuanXUPT:
1. topic name length is not greater then 255 , and in the Kafka 0.10 is 249
2. when add a new topic name should check validity


edenhill:
I dont want the topic name to be enforced in the client since an old client must work with a newer broker that might have less strict topic name requirements.

PR #1014, opened 2017-01-19
Add CMake

ruslo:


ruslo:
This is basic CMake configuration. Currently I have problems with test, even `./configure && make && cd tests && make` doesn't work for me. Builds fine on OSX and Linux, on Windows I got few problems. The first one is the way how to integrate [lz4](https://github.com/lz4/lz4) into project.

edenhill:
On Windows we make use of nuget for dependencies, so having msbuild automatically restore nuget packages should work.

CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+# This option doesn't affect build in fact, only C code
+# (see 'rd_kafka_version_str'). In CMake the build type feature usually used
+# (like Debug, Release, etc.).
+option(WITHOUT_OPTIMIZATION "Disable optimization" OFF)
+
+option(ENABLE_DEVEL "Enable development asserts, checks, etc" OFF)
+option(ENABLE_REFCNT_DEBUG "Enable refcnt debugging" OFF)
+option(ENABLE_SHAREDPTR_DEBUG "Enable sharedptr debugging" OFF)
+
+option(WITH_ZLIB "With ZLIB" ON)
```
edenhill: Shouldn't these be set after checking availability?

CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+option(WITH_SSL "With SSL" ON)
+
+# }
+
+option(RDKAFKA_BUILD_EXAMPLES "Build examples" ON)
+option(RDKAFKA_BUILD_TESTS "Build tests" ON)
+
+set(TRYCOMPILE_SRC_DIR "${CMAKE_CURRENT_LIST_DIR}/try_compile")
```
edenhill: Move this to "packaging/cmake/try_compile"

CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+# * TRYCOMPILE_SRC_DIR
+# Out:
+# * HAVE_ATOMICS_32
+# * HAVE_ATOMICS_32_SYNC
+# * HAVE_ATOMICS_64
+# * HAVE_ATOMICS_64_SYNC
+# * HAVE_REGEX
+# * HAVE_STRNDUP
+# * LINK_ATOMIC
+include("try_compile/cmake/rdkafka_setup.cmake")
```
edenhill: and you can switch the order here so that cmake files are in "packaging/cmake/...."

src/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+    rdlist.c
+    rdlog.c
+    rdports.c
+    rdrand.c
+    rdregex.c
+    rdstring.c
+    snappy.c
+    tinycthread.c
+)
+
```
edenhill: sasl seems to be missing.
Also do a master rebase, the sasl files changed names.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+set(
+    sources
+    0001-multiobj.c
```
edenhill: is there any way to wildcard this?

try_compile/atomic_32_test.cpp
```diff
@@ -0,0 +1,8 @@
+#include <inttypes.h>
```
edenhill: these should probable be .c, not .cpp?

CMakeLists.txt
ruslo: In my personal opinion - no. See for details:
* https://docs.hunter.sh/en/latest/creating-new/cmake.html#environment

ruslo:
> having msbuild automatically restore nuget packages should work

As far as I know CMake doesn't work well with nuget. Let me know if I'm missing something.

CMakeLists.txt
edenhill: The prefix might be unfortunate in librdkafka, but:
 * `WITH_foo` = foo was found and support for it will be compiled in (e.g., zlib compression)
 * `ENABLE_foo` = --enable-foo was specified to ./configure and we should try to set `WITH_foo` if the library is found. (this is mostly used in the reverse way, to --disable-ssl certain features).

This is probable the other way around in CMake, but we'll need to ignore that fact now since the rest of the code relies on the above assumptions.

Also: I think it is okay for some features to be left out if dependent libraries are not found, and that's how mklove and autoconf works today, so I dont think we should change that approach when cmake is used.

tests/CMakeLists.txt
ruslo: > is there any way to wildcard this?

Yes, but it's not recommended since may lead to problems:
```
We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.
```
* https://cmake.org/cmake/help/latest/command/file.html

edenhill: > "generated build system cannot know when to ask CMake to regenerate."

What does that mean in practice?
When is the generated build system asking CMake to generate?

ruslo: > What does that mean in practice?

Stable/predictable behavior, few scenarios described here:
* http://stackoverflow.com/q/1027247/2288008

src/CMakeLists.txt
ruslo: I've added `WITH_SASL` option

CMakeLists.txt
ruslo: Done

CMakeLists.txt
ruslo: Now it's `include("packaging/cmake/try_compile/rdkafka_setup.cmake")`

try_compile/atomic_32_test.cpp
ruslo: Done

CMakeLists.txt
ruslo: > --enable-foo was specified to ./configure and we should try to set WITH_foo if the library is found

If I do `./configure --enable-foo` and library is not found then `WITH_foo` will not be set?

edenhill: Correct (which is a bug in mklove, it should actually fail in that case)

tests/CMakeLists.txt
edenhill: I sort of agree, but in this case I think I actually prefer the glob approach for the tests.

edenhill:
Could you also add a doozer target for cmake so it is verified on CI?
It only needs to be run for one platform - xenial-amd64

tests/CMakeLists.txt
ruslo: Okay, added:
* https://github.com/edenhill/librdkafka/pull/1014/commits/bb7d7de1a59d96f33fa2c06daf8e8ee54465a2e6

However I have to warn you that it's really not about "I prefer" vs "I not prefer". This is about "plain wrong" vs "right".

CMakeLists.txt
ruslo: Does this one fits your requirements?

| User's WITH_foo | User's ENABLE_foo | Package found? | Effective       |
| --------------- | ----------------- | -------------- | --------------- |
| *not set*       | YES               | *no*           | WITH_foo = NO   |
| *not set*       | YES               | *yes*          | WITH_foo = YES  |
| YES             | YES               | *no*           | **FATAL_ERROR** |
| YES             | YES               | *yes*          | WITH_foo = YES  |
| *not set*       | NO                | *x*            | WITH_foo = NO   |

ruslo:
> Could you also add a doozer target for cmake so it is verified on CI?

Will take a look

ruslo:
> Will take a look

Target `cmake-xenial-amd64` added: https://doozer.io/edenhill/librdkafka/build/205

tests/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+#    We do not recommend using GLOB to collect a list of source files from your
+#    source tree. If no CMakeLists.txt file changes when a source is added or
+#    removed then the generated build system cannot know when to ask CMake to
+#    regenerate.
+#
+# Some extra information:
+# * http://stackoverflow.com/q/1027247/2288008
+# Forced here:
+# * https://github.com/edenhill/librdkafka/pull/1014#discussion_r97294390
+file(GLOB c_sources "0*-*.c")
```
edenhill: Okay, I think you are right, let's not glob these.
Sorry!

CMakeLists.txt
ruslo: Okay, sounds like this:
* https://github.com/edenhill/librdkafka/pull/1014/commits/c42e7f526c1eec53aaaba036cca847483c152147

tests/CMakeLists.txt
ruslo: Reverted: https://github.com/edenhill/librdkafka/pull/1014/commits/2bc318059b74ee605b3d889ed6896ca233190f3c

CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+# (like Debug, Release, etc.).
+option(WITHOUT_OPTIMIZATION "Disable optimization" OFF)
+
+option(ENABLE_DEVEL "Enable development asserts, checks, etc" OFF)
+option(ENABLE_REFCNT_DEBUG "Enable refcnt debugging" OFF)
+option(ENABLE_SHAREDPTR_DEBUG "Enable sharedptr debugging" OFF)
+
+option(WITH_ZLIB "With ZLIB" ON)
+option(WITH_SSL "With SSL" ON)
+option(WITH_SASL "With Sasl" OFF)
```
ruslo: Fixed: https://github.com/edenhill/librdkafka/pull/1014/commits/689415e8da1856296d827092efed176ef5341220

src/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+  list(APPEND sources rdgz.c)
+endif()
+
+if(NOT HAVE_REGEX)
+  list(APPEND sources regexp.c)
+endif()
+
+add_library(rdkafka ${sources})
+
+# Support '#include <rdkafka.h>'
```
ruslo: This is fix for this file: https://github.com/edenhill/librdkafka/blob/89ed67fabb3da66b6c0ecf7fd704c0e1381348fa/examples/rdkafka_performance.c#L43

src/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+    tinycthread.c
+)
+
+if(WITH_SASL)
+  list(
+      APPEND
+      sources
+      rdkafka_sasl.c
+      rdkafka_sasl.h
+      rdkafka_sasl_cyrus.c
```
ruslo: Fixed: https://github.com/edenhill/librdkafka/pull/1014/commits/2058da4f940188f94096f19e7104a8de8cfabf11

src/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+
+if(LINK_ATOMIC)
+  target_link_libraries(rdkafka PUBLIC "-latomic")
+endif()
+
+find_package(Threads REQUIRED)
+target_link_libraries(rdkafka PUBLIC Threads::Threads)
+
+if(WITH_SASL)
+  target_link_libraries(rdkafka PUBLIC "-lsasl2")
```
ruslo: See https://github.com/edenhill/librdkafka/pull/1014/commits/e8412ed44209a2d648758eede8df6424d9de12fe

Whissi:
Is librdkafa switching from mklove to CMake?

Please add an CMake option to disable optional usage of lz4, see #1025, if you haven't done this yet.

edenhill:
@Whissi No, CMake was contributed to provide an alternative, but mklove will remain as the primary build system.

edenhill:
Thanks for this!

PR #1025, opened 2017-01-27
configure: Add option to disable automagic dependency on liblz4

Whissi:
Previously, mklove activated lz4 support when lz4 was found. This added
a so called "automagic" dependency on liblz4 which is a problem from
distributions.

This commit will add an option which will allow you to explicit disable
lz4 usage.


edenhill:
Thanks!

PR #1053, opened 2017-02-13
Compatibility bugfix of a link issue due to change of RdKafka::Conf method declaration order. 

vin-d:
Method declaration issue initially introduced  by  c09efc160949cd3a1c0c8480ff22f6f7943c09e1

introduction of 

> virtual Conf::ConfResult set (const std::string &name, ConsumeCb *consume_cb,  std::string &errstr) = 0;

in the middle of existing methods of class definition, makes software using librdkafka that are compiled against previous version of librdkafka (pre c09efc160949cd3a1c0c8480ff22f6f7943c09e1), link wrong methods of the class when used with the library which has this modified interface. (...yes this sentence is a bit confusing -_-)


edenhill:
The diff shows up empty?

vin-d:
yes, i'm trying to findout what I did wrong.
i'm really not git-fluent.

PR #1054, opened 2017-02-13
Compatibility bugfix of a link issue due to change of RdKafka::Conf method declaration order.

vin-d:
Method declaration issue initially introduced by c09efc1

introduction of

> virtual Conf::ConfResult set (const std::string &name, ConsumeCb *consume_cb, std::string &errstr) = 0;

in the middle of existing methods of class definition, makes software using librdkafka that are compiled against previous version of librdkafka (pre c09efc1), link wrong methods of the class when used with the library which has this modified interface. (...yes this sentence is a bit confusing -_-)


vin-d:
ok @edenhill , should be better like that.

i'm not sure why it shows all my differents old commits from november in the list. of the PR, but at least d6fc13b is there.

vin-d:
I tested before/after, and it's really this part involved in d6fc13b that created this "wrong link" issue.

edenhill:
Thanks!

PR #1083, opened 2017-02-28
sasl: protection level and message size not sent

zyzil:
The protection level and message size was not being sent back to
the server during the authentication handshake.  This was causing
the first characters of the user principal to be misinterpreted
and could cause authentication errors due to an unknown protection
level if the first character of the username did not have bit 1 set


edenhill:
Thanks a lot for this, will review it next week.

edenhill:
Big thanks for this @zyzil

PR #1105, opened 2017-03-08
Adding Swift to Language Bindings

RockfordWei:
Perfect is the most popular Server Side Swift API library and HTTP
server framework and librdkafka was successfully imported to Swift on March, 2017.


edenhill:
Awesome! Thanks

PR #1116, opened 2017-03-14
Fix minor conf leak

mcandre:
Clean up after heap allocation of Kafka configuration struct pointer, so that developers have a more solid example code to work from.


edenhill:
I don't see the need to clutter the example code with cleanup on error, it will just exit(1) immediately anyway.

This change also has some problems with it, namely destroying things in the wrong order.

Thank you for your effort but I don't think this one will pay off, the example code is bloated as-is.

PR #1119, opened 2017-03-16
Feature/add error messages

llinsky:
This adds error messages for some ssl configuration errors which previously returned "No error"


edenhill:
Thanks for this!

PR #1121, opened 2017-03-16
Fixed high level consumer CPU usage issue

orthrus:
Fixed an issue that caused the high level consumer to use a lot of CPU due to a queue being continuously signaled while there are no new queue entries available.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -2929,7 +2929,7 @@ void rd_kafka_cgrp_serve (rd_kafka_cgrp_t *rkcg) {
         case RD_KAFKA_CGRP_STATE_UP:
 		/* Move any ops awaiting the coordinator to the ops queue
 		 * for reprocessing. */
-		rd_kafka_q_concat(rkcg->rkcg_ops, rkcg->rkcg_wait_coord_q);
+			rd_kafka_q_concat(rkcg->rkcg_ops, rkcg->rkcg_wait_coord_q);
 
                 /* Relaxed coordinator queries. */
                 if (rd_interval(&rkcg->rkcg_coord_query_intvl,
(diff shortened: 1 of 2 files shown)


edenhill:
Good find!

src/rdkafka_cgrp.c
```diff
@@ -2929,7 +2929,10 @@ void rd_kafka_cgrp_serve (rd_kafka_cgrp_t *rkcg) {
         case RD_KAFKA_CGRP_STATE_UP:
 		/* Move any ops awaiting the coordinator to the ops queue
 		 * for reprocessing. */
-		rd_kafka_q_concat(rkcg->rkcg_ops, rkcg->rkcg_wait_coord_q);
+			if (rd_kafka_q_len(rkcg->rkcg_wait_coord_q))
```
edenhill: I think this should be fixed in q_concat() instead, returning immediately if srcq is empty.

orthrus:
I've updated the fix. I couldn't use rd_kafka_q_len since it tries to lock the queue, which causes a simulated deadlock.

edenhill:
The srcq is not locked in q_concat() (it is assumed to already be locked, or not needed to be locked).
So I think you should just check srcq->rkq_len the first thing you do in q_concat, if it is zero you return immediately, before doing anything else.

orthrus:
Would that also work for forwarded queues?

edenhill:
You are absolutely right, it needs to be checked when srcq->rkq_fwdq is NULL

raskolnikoov:
Lets release this fix!

edenhill:
Thank's a lot of for identifying and fixing this issue @orthrus !

Since this PR would need another round or two of review (whitespace diffs, remove some srcq checks, etc) I took the liberty to push a commit with your fix, giving you proper attribution of course:
https://github.com/edenhill/librdkafka/commit/5eb6f1730fd2168c11223b623e2246e38aafb57f

PR #1134, opened 2017-03-22
Use SNI when connecting through SSL

vincentbernat:
Add TLS SNI (serve name) when connecting with SSL. If the broker is
behind a load balancer, the load balancer can use this information to
locate the right server.

I didn't find an easy way to just get the hostname. When I started, I thought I would just have to use `rd_kafka_broker_name()`, so the patch would have been quite minimal. Since some parsing is needed, maybe the patch is not worth it. As I have other non-SNI client (Scala client), I don't really need the patch yet. If you have doubts, feel free to drop it.

### DIFF
--- a/src/rdkafka_transport.c
+++ b/src/rdkafka_transport.c
@@ -490,6 +490,8 @@ static int rd_kafka_transport_ssl_connect (rd_kafka_broker_t *rkb,
 					   rd_kafka_transport_t *rktrans,
 					   char *errstr, int errstr_size) {
 	int r;
+	char name[RD_KAFKA_NODENAME_SIZE];
+	char *t;
 
 	rktrans->rktrans_ssl = SSL_new(rkb->rkb_rk->rk_conf.ssl.ctx);
 	if (!rktrans->rktrans_ssl)
@@ -497,6 +499,18 @@ static int rd_kafka_transport_ssl_connect (rd_kafka_broker_t *rkb,
 
 	if (!SSL_set_fd(rktrans->rktrans_ssl, rktrans->rktrans_s))
 		goto fail;
+
+#if (OPENSSL_VERSION_NUMBER >= 0x0090806fL) && !defined(OPENSSL_NO_TLSEXT)
+	/* If non-numerical hostname, send it for SNI */
+	rd_snprintf(name, sizeof(name), "%s", rkb->rkb_nodename);
+	if ((t = strrchr(name, ':')))
+		*t = '\0';
+	if (!(/*ipv6*/(strchr(name, ':') &&
+		       strspn(name, "0123456789abcdefABCDEF:.[]%") == strlen(name)) ||
+	      /*ipv4*/strspn(name, "0123456789.") == strlen(name)) &&
+	    !SSL_set_tlsext_host_name(rktrans->rktrans_ssl, name))
+		goto fail;
+#endif
 
 	r = SSL_connect(rktrans->rktrans_ssl);
 	if (r == 1) {


src/rdkafka_transport.c
```diff
@@ -498,6 +500,14 @@ static int rd_kafka_transport_ssl_connect (rd_kafka_broker_t *rkb,
 	if (!SSL_set_fd(rktrans->rktrans_ssl, rktrans->rktrans_s))
 		goto fail;
 
+	mtx_lock(&rkb->rkb_lock);
+	rd_snprintf(name, sizeof(name), "%s", rkb->rkb_nodename);
```
edenhill: Can't you use rkb_origname here?
(Also, you won't need the lock since rkb_nodename is only written to from the broker's thread.)

src/rdkafka_transport.c
```diff
@@ -498,6 +500,14 @@ static int rd_kafka_transport_ssl_connect (rd_kafka_broker_t *rkb,
...
+	mtx_unlock(&rkb->rkb_lock);
+	if ((t = strrchr(name, ':')))
+		*t = '\0';
+	if (!SSL_set_tlsext_host_name(rktrans->rktrans_ssl, name))
```
edenhill: Is this side-effect free? I don't want it to break existing users.

edenhill: Looking at [this](https://patches.libav.org/patch/17585/) we probably want this conditionalized.

src/rdkafka_transport.c
vincentbernat: `rkb_origname` is more difficult to parse? Couldn't it be `ssl://nodename:9094` for example? What difference would it be compared to `rkb_nodename`?

src/rdkafka_transport.c
vincentbernat: What is your lowest target for distributions? I can check if this is still needed on the latest CentOS 5.

edenhill: People are using really old stuff, so better play it safe and ifdef it out. (if libav needed that check then we might too)

src/rdkafka_transport.c
edenhill: Fair enough! But remove the locks

src/rdkafka_transport.c
edenhill: What about this:
> Is this side-effect free? I don't want it to break for existing users.

And I dont mean compilation-wise, but functionality-wise.

vincentbernat: In most cases, SNI would be not used and ignored by the receiving end. For compilation, the libav stuff is from 2012. In current OpenSSL version, the function itself is a define, so it would be more logical to use `#ifdef SSL_set_tlsext_hostname`. The problem with ifdefs is that they can break. If a newer version of OpenSSL changes the define by an actual function, this won't work any more and you will silently drop the ability to specify SNI. Idem for alternate implementations (LibreSSL, WolfSSL, BoringSSL). Hence my suggestion to check if a define is really needed.

src/rdkafka_transport.c
vincentbernat: Locks removed.

src/rdkafka_transport.c
edenhill: Fair enough, can we use `#if OPENSSL_VERSION_NUMBER  >= ..the-release-it-was-added-in` ?

vincentbernat: So, I have looked a bit what was the most common way to detect this feature.

There is:

 - `#if (OPENSSL_VERSION_NUMBER >= 0x0090806fL) && !defined(OPENSSL_NO_TLSEXT)`
 - `#if !defined(OPENSSL_NO_TLSEXT) && defined(SSL_set_tlsext_host_name)`
 - `#ifdef SSL_CTRL_SET_TLSEXT_HOSTNAME`

The first one seems more future proof for me, as if the implementation changes, we'll still get the feature implemented. The other cases rely on the current implementation. Alternate SSL implementation, like LibreSSL, also define `OPENSSL_VERSION_NUMBER`, sometimes with a bogus value large enough for our use.

So, I'll amend the commit with this change.

Another question: many implementations only set hostname for non-numeric value. Do you see a quick way to test we don't have just an IP address? If not, I don't think it's worth writing additional code for that.

edenhill: 👍  on #if check

Maybe it is safest not to enable this by default but provide a configuration option, this way we ensure not breaking functionality for existing users, and making the option explicit also allows us to document the hostname recommendation/requirement with the configuration option.
What do you think?

Otherwise, regarding host/ip, we'll need to dig up some RFCs to know exactly what constitutes a non-hostname (is digit . :  enough?)

vincentbernat: I don't think this is worth using an option. For example, the official Java client is enabling this automatically and this doesn't seem configurable. I suppose this is done automatically by some layer abstraction. Except for `s_client`, I never found a client where this option was configurable.

As for IP, [RFC](https://tools.ietf.org/html/rfc3546#section-3.1) says:

>  Literal IPv4 and IPv6 addresses are not permitted in "HostName".

I think your suggestion is enough. If IPv6, I suppose we'll get brackets (since it's extracted from host:port notation)?

edenhill: You had me at "the official Java client".

Re hostname check, something like this should be enough:
```
if (!(/*ipv6*/((strchr(hostname, ':') && strspn(hostname, "0123456789abcdefABCDEF:.[]%") == strlen(hostname)) ||
    /*ipv4*/strspn(hostname, "01234567890.") == strlen(hostname)))
  is a hostname;
```

vincentbernat: OK for IPv4. The `strspn()` for IPv6 seems superfluous since a hostname cannot have ":". But maybe you do this to match IPv4?

edenhill: Id rather test for what an address is supposed to look like than what doesnt look like a hostname, with all weird encodings of the latter it seems safer so.

vincentbernat: OK, I have updated the commit with your suggestion.

src/rdkafka_transport.c
```diff
@@ -498,6 +500,18 @@ static int rd_kafka_transport_ssl_connect (rd_kafka_broker_t *rkb,
 	if (!SSL_set_fd(rktrans->rktrans_ssl, rktrans->rktrans_s))
 		goto fail;
 
+#if (OPENSSL_VERSION_NUMBER >= 0x0090806fL) && !defined(OPENSSL_NO_TLSEXT)
+	/* If non-numerical hostname, send it for SNI */
+	rd_snprintf(name, sizeof(name), "%s", rkb->rkb_nodename);
+	if ((t = strrchr(name, ':')))
```
edenhill: This will truncate an IPv6 address without a port, which probably isnt right (not sure it matters much here though).

The hostname/port is already parsed in broker_add() so I suggest adding a new field to rd_kafka_broker_t:   `char *rkb_hostname;` which you set in broker_add() to the `name` parameter.
And then you can use rkb_hostname directly here.

edenhill: I can do these modifications if you want to, I understand that you might have more important things to do than factoring librdkafka :)

vincentbernat: My understanding is that `nodename` was already a canonicalized version of `origname` and we should always have a port. Isn't that right? If not, yes, an additional field would be helpful (but also, if we have an IPv6, we won't use it as hostname and an IPv6 cannot have a single colon, but I understand, this is not right to write buggy code just because it works).

edenhill: Since we're only interested in the hostname part here, and the port delimiter is also an ipv6 token, I think it makes more sense to use a specific rkb_hostname field.

edenhill: Fair enough, good point!

edenhill:
Are we good here?

vincentbernat:
Yep, OK for me!

edenhill:
Big thanks!

PR #1157, opened 2017-04-05
Builtin SASL PLAIN and SCRAM support

edenhill:


src/rdkafka.c
```diff
@@ -46,10 +46,7 @@
 #include "rdkafka_assignor.h"
 #include "rdkafka_request.h"
 #include "rdkafka_event.h"
-
-#if WITH_SASL
```
brefsdal: @edenhill Does this change conflict with use of `WITH_SASL=0`?  I'm seeing errors such as 

```
undefined symbol: rd_kafka_sasl_global_init
```

when building without sasl.

edenhill: Are you seeing that with a standard librdkafka build, or with the one embedded in node-rdkafka?

brefsdal: I'm using node-rdkafka, but it sounds like some confusion between `ENABLE_SASL` and `WITH_SASL`.

edenhill: Aye, I think Steve is on it

PR #1159, opened 2017-04-06
Add a Decribe Group Method

jbfuzier:
Expose a function that issue a DescribeGroupsRequest and generate a json string from it.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3047,6 +3047,8 @@ rd_kafka_describe_group_json (rd_kafka_t *rk, const char *group,
                 for (j = 0 ; j < gi->member_cnt ; j++) {
                         const struct rd_kafka_group_member_info *mi;
                         mi = &gi->members[j];
+                        if (j!=0)
+                            _st_printf(",");
                         _st_printf("\"%s\": { "
                                  "\"id\": \"%s\","
                                  "\"client_id\": \"%s\","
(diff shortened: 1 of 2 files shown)


edenhill:
Thanks for your contribution.

Some concerns I have:
 * I dont think librdkafka should provide JSON output for structured data, that's up to the application or binding to construct (correctly and safely) from the rich C structs.
 * You mention that ListGroups is not available for older brokers, which is true, but in that case neither is DescribeGroups, right? They were both added in the same commit.
 * There is probably not a real need for an additional API, we should instead fix list_groups() to be a bit smarter with the desired group id and skip the ListGroups step and go directly to DescribeGroup if a groupid is specified, this should cover the use-case that this PR seems to implement.

jbfuzier:
Thanks for tour reply,

* Regarding the JSON I thought it was ok to do this here because that is what is done for the stats ? For the python binding is it ok to build the JSON in the python c module and pass a JSON string to python ?

* When I asked my colleague to get listgroups and describegroup rights, he told me that there is no listgroups right available in the version we are using. I will double check with him.

* As both APIs where added at the same time there is no need for this, list_groups already has all the data needed.

Sorry for the trouble, I should have double checked before sending a PR

jbfuzier:
You were right, our cluster does indeed support ListGroup. 
The problem is in the java client that allows to change the ACL on the cluster, Describe option is missing for the cluster. If we set the cluster ACL to All it does work, so the issue is in the java client.

I have added a describe_groups method in the python client based on the example in librdkafka : https://github.com/jbfuzier/confluent-kafka-python/commit/68c9bde930c30b88043421856bf050b9e7f21c5a 

Does it seems ok to you ?

PR #1175, opened 2017-04-14
Small grammar nitpick

nyoung:
Just updated a couple of cases where the wrong word was used in the configuration description for topic.metadata.refresh.fast.interval.ms


edenhill:
Thanks!

PR #1178, opened 2017-04-17
Portability toward xcl on z

nthiebaud:
mostly cosmetic and portability changes to help toward a support for xlc compiler on z/OS

### DIFF
--- a/src/rdinterval.h
+++ b/src/rdinterval.h
@@ -1,5 +1,5 @@
-#pragma once
-
+#ifndef _RDINTRVAL_H
+#define _RDINTRVAL_H
 
 
 #include "rd.h"
@@ -114,3 +114,5 @@ static RD_INLINE RD_UNUSED void rd_interval_disable (rd_interval_t *ri) {
 static RD_INLINE RD_UNUSED int rd_interval_disabled (const rd_interval_t *ri) {
         return ri->ri_ts_last == 6000000000000000000LL;
 }
+
+#endif
--- a/src/rdposix.h
+++ b/src/rdposix.h
@@ -48,23 +48,37 @@
  * Annotations, attributes, optimizers
  */
 #ifndef likely
+#ifdef __370__
+#define likely(x) (x)
+#else
 #define likely(x)   __builtin_expect((x),1)
 #endif
+#endif
 #ifndef unlikely
+#ifdef __370__
+#define unlikely(x) (x)
+#else
 #define unlikely(x) __builtin_expect((x),0)
 #endif
+#endif
 
-#define RD_UNUSED   __attribute__((unused))
 #define RD_INLINE   inline
+#define RD_IS_CONSTANT(p)  __builtin_constant_p((p))
+#ifndef __370__
+#define RD_UNUSED   __attribute__((unused))
+#define RD_TLS      __thread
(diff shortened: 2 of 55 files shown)


src/rdposix.h
```diff
@@ -108,12 +126,22 @@
  */
 static RD_INLINE RD_UNUSED
 void rd_usleep (int usec, rd_atomic32_t *terminate) {
+#ifdef __370__
+    if(usec > 1000000)
+    {
```
edenhill: Since sleep() can be interrupted this needs to be wrapped in a similar loop as the one below.

Style guide promotes open-brace on same line.

src/rdinterval.h
```diff
@@ -1,5 +1,5 @@
-#pragma once
-
+#ifndef _RDINTRVAL_H
```
edenhill: _RDINTERVAL_H_

edenhill:
Did a quick scan through and it overall looked good.

Some early comments:
 * avoid white-space diffs (e.g., license cleanups, empty lines, tabs and spaces, etc). The PR should typically only cover one thing, and it will make it smaller and easier to handle. I will fix layout problems for the entire code base (the mix of spaces and tabs especially) in one go later when there is a style checker in place.
 * I believe "_<HEADER_NAME_H>_" is a more standard notation, e.g.: `#ifdef _RDKAFKA_BROKER_H_`, please also mark the trailing #endif with a `#endif /* _RDKAFKA_BROKER_H_ */` comment.
 * Add comments for weird defines, such as __370__, so that non-zos people knows what it means

src/rdposix.h
nthiebaud: ok fixing

wrt to open-brace.. yeah sorry.. finger memory is hard to fight sometimes :-)

src/rdinterval.h
nthiebaud: ok, fixing

nthiebaud:
sorry I had to change the base branch (to drop whitepace patches for now)  and I do not know on github how to change the source branch of the pull (or even if it is possible) so I'll close this one and open a new one :-(

edenhill:
>  (setq-default show-trailing-whitespace t) will make them stick-out like sore thumb

I do, along with the show-tabs-when-it-shouldvebeen-spaces-damnit, and I tell you that browsing the source hurts, but that's a good incentive to automate a style enforcer, that's my thinking :)

nthiebaud:
"show-tabs-when-it-shouldvebeen-spaces-damnit"
wait... I went out of my way to follow the 'linux' style here, which means using tab not space
what is your policy here:  I do not mind adjusting since it is foreign to me anyway, so foreign one way or the other : same difference to me :-)
PS: I normally use : no tabs and 4 indents)
 PS2: when I started messign with the code.. I did not have yet an emacs config to adjust the style automatically for this project.. I do now, but all I do is set the c-style to "linux"  is that correct ?

edenhill:
There is a .dir-locals.el in the top directory which sets the style to linux, but changes tabs to spaces, at least it should.

Also see https://github.com/edenhill/librdkafka/blob/master/CONTRIBUTING.md

nthiebaud:
"There is a .dir-locals.el"  ahhh sorry too old emacs on my side.. I'll tweak my config to do the tabs to spaces

PR #1183, opened 2017-04-18
Compiler compat for z/OS xlc

nthiebaud:


nthiebaud:
Note I have not addressed yet the 'explain __370__' part... bear with me.
Due to many header-order issue I have been having I'd like to introduce a
rdkafka_compat.h header that will be included _first_ in every c source and that
will deal with #define that need to be first and other platform specific define/system header quirk

src/rd.h
```diff
@@ -446,3 +447,5 @@ typedef struct rd_chariov_s {
         char  *ptr;
         size_t size;
 } rd_chariov_t;
+
+#endif /* NDef RD_H */
```
edenhill: This is a true nit, but I dont like "NDef", it made me wonder what it was for 2 seconds before I realized.
so to avoid that grave confusion can we go with just `#endif /* !_RD_H */` (including the prefixing underscore)?

src/rdinterval.h
```diff
@@ -1,5 +1,5 @@
-#pragma once
-
+#ifndef RDINTRVAL_H
```
edenhill: _RD_INTERVAL_H

src/rdkafka.h
```diff
@@ -81,10 +82,17 @@ typedef SSIZE_T ssize_t;
 #else
 #include <sys/socket.h> /* for sockaddr, .. */
 
+#ifndef __370__
```
edenhill: For clarity I'd like to see this reversed to: ` #ifdef __370__ /* add comment here too */` and then the zos stuff following, after that do the default #else case

src/rdkafka_sasl.h
```diff
@@ -26,8 +26,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#pragma once
-
+#ifndef RDKAFAK_SASL_H
```
edenhill: typo

src/rdkafka_sasl_int.h
```diff
@@ -26,7 +26,9 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#pragma once
+#ifndef RDKAFAK_SASL_INT_H
```
edenhill: typo

src/rdkafka_queue.h
```diff
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef RDKAFKA_QUEUE_H
+#define RDKAFAK_QUEUE_H
```
edenhill: typo

src/rdkafka_subscription.h
```diff
@@ -26,6 +26,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#pragma once
+#ifndef RDKAFAK_SUBSTRICPTION_H
```
edenhill: typo

src/rdposix.h
```diff
@@ -30,7 +29,8 @@
 /**
  * POSIX system support
  */
-#pragma once
+#ifndef RDPOSIX_H
```
edenhill: _ prefix

src/rdposix.h
```diff
@@ -48,23 +48,37 @@
  * Annotations, attributes, optimizers
  */
 #ifndef likely
+#ifdef __370__
```
edenhill: comments

src/rdposix.h
```diff
@@ -48,23 +48,37 @@
...
 #define unlikely(x) __builtin_expect((x),0)
 #endif
+#endif
 
-#define RD_UNUSED   __attribute__((unused))
 #define RD_INLINE   inline
-#define RD_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
-#define RD_NORETURN __attribute__((noreturn))
 #define RD_IS_CONSTANT(p)  __builtin_constant_p((p))
+#ifndef __370__
```
edenhill: reverse as instructed earlier

src/rdposix.h
```diff
@@ -48,23 +48,37 @@
...
+#define RD_UNUSED
+#define RD_WARN_UNUSED_RESULT
+#define RD_NORETURN
+#endif
 
 /**
 * Allocation
 */
-#if !defined(__FreeBSD__)
+#if !defined(__FreeBSD__) && !defined(__370__)
```
edenhill: comment

src/rdposix.h
```diff
@@ -108,12 +126,45 @@
...
+    while(usec > 0) {
+        unsigned long long stck_start;
+        unsigned long long stck_end;
+
+        __stckf(&stck_start);
+	/* if sleep return non 0 it was interupted
+	 * the return is teh number of second it had
+	 * _not_ slept
+	 */
+	if(left)
```
edenhill: see style guide: spaces, brace on same line.

src/rdposix.h
```diff
@@ -108,12 +126,45 @@
...
+	{
+	    left = sleep(left);
+	}
+	else
+	{
+	    usleep(usec);
+	}
+	if(terminate && rd_atomic32_get(terminate)) {
+	    break;
+	}
```
edenhill: no braces for single statements

src/rdposix.h
```diff
@@ -108,12 +126,45 @@
...
+        unsigned long long stck_start;
+        unsigned long long stck_end;
+
+        __stckf(&stck_start);
+	/* if sleep return non 0 it was interupted
+	 * the return is teh number of second it had
+	 * _not_ slept
+	 */
+	if(left)
+	{
```
edenhill: no braces for single statements..

src/rd.h
nthiebaud: Well I have a problem with confusing ! and NDef iow
#if FOO
#endif /* !FOO */
and
#ifndef FOO
#endif /* NDef FOO */

oh and fyi I tend to use

#if FOO
#else /* !FOO */
#endif /* !FOO */

when there are if /else construct.. iow the endif comment is related to the section it close not the section that opened the #if 

Using ! for def/ndef is ambiguous and confusing. I do not mind an alternative suggestion rather than NDef  but it must be different than the existing ! which already has an assigned meaning.

Now you are the maintainer of this codebase, if you insist on confusing convention, I'll comply, albeit under protest :-)

src/rdkafka.h
nthiebaud: Ahh I did not understood what you wanted the first time around.. ok will do

src/rdposix.h
nthiebaud: I got confused too I though you wanted #define <FILENAME> with s/./_/g and tr [a-z] [A-Z]
but now I'm getting that you want to keep the _ prefix I originally put _and_ add a _ for rd modules iow rd\[^.]*\)\.h -> _RD_\1_H (modulo uppercase) 
correct ?

src/rdposix.h
nthiebaud: yeah sorry, finger memory...

src/rdposix.h
nthiebaud: urgh.. _that_ is begging for bugs. I mean the place of the braces, ok that is a taste issue. but the no braces for single statement is asking of Mariner 1 like bugs. sigh.

src/rd.h
edenhill: Yeah, it is a matter of taste, and empirical evidence ;)
``` 
maglun@eden:/usr/include$ grep -R '^#endif.* _.*_H' . | wc -l
1425
maglun@eden:/usr/include$ grep -Ri '^#endif.*ndef' . | wc -l
134
```

src/rdposix.h
edenhill: `_<uppercase-headerfile_with_underscore_instead_of_dot>`

rd.h -> _RD_H
rdkafka_interval.h -> _RDKAFKA_INTERVAL_H

nthiebaud: _<uppercase-headerfile_with_underscore_instead_of_dot>
that rule would yield
_RDPOSIX_H
is that what you meant or do you want
_RD_POSIX_H ?

src/rd.h
nthiebaud: grep -R -i '^#endif.* .*_H' . | grep '!' | wc -l
425

nthiebaud: iow I'd be more comfortable with #endif /* <filename>_H */
presumably the format <filename>_H carry an implicite 'this is the end of the header guard'
and implicit notation is better than incorrect one :-)

src/rdposix.h
nthiebaud: see your comment above about rdinterval.h

edenhill: It should solely be based on filename: make uppercase, prefix with `_`, change dot to underscore.
Whatever else I said is wrong :)

src/rd.h
edenhill: skipping ! is fine

nthiebaud: ok, will do

src/rdposix.h
nthiebaud: ok will do

src/rdposix.h
```diff
@@ -108,12 +126,45 @@
...
+#ifdef __370__
+    int left = usec / 1000000;
+
+    while(usec > 0) {
+        unsigned long long stck_start;
+        unsigned long long stck_end;
+
+        __stckf(&stck_start);
+	/* if sleep return non 0 it was interupted
+	 * the return is teh number of second it had
```
edenhill: teh -> the

src/rdposix.h
```diff
@@ -108,12 +126,45 @@
...
+	    left = sleep(left);
+	}
+	else
+	{
+	    usleep(usec);
+	}
+	if(terminate && rd_atomic32_get(terminate)) {
+	    break;
+	}
+	__stckf(&stck_end);
```
edenhill: is this a monotonic clock? otherwise we need a check for time going backwards.

src/rdposix.h
```diff
@@ -108,12 +126,45 @@
...
+	stck_end -= stck_start;
+	/* a storeclock shifted right by 12 is a number of
+	 * microsecond since 1/1/1900
+	 */
+	stck_end >>= 12;
```
edenhill: why is this needed? the start and end clocks should have the same base (1900)

nthiebaud: I'm not sure I understand the question... the goal is to wait for a 'duration'
and that duration is expressed in micro-seconds... the comment above explain why we need to shift 12 to get that

PR #1191, opened 2017-04-25
Plugin and interceptor support

edenhill:
Ready for review


tests/plugin_test/plugin_test.c
```diff
@@ -0,0 +1,58 @@
...
+ */
+rd_kafka_resp_err_t conf_init (rd_kafka_conf_t *conf,
+                               void **plug_opaquep,
+                               char *errstr, size_t errstr_size) {
+        printf("plugin conf_init called!\n");
+        *plug_opaquep = my_opaque;
+        return RD_KAFKA_RESP_ERR_NO_ERROR;
+}
+
+void conf_destroy (const rd_kafka_conf_t *conf, void *plug_opaque) {
```
ewencp: Not sure if you feel the need to test it, but omitting `conf_destroy` is possibly worth testing as well.

src/rddl.c
```diff
@@ -0,0 +1,127 @@
...
+        return rd_strdup(buf);
+#endif
+}
+
+/**
+ * @brief Attempt to load library \p path.
+ * @returns the library handle (platform dependent, thus opaque) on success,
+ *          else NULL.
+ */
+void *rd_dl_open (const char *path, char *errstr, size_t errstr_size) {
```
edenhill: I always typedef things like this :)
But in this particular case it is supposed to be void * (see dlopen())

src/rdkafka.h
```diff
@@ -3179,6 +3202,267 @@ int rd_kafka_queue_poll_callback (rd_kafka_queue_t *rkqu, int timeout_ms);
...
+ *
+ * @brief A plugin interface that allows external runtime-loaded libraries
+ *        to integrate with a client instance without modifications to
+ *        the application code.
+ *
+ *        Plugins are loaded when referenced through the `plugin.library.paths`
+ *        configuration property and operates on the \c rd_kafka_conf_t
+ *        object prior \c rd_kafka_t instance creation.
+ *
+ * @warning Plugins require the application to link librdkafka dynamically
```
edenhill: Yeah, this might turn out to be a headache..

src/rdkafka.c
```diff
@@ -188,22 +189,29 @@ void rd_kafka_log_buf (const rd_kafka_t *rk, int level, const char *fac,
...
+                conf->log_cb(rk, level, fac, buf);
         }
 }
 
-void rd_kafka_log0 (const rd_kafka_t *rk, const char *extra, int level,
-		   const char *fac, const char *fmt, ...) {
+/**
+ * @brief Logger
+ *
+ * @remark conf must be set, but rk may be NULL
```
edenhill: I generally try to avoid checking/asserting for obvious and trivial stuff (they seldom pay off),b but reserve the right to assert to the more stateful checks..

src-cpp/rdkafkacpp.h
```diff
@@ -1049,6 +1049,23 @@ class RD_EXPORT Handle {
    * @returns ERR_NO_ERROR on success or an error code on error.
    */
   virtual ErrorCode set_log_queue (Queue *queue) = 0;
+
+  /**
+   * @brief Returns the ClusterId as reported in broker metadata.
+   *
+   * @param timeout_ms If there is no cached value from metadata retrieval
```
edenhill: Fixed some grammar, but this piece of documentation belongs here.
Can you clarify the confusion?

edenhill:
@ewencp clusterid() and latency() are needed by some interceptor libraries.

src/rdkafka_metadata.c
```diff
@@ -508,6 +508,20 @@ rd_kafka_parse_Metadata (rd_kafka_broker_t *rkb,
         rd_kafka_wrlock(rkb->rkb_rk);
         rkb->rkb_rk->rk_ts_metadata = rd_clock();
 
+        /* Update cached cluster id. */
+        if (RD_KAFKAP_STR_LEN(&cluster_id) > 0 &&
+            (!rkb->rkb_rk->rk_clusterid ||
+             rd_kafkap_str_cmp_str(&cluster_id, rkb->rkb_rk->rk_clusterid))) {
+                rd_rkb_dbg(rkb, BROKER|RD_KAFKA_DBG_GENERIC, "CLUSTERID",
```
ewencp: If you already had a non-null cluster ID, shouldn't this be an error? I see that it isn't in Java https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/Metadata.java#L247-L249 but I think that might be incorrect.

ewencp: Sounds like this was something punted on for the first version of cluster ID, so I guess matching the Java behavior will make sense. The idea is that if you really want to make sure you're working against a specific cluster, it's something you could specify via config and then validate here rather than looking for changes in the ID.

src/rdkafka.c
```diff
@@ -1324,6 +1344,8 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *conf,
                         rd_kafka_destroy_internal(rk);
                         rd_kafka_set_last_error(RD_KAFKA_RESP_ERR__INVALID_ARG,
                                                 EINVAL);
+                        if (!conf)
```
edenhill: Thanks!

src/rdkafka_metadata.c
edenhill: Aye, we can add a `cluster.id` configuration property later that is checked to match.

ewencp:
@edenhill This LGTM aside from the merge conflicts, though I still need to review the code that uses this functionality.

PR #1199, opened 2017-05-03
Partition count should be per topic in offset request buffer

### LINKED ISSUE #1190: offsetsForTimes failing for some TopicPartitions
Description
===========
`offsetsForTimes` fails to return offset repeatably for specific `TopicPartition`s.
`offsetsForTimes` returns `RdKafka::ERR_NO_ERROR`. Everything else looks fine, I can query the high watermark offset on all partitions for example.

Some debug output from my client (partition 1 fails):
```
Topic: multiparttest_events, partition: 0, time (milliseconds past epoch): 1493115342000, looked up offset as: 25196, current high watermark is: 31257
Topic: multiparttest_events, partition: 1, time (milliseconds past epoch): 1493115342000, looked up offset as: 1493115342000, current high watermark is: 12120
Topic: multiparttest_events, partition: 2, time (milliseconds past epoch): 1493115342000, looked up offset as: 6008, current high watermark is: 12018
```
Getting offset for partition 1 also fails when only calling `offsetsForTimes` on only that one `TopicPartition`.

I have reproduced this problem on more than one Kafka cluster. 

It looks like no offset reply is received for the problem TopicPartition; [this line](https://github.com/edenhill/librdkafka/blob/90efd246682d86e356b1789ce093d6ea1a214956/src/rdkafka_request.c#L191) is not hit for the failed partition.

Checklist
=========

 - [ ] librdkafka version (release number or git tag): 0.9.5
 - [ ] Apache Kafka version: 0.10.2.0
 - [ ] librdkafka client configuration:
```
  auto conf = std::unique_ptr<Conf>(Conf::create(Conf::CONF_GLOBAL));
  std::string errorMsg;
  conf->set("metadata.broker.list", brokerAddr, errorMsg);
  conf->set("session.timeout.ms", "10000", errorMsg);
  conf->set("group.id", "mantid", errorMsg);
  conf->set("message.max.bytes", "10000000", errorMsg);
  conf->set("fetch.message.max.bytes", "10000000", errorMsg);
  conf->set("replica.fetch.max.bytes", "10000000", errorMsg);
  conf->set("enable.auto.commit", "false", errorMsg);
  conf->set("enable.auto.offset.store", "false", errorMsg);
  conf->set("offset.store.method", "none", errorMsg);
  conf->set("api.version.request", "true", errorMsg);
```
 - [ ] Operating system: Ubuntu 16.04
 - [ ] Using the high-level KafkaConsumer
 - [ ] Critical issue

matthew-d-jones:
Fixes #1190 

`offsetsForTimes()` is failing when requesting offset for multiple partitions, which span multiple topics, on a single broker. This is because the number of partitions in the request was accumulated for all partitions, but should be per topic.

Note, there is no regression test in place yet, may want this before merging.

### DIFF
--- a/src/rdkafka_request.c
+++ b/src/rdkafka_request.c
@@ -289,6 +289,8 @@ void rd_kafka_OffsetRequest (rd_kafka_broker_t *rkb,
                         rd_kafka_buf_write_str(rkbuf, rktpar->topic, -1);
                         topic_cnt++;
                         last_topic = rktpar->topic;
+                        /* New topic so reset partition count */
+                        part_cnt = 0;
 
                         /* PartitionArrayCnt: updated later */
                         of_PartArrayCnt = rd_kafka_buf_write_i32(rkbuf, 0);


edenhill:
LGMT. thanks!

Issue #1156, opened 2017-04-04
Endian Problems in snappy on big endian systems


rthalley:
Description
===========

There are multiple problems with snappy on big endian systems.  The first problem is that this code in snappy-compat.h doesn't work

```#if __BYTE_ORDER == __LITTLE_ENDIAN
#define __LITTLE_ENDIAN__ 1
#endif
```

Neither __BYTE_ORDER nor __LITTLE_ENDIAN are defined, so it ends up comparing 0 == 0, which is true, and so it sets __LITTLE_ENDIAN__ when the system isn't little endian.  (This surprised me, but it's correct ISO C behavior, I checked!)  The patch below fixes that issue, and I've got all relevant patches from your "align branch" applied too, yet still there are problems writing though reading seems to work.  I do not currently know why writing fails; all I know is the broker complains with a " Broker: Invalid message" error.  Setting "compression.codec" to "none" allows writes to work, so I assume the problem is snappy related.  A colleague and I looked for other obvious endian problems in snappy but we didn't notice any.

I looked to see if there are any upstream fixes in snappy but didn't see any.  Upstream doesn't compile on SPARC at all, nor does it have a test suite, so I haven't tried further upstream debugging.

How to reproduce
================

Compile with the patch below on solaris sparc.   Send a message to a broker with snappy compression enabled; the broker will reject it.

Checklist
=========
Please provide the following information:

 - [ ] librdkafka version (release number or git tag): v0.9.4
 - [ ] Apache Kafka version: 0.10.2
 - [ ] librdkafka client configuration: compression.codec = snappy
 - [ ] Operating system: Solaris 11 SPARC
 - [ ] Using the legacy Consumer N/A
 - [ ] Using the high-level KafkaConsumer N/A
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka N/A
 - [ ] Provide broker log excerpts N/A
 - [ ] Critical issue yes

```
diff -u -r old/src/rdendian.h new/src/rdendian.h
--- old/src/rdendian.h  Fri Apr  8 08:01:44 2016
+++ new/src/rdendian.h  Tue Apr  4 07:15:38 2017
@@ -63,14 +63,30 @@
 #elif defined sun
   #include <sys/byteorder.h>
   #include <sys/isa_defs.h>
+#define __LITTLE_ENDIAN 1234
+#define __BIG_ENDIAN 4321
 #ifdef _BIG_ENDIAN
[...]

PR #1219, opened 2017-05-18
Fix endianism issues that were causing snappy to compress incorrectly on SPARC

rthalley:
See [Issue #1156] for more info.

The two key things are:

1) getting __LITTLE_ENDIAN, __BIG_ENDIAN, and __BYTE_ORDER defined
   properly

2) the (uint16_t) part of the

   #define htole16(x) ((uint16_t)BSWAP_16(x))

line.

The cast is essential as otherwise in this part of a
macro in snappy_compat.h

    typeof((v)) _v = (v); \
    memcpy((x), &_v, sizeof(*(x))); })

the value of 'v' has typeof int, not uint16_t, and it takes the 32-bit
extended value and copies the first two bytes, which being big endian,
are zero and not the correct swapped value.

### DIFF
--- a/src/rdendian.h
+++ b/src/rdendian.h
@@ -63,14 +63,30 @@
 #elif defined sun
   #include <sys/byteorder.h>
   #include <sys/isa_defs.h>
+#define __LITTLE_ENDIAN 1234
+#define __BIG_ENDIAN 4321
 #ifdef _BIG_ENDIAN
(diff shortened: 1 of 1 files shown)


edenhill:
LGTM.

Thank you! I'm trusting you to have verified this, because I can't :)

rthalley:
We definitely verified the patch :)

PR #1220, opened 2017-05-18
added serialization and deserialzation related error codes

mhowlett:
In [confluent-kafka-dotnet](https://github.com/confluentinc/confluent-kafka-dotnet) we would like to expose serialization errors via the `OnConsumeError` event, which will require corresponding error codes. It's probably best if these are specified in librdkafka as other clients may also want to use them and if so, it would be best if they were consistent across clients.

Rather than put them in the block associated with local error codes, we may want to create a new block for error codes that are only produced by higher level clients (and prefix reasons with Client: not Local:). The argument against that though is it would cause problems if in the future any codes in that block became relevant to librdkafka.


edenhill:
Thanks and welcome back to C!

PR #1221, opened 2017-05-18
Fix for issue#1211, random crash on aix

asharma339:
Issue was that we forgot to call rd_atomic32_init() for rd_kafka_op_cnt of rd_kafka_op.c. It mattered only for AIX as for Sun and Linux, rd_atomic32_add/sub operations do not go thru mtx_lock() and mtx_unlock().


src/rdkafka.c
```diff
@@ -98,6 +98,7 @@ static void rd_kafka_global_init (void) {
         atexit(rd_shared_ptrs_dump);
 #endif
 	mtx_init(&rd_kafka_global_lock, mtx_plain);
+	rd_atomic32_init(&rd_kafka_op_cnt, 0);
```
edenhill: Shouldn't this be in a `#if ENABLE_DEVEL` block?

asharma339: That's correct.

edenhill:
Thanks for this!

PR #1222, opened 2017-05-19
ISSUE-926: Add SSL CTX cb

panarchus:
Add in callback configuration for SSL CTX configuration. C and C++ APIs

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -426,6 +426,30 @@ class RD_EXPORT DeliveryReportCb {
 
 
 /**
+ * @brief Configures the SSL context using the given callback
+ *
+ * SSL contexts are usually created by providing the locations of the
+ * various files containing certificates and keys. With this set,
+ * the client has the ability to configure and set up the context
+ * to their needs (in addition to or instead of using file-based APIs)
+ *
+ * This is called once per rd_kafka_t instance initialization from rd_kafka_new()
+ * and can override any other SSL configuration if set (called after
+ * regular librdkafka SSL configuration).
+ *
+ */
+class RD_EXPORT SSLContextCb {
+ public:
+ /**
+  * @brief SSL context configuration callback
+  */
+  virtual RdKafka::ErrorCode ssl_ctx_cb(void *ssl_ctx) = 0;
+
+  virtual ~SSLContextCb() { }
+};
+
+
+/**
  * @brief Partitioner callback class
  *
  * Generic partitioner callback class for implementing custom partitioners.
@@ -809,6 +833,11 @@ class RD_EXPORT Conf {
   virtual Conf::ConfResult set (const std::string &name,
                                 const std::string &value,
                                 std::string &errstr) = 0;
+
+  /* @brief Use with \p name = \c \"ssl_ctx_cb\" */
+  virtual Conf::ConfResult set (const std::string &name,
+                               SSLContextCb *ssl_ctx_cb,
+                               std::string &errstr) = 0;
 
   /** @brief Use with \p name = \c \"dr_cb\" */
   virtual Conf::ConfResult set (const std::string &name,
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -1078,6 +1078,28 @@ void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
 
 
 /**
+ * @brief Configures the SSL context using the given callback
+ *
+ * SSL contexts are usually created by providing the locations of the
+ * various files containing certificates and keys. With this set,
+ * the client has the ability to configure and set up the context
+ * to their needs (in addition to or instead of using file-based APIs)
+ *
+ * This is called once per rd_kafka_t instance initialization from rd_kafka_new()
+ * and can override any other SSL configuration if set (called after
+ * regular librdkafka SSL configuration).
+ *
+ */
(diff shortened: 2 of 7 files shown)


src/rdkafka_conf.h
```diff
@@ -113,6 +113,11 @@ struct rd_kafka_conf_s {
 		char *ca_location;
 		char *crl_location;
 	} ssl;
+
```
edenhill: put this inside the ssl struct above, for grouping.

src/rdkafka_transport.c
```diff
@@ -762,6 +762,24 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+	 */
+	if (rk->rk_conf.ssl_ctx_cb) {
+		rd_kafka_dbg(rk, SECURITY, "SSL",
+		            "Using user-defined SSL_CTX callback");
+
+		r = rk->rk_conf.ssl_ctx_cb(rk,
+		                           (void *)ctx,
+		                           rk->rk_conf.opaque);
+
+		if (r < 0) {
```
edenhill: the return value type is rd_kafka_resp_err_t  (use properly typed variable) and should be checked for ` != RD_KAFKA_RESP_ERR_NO_ERROR`.
Also, in this case, print the error code in the errstr and indicate where the error came from:
```
rd_snprintf(errstr, errstr_size, "User-defined SSL_CTX callback failed: %s", rd_kafka_err2str(err));
```

src/rdkafka.h
```diff
@@ -1018,6 +1018,26 @@ void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
 
 
 /**
+ * @brief Creates/configures the SSL context using the given callback
+ *
+ * SSL contexts are usually created by providing the locations of the
+ * various files containing certificates and keys. With this set,
+ * the client has the ability to configure and set up the context
+ * to theirs needs.
```
edenhill: their

src/rdkafka.h
```diff
@@ -1018,6 +1018,26 @@ void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
 
 
 /**
+ * @brief Creates/configures the SSL context using the given callback
```
edenhill: "Creates": the context has already been created and setup by librdkafka, so this callback can only provide additional configuration (or perhaps replace previously set config?).

I'm also wondering if maybe this should be an interceptor instead, allowing multiple sources (plugins, app) to modify the CTX.

src/rdkafka.h
```diff
@@ -1018,6 +1018,26 @@ void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
...
+ *
+ * SSL contexts are usually created by providing the locations of the
+ * various files containing certificates and keys. With this set,
+ * the client has the ability to configure and set up the context
+ * to theirs needs.
+ *
+ * This is called once during initialization and can override any other
```
edenhill: "..called once per rd_kafka_t instance initialization from rd_kafka_new().."

src/rdkafka.h
```diff
@@ -1018,6 +1018,26 @@ void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
...
+ * ssl configuration if set (called after regular rdkafka ssl
```
edenhill: capital SSL
spell out librdkafka

src/rdkafka.h
```diff
@@ -1018,6 +1018,26 @@ void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
...
+ * configuration).
+ *
+ */
+RD_EXPORT
+void rd_kafka_conf_set_ssl_ctx_cb (rd_kafka_conf_t *conf,
+			      rd_kafka_resp_err_t (*ssl_ctx_cb) (rd_kafka_t *rk,
+			                                         void *ssl_ctx,
```
edenhill: The 'void *' type here is problematic, we would really like to have this typed but that would make the header file depend on openssl.h, which we dont want.

An alternative would be to ifdef-guard this function with some define from openssl.h, allowing an application to use it if they first include openssl.h.

src-cpp/rdkafkacpp.h
```diff
@@ -365,6 +365,21 @@ class RD_EXPORT DeliveryReportCb {
 
 
 /**
+ * @brief SSL Context Configuration callback class
+ *
+ */
+class RD_EXPORT SSLContextCb {
+ public:
+ /**
+  * @breif SSL context configuration callback
```
edenhill: "breif" typo.

This class needs to be fully documented (copying the blurb from rdkafka.h).

panarchus:
Thanks for review - I'll hopefully get to it this week (super busy atm though)

src/rdkafka.h
panarchus: If a user compiles librdkafka and dynamically links it later, would this ifdef-guard work? The define might not be in the CU.

edenhill: If librdkafka compiled SSL support then the conf_set_ssl_ctx_cb symbol will be emitted,
and the application will need to be compiled using the `LIBRDKAFKA_WITH_SSL_TYPES` define to make the declaration visible.
Should be fine.

panarchus:
Sorry it took so long to incorporate feedback, I've been entirely too busy and promptly forgot to update. I've made some changes now that I remembered: I hope this is alright!

PR #1245, opened 2017-05-31
Support for Message format v2

edenhill:
This PR contains a quite substantial refactoring of buffers, Produce & Fetch request and response handling, and addition of new MessageSet & Message reader and writer interfaces.

# Review instructions

Since this PR is a bit all over the place I'm trying to ease the reviewing burden by organizing the PR into smaller and isolated (where possible) reviewable chunks.

There are some recommended reviewers below.

Pick off this list and update/comment on what parts you are reviewing/reviewed so we know what's left.

## Small isolated changes (bite sized):
 * [x] minor modifications to how the CRC32C code is initialized and used: https://github.com/edenhill/librdkafka/pull/1245/commits/900669d8963b1b1f0f9531d2e22bcf1c57ec8eab  (but no changes to the algo parts of course)
 * [x] Kafka protocol and error code enum update: https://github.com/edenhill/librdkafka/pull/1245/commits/d676455e1a6dbf1eb050066b49a86d8b5bb77365
 * [ ] rd_kafka_op_new_fetch_msg() helper: https://github.com/edenhill/librdkafka/pull/1245/files#diff-ce1115b8fe0c6a6153c49786e23858db

## Medium sized:
 * [x] Varint implementation: varint.c: https://github.com/edenhill/librdkafka/pull/1245/files#diff-fa9f3afadb2f717c80b7e4e8c1aebb04  varint.h: https://github.com/edenhill/librdkafka/pull/1245/files#diff-7e91a1dee72599e69085285502eb6081
 * [x] ProduceRequest/Response refactoring: https://github.com/edenhill/librdkafka/pull/1245/commits/1cf8167ae8830a5a2c7562ba511dfe14d0d7192c (@hachikuji)
 * [ ] Fetch backoff improvements (per partition instead of only per broker): https://github.com/edenhill/librdkafka/pull/1245/commits/8321e37bb4715396824bb1b9a60180dfcd520c77 (@ewencp)
 * [ ] Transport/socket layer changes to use new buffers: https://github.com/edenhill/librdkafka/pull/1245/files#diff-226483c4c83ab9400939e815bcb19564 (@ewencp)
  * [ ] Changes to rdkafka_broker.c are mostly removed lines (due to refactoring to other files), but some changes were also made to use the new buffering API and the new MsgSet reader API. Ignore the chunks of removed code and focus on the actual changes and additions: https://github.com/edenhill/librdkafka/pull/1245/files#diff-214080cd229adac5b1596529b6eecfe4

## Large:
 * [x] New buffer implementation, fairly isolated. Focus on rdbuf.c (https://github.com/edenhill/librdkafka/pull/1245/files#diff-db5d6dd5081347e4b6d74434e1ab1971) and rdbuf.h (https://github.com/edenhill/librdkafka/pull/1245/files#diff-f7c9b1e98cb07b002b0da760787eaf91), and to a lesser degree the changes in rdkafka_buf.h (and its rippling affects across the code base): https://github.com/edenhill/librdkafka/pull/1245/commits/a07001f831fe9fcb9acedb59fbc84ed88fc3a838 @ewencp
 * [x] MessageSet reader interface (https://github.com/edenhill/librdkafka/pull/1245/files#diff-7e1c6d507dfcbd116ab122581cbe0c1c) @hachikuji
 * [x] MessageSet writer interface (https://github.com/edenhill/librdkafka/pull/1245/files#diff-458801fac9fb23317d7bc06b50d4c368) @hachikuji

## What can be safely (..) ignored:
 * the plain CRC32C implementation import in commit https://github.com/edenhill/librdkafka/pull/1245/commits/c3ed3556281f7530103ffcb9320326a3f1511e76  (Mark Adler's implementation)
 * rdkafka_performance tweak: https://github.com/edenhill/librdkafka/pull/1245/commits/15ec92ca360c0ef7c9d7913e8c7b49c76a8e327c
 * FreeBSD porting fix: https://github.com/edenhill/librdkafka/pull/1245/commits/ba0ff8a5920f0a80deb9bccec6280572328e89fa
 * the unittest framework: https://github.com/edenhill/librdkafka/pull/1245/commits/50c9a6624115dc6d45883f96f648dcc0b8f63b4b
 * minor cleanups: https://github.com/edenhill/librdkafka/pull/1245/commits/46df23d383cc3dda41420fb6d4d5c926a31c5a6f
 * a comment: https://github.com/edenhill/librdkafka/pull/1245/commits/8c381f61f466dd89db2da5a6f0ae94f2e4dea6d9
 * Makefile updates: https://github.com/edenhill/librdkafka/pull/1245/commits/de934dfb6862f780d69877b13dd98a9e3d1f6c0e
 * Snappy/LZ4 refactoring (mostly moving code around): https://github.com/edenhill/librdkafka/pull/1245/commits/1f788acf670a5e78521162066ff720a9aac806d9


mhowlett:
checked 'Added updated Kafka protocol and error enums' commit LGTM (checked details pretty carefully).

src/rdvarint.h
```diff
@@ -0,0 +1,169 @@
...
+ */
+
+/**
+ * @brief unsigned-varint encodes unsigned integer \p num into buffer
+ *        at \p dst of size \p dstsize.
+ * @returns the number of bytes written to \p dst, or 0 if not enough space.
+ */
+
+static RD_INLINE RD_UNUSED
+size_t rd_uvarint_enc_u64 (char *dst, size_t dstsize, uint64_t num) {
```
mhowlett: where did your varint implementation come from? it's a bit different to both grpc and core Kafka:

https://github.com/google/protobuf/blob/6b27c1f981a9a93918e4039f236ead27165a8e91/src/google/protobuf/io/coded_stream.h#L1191

https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/utils/ByteUtils.java#L233

What you have looks a bit more elegant but I would have followed the Kafka implementation as closely as possible because then checking 1:1 against that would have been a valuable (and relatively easy) thing to do.

given the above, it would be good to see a few more tests (where expected outputs have been generated using the java code). If you feel that is justified, let me know and I'll be happy to contribute.

Alternatively (but not as good), do we have integration test cases that implicitly test a wide range of values?

src/rdkafka_lz4.c
```diff
@@ -0,0 +1,430 @@
...
+
+        return RD_KAFKA_RESP_ERR_NO_ERROR;
+}
+
+
+
+/**
+ * @brief Decompress LZ4F (framed) data.
+ *        Kafka broker versions <0.10.0.0 breaks LZ4 framing checksum, if
+ *        \p proper_hc we assume the checksum is okay (broker version >=0.10.0)
```
ijuma: Nit: it should probably talk about message format versions instead of broker versions, no?

src-cpp/rdkafkacpp.h
```diff
@@ -305,7 +305,58 @@ enum ErrorCode {
...
+         *  concurrent operation on the same transaction was ongoing */
+        ERR_CONCURRENT_TRANSACTIONS = 51,
+        /** Indicates that the transaction coordinator sending a
+         *  WriteTxnMarker is no longer the current coordinator for a
+         *  given producer */
+        ERR_TRANSACTION_COORDINATOR_FENCED = 52,
+        /** Transactional Id authorization failed */
+        ERR_TRANSACTIONAL_ID_AUTHORIZATION_FAILED = 53,
+        /** Security features are disabled */
+        ERR_SECURITY_DISABLED = 54
```
hachikuji: We also added OPERATION_NOT_ATTEMPTED(55) recently.

src/rdkafka_msgset_reader.c
```diff
@@ -0,0 +1,1076 @@
...
+        rd_kafka_op_t *rko;
+        rd_kafka_msg_t *rkm;
+        /* Only log decoding errors if protocol debugging enabled. */
+        int log_decode_errors = (rkbuf->rkbuf_rkb->rkb_rk->rk_conf.debug &
+                                 RD_KAFKA_DBG_PROTOCOL);
+        size_t message_end;
+
+        rd_kafka_buf_read_varint(rkbuf, &hdr.Length);
+        message_end = rd_slice_offset(&rkbuf->rkbuf_reader) + hdr.Length;
+        rd_kafka_buf_read_varint(rkbuf, &hdr.MsgAttributes);
```
hachikuji: I will update the KIP to reflect this (if I have not already), but we moved the control flag to one the batch attributes. When handling messages on the replica, we found that we needed a way to know whether the batch contained transactional markers. With a message-level attribute, this would require decompressing the batch, so instead we moved the attribute to the batch so that _all_ messages contained in the batch are considered control messages.

src/rdkafka_msgset_writer.c
```diff
@@ -0,0 +1,1160 @@
...
+                            void (*)(void *)) = {
+                [0] = rd_kafka_msgset_writer_write_msg_v0_1,
+                [1] = rd_kafka_msgset_writer_write_msg_v0_1,
+                [2] = rd_kafka_msgset_writer_write_msg_v2
+        };
+        size_t actual_written;
+        size_t pre_pos;
+
+        if (likely(rkm->rkm_timestamp))
+                MsgAttributes |= RD_KAFKA_MSG_ATTR_CREATE_TIME;
```
hachikuji: Just to be clear, v2 does not have a timestamp type attribute at the message level; it is only supported in the message set attributes.

src/rdkafka_msgset_reader.c
```diff
@@ -0,0 +1,1076 @@
...
+
+
+
+struct msgset_v2_hdr {
+        int64_t BaseOffset;
+        int32_t Length;
+        int32_t PartitionLeaderEpoch;
+        int8_t  MagicByte;
+        int32_t Crc;
+        int16_t Attributes;
```
hachikuji: Might be helpful to document the attributes in use somewhere. This is from the java version:
```
 *  -------------------------------------------------------------------------------------------------
 *  | Unused (6-15) | Control (5) | Transactional (4) | Timestamp Type (3) | Compression Type (0-2) |
 *  -------------------------------------------------------------------------------------------------
```

src/rdvarint.h
edenhill: I checked the internets for good implementations and they were scarce, finally ended up with one based on Andoma's here:
https://github.com/andoma/libsvc/blob/master/ntv_binary.c#L48

I don't share your sentiment for copying the Kafka implementation, which may have flaws itself, for low-level bit-mangling stuff like this a native approach is warranted.

You are absolutely right that there should be more unittest-cases though.
There is no point in specific integration tests for varints since they are used a handful of times for each message produced and consumed, thus heavily excercised during normal testing.

src/rdkafka_msgset_reader.c
edenhill: @hachikuji 
The KIP does not seem to be updated in this regard, it still mentions the control flag as a Message-level attribute.
Can you update the KIP and I'll modify my code?

src/rdkafka_msgset_writer.c
edenhill: Ah, right, fixed (so that the v2 msg writer ignores the MsgAttributes)

src/rdkafka_msgset_reader.c
ijuma: @apurvam updated the KIP yesterday, seems like he missed this.

hachikuji: @edenhill I updated the KIP this morning. Take a look and let me know if it's clear.

apurvam: yes. I missed this. I included the list of changes from KAFKA-5317, but this wasn't on it.

PR #1255, opened 2017-06-06
Add cppkafka to language bindings

mfontanini:
I finally released cppkafka, my C++11 rdkafka wrapper, so I think it's a good idea to add it here.


edenhill:
Thanks!

PR #1264, opened 2017-06-15
Documented hard exit flag for rdkafka_performance

theojepsen:
`rdkafka_performance` had the undocumented flag `-x` which causes the tool to hard exit after sending <cnt> messages. I documented this flag in the tool's help message.


edenhill:
Thanks!

PR #1289, opened 2017-06-28
Call on_acknowledgement() interceptor from internal thread

edenhill:
To avoid the case where a buggy application does not call poll()
and thus not serving delivery reports, the on_ack..() interceptor
is now called immediately from the broker thread when the
final delivery result is known.


mfontanini:
Does this mean that an application that previously assumed that the delivery deport callback would be executed in the context of the thread calling `poll()` would now have a potential race condition as this is called from another one?

edenhill:
@mfontanini The delivery report was not changed, it is still triggered from poll(), what we did was move the on_acknowledgement() interceptor call to the internal librdkafka broker threads to make sure it gets called regardless of what the application is doing (like, not calling poll() for some time, or at all).

mfontanini:
Ah okay, I was worried for a second. Never mind then!

PR #1337, opened 2017-07-22
In-memory SSL configuration

mikeydee36:
Allow SSL user-certificate, CA-certificate and private key to be passed in-memory into the library, and configured against the OpenSSL context.

The newly-supported configuration settings are (following the same naming convention as their file-based counterparts):

- "ssl.key_inmemory" - In-memory client private key (ASN.1-encoded DSA or RSA bytes: specify with ssl.key_inmemory_type), used for authentication.
- "ssl.key_inmemory_type" - In-memory client private key type: DSA or RSA.
- "ssl.certificate.location_inmemory" - In-memory X509 certificate bytes used for authentication.
- "ssl.ca.location_inmemory" - In-memory X509 CA certificate bytes for verifying the broker's key.


mikeydee36:
Hi, getting an error on the cmake check - looks like the source-file "0067-empty_topic.cpp" needs to be added to tests/CMakeLists.txt.

I'm happy to make any suggested changes :)

PR #1387, opened 2017-08-21
Fix 1386, messed up Null/Empty handling in the Consumer

edenhill:
This also fixes an off-by-one error for the Producer's returned offsets when produce.offset.report=false (default)


mhowlett:
I don't see any problems - LGTM conditional on you having verified the dotnet integration tests pass.

webmakersteve:
Do you know when this will make it into a release?

edenhill:
@webmakersteve We don't have a set release date yet, but within the next 30 days.

patrixgdd:
Hi, I have an issue on the consumer which is running scala kstream with 0.11.0.0 client, but the producer is node-rdkafka with 0.9.5.0 that the consumer has shutdown once it tried to connect the kafka 0.11 server! I should not show the code here because it is not related to the topic.
Looking forward to this 0.11.0 release to see if it is the nodejs issue.
Thanks a lot.

PR #1399, opened 2017-08-29
Delete topics

edenhill:
I needed this to speed up testing runs (using the kafka-topics.sh script to delete multiple topics is too slow).

@mhowlett While reviewing this, think of what it would look like in the .NET client.

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3330,6 +3330,56 @@ int rd_kafka_queue_poll_callback (rd_kafka_queue_t *rkqu, int timeout_ms);
 
 
 /**
+ * @name Admin interface
+ *
+ */
+
+/**
+ * @brief Delete topic(s) in cluster.
+ *
+ * @param topics List of topic (names) to delete. Per-topic errors are
+ *        returned in the \c err field of each topic.
+ * @param timeout_ms Time for both controller lookup and actual topic deletion
+ *                   on the broker. If < 0 the absolute value is used
+ *                   for controller lookup timeout only and no timeout is
+ *                   used on the broker - the topic will only be marked for
+ *                   deletion. If > 0 the remaining time after controller
+ *                   lookup will be used as DeleteTopics request timeout
+ *                   on the broker, depicting how long the broker will
+ *                   wait for all replicas to have deleted the topic
+ *                   before responding to the client.
+ *                   Despite timeout setting, the topic will be marked
+ *                   for deletion.
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success,
+ *          RD_KAFKA_RESP_ERR__TRANSPORT if no controller was found in
+ *          \p timeout_ms, or any other error code if the request failed.
+ *          If all topics errored out with the same error code that error code
+ *          will be returned, otherwise RD_KAFKA_RESP_ERR_NO_ERROR is
+ *          returned and the application will need to check for individual
+ *          topic error codes.
+ *
+ * @warning Deleting a topic will delete all messages from all replicas.
+ *
+ * @remark It may take several seconds after the call returns for the deletes to
+ *         propagate to all brokers in the close.
+ *
+ * @remark Requires broker version 0.10.1.0 or higher and
+ *         delete.topic.enable broker configuration set to true.
+ *
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_admin_delete_topics (rd_kafka_t *rk,
+                              rd_kafka_topic_partition_list_t *topics,
+                              int timeout_ms);
+
+
+/**@}*/
+
+
+
+
+/**
  * @name Plugin interface
  *
  * @brief A plugin interface that allows external runtime-loaded libraries
@@ -3813,7 +3863,6 @@ rd_kafka_interceptor_add_on_commit (
 
 
 /**@}*/
-
 
 #ifdef __cplusplus
 }
--- a/src/rdkafka_request.c
+++ b/src/rdkafka_request.c
@@ -1845,3 +1845,114 @@ int rd_kafka_ProduceRequest (rd_kafka_broker_t *rkb, rd_kafka_toppar_t *rktp) {
 
         return cnt;
 }
+
+
+
+rd_kafka_resp_err_t
+rd_kafka_handle_DeleteTopics (rd_kafka_t *rk,
+                              rd_kafka_broker_t *rkb,
+                              rd_kafka_resp_err_t err,
+                              rd_kafka_buf_t *rkbuf,
+                              rd_kafka_buf_t *request,
+                              rd_kafka_topic_partition_list_t *topics) {
+        const int log_decode_errors = 1;
+        int actions;
+        int32_t TopicArrayCnt;
+        int i = 0;
+        rd_kafka_resp_err_t same_err = RD_KAFKA_RESP_ERR_NO_ERROR;
+        int same_err_cnt = 0;
+
+        if (err)
+                goto err;
+
+        rd_kafka_buf_read_i32(rkbuf, &TopicArrayCnt);
+        if (TopicArrayCnt != topics->cnt)
+                rd_kafka_buf_parse_fail(rkbuf,
+                                        "TopicArrayCnt %"PRId32" should be "
+                                        "same as %d requested topic(s)",
+                                        TopicArrayCnt, topics->cnt);
+
+        for (i = 0 ; i < (int)TopicArrayCnt ; i++) {
+                rd_kafkap_str_t topic;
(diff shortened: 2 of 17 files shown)


src/rdkafka.h
```diff
@@ -3330,6 +3330,56 @@ int rd_kafka_queue_poll_callback (rd_kafka_queue_t *rkqu, int timeout_ms);
...
+ *                   on the broker. If < 0 the absolute value is used
+ *                   for controller lookup timeout only and no timeout is
+ *                   used on the broker - the topic will only be marked for
+ *                   deletion. If > 0 the remaining time after controller
+ *                   lookup will be used as DeleteTopics request timeout
+ *                   on the broker, depicting how long the broker will
+ *                   wait for all replicas to have deleted the topic
+ *                   before responding to the client.
+ *                   Despite timeout setting, the topic will be marked
+ *                   for deletion.
```
treziac: Specify that with timeout < 0, we receive an RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT for partition where topics where effectively marked as deletion, where we receive a different error if deletion is not possible (unknown topic, etc).

src/rdkafka.h
```diff
@@ -3330,6 +3330,56 @@ int rd_kafka_queue_poll_callback (rd_kafka_queue_t *rkqu, int timeout_ms);
...
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success,
+ *          RD_KAFKA_RESP_ERR__TRANSPORT if no controller was found in
```
treziac: RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT  doesn't describe the request failed, topic, in most case (if not all), should effectively have be marked as deletion and deleted when possible

mhowlett:
I like that there is no new handle type (i.e. RD_KAFKA_ADMIN).

In the bindings, I think we should have another Class `AdminClient` that exposes this method. This could be constructed using a bootstrap server address, or a `Handle` exposed as a property on a Producer or Consumer (which is opaque from a users perspective). I also think we should move the generic methods like `GetWatermarkOffsets` currently on `Producer` / `Consumer` to this new `AdminClient`.

src/rdkafka.h
mhowlett: I'm not into the idea of using negative numbers here - I'd have an additional parameter (probably a boolean, possibly an enum) to indicate the meaning of the timeout parameter.

treziac: A "getCallback" boolean may be great indeed (with or without a C callback). And perhaps delete the `RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT` in case it is false then (as it is the response when there is no error)?

edenhill: @mhowlett The thing is that it is actually two timeouts baked into one:
 * controller lookup timeout, i.e., how long to wait for cluster metadata. This is mainly a problem for when this API is called right after client instantiation.
 * DeleteTopics protocol request which has its own **broker side** timeout (included in the request), telling the broker how long to wait for full delete propagation in the cluster before responding to the client.

From a user perspective though, there is typically only one timeout of concern: how long can my application wait for DeleteTopics to return, it is not really relevant if part of that time is spent looking up the controller.
And for power-users that understand that topic deletion is a best-effort operation time-wise, they might want to make sure the topic is marked for deletion but understand that there is no practical difference in waiting 100ms or 100seconds for the topic deletion to propagate, so they decide that the DeleteTopics protocol request should return immedatiately, only marking the topics for deletion, but they still want to allow X ms for looking up the controller - this is where the negative form comes in.
Yes, we could split that up into two parameters: lookup_timeout_ms and delete_timeout_ms, but I feel that is API bloat for something that will rarely be used.

edenhill: @treziac the positive timeout (or what remains of it after controller lookup) controls how long the request is blocked on the broker, if the topic has not been deleted by that time there will be no later propagation of the fact that the topics were deleted (or not), so a callback will not be much good.
Yes, we could use metadata, but that would require us to keep state between delete_topics() and future metadata calls, and there is still the risk that the topic is re-created between the delete and the next metadata query, in which case it will seem as if the topic was never deleted, but it was.

treziac: I understand the use of negative timestamp, just that negative timestamp looks weird compared to only positive timestamp and bool saying if we wait for broker response during remaining timeout (and set request timeout to 0 or remaining accordingly as you do)
```
rd_kafka_admin_delete_topics (rd_kafka_t *rk,
                               rd_kafka_topic_partition_list_t *topics,
                               int timeout_ms, //always positive, negative will fail if controller lookup fail
                               bool use_remaining_timeout_for_delete_callback );
```
 but I may not be used on C way of dealing with time :) . 

I assume this mean we would only call it with negative value to return as soon as possible (with max timeout we want to give to controller lookup), or a kind of high (say min) if we want a "success" response assured from broker (and if no success, we will have to rely on metadata or check other  metrics like monitoring number of topics on broker or znode, but this is out of kafka client scope).

Here setting a non negative rather small timeout (eg  200ms) just mean "I marked topic for deletion in 20 ms, but wait some more time for response which will surely be a broker timeout")

It will be easy to map this on .NET anyway, so no issue here if it fits better like this in C.

mhowlett: @edenhill - I think clarity/straightforwardness is more important to prioritize on than terseness of API. Maybe here you could have one overload with timeout_ms corresponding to the current implementation with positive timeout_ms, and another with both controller_timeout_ms and lookup_timeout_ms (as you suggest). I find this straightforward. I find the current setup requires mental energy to overcome the negative number weirdness.

edenhill: From a maintenance perspective it is imperative to keep the API surface at a bare minimum since these APIs must age well and be supported for the foreseeable future.
Adding multiple variations of the same API is not best practice in the C world, but rather a reluctant afterthought to fix an initial API design failure.
C users are typically accustomed to funky APIs that do require some documentation reading.

I do agree that this particular API is indeed somewhat funky with the negative timeout, encoding rich information in the signedness of a type, and from that perspective it might be better to have two timeouts: lookup_timeout_ms and request_timeout_ms, but does it end there? Should there be a network_timeout_ms and queue_timeout_ms as well? There are so many moving parts under the hood and we really just want to expose an API that is easy to use, but still somewhat flexible for hard-core users. And having developers understand and decide on proper timeouts has proven to be quite problematic, so the fewer things they need to be concerned with, the better.

In this case that means that for 99% of the users they just need to set a maximum timeout allowed for this call to do its thing, and if the return value is XYZ it means that topics were at least marked for deletion.
For the 1% power users, they can get the completely asynchronous request behaviour (instant timeout) without having to busy-loop until the controller is available (which is something we want to abstract from the user).

There are a couple of use-cases:
 1. lookup_timeout=X, request_timeout=0
 2. lookup_timeout=X, request_timeout=remaining(X)
 3. lookup_timeout=X, request_timeout=Y

The current scheme supports 1 and 2. If we believe there is a specific need for case 3 we'll need to add the second timeout as well, at the expense of an API that is harder to use for most users.

mhowlett: I think the two method approach it's easier to comprehend and therefore easier to use and support, but this is just my 2c and API is ultimately up to you. In C# I will want to expose a less confusing API though, and method overloading is very natural there.

src/rdkafka_request.c
```diff
@@ -1845,3 +1845,114 @@ int rd_kafka_ProduceRequest (rd_kafka_broker_t *rkb, rd_kafka_toppar_t *rktp) {
...
+                        same_err = ErrorCode;
+
+                if (same_err == ErrorCode)
+                        same_err_cnt++;
+        }
+
+        /* If all topics have the same error (or success), propagate that
+         * error as the request-level error (or success). */
+        if (same_err_cnt == TopicArrayCnt)
+                err = same_err;
```
treziac: On this again, returning a broker_timeout error when timeout is negative seems weird (but this is more the DeleteTopicsResponse from protocol which is weird)

Also, resending a common error seems inconsitent with other method as offset handle. I would prefer error is only for general error (could not send to broker...) 
If I don't have authorization to delete topics for example, I would receive the error when all topics are valid, but a success if one inexistant topic was in the list (as all error wouldn't be the same)

src/rdkafka.h
edenhill: We should probably go the Java way of providing an extensible Options type to Admin APIs.
E.g.;

```
 rd_kafka_admin_option_t *opt = rd_kafka_admin_option_new();
 rd_kafka_admin_option_set_timeout(opt, 1234); // block no longer than 1.234 seconds on the broker

 err = rd_kafka_admin_delete_topics(rk, myListOfTopics, opt, my_total_timeout_ms);
...
```

Issue #1277, opened 2017-06-22
Add parameter for setting apiVersion timeout request


vinipuh5:
Description
===========
In rd_kafka_ApiVersionRequest the minimized timeout is hardcoded to 10 seconds. For my application, I forced to change it to 5 seconds because of some needs, but this means that I'll be forced to change it in each new version.
I suggest as an enhancement to add a parameter to customize this property.

PR #1402, opened 2017-08-30
Issue #1277: Add parameter for setting apiVersion timeout request

vinipuh5:
Allow to config broker version request timeout by setting "api.version.request.timeout.ms"


src/rdkafka_conf.c
```diff
@@ -392,6 +392,10 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  "**NOTE**: Depends on broker version >=0.10.0. If the request is not "
 	  "supported by (an older) broker the `broker.version.fallback` fallback is used.",
 	  0, 1, 1 },
+	{ _RK_GLOBAL, "api.version.request.timeout.ms", _RK_C_INT,
+	  _RK( api_version_request_timeout_ms ),
+	  "Timeout for broker version request.",
```
edenhill: "Timeout for broker API version requests."

src/rdkafka_conf.c
```diff
@@ -392,6 +392,10 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  "**NOTE**: Depends on broker version >=0.10.0. If the request is not "
 	  "supported by (an older) broker the `broker.version.fallback` fallback is used.",
 	  0, 1, 1 },
+	{ _RK_GLOBAL, "api.version.request.timeout.ms", _RK_C_INT,
+	  _RK( api_version_request_timeout_ms ),
```
edenhill: stray space after timeout_ms

src/rdkafka_conf.c
```diff
@@ -392,6 +392,10 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
+	  "Timeout for broker version request.",
+	  1000, 10 * 1000, 10 * 1000 },
```
edenhill: This doesn't need to be that restricted.
Let it be 1ms to 5 minutes.

src/rdkafka_request.c
```diff
@@ -1560,8 +1560,8 @@ void rd_kafka_ApiVersionRequest (rd_kafka_broker_t *rkb,
 	/* 0.9.0.x brokers will not close the connection on unsupported
 	 * API requests, so we minimize the timeout to 10s for the request.
 	 * This is a regression on the broker part. */
-	if (rkb->rkb_rk->rk_conf.socket_timeout_ms > 10*1000)
-		rkbuf->rkbuf_ts_timeout = rd_clock() + (10 * 1000 * 1000);
+	if (rkb->rkb_rk->rk_conf.socket_timeout_ms > rkb->rkb_rk->rk_conf.api_version_request_timeout_ms)
```
edenhill: I don't think we need this check anymore, it was only there to make a low socket.timeout.ms take precedence over the hard-coded 10s timeout, but with a specific config property we can just set rkbuf_ts_timeout to the api_version_request_timeout.ms.

src/rdkafka_request.c
```diff
@@ -1596,8 +1596,8 @@ void rd_kafka_SaslHandshakeRequest (rd_kafka_broker_t *rkb,
 	 * API requests, so we minimize the timeout of the request.
 	 * This is a regression on the broker part. */
 	if (!rkb->rkb_rk->rk_conf.api_version_request &&
-            rkb->rkb_rk->rk_conf.socket_timeout_ms > 10*1000)
```
edenhill: Revert this change, this is for SaslHandshake, not ApiVersionRequest

vinipuh5:
@edenhill Done

vinipuh5:
@edenhill Is it ok that there's a test which failed in doozer/target/cmake-xenial-amd64? I noticed that this test fails in a lot of pull requests.

edenhill:
Yeah, I'm not maintaing the contributed cmake build system, I leave that to the community. So any cmake errors can be ignored for now.

Will look into this PR during this week.

Thanks

PR #1406, opened 2017-09-02
Remove unused parameters

Jose123456:
There is no "-s" command line option. The opaque pointer passed to
msg_consume is not used by that function.


edenhill:
Thanks for this

Jose123456:
You're welcome, and thank you for the speedy merge

PR #1417, opened 2017-09-11
property description clarifications

mhowlett:
@edenhill - some things in configuration.md I think would benefit from clarification (... and which I may or may not be interpreting correctly... please check carefully!).

Also: `max.in.flight.requests.per.connection` is a [*] property, but it's only producers that will ever have more than one request in-flight right? (because you want the last retrieved offset to construct the next response on the consumer). I think that would be worth noting if my understanding is correct here.


edenhill:
So `max.in.flight..` is a generic property that is applied to all broker communication, but it doesn't do that much of a difference for anything but the Producer since other mechanisms (the fetcher) make sure there is only one outstanding Fetch request for the Consumer anyway.
It would affect things like asynchronous high-rate Offset commits, but it is highly questionable if there is any use to limit those by this property.

CONFIGURATION.md
```diff
@@ -77,8 +77,8 @@ coordinator.query.interval.ms            |  *  | 1 .. 3600000    |        600000
...
 enable.auto.offset.store                 |  C  | true, false     |          true | Automatically store offset of last message provided to application. <br>*Type: boolean*
-queued.min.messages                      |  C  | 1 .. 10000000   |        100000 | Minimum number of messages per topic+partition in the local consumer queue. <br>*Type: integer*
-queued.max.messages.kbytes               |  C  | 1 .. 1000000000 |       1000000 | Maximum number of kilobytes per topic+partition in the local consumer queue. This value may be overshot by fetch.message.max.bytes. <br>*Type: integer*
+queued.min.messages                      |  C  | 1 .. 10000000   |        100000 | Minimum number of messages per topic+partition librdkafka tries to maintain in the local consumer queue. <br>*Type: integer*
+queued.max.messages.kbytes               |  C  | 1 .. 1000000000 |       1000000 | Maximum number of kilobytes per topic+partition in the local consumer queue. This value may be overshot by fetch.message.max.bytes. This property has higher priority than queued.min.messages. <br>*Type: integer*
```
mhowlett: in particular, verify this is correct.

CONFIGURATION.md
```diff
@@ -91,7 +91,7 @@ offset_commit_cb                         |  C  |                 |
 enable.partition.eof                     |  C  | true, false     |          true | Emit RD_KAFKA_RESP_ERR__PARTITION_EOF event whenever the consumer reaches the end of a partition. <br>*Type: boolean*
 check.crcs                               |  C  | true, false     |         false | Verify CRC32 of consumed messages, ensuring no on-the-wire or on-disk corruption to the messages occurred. This check comes at slightly increased CPU usage. <br>*Type: boolean*
 queue.buffering.max.messages             |  P  | 1 .. 10000000   |        100000 | Maximum number of messages allowed on the producer queue. <br>*Type: integer*
-queue.buffering.max.kbytes               |  P  | 1 .. 2097151    |       4000000 | Maximum total message size sum allowed on the producer queue. <br>*Type: integer*
+queue.buffering.max.kbytes               |  P  | 1 .. 2097151    |       4000000 | Maximum total message size sum allowed on the producer queue. This property has higher priority than queue.buffering.max.messages. <br>*Type: integer*
```
mhowlett: in particular, verify this is correct.

src/rdkafka_conf.c
```diff
@@ -178,16 +178,19 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
 	{ _RK_GLOBAL, "max.in.flight.requests.per.connection", _RK_C_INT,
 	  _RK(max_inflight),
-	  "Maximum number of in-flight requests the client will send. "
-	  "This setting applies per broker connection.",
+		"Maximum number of in-flight requests per broker connection. "
+		"This is a generic property applied to all broker communication, "
+					"however it is primarily relevant to produce requests. "
+					"In particular, note that other mechanisms limit the number "
+					"of outstanding consumer fetch request per broker to one.",
 	  1, 1000000, 1000000 },
```
edenhill: Indentation is off here.

src/rdkafka_conf.c
```diff
@@ -679,7 +683,8 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  1, 10000000, 100000 },
 	{ _RK_GLOBAL|_RK_PRODUCER, "queue.buffering.max.kbytes", _RK_C_INT,
 	  _RK(queue_buffering_max_kbytes),
-	  "Maximum total message size sum allowed on the producer queue.",
+		"Maximum total message size sum allowed on the producer queue. "
+					"This property has higher priority than queue.buffering.max.messages.",
```
edenhill: indentation

src/rdkafka_conf.c
```diff
@@ -602,14 +605,15 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
-          "local consumer queue.",
+	  "Minimum number of messages per topic+partition "
+          "librdkafka tries to maintain in the local consumer queue.",
 	  1, 10000000, 100000 },
 	{ _RK_GLOBAL|_RK_CONSUMER, "queued.max.messages.kbytes", _RK_C_INT,
 	  _RK(queued_max_msg_kbytes),
           "Maximum number of kilobytes per topic+partition in the "
           "local consumer queue. "
-          "This value may be overshot by fetch.message.max.bytes.",
+					"This value may be overshot by fetch.message.max.bytes. "
```
edenhill: indentation

mhowlett:
Yep, I noticed the bad indentation after pushing and did an additional commit to fix this up - did you miss that, or was this wrong as well? (it looks ok to me).

PR #1426, opened 2017-09-21
Improved release automation and instructions

edenhill:
Semi-automation for NuGet and Homebrew


packaging/nuget/artifact.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env python
+#
+#
+# Collects CI artifacts from S3 storage, downloading them
+# to a local directory.
+#
+# The artifacts' folder in the S3 bucket must have the following token
```
mageshn: are we releasing all the clients in this script?

edenhill: This script was taken from confluent-kafka-python :|, but will in this incarnation only grab librdkafka artifacts.
The S3 layout is generic for all our clients needs.

mageshn:
Don't find anything glary...LGTM

PR #1451, opened 2017-10-09
tell the os about the names of threads

tbsaunde:


edenhill:
The thread name is also used in librdkafka logging and there is no length restriction there, and having unique thread names is critical from a logging/troubleshooting perspective (otherwise the thread name is irrelevant).

The only practical way to keep broker thread names unique and meaningful is to use the broker name and port (i.e., rkb_name), which will most of the time be longer than 15 characters.

I understand that it is nice to have somewhat meaningful thread names in ps/top output, are there any other reasons?

Maybe we should have two thread names per thread:
 * the existing rd_kafka_thread_name[] unchanged
 * a short form descriptive non-unique thread name for pthread: "rdkafka:main", "rdkafka:broker"

edenhill:
Thanks again!

edenhill:
Example output from `top -H`:
```
29341 user    20   0  646112 289348   5456 R 80,0  0,9   0:05.04 rdkafka_perform                           
29345 user    20   0  646112 289348   5456 R 13,3  0,9   0:01.04 rdk:broker3                               
29346 user    20   0  646112 289348   5456 R 13,3  0,9   0:00.63 rdk:broker4                               
29344 user    20   0  646112 289348   5456 R  6,7  0,9   0:00.85 rdk:broker2                               
29342 user    20   0  646112 289348   5456 S  0,0  0,9   0:00.00 rdk:main                                  
29343 user    20   0  646112 289348   5456 S  0,0  0,9   0:00.00 rdk:broker-1                  
```

PR #1452, opened 2017-10-10
avoid holding the toppar lock across a write syscall

tbsaunde:
Holding locks across syscalls is generally not a great idea since it
means spending longer with the lock and increases the chances the
process will spend time holding the lock but descheduled.  In this case
it is especially unfortunate because the write will make a broker thread
runnable if it is blocked on poll, but if the broker thread runs before
the thread calling write() it will immediately block again trying to
take the toppar lock.  This appears to reduce worst case latency by at
least 1ms, and quiet possibly much more, its hard to know exactly how
much should be attributed to this instead of other freek events hurting
worst case latency.  Its somewhat unfortunate we need to do a little
more work unconditionally, but the atomic accesses can be optimized
somewhat, and it may be a good idea to use the wakeup fd system more.

### DIFF
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -609,26 +609,28 @@ void rd_kafka_toppar_desired_del (rd_kafka_toppar_t *rktp) {
  * Append message at tail of 'rktp' message queue.
  */
 void rd_kafka_toppar_enq_msg (rd_kafka_toppar_t *rktp, rd_kafka_msg_t *rkm) {
+	int wakeup_fd, queue_len;
 
 	rd_kafka_toppar_lock(rktp);
+	wakeup_fd = rktp->rktp_msgq_wakeup_fd;
 	rd_kafka_msgq_enq(&rktp->rktp_msgq, rkm);
+	queue_len = rd_kafka_msgq_len(&rktp->rktp_msgq);
+        rd_kafka_toppar_unlock(rktp);
 #ifndef _MSC_VER
-        if (rktp->rktp_msgq_wakeup_fd != -1 &&
-            rd_kafka_msgq_len(&rktp->rktp_msgq) == 1) {
+        if (wakeup_fd != -1 && queue_len == 1) {
                 char one = 1;
                 int r;
-                r = rd_write(rktp->rktp_msgq_wakeup_fd, &one, sizeof(one));
+                r = rd_write(wakeup_fd, &one, sizeof(one));
                 if (r == -1)
                         rd_kafka_log(rktp->rktp_rkt->rkt_rk, LOG_ERR, "PARTENQ",
                                      "%s [%"PRId32"]: write to "
                                      "wake-up fd %d failed: %s",
                                      rktp->rktp_rkt->rkt_topic->str,
                                      rktp->rktp_partition,
-                                     rktp->rktp_msgq_wakeup_fd,
+                                     wakeup_fd,
                                      rd_strerror(errno));
         }
 #endif
-        rd_kafka_toppar_unlock(rktp);
 }
 
 
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -613,8 +613,7 @@ void rd_kafka_toppar_enq_msg (rd_kafka_toppar_t *rktp, rd_kafka_msg_t *rkm) {
 
 	rd_kafka_toppar_lock(rktp);
 	wakeup_fd = rktp->rktp_msgq_wakeup_fd;
-	rd_kafka_msgq_enq(&rktp->rktp_msgq, rkm);
-	queue_len = rd_kafka_msgq_len(&rktp->rktp_msgq);
+	queue_len = rd_kafka_msgq_enq(&rktp->rktp_msgq, rkm);
         rd_kafka_toppar_unlock(rktp);
 #ifndef _MSC_VER
         if (wakeup_fd != -1 && queue_len == 1) {
--- a/src/rdkafka_msg.h
+++ b/src/rdkafka_msg.h
@@ -260,11 +260,15 @@ static RD_INLINE RD_UNUSED void rd_kafka_msgq_insert (rd_kafka_msgq_t *rkmq,
 /**
  * Append message to tail of message queue.
  */
-static RD_INLINE RD_UNUSED void rd_kafka_msgq_enq (rd_kafka_msgq_t *rkmq,
+static RD_INLINE RD_UNUSED int rd_kafka_msgq_enq (rd_kafka_msgq_t *rkmq,
 						rd_kafka_msg_t *rkm) {
+	int len;
+
 	TAILQ_INSERT_TAIL(&rkmq->rkmq_msgs, rkm, rkm_link);
-	rd_atomic32_add(&rkmq->rkmq_msg_cnt, 1);
+	len = rd_atomic32_add(&rkmq->rkmq_msg_cnt, 1);
 	rd_atomic64_add(&rkmq->rkmq_msg_bytes, rkm->rkm_len+rkm->rkm_key_len);
+	
+	return len;
 }
 
 


src/rdkafka_partition.c
```diff
@@ -609,26 +609,23 @@ void rd_kafka_toppar_desired_del (rd_kafka_toppar_t *rktp) {
  * Append message at tail of 'rktp' message queue.
  */
 void rd_kafka_toppar_enq_msg (rd_kafka_toppar_t *rktp, rd_kafka_msg_t *rkm) {
+	int wakeup_fd, queue_len;
+	rd_kafka_t *rk;
```
edenhill: You dont need to keep `rk`, use rd_assert() which does not take an rk argument.

src/rdkafka_partition.c
```diff
@@ -609,26 +609,23 @@ void rd_kafka_toppar_desired_del (rd_kafka_toppar_t *rktp) {
...
+        rd_kafka_toppar_unlock(rktp);
 #ifndef _MSC_VER
-        if (rktp->rktp_msgq_wakeup_fd != -1 &&
-            rd_kafka_msgq_len(&rktp->rktp_msgq) == 1) {
+        if (wakeup_fd >= 0 && queue_len == 1) {
                 char one = 1;
                 int r;
-                r = rd_write(rktp->rktp_msgq_wakeup_fd, &one, sizeof(one));
-                if (r == -1)
-                        rd_kafka_log(rktp->rktp_rkt->rkt_rk, LOG_ERR, "PARTENQ",
```
edenhill: why was this removed?

src/rdkafka_partition.c
```diff
@@ -609,26 +609,23 @@ void rd_kafka_toppar_desired_del (rd_kafka_toppar_t *rktp) {
...
 	rd_kafka_toppar_lock(rktp);
+	wakeup_fd = rktp->rktp_msgq_wakeup_fd;
+	rk = rktp->rktp_rkt->rkt_rk;
 	rd_kafka_msgq_enq(&rktp->rktp_msgq, rkm);
+	queue_len = rd_kafka_msgq_len(&rktp->rktp_msgq);
+        rd_kafka_toppar_unlock(rktp);
 #ifndef _MSC_VER
-        if (rktp->rktp_msgq_wakeup_fd != -1 &&
-            rd_kafka_msgq_len(&rktp->rktp_msgq) == 1) {
+        if (wakeup_fd >= 0 && queue_len == 1) {
```
edenhill: This is not completely without side-effects:

Having the queue len + write outside of holding the lock risks two (or more) threads calling write() simultaneously which will cause additional wake-ups in the broker thread and more context switches in these enq threads, all resulting in increased CPU usage.

edenhill: Please try to keep as much of the existing code as possible if a change is not strictly needed.
E.g.. `wakeup_fd == -1 ..`

I think we should change msgq_len() to return the return-value from atomic_add() so that we don't need the msgq_len() call at all.

tbsaunde: the change to the comparison was an accident, and I'll revert that, though I do sort of think >= 0 is a better test given that other negative return values don't mean anything.
I was planning on doing that change to msgq_enq, but wanted to keep it separate, and didn't think of it till this was written.

src/rdkafka_partition.c
tbsaunde: I replaced it with the assert because I can't see an error that can actually happen except EINTER maybe which maybe we should separately handle.  I don't have a problem with the logging in general, but in this case it makes it hard to drop the lock earlier.  Since the log message includes the topic name we somehow need to have access to that in the exceptional case.  If we drop the lock before write() we don't know that string will even still be around right? so the alternative I can see is copying the string every time while holding the lock and that seems really bad.  I guess we could drop the topic name / partition from the message, but that would mean still fetching rk while holding the lock.

src/rdkafka_partition.c
tbsaunde: I don't think so?  The key thing is that the queue length is read while we hold the lock.  That means if two threads read a queue length of 1 in the time between when those two threads held the lock another thread must have dequeued messages to set the queue length to 0.  Another way to think about this is that if you put the unlock after the if the additional operations that take place with the lock held  only depend on values local to the executing thread.

src/rdkafka_partition.c
tbsaunde: ah thanks, though this may conflict with your comments on the logging change below?

tbsaunde:
@edenhill, I believe I've addressed all your comments, can you please take another look?

src/rdkafka_partition.c
edenhill: We can't make any safe assumptions on what error we get from a system call (since the code runs on a plethora of different operating systems and versions), so we must be very careful in not doing anything drastic in these rare unknown cases.
Thus, we absolute don't want to crash the application if a non-essential wakeup file descriptor is not writable :)

Generally, asserts should only be used to guard for internal state and should never be used for external input.

Regarding the log locking:
The log message is safe without locking, because:
 * since all relevant objects (rktp, rkt, rk) are refcounted and our caller must hold a reference these objects will not go away from under our feet
 * `rktp->rktp_rkt` and `rkt->rkt_rk` are constant, they are never changed during the lifetime of these objects
 * same with `rkt_topic`, `rktp_partition`, etc.

The only thing that may change is `rktp->rktp_msgq_wakeup_fd` which must be protected by a lock, but since you've already extracted the fd into a variable we can use that instead.
This also shows us a valid scenario where the write may fail:
 * thread 1 is in enq_msg() having extracted the fd and getting ready to write
 * thread 2 removes the io_event fd from the queue and closes the fd
In this case thread 1's write() will fail, but the program should continue to operate since on the next call the enq_msg() the fd will no longer be set.

src/rdkafka_topic.c
```diff
@@ -657,7 +657,7 @@ static void rd_kafka_topic_assign_uas (rd_kafka_itopic_t *rkt,
 	rd_kafka_msg_t *rkm, *tmp;
 	rd_kafka_msgq_t uas = RD_KAFKA_MSGQ_INITIALIZER(uas);
 	rd_kafka_msgq_t failed = RD_KAFKA_MSGQ_INITIALIZER(failed);
-	int cnt;
+	int cnt, failed_cnt = 0;
```
edenhill: This is the slow-path so you don't really need to do this optimization here. Suggest removing this patch.

src/rdkafka_msg.c
```diff
@@ -482,17 +482,18 @@ int rd_kafka_msgq_age_scan (rd_kafka_msgq_t *rkmq,
...
 	TAILQ_FOREACH_SAFE(rkm, &rkmq->rkmq_msgs, rkm_link, tmp) {
 		if (likely(rkm->rkm_ts_timeout > now))
 			break;
 
 		rd_kafka_msgq_deq(rkmq, rkm, 1);
-		rd_kafka_msgq_enq(timedout, rkm);
+		timedout_cnt = rd_kafka_msgq_enq(timedout, rkm);
 	}
 
-	return rd_atomic32_get(&timedout->rkmq_msg_cnt) - cnt;
```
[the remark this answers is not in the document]
tbsaunde: Its O(N), but one would hope in practice nothing times out and its ~O(1).  aiui atomic operations are especially costly, so if we generally don't loop it seems like it may actually matter.  I'm pretty sure the calculation is already racey and shouldn't be used for something that must know exactly how many elements are in the queue at this exact moment, I think to know exactly you need to hold the queue lock.  The problem is that between when the atomic fetch is performed and the value is returned to the caller another thread could modify timedout.  Looking at the rdkafka_broker.c caller it seems ok with this raceyness, in the worst case you miss the new queue elements, but get them the next time through.  It looks like the only other caller only uses the return value in logging, so that's not terribly important.

That all said I'm happy to revert it doing the conservatively correct thing until we have numbers showing it matters.  Mostly this second patch was because I think you suggested it earlier.

src/rdkafka_topic.c
tbsaunde: Well, the if (failed_cnt > 0) check at the very end is always executed, so that matters somewhat I guess.  It also moves us closer to a world where the count is just protected by the queue lock instead of using atomic operations.  I'm not sure what downside you see, but I'll remove this files changes for now to get the rest in.

edenhill:
Thanks for this!

Issue #1435, opened 2017-09-28
Consuming using auto.offset.reset=latest doesn't emit EOF for partition 


mfontanini:
Description
===========
This happened a couple of times but I can't get any logs. I basically ended up using auto.offset.reset=latest to bypass #1307 but now it seems like once in a while I hit a case where I'm loading messages right after subscribing to a topic where there are messages in already and I never get an EOF for one partition. I may add that I'm pausing consumption on every partition when I reach EOF. My consumption looks like:

```
consumer.subscribe()
while (not reached eof/filled an internal queue on all partitions) {
     msg = consumer.poll()
     if (msg.eof()) {
        consumer.pause(msg.partition)
     }
     else {
         // store message in some queue
         if (some condition) {
              consumer.pause(msg.partition)
         }
     }
}
```

And that is pretty much it. In this case I managed to attach a debugger and I can see that for one of the 10 partitions that were assigned I never reached EOF nor did I get any messages so I'm basically looping forever just waiting for that one partition to give me something. 

I think this is related to auto.offset.reset=latest because I have other consumers using the same code but using offset reset smallest which have never seen the bug and have been running for months now.  

How to reproduce
================
Consume and wait for an EOF that never comes, possibly using auto offset reset latest policy. 

Checklist
=========
Please provide the following information:

 - [x] librdkafka version (release number or git tag): 0.9.5. Note that I'd like to use the latest version but I've been tracking the issues and noticed there's a couple of things I wouldn't want to be affected by so I'd rather wait for the next release before doing so.
 - [x] Apache Kafka version: 0.10.1.0
 - [x] librdkafka client configuration: `enable.auto.commit=false, auto.offset.reset=latest`
 - [x] Operating system: GNU/Linux x64
 - [ ] Using the legacy Consumer
 - [x] Using the high-level KafkaConsumer
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue: not sure if this qualifies as critical, but the fact that I can consume and not get a single message from a partition, not even one that I'm supposed to get (the EOF one) seems undesirable and in this case completely stalls my consumption. Now I need to apply some sort of timeout so that if I'm looping for more than X seconds and I don't get any message, I just kill my application.

edenhill:
Is there any difference between the working and non-working partitions in regards to message rate, broker leader, etc?

[...]

PR #1453, opened 2017-10-11
Don't store invalid offset as next one when pausing

### LINKED ISSUE #1307: Pausing consumption resets offsets
Description
===========
I have some application which pauses and resumes consumption on one topic so as to throttle each partition individually. This works fine but I'm coming across some pretty bad issue where:

* The consumer offset lags are 0 when the consumer comes up
* I start consumption, get EOF on all partitions (no messages at all as there's no lag).
* I pause consumption on all partitions. This is done on purpose to fill some local queues for each partition.
* Then I resume consumption on all partitions. At this point, all my local queues are filled and I can actually start processing messages.

Normally this works fine, but when I start and get EOF on all partitions, at the point where I resume consumption, instead of consuming from the next offset, the `auto.offset.reset` policy, which I set to smallest, kicks in and I'm not consuming the entire topic from scratch, which is undesirable to say the least.

How to reproduce
================

I can reproduce this 100% of the times doing the following:

* Create a topic which has some messages.
* Consume all of them using consumer group id X and commit their offsets.
* Stop your consumer app
* Start a new consumer (using the same group id X) using `auto.offset.reset == smallest` and subscribe to the topic.
* Wait until partitions are assigned.
* Once they're assigned, pause consumption on all of them (e.g. `consumer.pause_partitions(consumer.get_assignment())`).
* Resume consumption on all of them (e.g. `consumer.resume_partitions(consumer.get_assignment())`).

Expected behavior: nothing happens as there are no messages to be consumed
Actual behavior: all of the messages are consumed again.

See the relevant log lines [here](https://gist.github.com/mfontanini/90ff98a79ab272a5fa7887a8c747fd9c).

So, without knowing much about the internals of librdkafka, it seems like since the library doesn't know any offset (because it got EOF initially), it somehow discards what the `OffsetFetchResponse` said and thinks it has an INVALID offset, which makes it end up going back to the beginning as the policy indicates.

Checklist
=========

 - [x] librdkafka version (release number or git tag): 0.9.4
 - [x] Apache Kafka version: 0.10.1.0
 - [x] librdkafka client configuration: `enable.auto.commit=false, auto.offset.reset=smallest`
 - [x] Operating system: GNU/Linux (Ubuntu Xenial 64 bits)
 - [x] Using the high-level KafkaConsumer
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka provided in the "how to reproduce" section
 - [ ] Critical issue: as much as you'll likely think this is some corner case, this affects me very badly as I rely on this behavior. Also, I already had to patch the code a bit since I've had issues with librdkafka ignoring some pause/resume operations if I don't poll in between, but that's another story.

mfontanini:
This fixes the issue where a consumer subscribes to a topic and then pause consumption where all consumer offsets are initially already EOF. Without this check, the offset that's used after resuming consumption is invalid, which ends up falling back to whatever `auto.reset.policy` is being used, which in turn makes the consumer "forget" about which offset was actually already committed.

This fixes #1307. I'm not sure if this has any undesired side effects but at least I'm sure I'm not hitting the issue anymore.

### DIFF
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -1531,7 +1531,9 @@ static void rd_kafka_toppar_pause_resume (rd_kafka_toppar_t *rktp,
 		if (rk->rk_type == RD_KAFKA_CONSUMER) {
 			/* Save offset of last consumed message+1 as the
 			 * next message to fetch on resume. */
-			rktp->rktp_next_offset = rktp->rktp_app_offset;
+			if (rktp->rktp_app_offset != RD_KAFKA_OFFSET_INVALID) {
+				rktp->rktp_next_offset = rktp->rktp_app_offset;
+			}
 
 			rd_kafka_dbg(rk, TOPIC, pause?"PAUSE":"RESUME",
 				     "%s %s [%"PRId32"]: at offset %s "


mfontanini:
Any updates on this one? I haven't hit the issue again after applying this fix.

edenhill:
Thank you for this clean fix!
I'm adding a test-case separately

edenhill:
Is this fixing #1435  as well you think?

PR #1480, opened 2017-10-23
Message headers support (C)

edenhill:


examples/rdkafka_example.c
```diff
@@ -187,6 +189,27 @@ static void msg_consume (rd_kafka_message_t *rkmessage,
 				!timestamp ? 0 :
 				(int)time(NULL) - (int)(timestamp/1000));
 		}
+
+                if (!rd_kafka_message_headers(rkmessage, &hdrs)) {
```
mhowlett: some n00b commentary for your entertainment. I find this difficult to read, though I guess it's a Cism and so no problem.  I'd have written this here: `if (rd_kafka_message_headers(rkmessage, &hdrs) == RD_KAFKA_RESP_ERR_NO_ERROR)`.

examples/rdkafka_example.c
```diff
@@ -187,6 +189,27 @@ static void msg_consume (rd_kafka_message_t *rkmessage,
...
+                        size_t idx = 0;
+                        const char *name;
+                        const void *val;
+                        size_t size;
+
+                        fprintf(stdout, "%% Headers:");
+
+                        while (!rd_kafka_header_iter_all(hdrs, idx++,
```
mhowlett: I'd remove the word `iter` from this method name since you're iterating outside the method call (with the `idx++`).  also the word `all`. perhaps just `rd_kafka_header_get` or something?

examples/rdkafka_example.c
```diff
@@ -370,6 +395,31 @@ int main (int argc, char **argv) {
...
+                        val = strchr(name, '=');
+                        if (val) {
+                                name_sz = (size_t)(val-name);
+                                val++; /* past the '=' */
+                        }
+
+                        if (!hdrs)
+                                hdrs = rd_kafka_headers_new(8);
+
+                        err = rd_kafka_header_add(hdrs, name, name_sz, val, -1);
```
mhowlett: this is a bit tricky (dual meaning of the -1), but I can't think of anything I prefer and I guess it's ok.

src/rdkafka.h
```diff
@@ -845,10 +862,164 @@ typedef enum rd_kafka_vtype_t {
...
+ * @param name       Header name.
+ * @param name_size  Header name size (not including the null-terminator).
+ *                   If -1 the \p name length is automatically acquired using
+ *                   strlen().
+ * @param value      Pointer to header value, or NULL (set size to 0 or -1).
+ * @param value_size Size of header value. If -1 the \p value is assumed to be a
+ *                   null-terminated string and the length is automatically
+ *                   acquired using strlen().
+ *
+ * @returns RD_KAFKA_RESP_ERR__READ_ONLY if the headers are read-only,
```
mhowlett: RD_KAFKA_RESP_ERR__READ_ONLY is never returned I think? do you have plans for this? is it a bit odd for a C api to try and enforce this sort of thing?

src/rdkafka.h
```diff
@@ -845,10 +862,164 @@ typedef enum rd_kafka_vtype_t {
...
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_header_add (rd_kafka_headers_t *hdrs,
+                     const char *name, ssize_t name_size,
+                     const void *value, ssize_t value_size);
+
+/**
+ * @brief Remove all headers for the given key (if any).
+ *
+ * @returns RD_KAFKA_RESP_ERR__READ_ONLY if the headers are read-only,
```
mhowlett: same comment about readonly as above

src/rdkafka.h
```diff
@@ -845,10 +862,164 @@ typedef enum rd_kafka_vtype_t {
...
+rd_kafka_header_remove (rd_kafka_headers_t *hdrs, const char *name);
+
+
+/**
+ * @brief Find last header in list \p hdrs matching \p name.
+ *
+ * @param name   Header to find (last match).
+ * @param valuep (out) Set to a (null-terminated) const pointer to the value
+ *               (may be NULL).
+ * @param sizep  (out) Set to the value's size (not including null-terminator).
```
mhowlett: perhaps include a note about what it means if sizep == -1

src/rdkafka.h
```diff
@@ -845,10 +862,164 @@ typedef enum rd_kafka_vtype_t {
...
+ * @param hdrs   Headers to iterate.
+ * @param idx    Iterator index, start at 0 and increment by one for each call
+ *               as long as RD_KAFKA_RESP_ERR_NO_ERROR is returned.
+ * @param name   Header name to match.
+ * @param valuep (out) Set to a (null-terminated) const pointer to the value
+ *               (may be NULL).
+ * @param sizep  (out) Set to the value's size (not including null-terminator).
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_header_iter (const rd_kafka_headers_t *hdrs, size_t idx,
```
mhowlett: don't like `iter` in the name here either. maybe `rd_kafka_header_get` is appropriate for this as well (not sure of the best name).

src/rdkafka_buf.h
```diff
@@ -195,8 +195,6 @@ rd_tmpabuf_write_str0 (const char *func, int line,
                                 "expected %"PRIusz" bytes > %"PRIusz    \
                                 " remaining bytes",                     \
                                 __len0, rd_kafka_buf_read_remain(rkbuf)); \
-                        (rkbuf)->rkbuf_err = RD_KAFKA_RESP_ERR__BAD_MSG; \
```
mhowlett: i'm not going to bother understanding why you took this out, I assume you know what you're doing.

src/rdkafka_header.c
```diff
@@ -0,0 +1,220 @@
...
+
+        return RD_KAFKA_RESP_ERR_NO_ERROR;
+}
+
+
+/**
+ * @brief header_t(name) to char * comparator
+ */
+static int rd_kafka_header_cmp_str (void *_a, void *_b) {
+        const rd_kafka_header_t *a = _a;
```
mhowlett: not sure why you have `void*` parameters then immediately cast?

mhowlett:
LGTM after you've addressed comments as you see fit.

src/rdkafka_header.c
edenhill: The comparator signature is `int () (void *, void *)`, passing a function with another signature would cause a compiler warning

src/rdkafka_buf.h
edenhill: parse_fail() already sets it

src/rdkafka.h
edenhill: sizep will not be set to -1 (null values have a size of 0)

src/rdkafka.h
edenhill: I got this from the Java API, but it turned out to be an internal API that we dont need to mimic. Will remove it.

almazik:
Is there a progress on this merge request? This feature is important for us...

johnistan:
I have played around with python support for this feature. I haven't found a public API to find out how many headers an msg has. leading to hacks like:

```       
        //HACK to figure out size of headers
        while (!rd_kafka_header_iter_all(hdrs, header_size++,
                                         &header_key, &header_value, &header_value_size)) {
        }
```

Can we add a `rd_kafka_header_length` to the public? Or is there an existing approach I am not seeing.

edenhill:
@johnistan Good idea, will add.

examples/rdkafka_example.c
edenhill: Okay, what about `rd_kafka_header_get()` for the key-specific one, and `rd_kafka_header_enum()` or `..list()` or `..get_all()` for the one that iterates all keys?

src/rdkafka.h
edenhill: Second thought I think we'll keep it so that headers can be "closed" when final message serialization has taken place, this would stop (future) late-stage interceptors from thinking they could modify the headers after serialization.

mhowlett:
@edenhill - go for it. lgtm.

PR #1481, opened 2017-10-23
stop calling cnd_timedwait() with a timeout of 0h

tbsaunde:
Waiting 0ms to be signaled the queue now contains elements doesn't
really mean anything.  However per the POSIX specification of
pthread_cond_timedwait() the lock must be released and reacquired.  As a
result if we skip calling cnd_timedwait() when the timeout is 0 we
improve average and best case latency in various conditions by between
0.1 and 0.2ms.

### DIFF
--- a/src/rdkafka_queue.c
+++ b/src/rdkafka_queue.c
@@ -359,7 +359,10 @@ rd_kafka_op_t *rd_kafka_q_pop_serve (rd_kafka_q_t *rkq, int timeout_ms,
                                         break; /* Proper op, handle below. */
                         }
 
-                        /* No op, wait for one */
+                        /* No op, wait for one if we have time left */
+			if (timeout_ms == 0)
+				break;
+
                         pre = rd_clock();
 			if (cnd_timedwait_ms(&rkq->rkq_cond,
 					     &rkq->rkq_lock,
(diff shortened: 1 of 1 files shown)


src/rdkafka_queue.c
```diff
@@ -361,7 +361,8 @@ rd_kafka_op_t *rd_kafka_q_pop_serve (rd_kafka_q_t *rkq, int timeout_ms,
 
                         /* No op, wait for one */
                         pre = rd_clock();
-			if (cnd_timedwait_ms(&rkq->rkq_cond,
+			if (timeout_ms == 0 ||
```
edenhill: I think we can simply unlock and return NULL if timeout_ms is 0 instead of trying again, that's what will happen below at the `while (timeout_ms != ..NOWAIT)` anyway,

Put the check above the `pre = ..` assignment.

edenhill:
Nice find :)

src/rdkafka_queue.c
tbsaunde: @edenhill, actually that's what this already does, because in the case cnd_timedwait() returns timedout we just unlock and return null.  That said we could avoid the call to clock_gettime() which would save us a couple of microseconds while holding a lock.  So I'll make this a separate check.

PR #1483, opened 2017-10-23
Allow for calling rd_kafka_queue_io_event_enable() from the C++ world

akhi3030:
Hi,

As discussed in, https://github.com/edenhill/librdkafka/issues/617, I've created a pull request.  Awaiting feedback.

Thanks,
Akhi

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -1429,6 +1429,23 @@ class RD_EXPORT Queue {
   virtual int poll (int timeout_ms) = 0;
 
   virtual ~Queue () = 0;
+
+  /**
+   * @brief Enable IO event triggering for queue.
+   *
+   * To ease integration with IO based polling loops this API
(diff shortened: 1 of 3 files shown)


src-cpp/rdkafkacpp.h
```diff
@@ -1428,6 +1428,8 @@ class RD_EXPORT Queue {
    */
   virtual int poll (int timeout_ms) = 0;
 
+  virtual void io_event_enable (int fd, const void *payload, size_t size) = 0;
```
edenhill: Need a doc string here. Copy (and reformat) the one from rdkafka.h

edenhill: Move this declaration to the end of the class for ABI stability

akhi3030:
Thank you very much for the feedback.  This is my first PR so I hope I've done things correctly.

edenhill:
Thanks for this!

PR #1491, opened 2017-10-27
Fix DNS cache logic and update default from 1s to 60s

dacjames:
To my eyes, the dns caching logic was not fully implemented. 

This converts to using rd_clock() and sets the rkb_t_rsal_last timestamp after the value is loaded.

I also adjust the default dns cache time to 60s. 1s is very low in practice; when you have a large number of clients, a low DNS caching time leads to a flood of DNS requests when a broker is restarted. Could split out the config change if necessary

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -27,7 +27,7 @@ socket.receive.buffer.bytes              |  *  | 0 .. 100000000  |             0
 socket.keepalive.enable                  |  *  | true, false     |         false | Enable TCP keep-alives (SO_KEEPALIVE) on broker sockets <br>*Type: boolean*
 socket.nagle.disable                     |  *  | true, false     |         false | Disable the Nagle algorithm (TCP_NODELAY). <br>*Type: boolean*
 socket.max.fails                         |  *  | 0 .. 1000000    |             3 | Disconnect from broker when this number of send failures (e.g., timed out requests) is reached. Disable with 0. NOTE: The connection is automatically re-established. <br>*Type: integer*
-broker.address.ttl                       |  *  | 0 .. 86400000   |          1000 | How long to cache the broker address resolving results (milliseconds). <br>*Type: integer*
+broker.address.ttl                       |  *  | 0 .. 86400000   |         60000 | How long to cache the broker address resolving results (milliseconds). <br>*Type: integer*
 broker.address.family                    |  *  | any, v4, v6     |           any | Allowed broker IP address families: any, v4, v6 <br>*Type: enum value*
 reconnect.backoff.jitter.ms              |  *  | 0 .. 3600000    |           500 | Throttle broker reconnection attempts by this value +-50%. <br>*Type: integer*
 statistics.interval.ms                   |  *  | 0 .. 86400000   |             0 | librdkafka statistics emit interval. The application also needs to register a stats callback using `rd_kafka_conf_set_stats_cb()`. The granularity is 1000ms. A value of 0 disables statistics. <br>*Type: integer*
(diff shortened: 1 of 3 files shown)


src/rdkafka_broker.c
```diff
@@ -596,17 +596,18 @@ rd_kafka_broker_send (rd_kafka_broker_t *rkb, rd_slice_t *slice) {
 static int rd_kafka_broker_resolve (rd_kafka_broker_t *rkb) {
 	const char *errstr;
 
+	rd_ts_t now = rd_clock();
```
edenhill: No need for temp variable here, only used once.

src/rdkafka_conf.c
```diff
@@ -284,7 +284,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  _RK(broker_addr_ttl),
 	  "How long to cache the broker address resolving "
           "results (milliseconds).",
-	  0, 86400*1000, 1*1000 },
+	  0, 86400*1000, 60*1000 },
```
edenhill: DNS caching is performed by the resolver (typically libc) based on the records TTL setting, adding another layer of caching is inadvisable and will slow down the propagation of DNS record changes.

The initial default of broker.address.ttl was actually 5 minutes, but was decreased to 1s for this reason:
a0aff233e0be2f295e2e32e468ab22498c86c62c
https://github.com/edenhill/librdkafka/issues/494

The current default of 1s provides a bit of dampening, rather than caching.
A user that wants to dampen further can increase the configuration property.

src/rdkafka_broker.c
```diff
@@ -624,6 +625,8 @@ static int rd_kafka_broker_resolve (rd_kafka_broker_t *rkb) {
                                              "Failed to resolve '%s': %s",
                                              rkb->rkb_nodename, errstr);
 			return -1;
+		} else {
+			rkb->rkb_t_rsal_last = rd_clock();
```
edenhill: use 8 whitespace tabs.

dacjames:
> DNS caching is performed by the resolver (typically libc) based on the records TTL setting.

That’s not true. In general, libc does not perform DNS caching and major Linux distributions do not enable it by default.

I’ll remove it if you want but the current default is bad and will overload DNS servers during broker maintenance once you have a few hundred clients.

A more robust form of backoff when restablishing the broker connection may be a better solution to the DNS flooding problem, though. Even with a longer DNS cache, the current retry loop makes decommissioning a broker from a large, live system impossible.

src/rdkafka_broker.c
dacjames: If that’s your preference. As someone who recently spent a lot of time reading librdkafka from a place of relative ignorance, more temporary variables would have been very helpful.

src/rdkafka_broker.c
dacjames: Do you mean to use tabs instead of spaces? The block above was inconsistent (the rd_kafka_broker_fail call is indented with spaces) so I wasn't sure what to do.

dacjames:
@edenhill Applied changes per your review.

Tried to fix the whitespace but ending up confirming that it is already using tabs.

[tabs](https://i.imgur.com/XKGv4hA.png)

src/rdkafka_broker.c
edenhill: librdkafka has mixed indents due to weird historical reasons, but all new code, and any changes to existing code, should use 8 whitespaces for tabs.

src/rdkafka_conf.c
```diff
@@ -284,7 +284,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  _RK(broker_addr_ttl),
 	  "How long to cache the broker address resolving "
           "results (milliseconds).",
-	  0, 86400*1000, 1*1000 },
```
edenhill: avoid unneccessary changes (for git blame to be convenient)

CONFIGURATION.md
```diff
@@ -27,7 +27,7 @@ socket.receive.buffer.bytes              |  *  | 0 .. 100000000  |             0
 socket.keepalive.enable                  |  *  | true, false     |         false | Enable TCP keep-alives (SO_KEEPALIVE) on broker sockets <br>*Type: boolean*
 socket.nagle.disable                     |  *  | true, false     |         false | Disable the Nagle algorithm (TCP_NODELAY). <br>*Type: boolean*
 socket.max.fails                         |  *  | 0 .. 1000000    |             3 | Disconnect from broker when this number of send failures (e.g., timed out requests) is reached. Disable with 0. NOTE: The connection is automatically re-established. <br>*Type: integer*
-broker.address.ttl                       |  *  | 0 .. 86400000   |          1000 | How long to cache the broker address resolving results (milliseconds). <br>*Type: integer*
+broker.address.ttl                       |  *  | 0 .. 86400000   |         60000 | How long to cache the broker address resolving results (milliseconds). <br>*Type: integer*
```
edenhill: this file was not re-generated after you changed the default back so it still contains the original change.
Remove from PR

dacjames:
@edenhill I think this resolves said technicalities.

edenhill:
Big thanks for this!

PR #1496, opened 2017-10-31
Make librdkafka compiles with ccache.

yangyingchao:
Ccache can be used to speed up compiling. When using ccache, we usually
set following environments variables:

```
export CC="ccache gcc"
export CXX="ccache g++"
```

In current version of configure.cc, we pass $CXX to function
mkl_mkvar_set without quoting marks:

```
mkl_mkvar_set "CXX" CXX $CXX
```
mkl_mkvar_set actually receives 4 variables: CXX, CXX, ccache and g++,
then mkl_mkvar_set will set CXX to ccache, but not "ccache g++", thus
causes errors when compiling files in folder src-cpp.

### DIFF
--- a/mklove/modules/configure.cc
+++ b/mklove/modules/configure.cc
@@ -39,7 +39,7 @@ function checks {
             fi
         fi
         export CXX="${CXX}"
-        mkl_mkvar_set "CXX" CXX $CXX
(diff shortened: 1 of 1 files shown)


edenhill:
Good stuff, thanks!

PR #1505, opened 2017-11-06
Fix #1397

### LINKED ISSUE #1397: Producer thread stuck in infinite loop
Hi,

Description
===========
After running on production for some time, I'm seeing the librdkafka thread on 100% cpu. 
Call stack is:
```
#0  rd_kafka_broker_ua_idle (rkb=rkb@entry=0x7f20fc008290, timeout_ms=0, timeout_ms@entry=-1) at rdkafka_broker.c:3316
#1  0x00007f21137b7afd in rd_kafka_broker_thread_main (arg=arg@entry=0x7f20fc008290) at rdkafka_broker.c:4731
#2  0x00007f21137fdc77 in _thrd_wrapper_function (aArg=<optimized out>) at tinycthread.c:624
#3  0x00007f2113c8f6ba in start_thread (arg=0x7f210de3e700) at pthread_create.c:333
#4  0x00007f21130433dd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```
From what I saw, rkb->rkb_state is RD_KAFKA_BROKER_STATE_CONNECT and rkb->rkb_blocking_max_ms is 0. So, rd_kafka_broker_ua_idle returns immediately, and rd_kafka_broker_thread_main doesn't really do anything in this state.
If I understand correctly, this means that librdkafka issued some connect request that got stuck for some reason, and it's polling for it indefinitely. If that is correct - is it possible to configure some timeout on connect?
Other than that, the log is full with `kafka message delivery failed: Local: Message timed out`, getting hundreds of these per second.
I'm leaving the process in this state for now, in case there any further info you want me to extract from it.

How to reproduce
================
Unfortunately, I have no way to reproduce this, it happens only after running for a long time in production.

Checklist
=========

 - [X] librdkafka version (release number or git tag): I took master couple of months ago, I'm standing on commit 2adde005561089ececefb28270b644b0344e7180
 - [X] Apache Kafka version: kafka_2.11-0.10.1.1
 - [X] librdkafka client configuration: setting these parameters
```
bootstrap.servers = ...
request.required.acks = 0
compression.codec = snappy
log_level = 6
message.send.max.retries = 0
queue.buffering.max.messages = 100000
retry.backoff.ms = 10
```
 - [X] Operating system: Ubuntu 16.04.2 LTS
 - [ ] Using the high-level KafkaConsumer
 - [X] Provide logs (with `debug=..` as necessary) from librdkafka
As mentioned above, getting hundreds of these per second - `kafka message delivery failed: Local: Message timed out`
 - [ ] Critical issue

Thank you

Eran

edenhill:
This is two fixes for the same thing, but they are valuable each on their own.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -343,6 +343,15 @@ void rd_kafka_broker_fail (rd_kafka_broker_t *rkb,
 		rkb->rkb_recv_buf = NULL;
 	}
 
+        /* Reset max blocking time back to the default to avoid busy-looping
+         * on reconnect if blocking=0 (#1397).
+         * But honour the lower on-termination blocking time. */
+        if (rd_kafka_terminating(rkb->rkb_rk))
+                rkb->rkb_blocking_max_ms = 1;
+        else
+                rkb->rkb_blocking_max_ms =
+                        rkb->rkb_rk->rk_conf.socket_blocking_max_ms;
+
 	rd_kafka_broker_lock(rkb);
 
 	/* The caller may omit the format if it thinks this is a recurring
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -2285,14 +2285,16 @@ static void rd_kafka_broker_ua_idle (rd_kafka_broker_t *rkb, int timeout_ms) {
          * in state ..BROKER_STATE_CONNECT we only run this loop
          * as long as the state remains the same as the initial, on a state
          * change - most likely to UP, a correct serve() function
-         * should be used instead. */
-        while (!rd_kafka_broker_terminating(rkb) &&
-               (int)rkb->rkb_state == initial_state &&
-               !rd_timeout_expired(rd_timeout_remains(abs_timeout))) {
-
+         * should be used instead.
+         * Regardless of constraints (terminating, timeouts), poll at
+         * least once. The state will not have changed on the first iteration.
+         */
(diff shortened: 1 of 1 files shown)


src/rdkafka_broker.c
```diff
@@ -343,6 +343,15 @@ void rd_kafka_broker_fail (rd_kafka_broker_t *rkb,
 		rkb->rkb_recv_buf = NULL;
 	}
 
+        /* Reset max blocking time back to the default to avoid busy-looping
+         * on reconnect if blocking=0 (#1397).
+         * But honour the lower on-termination blocking time. */
+        if (rd_kafka_terminating(rkb->rkb_rk))
+                rkb->rkb_blocking_max_ms = 1;
+        else
```
mhowlett: my understanding is limited, but this looks good to me (resetting back to what it would be on a new connection). I wonder if you need to explicitly set this to 1 in the termination case though, but doesn't hurt.

mhowlett:
looks good to the extent of my knowledge. I don't completely understand the reasoning behind the change in `rd_kafka_broker_ua_idle`.

src/rdkafka_broker.c
edenhill: Worst case the thread will block on io-wait for blocking_max_ms (def 1000ms), this slows down termination, so we minimize the blocking time when we know that we are terminating.

Issue #1502, opened 2017-11-01
SASL Kerberos using wrong principal on Windows


edenhill:
The src/rdkafka_sasl_win32.c code is using the client principal `sasl.kerberos.principal` for the target service principal `sasl.kerberos.service.name`, it should use the latter.

PR #1507, opened 2017-11-07
win32: Use sasl.kerberos.service.name for broker principal, not sasl.…

edenhill:
…kerberos.principal (#1502)


mhowlett:
LGTM

CONFIGURATION.md
```diff
@@ -59,8 +59,8 @@ ssl.certificate.location                 |  *  |                 |
 ssl.ca.location                          |  *  |                 |               | File or directory path to CA certificate(s) for verifying the broker's key. <br>*Type: string*
 ssl.crl.location                         |  *  |                 |               | Path to CRL for verifying broker's certificate validity. <br>*Type: string*
 sasl.mechanisms                          |  *  |                 |        GSSAPI | SASL mechanism to use for authentication. Supported: GSSAPI, PLAIN, SCRAM-SHA-256, SCRAM-SHA-512. **NOTE**: Despite the name only one mechanism must be configured. <br>*Type: string*
-sasl.kerberos.service.name               |  *  |                 |         kafka | Kerberos principal name that Kafka runs as. <br>*Type: string*
-sasl.kerberos.principal                  |  *  |                 |   kafkaclient | This client's Kerberos principal name. <br>*Type: string*
+sasl.kerberos.service.name               |  *  |                 |         kafka | Kerberos principal name that Kafka runs as, not including /hostname@REALM <br>*Type: string*
```
rnpridgeon: Is this to say you can only contact Kafka if it resides in the same realm? How does this work if you have cross-realm trusts enabled and you want to communicate with a broker that lives in a realm other than your own? I ask because this seems like a fairly plausible situation since the client is always the login user.

edenhill: Yes, we'll expose the realm and client principal name in a later version (on demand).

PR #1513, opened 2017-11-09
new nuget package layout + debian9 librdkafka build + simplifying

mhowlett:
@edenhill - the nuget package generated from this should be fine, but I should finish testing all usage scenarios before merging.

### DIFF
--- a/packaging/nuget/packaging.py
+++ b/packaging/nuget/packaging.py
@@ -307,18 +307,22 @@ class NugetPackage (Package):
             [{'arch': 'x64', 'plat': 'linux', 'fname_startswith': 'librdkafka.tar.gz'}, './lib/librdkafka.so.1', 'runtimes/linux-x64/native/librdkafka.so'],
 
             [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'msvcr120.zip'}, 'msvcr120.dll', 'runtimes/win7-x64/native/msvcr120.dll'],
+            # matches librdkafka.redist.{VER}.nupkg
             [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'librdkafka.redist'}, 'build/native/bin/v120/x64/Release/librdkafka.dll', 'runtimes/win7-x64/native/librdkafka.dll'],
             [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'librdkafka.redist'}, 'build/native/bin/v120/x64/Release/librdkafkacpp.dll', 'runtimes/win7-x64/native/librdkafkacpp.dll'],
             [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'librdkafka.redist'}, 'build/native/bin/v120/x64/Release/zlib.dll', 'runtimes/win7-x64/native/zlib.dll'],
+            # matches librdkafka.{VER}.nupkg
             [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'librdkafka', 'fname_excludes': ['redist', 'symbols']},
              'build/native/lib/v120/x64/Release/librdkafka.lib', 'build/native/lib/win7/x64/win7-x64-Release/v120/librdkafka.lib'],
             [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'librdkafka', 'fname_excludes': ['redist', 'symbols']},
              'build/native/lib/v120/x64/Release/librdkafkacpp.lib', 'build/native/lib/win7/x64/win7-x64-Release/v120/librdkafkacpp.lib'],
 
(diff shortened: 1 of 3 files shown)


packaging/nuget/packaging.py
```diff
@@ -523,74 +298,80 @@ def xlayout (self):
...
+        mappings = [
+            [{'arch': 'x64', 'plat': 'linux', 'fname_startswith': 'librdkafka.tar.gz'}, './include/librdkafka/rdkafka.h', 'build/native/include/librdkafka/rdkafka.h'],
+            [{'arch': 'x64', 'plat': 'linux', 'fname_startswith': 'librdkafka.tar.gz'}, './include/librdkafka/rdkafkacpp.h', 'build/native/include/librdkafka/rdkafkacpp.h'],
+
+            [{'arch': 'x64', 'plat': 'osx', 'fname_startswith': 'librdkafka.tar.gz'}, './lib/librdkafka.dylib', 'runtimes/osx-x64/native/librdkafka.dylib'],
+            [{'arch': 'x64', 'plat': 'linux', 'fname_startswith': 'librdkafka-debian9.tgz'}, './lib/librdkafka.so.1', 'runtimes/linux-x64/native/debian9-librdkafka.so'],
+            [{'arch': 'x64', 'plat': 'linux', 'fname_startswith': 'librdkafka.tar.gz'}, './lib/librdkafka.so.1', 'runtimes/linux-x64/native/librdkafka.so'],
+
+            [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'msvcr120.zip'}, 'msvcr120.dll', 'runtimes/win7-x64/native/msvcr120.dll'],
+            [{'arch': 'x64', 'plat': 'win7', 'fname_startswith': 'librdkafka.redist'}, 'build/native/bin/v120/x64/Release/librdkafka.dll', 'runtimes/win7-x64/native/librdkafka.dll'],
```
edenhill: It would be good with a small comment what filename it expects when only specifying a prefix.
E.g.,
`# from librdkafka.redist.VER.nupkg`

For the one below saying just `librdkafka` in particular.

Alternatively if a glob name was used instead.

packaging/nuget/packaging.py
```diff
@@ -523,74 +298,80 @@ def xlayout (self):
...
+                for exclude in fname_excludes:
+                    if exclude in a.fname:
+                        found = False
+                        break
+
+                if found:
+                    artifact = a
+                    break
+
+            if artifact == None:
```
edenhill: Should typically be `.. is None`

packaging/nuget/packaging.py
```diff
@@ -523,74 +298,80 @@ def xlayout (self):
...
+                raise Exception('unable to find file in archive %s with tags %s that ends with "%s"' % (a.fname, str(attributes), fname_startswith))
```
edenhill: "that *starts* with"

packaging/nuget/packaging.py
```diff
@@ -523,74 +298,80 @@ def xlayout (self):
...
+
+            outf = os.path.join(self.stpath, m[2])
+            member = m[1]
+            try:
+                zfile.ZFile.extract(artifact.lpath, member, outf)
+            except KeyError as e:
+                raise Exception('file not found in archive %s: %s. Files in archive are: %s' % (artifact.lpath, e, zfile.ZFile(artifact.lpath).getnames()))
 
-    def verify (self, path):
```
edenhill: What's the reason for removing the verifier?

packaging/nuget/packaging.py
mhowlett: yep

packaging/nuget/packaging.py
mhowlett: well spotted, originally i had it as ending with

packaging/nuget/packaging.py
mhowlett: i thought it was superfluous - if an artifact isn't found during build, the program will exit.

packaging/nuget/packaging.py
mhowlett: yep, good idea. it's unfortunate the naming of the artifacts makes differentiating them tricky.

packaging/nuget/packaging.py
edenhill: It is the only test we have for the package, and it is already written, so I believe it should stay in.

mhowlett: ok, it's back in.

packaging/nuget/release.py
```diff
@@ -76,10 +76,8 @@
 
     print('')
 
-    if len(pkgfiles) > 0:
-        print('Created package:')
-        for pkg in pkgfiles:
-            print(pkg)
-    else:
-        print('Package not created.')
+    if not p.verify(pkgfile):
```
edenhill: Did you verify that verify works with the new package builds?

mhowlett: yes. note also that the package contents are different from before.

PR #1523, opened 2017-11-14
Offsets store err fix

edenhill:
For https://github.com/confluentinc/confluent-kafka-dotnet/pull/339

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -1718,7 +1718,10 @@ public:
    *
    * @remark \c enable.auto.offset.store must be set to \c false when using this API.
    *
-   * @returns RdKafka::ERR_NO_ERROR on success or an error code on error.
+   * @returns RdKafka::ERR_NO_ERROR on success, or
+   *          RdKafka::ERR___UNKNOWN_PARTITION if none of the offsets could
+   *          be stored, or
+   *          RdKafka::ERR___INVALID_ARG if \c enable.auto.offset.store is true.
    */
   virtual ErrorCode offsets_store (std::vector<TopicPartition*> &offsets) = 0;
 };
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -2357,7 +2357,7 @@ rd_kafka_resp_err_t rd_kafka_offset_store(rd_kafka_topic_t *rkt,
 
 
 /**
- * @brief Store offsets for one or more partitions.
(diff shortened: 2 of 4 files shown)


src-cpp/rdkafkacpp.h
```diff
@@ -1718,7 +1718,10 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
    *
    * @remark \c enable.auto.offset.store must be set to \c false when using this API.
    *
-   * @returns RdKafka::ERR_NO_ERROR on success or an error code on error.
+   * @returns RdKafka::ERR_NO_ERROR on success, or
+   *          RdKafka::ERR___UNKNONW_PARTITION if none of the offsets could
```
mhowlett: UNKNONW_PARTITION -> UNKNOWN_PARTITION

src/rdkafka.h
```diff
@@ -2367,8 +2367,10 @@ rd_kafka_resp_err_t rd_kafka_offset_store(rd_kafka_topic_t *rkt,
  *
  * @remark \c `enable.auto.offset.store` must be set to "false" when using this API.
  *
- * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success or an error code if
- *          none of the offsets could be stored.
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success, or
+ *          RD_KAFKA_RESP_ERR__UNKNONW_PARTITION if none of the
```
mhowlett: UNKNONW_PARTITION -> UNKNOWN_PARTITION

src-cpp/rdkafkacpp.h
edenhill: good spottage!

src/rdkafka_offset.c
```diff
@@ -715,7 +718,7 @@ rd_kafka_offsets_store (rd_kafka_t *rk,
                 ok_cnt++;
         }
 
-        return offsets->cnt > 0 && ok_cnt < offsets->cnt ?
+        return offsets->cnt > 0 && ok_cnt == 0 ?
```
mhowlett: The alternative is to remove this check altogether. I think I agree what you have is the best option.

mhowlett:
asside from spelling errors, LGTM.

src/rdkafka_offset.c
edenhill: This is sort of an optimization on the caller/user's behalf: if all partitions failed the user won't have to check the error code for each individual partition.
Questionable value, but it is there now.

PR #1535, opened 2017-11-20
Mention rdkafka-ruby in README

jurre:
I heard a colleague mention this lib but hard a hard time finding it, figured it would be nice to mention in the README.


edenhill:
Thank you!

PR #1592, opened 2017-12-12
use #if instead of #ifdef / defined() for atomics

vavrusa:
The #cmakedefine01 always defines the names e.g. HAVE_ATOMICS_32 with
either 0 or 1, but the code in https://github.com/edenhill/librdkafka/blob/master/src/rdatomic.h makes checks such as:

```c
#ifndef HAVE_ATOMICS_32
	mtx_t lock;
#endif
```

This will always compile support for atomics regardless of whether
it's supported or not.

### DIFF
--- a/src/rdatomic.h
+++ b/src/rdatomic.h
@@ -31,14 +31,14 @@
 
 typedef struct {
 	int32_t val;
-#ifndef HAVE_ATOMICS_32
+#if !HAVE_ATOMICS_32
 	mtx_t lock;
 #endif
 } rd_atomic32_t;
 
 typedef struct {
 	int64_t val;
-#ifndef HAVE_ATOMICS_64
+#if !HAVE_ATOMICS_64
 	mtx_t lock;
 #endif
 } rd_atomic64_t;
@@ -46,7 +46,7 @@ typedef struct {
 
 static RD_INLINE RD_UNUSED void rd_atomic32_init (rd_atomic32_t *ra, int32_t v) {
 	ra->val = v;
-#if !defined(_MSC_VER) && !defined(HAVE_ATOMICS_32)
+#if !defined(_MSC_VER) && !HAVE_ATOMICS_32
 	mtx_init(&ra->lock, mtx_plain);
 #endif
 }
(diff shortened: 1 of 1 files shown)


edenhill:
Good catch!

I think it might be better to change the #ifdefs to #if though, it is only the ATOMIC checks that do ifdef/defined():
```
$ git grep '#if.*HAVE'
packaging/cmake/config.h.in:#if (HAVE_ATOMICS_32)
packaging/cmake/config.h.in:#if (HAVE_ATOMICS_64)
src/rd.h:#if HAVE_STRNDUP
src/rdatomic.h:#ifndef HAVE_ATOMICS_32
src/rdatomic.h:#ifndef HAVE_ATOMICS_64
src/rdatomic.h:#if !defined(_MSC_VER) && !defined(HAVE_ATOMICS_32)
src/rdatomic.h:#if !defined(_MSC_VER) && !defined(HAVE_ATOMICS_64)
src/rdposix.h:#if HAVE_STRERROR_R
src/rdregex.c:#if HAVE_REGEX
src/rdregex.c:#if HAVE_REGEX
src/rdregex.c:#if HAVE_REGEX
src/rdregex.c:#if HAVE_REGEX
src/rdregex.c:#if HAVE_REGEX  /* use libc regex */
win32/wingetopt.h:#ifndef HAVE_DECL_GETOPT
```

vavrusa:
Makes sense, amended the PR.

src/rdatomic.h
```diff
@@ -31,22 +31,22 @@
 
 typedef struct {
 	int32_t val;
-#ifndef HAVE_ATOMICS_32
+#if HAVE_ATOMICS_32
```
edenhill: #if !HAVE..

src/rdatomic.h
```diff
@@ -31,22 +31,22 @@
...
 	mtx_t lock;
 #endif
 } rd_atomic32_t;
 
 typedef struct {
 	int64_t val;
-#ifndef HAVE_ATOMICS_64
+#if HAVE_ATOMICS_64
```
edenhill: #if !HAVE..

src/rdatomic.h
vavrusa: 😞 Fixed

vavrusa:
Do you want me to make more  changes or does it look okay? The CI build seems to have failed for unrelated reasons and I can't rerun it.

edenhill:
Thank you for this!

PR #1602, opened 2017-12-21
Allow build static library

proller:


src/CMakeLists.txt
```diff
@@ -78,7 +78,15 @@ if(NOT HAVE_REGEX)
   list(APPEND sources regexp.c)
 endif()
 
-add_library(rdkafka SHARED ${sources})
+option(RDKAFKA_BUILD_STATIC "Build static rdkafka library" OFF)
```
edenhill: Can you add instructions for how to build static libraries to packaging/cmake/README.md?

proller: now ok?

edenhill:
Thanks for this!

PR #1612, opened 2018-01-02
Add "partitioner" topic config property

edenhill:
This makes the builtin partitioners (including the new Java-compatible murmur2 partitioners) available to any application or language binding that exposes the string-based key=value configuration property interface.

E.g.:
 `p = confluent_kafka.Producer({.., 'default.topic.config': {'partitioner':'murmur2_random'}})`


CONFIGURATION.md
```diff
@@ -115,7 +115,8 @@ acks                                     |  P  |                 |
...
 produce.offset.report                    |  P  | true, false     |         false | Report offset of produced message back to application. The application must be use the `dr_msg_cb` to retrieve the offset from `rd_kafka_message_t.offset`. <br>*Type: boolean*
-partitioner_cb                           |  P  |                 |               | Partitioner callback (set with rd_kafka_topic_conf_set_partitioner_cb()) <br>*Type: pointer*
+partitioner                              |  P  |                 | consistent_random | Partitioner: `random` - random distribution, `consistent` - CRC32 hash of key (Empty and NULL keys are mapped to single partition), `consistent_random` - CRC32 hash of key (Empty and NULL keys are randomly partitioned), `murmur2` - Java Producer compatible Murmur2 hash of key (NULL keys are mapped to single partition), `murmur2_random` - Java Producer compatible Murmur2 hash of key (NULL keys are randomly partitioned). <br>*Type: string*
```
mhowlett: It would be good to explicitly note which partitioner is exactly the same as the java default (I believe that is `murmur2_random`, but I didn't check).

src/rdkafka_msg.c
```diff
@@ -543,33 +543,29 @@ int32_t rd_kafka_msg_partitioner_consistent_random (const rd_kafka_topic_t *rkt,
...
-								   key,
-								   keylen,
-								   partition_cnt,
-								   rkt_opaque,
-								   msg_opaque);
+                                                 const void *key, size_t keylen,
+                                                 int32_t partition_cnt,
+                                                 void *rkt_opaque,
+                                                 void *msg_opaque) {
+        if (!key)
```
mhowlett: minor: in `rd_kafka_msg_partitioner_consistent_random` the test is keylen == 0. I assume you organize for this to be the case in NULL case as well but you could consider explicitly doing a !key check there to make it easy to read.

mhowlett: but for this method, yep I verified this logic matches the java default partitioner.

src/rdmurmur2.c
```diff
@@ -30,31 +30,128 @@
  #include "rdunittest.h"
  #include "rdmurmur2.h"
 
-int unittest_murmurhashneutral2 (void) {
+/* MurmurHash2, by Austin Appleby
```
mhowlett: fine with all this assuming you're sure it all aligns with what java is doing.

tests/0048-partitioner.c
```diff
@@ -87,7 +88,200 @@ static void do_test_failed_partitioning (void) {
...
+#define _PART_CNT 17
+#define _MSG_CNT 5
+        const char *unaligned = "123456";
+        /* Message keys */
+        const char *keys[_MSG_CNT] = {
+                NULL,
+                "", // empty
+                unaligned+1,
+                "this is another string with more length to it perhaps",
+                "hejsan"
```
mhowlett: hello!

src/rdmurmur2.c
mhowlett: ok, I see you're testing that well enough below.

mhowlett:
LGTM, only minor comments.

CONFIGURATION.md
edenhill: Will do

src/rdkafka_msg.c
edenhill: This is a legacy mistake, the consistent_random partitioner treats NULL and empty keys the same way, but only NULL keys should be randomized.

I'm however reluctant to change this since it will change the behaviour of existing applications.

The murmur2_random partitioner follows the Java default partitioner and only checks for NULL.

PR #1621, opened 2018-01-04
Fix When rebalancing and there is uncommited msg, the consumer will s…

### LINKED ISSUE #1605: Consumer stops consuming if commit occurs during broker rebalancing
Description
===========
When the brokers are getting rebalanced, the consumer process throws an `Broker: Not coordinator for group` error and gets stuck and stops consuming.

How to reproduce
================
1. Setup a cluster of 2 brokers (Replication factor is 2).
2. Keep 1 broker up and 2nd broker down
3. Publish messages to a partition whose preferred leader is the 2nd broker which is down. (request.required.acks is 1)
4. Start a consumer. ("session.timeout.ms": 15000, "enable.auto.commit": false}. I am using synchronous commits 
5. Bring the 2nd broker up
6. Once rebalance is done, I immediately get an error `Broker: Not coordinator for group` and the consumer process gets stuck and doesn't get any further messages although requests are going to the broker

Broker logs during the error -

ERROR [ReplicaFetcherThread-0-3]: Error for partition [<topic>,<partition>] to broker 3:org.apache.kafka.common.errors.NotLeaderForPartitionException: This server is not the leader for that topic-partition. (kafka.server.ReplicaFetcherThread)

Checklist
=========

 - librdkafka version - 0.11.1
 - Apache Kafka version: 0.11.0.0
 - librdkafka client configuration:  "session.timeout.ms": 15000, "partition.assignment.strategy": "range", "enable.auto.commit": false, "auto.offset.reset": "earliest"
 - Operating system: Ubuntu 14.04

It looks like the message is read from broker 1 and committed to broker 1 but the read happens before rebalance and commit occurs after rebalance and hence the issue.

ChenyuanHu:
…top consume, Close #1605

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -1864,6 +1864,15 @@ static void rd_kafka_cgrp_op_handle_OffsetCommit (rd_kafka_t *rk,
 	rd_kafka_assert(NULL, rkcg->rkcg_wait_commit_cnt > 0);
 	rkcg->rkcg_wait_commit_cnt--;
 
+        if (err == RD_KAFKA_RESP_ERR_NO_ERROR) {
+                rd_kafka_dbg(rk, CGRP, "COMMIT", "All commited, call fetch_start.");
+                if (rkcg->rkcg_wait_commit_cnt == 0 &&
+                    rkcg->rkcg_assignment &&
+                    RD_KAFKA_CGRP_CAN_FETCH_START(rkcg))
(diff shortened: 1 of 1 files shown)


ChenyuanHu:
@edenhill Please review.
https://github.com/edenhill/librdkafka/issues/1605

andreycha:
Hi @edenhill, is this request going to be checked and reviewed any time soon?

edenhill:
Yes, thank you for this, we'll tend to it shortly!

edenhill:
Thank you for this!

PR #1637, opened 2018-01-16
Fix build error on Windows

raulbocanegra:
Fix C2491: definition of dllimport static data member not allowed


edenhill:
Thanks for this!

PR #1638, opened 2018-01-16
cmake build improve on windows

raulbocanegra:
In Windows, the option WITH_PLUGINS is set to OFF, but in src/win32_config.h you can find #define WITH_PLUGINS 1.
This cause link errors when you link against librdkafka. So you have to manually set #define WITH_PLUGINS 0 for the cmake build to work properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,9 @@ endif()
 
 option(RDKAFKA_BUILD_EXAMPLES "Build examples" ON)
 option(RDKAFKA_BUILD_TESTS "Build tests" ON)
-
+if(WIN32)
+    option(WITHOUT_WIN32_CONFIG "Avoid including win32_config.h on cmake builds" ON)
+endif(WIN32)
 
 # In:
 # * TRYCOMPILE_SRC_DIR
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,20 +1,30 @@
-add_executable(rdkafka_example rdkafka_example.c)
-target_link_libraries(rdkafka_example PUBLIC rdkafka)
+if(WIN32)
+    set(win32_sources ../win32/wingetopt.c ../win32/wingetopt.h)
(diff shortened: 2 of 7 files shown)


raulbocanegra:
I'm not sure how this can be fixed to make cmake builds and regular Windows builds both to work fine.
Probably we could make a configured cmake header, but not sure how to do this without breaking Windows build without cmake

edenhill:
WITH_PLUGINS should be enabled on Windows, it is using LoadLibrary which is available in the standard library.

raulbocanegra:
And should WITH_PLUGINS also be enabled when building static version ol librdkafka? I don't know what plugins are used for.

edenhill:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L3650

raulbocanegra:
Thanks!
So, if I'm not going to use any plugin I can safely turn this off, right?

edenhill:
Yep

edenhill:
Could you have CMake emit a command line define for "WITH_PLUGINS=0" when building static library and then add an `#ifndef WITH_PLUGINS` guard to the define in win32_config.h?

rbocanegra:
Yep I will try to do it during the day.
I have been trying to compile with cmake with -DWITH_PLUGINS=ON on Windows and it has som link errors.
error LNK2001: unresolved external symbol rd_dl_open
error LNK2001: unresolved external symbol rd_dl_close
error LNK2001: unresolved external symbol rd_dl_sym

I will try your last suggestion to see if that works

raulbocanegra:
I think it is done.
The link errors I reported in my previous message:
error LNK2001: unresolved external symbol rd_dl_open
error LNK2001: unresolved external symbol rd_dl_close
error LNK2001: unresolved external symbol rd_dl_sym
were due to the fact that rddl.c was not added to cmake project on Windows.
I have also make some other changes to compile shared and statically on Windows. 
Feel free to ask if you need any clarification.

edenhill:
Looks good!

Is this ready to be merged?

raulbocanegra:
yep

Issue #1310, opened 2017-07-07
API to set custom memory allocator in librdkafka


roopabayar:
Description
===========
I could not locate an API that could be used to set custom memory alloc/free function in librdkafka.
How do we achieve it?

How to reproduce
================

Checklist
=========
Please provide the following information:

 - [ ] librdkafka version (release number or git tag):
4da1061ad1d993f446e54356b9dbc82f4ef52583
 - [ x] Apache Kafka version: 
0.10.2
 - [ ] librdkafka client configuration:
 - [ ] Operating system:
Linux/Ubuntu
 - [ ] Using the legacy Consumer
 - [ ] Using the high-level KafkaConsumer
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
Can you explain how you are planning to use this?

Do you want to use separate allocator instances for each rd_kafka_t instance?

Do you want the allocator to be used for configuration settings as well (rd_kafka_conf_t ..)?

roopabayar:
If i could register my malloc and free function with the library, i will
have better visibility to the memory used. This would work for any of the
objects.

For ex: I am calling the rd_kafka_produce function for which i am passing
the buffer pointer and asking library to free it by setting flags
RD_KAFKA_MSG_F_FREE. My memory allocator is accounting the allocations, but
is not able to account for the free as the free is done using RD_FREE
(free).


Thanks
Roopa

eugpermar:
Hi, @roopabayar ! I have a similar use case (many messages in the same memory chunk buffer, with reference counter). I use the delivery report callback to reduce reference counter & free.

You could add bytes to your counter just after the successfully produce(), and subtract it in delivery report callback. What you can't account this way is librdkafka internal allocations, but they should be insignificant against a lot of messages memory.

PR #1645, opened 2018-01-22
Create a new rd_kafka_conf_set_free_payload_cb API to set the callbac…

4alexey:
Create a new rd_kafka_conf_set_free_payload_cb API to set the callback to free memory pointer passed to rd_kafka_produce API when RD_KAFKA_MSG_F_FREE flag is set. This is a common scenario when a custom application memory management is used. For performance-critical applications passing RD_KAFKA_MSG_F_FREE will eliminate extra malloc and memcpy calls to allocate and copy message payload memory. Librdkafka currently calls free(payload) to free memory in rd_kafka_msg_destroy when RD_KAFKA_MSG_F_FREE is set. This will result in undefined behavior when a custom memory allocation is used in the application and the payload pointer memory is not allocated with calloc, malloc, or realloc. To enable this scenario following code changes are proposed:

1. Create a new  rd_kafka_conf_set_free_payload_cb to set a free payload callback intended to call  custom memory deallocation application function.
2. Create a new RD_KAFKA_OP_PAYLOAD_FREE operation.
3. In rd_kafka_msg_destroy enqueue  RD_KAFKA_OP_PAYLOAD_FREE operation to the tail of the 'rkq' queue with RD_KAFKA_PRIO_MEDIUM priority. This will result in d_kafka_poll_cb call.
4. In rd_kafka_poll_cb free the payload by calling the application callback ( rk→rk_conf.free_payload_cb).

Only C code is currently implemented. C++ changes are proposed.

---
rdkafka_op.h
rdkafka_msg.c
rdkafka_event.c
rdkafka_conf.h
rdkafka_conf.c
rdkafka.h
rdkafka.c

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -1409,6 +1409,9 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *app_conf,
 		rk->rk_conf.enabled_events |= RD_KAFKA_EVENT_REBALANCE;
 	if (rk->rk_conf.offset_commit_cb)
 		rk->rk_conf.enabled_events |= RD_KAFKA_EVENT_OFFSET_COMMIT;
+	if (rk->rk_conf.free_payload_cb)
+		rk->rk_conf.enabled_events |= RD_KAFKA_EVENT_FREE_PAYLOAD;
+
 
 	/* Convenience Kafka protocol null bytes */
 	rk->rk_null_bytes = rd_kafkap_bytes_new(NULL, 0);
@@ -2720,6 +2723,11 @@ rd_kafka_poll_cb (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko,
         case RD_KAFKA_OP_TERMINATE:
                 /* nop: just a wake-up */
                 break;
+
+        case RD_KAFKA_OP_PAYLOAD_FREE:
+            if (likely(rk->rk_conf.free_payload_cb != NULL))
+                    rk->rk_conf.free_payload_cb(rk, rko->rko_u.payload_free.payload);
+            break;
 
         default:
                 rd_kafka_assert(rk, !*"cant handle op type");
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -1329,6 +1329,11 @@ void rd_kafka_conf_set_dr_msg_cb(rd_kafka_conf_t *conf,
                                                      rkmessage,
                                                      void *opaque));
 
+RD_EXPORT
+void rd_kafka_conf_set_free_payload_cb (rd_kafka_conf_t *conf,
+			      void (*free_payload_cb) (rd_kafka_t *rk,
+					     void *payload));
+
 
 /**
  * @brief \b Consumer: Set consume callback for use with rd_kafka_consumer_poll()
@@ -3448,6 +3453,7 @@ typedef int rd_kafka_event_type_t;
 #define RD_KAFKA_EVENT_REBALANCE     0x10 /**< Group rebalance (consumer) */
 #define RD_KAFKA_EVENT_OFFSET_COMMIT 0x20 /**< Offset commit result */
 #define RD_KAFKA_EVENT_STATS         0x40 /**< Stats */
+#define RD_KAFKA_EVENT_FREE_PAYLOAD  0x50 /**< Free message payload when RD_KAFKA_MSG_F_FREE is used */
 
 
 typedef struct rd_kafka_op_s rd_kafka_event_t;
--- a/src/rdkafka_conf.c
+++ b/src/rdkafka_conf.c
@@ -761,6 +761,9 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  _RK(dr_err_only),
 	  "Only provide delivery reports for failed messages.",
 	  0, 1, 0 },
+	{ _RK_GLOBAL|_RK_PRODUCER, "free_payload_cb", _RK_C_PTR,
+	      _RK(free_payload_cb),
+	      "Free payload callback (set with rd_kafka_conf_set_free_payload_cb())" },
 	{ _RK_GLOBAL|_RK_PRODUCER, "dr_cb", _RK_C_PTR,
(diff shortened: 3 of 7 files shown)


edenhill:
Thanks for your contirbution.

This API makes sense but I believe it is already covered by the functionality of the delivery report callback which is not only meant for delivery signaling but also to allow the application to free its own payload.

4alexey:
Hi,

The delivery report callback could have been used to call the custom memory management deallocator.  The problem is with rd_kafka_msg_destroy function though, when RD_KAFKA_MSG_F_FREE flag is set, as it calls rd_free(rkm->rkm_payload). This results in a crush when the rkm_payload pointer is not allocated with calloc, malloc, or realloc, but points to the address pre-allocated with the custom memory allocator. 
The goal of using RD_KAFKA_MSG_F_FREE flag is to avoid extra malloc and memcpy calls for the payload, as it's already stored on the custom memory management pool.

Another way to solve this is by calling rd_free(rkm->rkm_payload) conditionally, but other changes in the rdkafka_conf.h are need then to signal the custom memory allocator is used. Since callbacks are the common way in the library design, I implemented it using the callback. 

The issue has already been discussed in https://github.com/edenhill/librdkafka/issues/1310 before.

void rd_kafka_msg_destroy (rd_kafka_t *rk, rd_kafka_msg_t *rkm) {
...
if (rkm->rkm_flags & RD_KAFKA_MSG_F_FREE && rkm->rkm_payload)
	rd_free(rkm->rkm_payload);
...
}

edenhill:
So what you do is not to pass FREE or COPY to produce(), that way it will not copy your payload and it relies on you freeing it in your delivery report callback.

4alexey:
Hi, thanks for the tip. When neither FREE nor COPY is passed the payload memory is not allocated nor freed indeed, and it is possible to free the payload in the delivery report callback. When compression is used, rd_kafka_msgset_writer_compress_snappy allocates memory in the code below. Is original payload memory used to copy over compressed message over, or it can be freed after message is compressed? In that case there is no need to hold on the original payload, as it may take a while to deliver the message.

        /* Calculate maximum compressed size and
         * allocate an output buffer accordingly. */
        ciov->iov_len = rd_kafka_snappy_max_compressed_length(len);
        ciov->iov_base = rd_malloc(ciov->iov_len);

edenhill:
The payload memory is never modified, so when compression is enabled a new internal buffer is allocated.

4alexey:
In that case the free_payload_cb is useful, as it makes sense to free payload memory as soon as possible. I'll look at how to implement this as we need to improve the performance and memory utilization for my current project work.
I'd propose one more feature: a callback to allocate memory using a custom memory allocator. I'd use it in the above rd_malloc call when memory is allocated for compressed payload buffer.

edenhill:
We do appreciate your effort but I don't think a free_payload_cb is really warranted since in the typical case it will differ in allocation lifetime by approximately the round-trip-time of the broker (e.g., a couple of milliseconds).

As for custom allocator, there is a previous issue #1310 to track that feature request.

4alexey:
I appreciate your response. I'll benchmark the performance of our application using different implementations: (a) by freeing memory in the delivery report callback, (b) freeing memory  in the free_payload_callback after compression happens, and (c) by overriding rd_malloc(ciov->iov_len) for the compressed memory with our custom memory allocator. Possibly the perf gain, if any, doesn't worth the effort.

4alexey:
When passing 0 flag value to rd_kafka_produce and freeing the memory in the delivery report callback, we occasionally see the library crashing. Attaching the core dump stack trace. Any idea why this is happening?

(gdb) where
#0  rd_crc32_update (rkbuf=0x7f2694804650, buf=<value optimized out>,
    len=<value optimized out>, allow_crc_calc=1, free_cb=<value optimized out>)
    at rdcrc32.h:108
#1  rd_kafka_buf_push0 (rkbuf=0x7f2694804650, buf=<value optimized out>,
    len=<value optimized out>, allow_crc_calc=1, free_cb=<value optimized out>)
    at rdkafka_buf.c:87
#2  0x00000000005ba3fe in rd_kafka_msgset_writer_write_msg_payload (
    msetw=0x7f2675ff32a0, rkm=0x7f2688000be0, Offset=<value optimized out>,
    MsgAttributes=0 '\000', free_cb=0) at rdkafka_msgset_writer.c:440
#3  rd_kafka_msgset_writer_write_msg_v0_1 (msetw=0x7f2675ff32a0,
    rkm=0x7f2688000be0, Offset=<value optimized out>, MsgAttributes=0 '\000',
    free_cb=0) at rdkafka_msgset_writer.c:502
#4  0x00000000005b9dae in rd_kafka_msgset_writer_write_msg (
    msetw=0x7f2675ff32a0, rkm=0x7f2688000be0, Offset=<value optimized out>,
    MsgAttributes=<value optimized out>, free_cb=<value optimized out>)
    at rdkafka_msgset_writer.c:635
#5  0x00000000005bad51 in rd_kafka_msgset_writer_write_msgq (
    rkb=<value optimized out>, rktp=0x7f26a0001a90,
    MessageSetSizep=0x7f2675ff3398) at rdkafka_msgset_writer.c:706
#6  rd_kafka_msgset_create_ProduceRequest (rkb=<value optimized out>,
    rktp=0x7f26a0001a90, MessageSetSizep=0x7f2675ff3398)
    at rdkafka_msgset_writer.c:1158
#7  0x000000000059391a in rd_kafka_ProduceRequest (rkb=0x7f2688002b70,

edenhill:
Try to reproduce with valgrind or asan, it will tell you what the problem is

4alexey:
I could not reproduce the exact problem in the valgrind. However, by our application design, the threads where kafka handle and topic were created, get killed occasionally, and the new thread is created, instantiating another kafka handle and topic. When the thread is killed, the kafka handle/topic do was not cleaned up with rd_kafka_topic_destroy and rd_kafka_destroy. Could this cause the crash represented above i.e. not cleared instance of kafka hande/topic cause a crash in the newly created instances.

edenhill:
Killing threads, leaking resources and leaving kafka client instances running in the background, is obviously neither recommended or supported.

PR #1646, opened 2018-01-23
Fixes doc for the acks property to refer to the correct broker property

solsson:
https://kafka.apache.org/documentation/#brokerconfigs


edenhill:
CONFIGURATION.md is automatically generated based on rdkafka_conf.c, that's where you need to change it, rebuild (to regen CONFIGURATION.md), and file a PR containing changes to both those files.

edenhill:
fixing this in separate commit. Thanks!

PR #1659, opened 2018-01-30
Connection on demand to brokers

gduranceau:
Define a new global property connect.on.demand. When set, connections to
brokers are not established immediately, but delayed until they are
about to be used. Another property inactivity.timeout.ms controls the
closure of connections to brokers, after they have been inactive for a
while.

This partly addresses:
https://github.com/edenhill/librdkafka/issues/825


gduranceau:
Hi @edenhill ,

I implemented connection on demand scheme in librdkafka, following the discussion on https://github.com/edenhill/librdkafka/issues/825

The patch adds 2 new configuration settings to librdkafka:
-   connect.on.demand (can be 0 or 1) to activate connection on demand (default=0)
-   inactivity.timeout.ms (default = 1 min), after which inactive sockets to broker are closed.

When using connection on demand, a single connection to one of the bootstrap nodes is established initially when a metadata request has to be sent. This connection is closed after the inactivity timeout if this broker isn’t used anymore afterwards. New connections to targeted leader brokers are opened when required (and closed in case of inactivity).

I didn't change the threading model. This patch only limits the number of connections opened to Kafka brokers.                 

Implementation details:

- When connection on demand is set, brokers are initially put in STATE_UP state (rather than STATE_INIT), so as to be selected by calls to rd_kafka_broker_any (state=STATE_UP). Elements can immediately be queued for any broker.
- The status for the connection (up/being-established or down) is identified by the rkb_transport attribute of the broker.
- If rd_kafka_broker_serve identifies that the connection is down, but elements are queued, it triggers connection establishment (by switching to STATE_DOWN).
- When any usable broker is needed (for metadata request for example), rd_kafka_broker_any_usable (and rd_kafka_cgrp_coord_query) gives priority to connected brokers. If there isn't any, it triggers connection for one of them, which should eventually become available. If this connection fails, another broker will eventually be selected.
- Activity on connection is monitored in rd_kafka_transport_io_serve, when polling events are retrieved. Inactivity is checked in the broker producer_serve and consumer_server loops.

I validated the patch in a number of producer/consumer scenarios (mostly with kafkacat), but of course, it's possible that I missed things. Any comment is welcome.

Cheers  -  Guillaume

edenhill:
This is great, will review in a couple of days!

The long term plan is to make thread creation on-demand, and maybe worker-thread-pooled, but this is a great medium term solution for what is essentially the real problems - the number of connections. The number of threads is in most cases an aesthetic.

Do you think we could rely on the broker's idle connection reaper instead of our own? This would make it this even simpler:
 * connect to a broker when we need to
 * don't automatically reconnect unless we need to

gduranceau:
Hi Magnus,

> The long term plan is to make thread creation on-demand, and maybe worker-thread-pooled, but this is a great medium term solution for what is essentially the real problems - the number of connections. The number of threads is in most cases an aesthetic.

I fully agree. I initially thought about implementing thread creation on demand when I started this, but I quickly gave up as it looked like impact was much bigger.

> Do you think we could rely on the broker's idle connection reaper instead of our own? This would make it this even simpler

I didn't think at all about relying on the broker closing idle connections. This is a good point.

I removed today the `inactivity.timeout.ms` property that I added, and also don't reconnect anymore to the broker in `rd_kafka_broker_fail` in case connection on demand is activated (and previous state was UP).

I'm still not entirely satisfied with this whole change though. It works well for producers, but for some reason, consumers still open at least 2 bootstrap connections (which are closed after a while if idle). I'll try to understand this tomorrow.

There's  one other annoying thing when consuming in a consumer group. Leader brokers from revoked partitions never get disconnected. It looks like the broker never consider those connections as idle, although librdkafka logs don't seem to show any sort of activity on those connections. This is something that I will look up as well.

Cheers  -  Guillaume

gduranceau:
Current state on this pull request after the last commit today:
Connection on demand is working pretty well, but connection handling is still not optimum when dealing with consumer groups.

gduranceau:
I fixed connection on demand when used with consumers in a group. With the latest change, the connection to a random broker to query for the coordinator is sent immediately (no more idle time of 1 second), and the connection to the coordinator once known is also established immediately (in both cases, by making use of `rd_kafka_broker_wakeup`).

Note that I squashed all the commits in the pull request.

This is now in a state where I have no pending change. Feel free to comment and ask for questions or clarifications.

Cheers  -  Guillaume

edenhill:
Great work, Guillaume!
I'll review this shortly.
Unfortunately this will not make the 0.11.4 release (end of february), since it is a drastic change that will require a lot of test time, but we'll schedule it for the release after thta.

edenhill:
Sorry for the delay, been busy driving the v0.11.4 client releases but that is almost out of the way now.

After the current work on the Topic Admin API (KIP-4) is done I'll start focusing on Idempotent Producer, followed by EoS.
EoS is a big task that will require a lot of testing, so I'm thinking it will make sense to incorporate this PR in that development branch to make sure this PR gets thorough testing.
I'm suspecting there are some corner cases with regards to the current assumption of brokers always being available that needs to be vetted out, and the additional testing load of EoS will give this enough runtime to find any regressions from the connection changes.

gduranceau:
For information. there is a problem / limitation that I identified with this implementation of connection on demand when producing data, which can lead to delivery failure for few messages when a broker comes down.

Connection to a leader broker is established only when the producer queues messages in the broker queue. Messages queued just before a leader broker shuts down and which had no time to be delivered trigger regular re-connection attempts (until delivery timeout), which all fail. Even though a new leader is elected in the meantime, those queued messages are not re-assigned to the new leader queue. As a consequence, they are reported to the client as undelivered after the delivery timeout.

It would be much better to be able to re-push those messages to the new elected leader queue, but I have not found out how to do that in a simple way yet.

billygout:
@gduranceau
> It would be much better to be able to re-push those messages to the new elected leader queue, but I have not found out how to do that in a simple way yet.

I wonder whether that is a result of the on-demand connections, or the same would have happened without on-demand connections? In other words, is this really a regression your patch is introducing?

gduranceau:
Maybe the problem is possible without my changes, but connection on demand makes it worse in any case.

When the connection is broken, without connection on demand, this immediately sets the broker state to DOWN, preventing any message to be queued to this broker. With connection on demand, the state is actually reset to UP, precisely so that messages can still be queued to this broker, and allow a re-connection when required (note that a connection can be reset even if the remote broker is up, when it has been idle for a while. In this case, it's important that the state is set to UP indeed).

When a broker fails with connection on demand, all messages queued while the new leader isn't elected (and this new leader is unknown to the client) are queued to the failed broker, and not delivered.

> So in theory this should work even with your modifications, given that the cluster elects a new leader in a timely manner.

I have to look this up again (I've run those tests a couple of weeks ago), but with a steady flow of messages produced, some are systematically not delivered as they remain in the failed broker queue. When the delivery timeout finally expires, the re-connection attempts are stopped (this was the point of commit https://github.com/edenhill/librdkafka/pull/1659/commits/d098e94dc32c6908136c8ff60f405f746faa265c).

gduranceau:
> The LEAVE causes any messages queued on the broker's xmitq to be put back on the partition queue:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_broker.c#L2159

Yes, this is ok for messages in the xmitq queue. But here, I end up with a MessageSet already generated and queued in the rkb_outbufs queue of the failed broker (this is actually the condition which allows to trigger a new connection). This one isn't moved on partition LEAVE. Ideally, it should be pushed in the rkb_outbufs queue (or maybe rkb_retrybufs) of the new leader (on partition JOIN).

edenhill:
@gduranceau Could you try adding RD_KAFKAP_Produce to this switch?
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_buf.c#L266

I'm not sure what error code will be the most appropriate, but probably ERR__TRANSPORT.

gduranceau:
@edenhill It looks like it's working with your suggestion. I've just ran a few quick tests. I'll do a proper validation soon, and update the pull request.

billygout:
Hi @gduranceau and @edenhill : I am still very interested in this feature, and would like to inquire what remains to be done to merge this PR? If no more changes are expected and this is being help up for lack of testing, I can offer to test this on a large-scale kafka deployment. Thanks!

gduranceau:
Hi @billygout ,

Magnus and I had a discussion a few weeks ago about this. To make sure this is properly tested, this will be integrated in the branch to be created soon for Exactly-once Semantic developments. Integration will come when this branch will be merged into master (likely not before end of Q3 2018).

In the meantime, this patch will make it way to production applications in my company in the coming weeks/months (several dozen of thousands of processes, mostly producer oriented). I'll report any issue (or lack of issue) here.

billygout:
@gduranceau Thank you for the detailed information! Q3 of 2018 is a bit later than I'd like to have this, but beggars can't be choosers :) I see there are merge conflicts, but do we have any reasonably up-to-date branch that this has been applied to that I can try this out on, even if I'm willing to void my warranty? thanks!

gduranceau:
@billygout 
I rebased the patch on master.

billygout:
Thanks so much @gduranceau !

billygout:
@gduranceau We've been running on thousands of producers using the previously rebased branch you had, at commit ed23da66ed21dac4787f50d73afaaee531e97889
So far, we've experienced no new problems that we've seen. Slight memory usage drop. No cpu changes. No reliability changes. Just major drop in connections. We also haven't noticed any regression in terms of latency which you might expect due to the need to connect before sending to a broker, but we are looking at latency on the order of seconds, so we probably wouldn't have noticed this unless it became eggregious.

gduranceau:
@billygout Thanks for sharing your (positive) experience with this patch. On my side, it has also been integrated lately in our producers (actually, a backport for branch 0.9), but it's not yet sufficiently widely deployed to really comment on it.

I'm trying to rebase the patch regularly on the master branch. When 0.11.5 will be out, I'll provide a tag in my repo for 0.11.5+this patch.

billygout:
@gduranceau can you please provide an update on the progress of your testing. I'm enjoying your code in my own fork for now, but would love to get into the mainline :)

gduranceau:
@billygout So far, everything is good. No problem reported (producer clients only).

For info, I created the branch [v0.11.5_connect_on_demand](https://github.com/gduranceau/librdkafka/commits/v0.11.5_connect_on_demand) in my repo with just this patch on top of v0.11.5 tag.

jonefeewang:
hi guys, this feature is really nice. We have a large amount of clients, probably 10k-80k , but not all of them produce message all the time. On demand connection will really decrease many "idle" connection to kafka broker. Is this feature released ?

billygout:
Yes, it's on by default as of librdkafka 1.0.0

PR #1667, opened 2018-02-02
Fix broker wake up

gduranceau:
A WAKEUP operation in a broker queue didn't trigger immediate wakeup (`rd_kafka_q_pop` didn't return on WAKEUP). This fix makes sure that the broker immediately stops waiting on its operation queue in case of wake-up, instead of waiting the timeout in `rd_kafka_broker_ops_serve`.

I noticed this problem while working on https://github.com/edenhill/librdkafka/pull/1659.


edenhill:
Good stuff!

edenhill:
Thank you for this!

PR #1675, opened 2018-02-07
metadata: Expose broker rack information from metadata request.

spasam:
Currently rack information is read (if available) and discarded.
This change preserves that and makes it available to the caller.

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3106,6 +3106,7 @@ typedef struct rd_kafka_metadata_broker {
         int32_t     id;             /**< Broker Id */
         char       *host;           /**< Broker hostname */
         int         port;           /**< Broker listening port */
+        char       *rack;           /**< Broker rack (if configured) */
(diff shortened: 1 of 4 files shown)


src/rdkafka.h
```diff
@@ -3106,6 +3106,7 @@ typedef struct rd_kafka_metadata_broker {
         int32_t     id;             /**< Broker Id */
         char       *host;           /**< Broker hostname */
         int         port;           /**< Broker listening port */
+        char       *rack;           /**< Broker rack (if configured) */
```
edenhill: Public types can't be modified since that would break the ABI guarantees of the library and break existing applications.
It was a mistake to make the metadata API use public types..

We're looking into an alternative metadata API with private types with accessor methods to be able to evolve the interface along with the broker API.

spasam: 😶 Thanks for the info.

PR #1688, opened 2018-02-14
replace use of #pragma once with header guards

tbsaunde:
Being a pragma #pragma once is non standard, and in practice the Sun
Studio compiler does not support it.  We can fix that by just using
standard #ifndef / #define / #endif header guards.  Since the
rdkafka_subscription.h header is empty other than the license text it is
simply removed.


CONFIGURATION.md
```diff
@@ -3,7 +3,7 @@
 
 Property                                 | C/P | Range           |       Default | Description              
 -----------------------------------------|-----|-----------------|--------------:|--------------------------
-builtin.features                         |  *  |                 | gzip, snappy, ssl, sasl, regex, lz4, sasl_gssapi, sasl_plain, sasl_scram, plugins | Indicates the builtin features for this build of librdkafka. An application can either query this value or attempt to set it with its list of required features to check for library support. <br>*Type: CSV flags*
```
edenhill: remove this file from the PR.

mklove/modules/configure.base
```diff
@@ -521,10 +521,12 @@ function mkl_generate {
 
     # Generate config.h
     mkl_write_h "// Automatically generated by $0 $*"
-    mkl_write_h "#pragma once"
+    mkl_write_h "#ifndef _CONFIG_H_"
+    mkl_write_h "#define _CONFIG_H_ 1"
```
edenhill: No need for defining it to 1, just an empty define is more standard.
This goes for all of these

src/crc32c.h
```diff
@@ -26,10 +26,13 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#pragma once
+#ifndef _CRC32C_H_
```
edenhill: prefix non rd-prefixed files with `_RD_<FILENAME>_H_` to avoid collision with system / 3rd-party headers.

src/rdatomic.h
```diff
@@ -25,7 +25,8 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-#pragma once
+#ifndef _RD_ATOMIC_H_
```
edenhill: to be consistent remove second underscore: `_RDATOMIC_H_`
This goes for the ones below as well.

tests/sockem.h
```diff
@@ -26,7 +26,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#pragma once
+#ifndef _TEST_SOCK_EM_H_
+#define _TEST_SOCK_EM_H_ 1
```
edenhill: `_SOCKEM_H_`

tests/testcpp.h
```diff
@@ -25,7 +25,8 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-#pragma once
+#ifndef _TEST_CPP_H_
```
edenhill: `_TESTCPP_H_`

tests/testshared.h
```diff
@@ -25,7 +25,8 @@
 * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 * POSSIBILITY OF SUCH DAMAGE.
 */
-#pragma once
+#ifndef _TEST_SHARED_H_
```
edenhill: `_TEST_SHARED_H_`

CONFIGURATION.md
tbsaunde: sorry, that was obviously a mistake, the way it gets regenerated by the build system and can get added to a commit with something like git commit -a is not great.

mklove/modules/configure.base
tbsaunde: I feel like I commonly see both, but it doesn't matter to me.

edenhill: Just this one (pun!) seems to be left as defining to 1

src/rdavl.h
```diff
@@ -32,7 +32,8 @@
  * Inspired by Ian Piumarta's tree.h implementation.
  */
 
-#pragma once
+#ifndef _RDAVL_H_
+#define _RDAVL_H_ 1
```
edenhill: and this one

src/rdinterval.h
```diff
@@ -1,6 +1,5 @@
-#pragma once
-
-
+#ifndef _RDINTERNAL_H_
```
edenhill: interval, not internal

tests/interceptor_test/interceptor_test.h
```diff
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef _INTERCEPTOR_TEST_H_
+#define _INTERCEPTOR_TEST_H_ 1
```
edenhill: one

edenhill:
LGTM.
Can you fix the merge conflicts and then I can merge it

PR #1708, opened 2018-02-27
stop using typeof in rdkafka_transport.c

tbsaunde:
The typeof operator is non standard, and at least XLC doesn't support
it.  While the type of msghdr::msg_iovlen is a bit of a mess the largest
value we will ever want to store in it is IOV_MAX. IOV_MAX should be
much smaller than INT_MAX, and in both cases the array whose length we
are providing is on the stack, so it should easily fit in an int.  We
know that the length should be non negative, so it should be safe to
cast to int, and then allow C promotion rules to widen to the actual
type of msg_iovlen if larger.


edenhill:
Looks good, thanks!

PR #1711, opened 2018-03-01
Win32 fix

mattclarke:


src/win32_config.h
```diff
@@ -31,10 +31,12 @@
  */
 #pragma once
 
+#ifndef WITHOUT_WIN32_CONFIG
```
edenhill: Maybe it is better to do `#ifndef WITH_SSL` and so on, that would allow non-CMake users to also override these settings on Win32.

src/rdkafka.h
```diff
@@ -137,7 +137,7 @@ typedef SSIZE_T ssize_t;
  * @remark This value should only be used during compile time,
  *         for runtime checks of version use rd_kafka_version()
  */
-#define RD_KAFKA_VERSION  0x000b01ff
+#define RD_KAFKA_VERSION  0x000b03ff
```
edenhill: remove from PR

src-cpp/rdkafkacpp.h
```diff
@@ -91,7 +91,7 @@ namespace RdKafka {
  * @remark This value should only be used during compile time,
  *         for runtime checks of version use RdKafka::version()
  */
-#define RD_KAFKA_VERSION  0x000b01ff
```
edenhill: remove from PR

PR #1715, opened 2018-03-07
Small change to proprogate the WITHOUT_WIN32_CONFIG flag to the build

mattclarke:
I think this small change is needed to enable overriding of the settings in src/win32_config.h.

Perhaps @raulbocanegra could check this?


raulbocanegra:
You are right. I should have added this on my last PR. Thanks! :-)

edenhill:
Are you guys happy with this?

raulbocanegra:
Yep

shilf:
What does WITHOUT_WIN32_CONFIG do?Is WITHOUT_WIN32_CONFIG necessary?

mattclarke:
It allows one to override/replace the settings in win32_config.h without editing that file directly.

If you are happy with the defaults in win32_config.h then it can be ignored.

PR #1718, opened 2018-03-09
Avoid allocating BIOs and copying for base64 processing.

agl:
OpenSSL offers a more direct API for base64 encoding and decoding.
Previously the code created a couple of BIOs and used the streaming
interface, this results in unneccessary allocations.

Also, we've received a request to integrate this library with TensorFlow
and the base64 BIO interface is one of the things we're trying to trim
from our version of OpenSSL.

Note: if there's an easy alternative to using OpenSSL for base64
processing then I would recommand that. OpenSSL has some odd behaviours
around ignoring whitespace etc. While they don't immediately appear to
matter here, something simplier would be preferable.

### DIFF
--- a/src/rdkafka_sasl_scram.c
+++ b/src/rdkafka_sasl_scram.c
@@ -140,66 +140,69 @@ static char *rd_kafka_sasl_scram_get_attr (const rd_chariov_t *inbuf, char attr,
 
 /**
  * @brief Base64 encode binary input \p in
- * @returns a newly allocated base64 string
+ * @returns a newly allocated, base64-encoded string or NULL on error.
  */
 static char *rd_base64_encode (const rd_chariov_t *in) {
-        BIO *buf, *b64f;
-        BUF_MEM *ptr;
-        char *out;
+        char *ret;
+        size_t ret_len, max_len;
 
-        b64f = BIO_new(BIO_f_base64());
-        buf = BIO_new(BIO_s_mem());
-        buf = BIO_push(b64f, buf);
+        /* OpenSSL takes an |int| argument so the input cannot exceed that. */
+        if (in->size > INT_MAX) {
+                return NULL;
+        }
 
-        BIO_set_flags(buf, BIO_FLAGS_BASE64_NO_NL);
-        BIO_set_close(buf, BIO_CLOSE);
-        BIO_write(buf, in->ptr, (int)in->size);
-        BIO_flush(buf);
+        /* This does not overflow given the |INT_MAX| bound, above. */
+        max_len = (((in->size + 2) / 3) * 4) + 1;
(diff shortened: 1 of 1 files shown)


agl:
Please note: I was unable to get the SASL tests to run (before or after this change):

```
#### Version 0.10.2.0, suite SASL PLAIN: STARTING
### /agl/trivup/build/lib.linux-x86_64-2.7/trivup/apps/KafkaBrokerApp/deploy.sh: Downloading 0.10.2.0 from http://apache.mirrors.spacedump.net/kafka/0.10.2.0/kafka_2.11-0.10.2.0.tgz

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

edenhill:
Thanks for this!
I'm definitely open to use something leaner than OpenSSL for base64 and I'll look into viable alternatives (I think I did when I implemented SCRAM and didn't really find something I liked, but I'll have another look).

As for tests, it seems that mirror has gone away, you can override the download URL by setting env `KAFKA_URL=https://archive.apache.org/dist/kafka/0.10.2.0/kafka_2.12-0.10.2.0.tgz`

edenhill:
On a tangent:
depending on OpenSSL has proven problematic: the version on OSX is outdated and lacks devel files, the Windows solution with a 3rd party build is far from optimal, and relying on external dependencies is problematic for packaging the higher-level clients (confluent-kafka-{python,go,dotnet}).

What we would like to do, as an opt-in, is use an alternate TLS library that is portable and is easy to link statically - there are a number of OpenSSL forks and alternatives, but it is hard to tell which one to use.

agl:
> What we would like to do, as an opt-in, is use an alternate TLS library that is portable and is easy to link statically - there are a number of OpenSSL forks and alternatives, but it is hard to tell which one to use.

Could you just use OpenSSL? It can be built statically. The only issue arises with possible symbol collisions if something else tries to link OpenSSL too. Generally this is resolved by making your library a shared-object and [using symbol visibility](https://www.akkadia.org/drepper/dsohowto.pdf).

(In TensorFlow we'll use BoringSSL, but we generally only recommend BoringSSL to Google-related projects because we don't attempt to be generally compatible like OpenSSL does.)

> As for tests, it seems that mirror has gone away, you can override the download URL by setting env `KAFKA_URL=https://archive.apache.org/dist/kafka/0.10.2.0/kafka_2.12-0.10.2.0.tgz`

Thank you for that. The tests appeared to run with that, although quite a number failed. However, the same tests failed both before and after the change. (Thus if you can run the tests successfully then I'd strongly recommend doing so before committing this.)

edenhill:
Thank you for this!

Issue #1616, opened 2018-01-03
Receive failed: Invalid response size 100000039 (0..100000000): increase receive.message.max.bytes


du-do:
Description
===========
librdkafka  v0.11.1
kafka broker v0.10.1.0
ReadHat 5 (x64)
consumer config: `receive.message.max.bytes=100000000 ` 
message size : 5242880(5M)

I have three consumers, they are all getting messages from one broker. My consumers is slow, but the producer is fast.  So over time, the messages will pile up on the broker. 
Then i get this from the consumer all the time, and my consumer can not receive messages any more:
`%3|1514941148.962|FAIL|rdkafka#consumer-1| [thrd:172.20.51.38:9092/bootstrap]: 172.20.51.38:9092/1: Receive failed: Invalid response size 100000039 (0..100000000): increase receive.message.max.bytes`

I know that messaging should be balanced. However, the accumulation of messages led to the consumer failed to receive I think it is a problem.

I think broker send messages using `receive.message.max.bytes` ,But I do not understand why 39 bytes are added. Maybe it is the length of the head or something else. When i increase  `receive.message.max.bytes`  to 200000000, the problem is still not solved. Invalid response size is still 39 bytes more than my receive.message.max.bytes.

`%3|1514947746.873|FAIL|rdkafka#consumer-1| [thrd:172.20.51.38:9092/bootstrap]: 172.20.51.38:9092/1: Receive failed: Invalid response size 200000039 (0..200000000): increase receive.message.max.bytes
%3|1514947746.873|ERROR|rdkafka#consumer-1| [thrd:172.20.51.38:9092/bootstrap]: 172.20.51.38:9092/1: Receive failed: Invalid response size 200000039 (0..200000000): increase receive.message.max.bytes`
I think this may be related to internal judgment.

When i use another topic, i found  the  Invalid response size changed too.It became 100000038!
`3|1514873845.371|FAIL|rdkafka#consumer-1| [thrd:172.20.51.39:9092/bootstrap]: 172.20.51.39:9092/2: Receive failed: Invalid response size 100000038 (0..100000000): increase receive.message.max.bytes`

So i believe  Invalid response size = `receive.message.max.bytes` + 36 + length of the topic
Should this length be deducted when judging?
I am not sure whether my guess is correct or not. How to deal with this imbalance in sending and receiving information？

How to reproduce
================
Just make sure the producer generates messages faster than the consumer consumes

Checklist
=========

[...]

PR #1731, opened 2018-03-20
Add and use `fetch.max.bytes` to limit total Fetch response size (KIP-74, #1616)

edenhill:
`receive.message.max.bytes` was used in FetchRequest.max_bytes to tell
the broker to limit the response size, but since the broker may overshoot
this value up to one message.max.bytes it resulted in an endless config
chase where receive.message.max.bytes was increased and the response
size would follow + a couple of extra bytes.
Using the new `fetch.max.bytes` property and making sure it is lower than
`receive.message.max.bytes` fixes the issue.

This also adds verification for these configuration values to
the rd_kafka_new() call.


ijuma:
To be clear, the broker will only exceed max bytes if a single batch is larger than the limit (i.e. it would need return two batches where the second would cause it to go over). Given that, it seems like a config bug is needed to trigger this issue. I don't quite understand how this PR fixes that. It would be good to talk about a specific example of how it would break previously and now would work.

edenhill:
This is a typical error message from this case:
`Receive failed: Invalid response size 1000000040 (0..1000000000): increase receive.message.max.bytes`

receive.message.max.bytes is set to 1000000000, and reused as FetchRequest.max_bytes, the response has the overhead of FetchResponse framing and thus exceeds receive.message.max.bytes.
The problem typically surfaces on high-volume topics when fetching multiple partitions and there is no easy workaround.

src/rdkafka_broker.c
```diff
@@ -3038,7 +3038,7 @@ static int rd_kafka_broker_fetch_toppars (rd_kafka_broker_t *rkb, rd_ts_t now) {
         if (rd_kafka_buf_ApiVersion(rkbuf) == 4) {
                 /* MaxBytes */
                 rd_kafka_buf_write_i32(rkbuf,
-                                       rkb->rkb_rk->rk_conf.recv_max_msg_size);
+                                       rkb->rkb_rk->rk_conf.fetch_max_bytes);
```
edenhill: This is the essence of the fix, all the rest is config property handling.

mhowlett:
Wondering if there enough value in the extra configuration property to justify adding it? (making the library overly tunable acts to make it harder to use). Can you just use a const internal to librdkafka corresponding to the protocol overhead instead? 

If you definitely believe the extra config property is justified, I think the doc string could do with a bit of re-wording to really highlight the difference with `receive.message.max.bytes` (and implicitly justify it's existence).

ijuma:
Yeah, if it's just the framing issue, we should consider whether we can make it work without an additional config.

edenhill:
Note that fetch.max.bytes is the new config property, receive.message.max.bytes on the other is an existing property that we can't remove - in hindsight there might be little need for it.

fetch.max.bytes does have some use for people with certain performance requirements or unusually large messages.

edenhill:
Will reconstruct PR as to other external discussion:
 * automatically adjust receive.message.max.bytes to be at least fetch.max.bytes + framing
 * remove reference to receive.message.max.bytes in fetch.max.bytes config doc to avoid confusion.

mhowlett:
LGTM

PR #1733, opened 2018-03-21
Log the value that couldn't be found for s2f configuration options

ankon:
This helped me to find an issue in our configuration: node-rdkafka is lagging behind master (obviously), and the `consumer` value for the `debug` configuration option wasn't supported in that version yet.

### DIFF
--- a/src/rdkafka_conf.c
+++ b/src/rdkafka_conf.c
@@ -1330,8 +1330,9 @@ rd_kafka_anyconf_set_prop (int scope, void *conf,
 
 			/* No match */
 			rd_snprintf(errstr, errstr_size,
-				 "Invalid value for "
-				 "configuration property \"%s\"", prop->name);
+                                "Invalid value \"%.*s\" for "
+                                "configuration property \"%s\"",
+                                (int)(t-s), s, prop->name);
 			return RD_KAFKA_CONF_INVALID;
 
 		}
--- a/tests/0004-conf.c
+++ b/tests/0004-conf.c
@@ -381,6 +381,30 @@ int main_0004_conf (int argc, char **argv) {
 		rd_kafka_conf_destroy(conf);
 	}
 
+        {
+                rd_kafka_conf_res_t res;
+
+                TEST_SAY("Error reporting for S2F properties\n");
+                conf = rd_kafka_conf_new();
+
+                res = rd_kafka_conf_set(conf, "debug",
+                        "cgrp,invalid-value,topic", errstr, sizeof(errstr));
+
+                TEST_ASSERT(res == RD_KAFKA_CONF_INVALID,
+                        "expected 'debug=invalid-value' to fail with INVALID, "
+                        "not %d", res);
+                TEST_ASSERT(strstr(errstr, "invalid-value"),
+                        "expected invalid value to be mentioned in error, "
+                        "not \"%s\"", errstr);
+                TEST_ASSERT(
+                        !strstr(errstr, "cgrp") && !strstr(errstr, "topic"),
+                        "expected only invalid value to be mentioned, "
+                        "not \"%s\"", errstr);
+                TEST_SAY(_C_GRN "Ok: %s\n" _C_CLR, errstr);
+
+                rd_kafka_conf_destroy(conf);
+        }
+
 	/* Canonical int values, aliases, s2i-verified strings */
 	{
 		static const struct {


src/rdkafka_conf.c
```diff
@@ -1329,9 +1329,11 @@ rd_kafka_anyconf_set_prop (int scope, void *conf,
 				continue;
 
 			/* No match */
+			v = rd_strndup(s, (size_t)(t-s));
 			rd_snprintf(errstr, errstr_size,
-				 "Invalid value for "
-				 "configuration property \"%s\"", prop->name);
+				 "Invalid value \"%s\" for "
```
edenhill: instead of the heap strdup you could either use rd_strndupa() (stack-allocated) or simply:
`"Invalid value \"%.*s\" for ..", (int)(t-s), s, ..`

ankon: Thanks, learned something new here :)

I updated/force-pushed the change with that, and a test for this situation (based on my rough understanding how the tests are done)

ankon:
Hmm. AppVeyor's failure doesn't look like it is related to this PR:
~~~~
Using credentials from config. UserName: nuget
Please provide credentials for: https://ci.appveyor.com/nuget/edenhill
UserName: 
~~~~

tests/0004-conf.c
```diff
@@ -381,6 +381,25 @@ int main_0004_conf (int argc, char **argv) {
 		rd_kafka_conf_destroy(conf);
 	}
 
+	{
```
edenhill: I know the source is messy with tab and whitespace indentation, but all new code, and changed lines, should use 8 whitespaces for tab rather than the tab character.

tests/0004-conf.c
```diff
@@ -381,6 +381,25 @@ int main_0004_conf (int argc, char **argv) {
...
+		rd_kafka_conf_res_t res;
+
+		TEST_SAY("Error reporting for S2F properties\n");
+		conf = rd_kafka_conf_new();
+
+		res = rd_kafka_conf_set(conf, "debug", "invalid-value", errstr, sizeof(errstr));
```
edenhill: maybe strap some valid flags in there as well:
"cgrp,invalid-value,topic"

and then add another assertation:
`TEST_ASSERT(!strstr(errstr, "cgrp") && !strstr(errstr, "topic"))

src/rdkafka_conf.c
```diff
@@ -1330,8 +1330,8 @@ rd_kafka_anyconf_set_prop (int scope, void *conf,
 
 			/* No match */
 			rd_snprintf(errstr, errstr_size,
-				 "Invalid value for "
-				 "configuration property \"%s\"", prop->name);
+				 "Invalid value \"%.*s\" for "
+				 "configuration property \"%s\"", (int)(t-s), s, prop->name);
```
edenhill: Move the three arguments after `..configuration property..", ` to a new line to keep < 80 column width.

tests/0004-conf.c
ankon: It did have me thinking to do a reformat to be honest ... but I couldn't actually work out what the current settings were. Maybe providing an [`.editorconfig` file](http://editorconfig.org/) could help?

tests/0004-conf.c
ankon: Good idea!

ankon:
Updated with the review comments & rebased onto current master.

tests/0004-conf.c
edenhill: It is mentioned in the contribution guidelines:
https://github.com/edenhill/librdkafka/blob/master/CONTRIBUTING.md#indenting

The plan is to add something like [astyle](http://astyle.sourceforge.net/astyle.html) to enforce the coding guidelines.

CONFIGURATION.md
```diff
@@ -3,7 +3,7 @@
 
 Property                                 | C/P | Range           |       Default | Description              
 -----------------------------------------|-----|-----------------|--------------:|--------------------------
-builtin.features                         |  *  |                 | gzip, snappy, ssl, sasl, regex, lz4, sasl_gssapi, sasl_plain, sasl_scram, plugins | Indicates the builtin features for this build of librdkafka. An application can either query this value or attempt to set it with its list of required features to check for library support. <br>*Type: CSV flags*
+builtin.features                         |  *  |                 | gzip, snappy, ssl, sasl, regex, lz4, sasl_plain, sasl_scram, plugins | Indicates the builtin features for this build of librdkafka. An application can either query this value or attempt to set it with its list of required features to check for library support. <br>*Type: CSV flags*
```
edenhill: remove CONFIGURATION.md from your patch, it is automatically generated based on your build and there are missing feature flags.

src/rdkafka_conf.c
```diff
@@ -1330,8 +1330,9 @@ rd_kafka_anyconf_set_prop (int scope, void *conf,
 
 			/* No match */
 			rd_snprintf(errstr, errstr_size,
-				 "Invalid value for "
-				 "configuration property \"%s\"", prop->name);
+				 "Invalid value \"%.*s\" for "
```
edenhill: tab -> spaces

ankon:
Done. I only modified the tabs->spaces for the touched lines in rdkafka_conf.c though, so now things are a bit more wonky around that area than before.

edenhill:
The wonkiness is okay at this point :)

edenhill:
This looks good now! Will wait with merging this until after the upcoming release.

edenhill:
Thank you!

PR #1734, opened 2018-03-21
updated upload version

adyach:
updated upload version of librdkafka to be consistent with what is in the homebrew


packaging/homebrew/README.md
```diff
@@ -7,9 +7,9 @@ You should run it in two steps, first an implicit dry-run mode
 to check that things seem correct, and if that checks out a
 live upload mode which actually pushes the PR.
 
-    # Do a dry-run first, v0.11.0 is the librdkafka tag:
```
edenhill: This version is just an example, we don't want to update this file for each new version to release.

adyach: okay, fair enough

PR #1740, opened 2018-03-27
more compatibility for AIX and Solaris

tbsaunde:


src/rdendian.h
```diff
@@ -124,12 +124,36 @@
...
-         (((x) & 0xff00) << 8) |                \
-         (((x) & 0xff0000) >> 8) |              \
-         (((x) & 0xff000000) >> 24))
-
+#ifdef __GNUC__
+#define le32toh(x) __builtin_bswap32(x)
+#define le64toh(x) __builtin_bswap64(x)
+#else
+static inline uint32_t
+le32toh(uint32_t le)
```
edenhill: Why not keep the original macro?
I believe the compiler has a bigger chance of optimizing the macro than the function.

The function needs RD_UNUSED to avoid unused warnings.

tbsaunde: the point of the builtin is so that the compiler has a better idea what's going on and can produce a bswap instruction or whatever else is best for the hardware, so for gcc like compilers it should be at least as good.  Without reading generated code and benchmarking I'm really not sure what is best for xlc, but I also don't care that much about xlc, and the function certainly seems more readable to me.

Given the function is inline gcc's docs claim the unused attribute shouldn't be required, and infact it doesn't warn for me.

edenhill: Yes, using the gcc _builtin alternative makes sense, I was more thinking why change the existing macro to a function?

tbsaunde: mostly just wanting to write the le64toh as a function, and then being consistant between the two.

PR #1747, opened 2018-03-29
dynamically initialize rd_kafka_err_descs on solaris

tbsaunde:
There is a bug in the way Sun Studio generates relocations for pointers
into the middle of strings.  That means the name field of all the
rd_kafka_err_desc objects ends up being null instead of the correct
name.  Basically you can't have a compile time initialized pointer to
any part of a string literal other than the beginning.  Since the layout
of rd_kafka_err_desc is part of the ABI and we can't compile time
initialize it on Solaris we need to initialize it as part of the global
library initialization.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -122,6 +122,10 @@ void rd_kafka_set_thread_sysname (const char *fmt, ...) {
         thrd_setname(rd_kafka_thread_sysname);
 }
 
+#ifdef __sun
+static void init_err_descs(void);
+#endif
+
 static void rd_kafka_global_init (void) {
 #if ENABLE_SHAREDPTR_DEBUG
         LIST_INIT(&rd_shared_ptr_debug_list);
@@ -133,6 +137,9 @@ static void rd_kafka_global_init (void) {
 	rd_atomic32_init(&rd_kafka_op_cnt, 0);
 #endif
         crc32c_global_init();
+#ifdef __sun
+        init_err_descs();
+#endif
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/rdkafka.c
```diff
@@ -318,228 +325,246 @@ static const char *rd_kafka_type2str (rd_kafka_type_t type) {
...
+        _ERR_DESC(RD_KAFKA_RESP_ERR__VALUE_SERIALIZATION, \
+                  "Local: Value serialization error") \
+        _ERR_DESC(RD_KAFKA_RESP_ERR__KEY_DESERIALIZATION, \
+                  "Local: Key deserialization error") \
+        _ERR_DESC(RD_KAFKA_RESP_ERR__VALUE_DESERIALIZATION, \
+                  "Local: Value deserialization error") \
+        _ERR_DESC(RD_KAFKA_RESP_ERR__PARTIAL, \
+                  "Local: Partial response") \
+            _ERR_DESC(RD_KAFKA_RESP_ERR__READ_ONLY, \
+                              "Local: Read-only object") \
```
edenhill: indentation is off (use only whitespaces)

src/rdkafka.c
```diff
@@ -318,228 +325,246 @@ static const char *rd_kafka_type2str (rd_kafka_type_t type) {
...
+static struct rd_kafka_err_desc rd_kafka_err_descs[(RD_KAFKA_RESP_ERR_END_ALL - RD_KAFKA_RESP_ERR__BEGIN) + 1];
+
+static void init_err_descs(void)
+{
+#define _ERR_DESC(e, c) \
+    rd_kafka_err_descs[((int)e) - RD_KAFKA_RESP_ERR__BEGIN].code = e; \
+    rd_kafka_err_descs[((int)e) - RD_KAFKA_RESP_ERR__BEGIN].name = #e + 18; \
+    rd_kafka_err_descs[((int)e) - RD_KAFKA_RESP_ERR__BEGIN].desc = c;
+
+    ERR_LIST
```
edenhill: add ; to not mess up indent

src/rdkafka.c
```diff
@@ -318,228 +325,246 @@ static const char *rd_kafka_type2str (rd_kafka_type_t type) {
...
-                  "Broker: Operation not attempted"),
-
-	_ERR_DESC(RD_KAFKA_RESP_ERR__END, NULL)
+    ERR_LIST
 };
+#undef _ERR_DESC
+#else
+static struct rd_kafka_err_desc rd_kafka_err_descs[(RD_KAFKA_RESP_ERR_END_ALL - RD_KAFKA_RESP_ERR__BEGIN) + 1];
+
+static void init_err_descs(void)
```
edenhill: open brace on same line as function

src/rdkafka.c
```diff
@@ -318,228 +325,246 @@ static const char *rd_kafka_type2str (rd_kafka_type_t type) {
 	return types[type];
 }
 
+#define ERR_LIST \
```
edenhill: _ERR_LIST

src/rdkafka.c
```diff
@@ -133,6 +137,9 @@ static void rd_kafka_global_init (void) {
 	rd_atomic32_init(&rd_kafka_op_cnt, 0);
 #endif
         crc32c_global_init();
+#ifdef __sun
+        init_err_descs();
```
edenhill: rd_kafka_init_err_descs_suno();

src/rdkafka.c
edenhill: Add a comment saying why this is necessary (pretty much your PR comment)
```
/**
 * @brief .........
 *      ........
 *
 */
```

PR #1749, opened 2018-04-02
Fix a segment fault error in rdkafka_buf with zero-length string.

sunny1988:
(RD_KAFKAP_STR_LEN0(_klen) == 0) will treat zero-length as true as well,
and then assign NULL to kstr->str to cause segment fault in future access.
Fixed it by change it to (RD_KAFKAP_STR_IS_NULL(kstr))


edenhill:
Where was the segmentation fault seen?

sunny1988:
We have a coverage test to test edge cases by passing zero-length/empty string as host string or topic strings, etc. This is passed in previous versions like 0.9.2 and failed when we upgrade to 0.11.3. This simple change would preserve the old behavior. 
Definitely user can check the value before they pass them into librdkafka, but empty string instead of null should be a valid value and should only case a warning but not segfault/crash the process.

Thanks

edenhill:
Thank you for this!

PR #1750, opened 2018-04-02
Added generic pollable interface

accelerated:
**Description:**
- Refactored poll/yield into base class for generic programming
- Removed unwanted virtual derivation in some classes as it makes static up-casting impossible (and has heavy runtime performance cost)
- Removed superfluous destructors (much better to leave compiler defaulted ones in place until specific logic needs to be inserted in them)
- Queue class now also has a **yield()** method. This is actually desirable as **Queue::poll()** calls **rd_kafka_q_serve()** which does check for the yield flag to be set. 
- Fixed a linker dependency on _librt_ (CMake would not compile on Redhat 7)

Example use of Pollable interface:

```
using poll_vector = std::vector<std::shared_ptr<RdKafka::Pollable>>;
poll_vector polls;

for (size_t i = 0; i < 10; ++i)
{
    polls.emplace_back(Producer::Create(...));
    polls.emplace_back(Consumer::Create(...));
}

//generic poll loop
void pollLoop(const poll_vector& polls)
{
    while (1)
    {
        for (auto&& pollable : polls) pollable->poll();
        std::this_thread::sleep_for(std::chrono::milliseconds(100));
    }
}

std::thread t(pollLoop, polls); //run loop
```

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -26,5 +26,5 @@ if(NOT WIN32)
     target_link_libraries(rdkafka_consumer_example PUBLIC rdkafka)
     
     add_executable(kafkatest_verifiable_client kafkatest_verifiable_client.cpp)
-    target_link_libraries(kafkatest_verifiable_client PUBLIC rdkafka++)
+    target_link_libraries(kafkatest_verifiable_client PUBLIC rdkafka++ rt) #librt for clock_gettime
 endif(NOT WIN32)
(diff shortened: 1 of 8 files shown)


edenhill:
MSVC doesn't like this:
https://ci.appveyor.com/project/edenhill/librdkafka/build/0.11.4-R-pre451-sxbiosha/job/o6drorbape99s021#L263

accelerated:
Good reference topic on casting with diamond structure [here](http://deepakgiri123.blogspot.com/2011/05/c-up-casting-down-casting-in.html)

edenhill:
Outdated. Not sure if this is still of interest, if so, reopen it.

PR #1756, opened 2018-04-05
Runtime instrospection of config options

accelerated:
Description: Added two methods for obtaining runtime metadata info about configuration options. This will enable clients such as CppKafka to perform validation of user inputs (type and scope). No change in logic was made with the exception of exposing two private enums to the public header.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -678,7 +678,7 @@ void rd_kafka_destroy_final (rd_kafka_t *rk) {
         rd_kafkap_str_destroy(rk->rk_client_id);
         rd_kafkap_str_destroy(rk->rk_group_id);
         rd_kafkap_str_destroy(rk->rk_eos.TransactionalId);
-	rd_kafka_anyconf_destroy(_RK_GLOBAL, &rk->rk_conf);
+	rd_kafka_anyconf_destroy(RD_GLOBAL, &rk->rk_conf);
         rd_list_destroy(&rk->rk_broker_by_id);
 
 	rd_kafkap_bytes_destroy((rd_kafkap_bytes_t *)rk->rk_null_bytes);
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -678,7 +678,7 @@ void rd_kafka_destroy_final (rd_kafka_t *rk) {
         rd_kafkap_str_destroy(rk->rk_client_id);
         rd_kafkap_str_destroy(rk->rk_group_id);
         rd_kafkap_str_destroy(rk->rk_eos.TransactionalId);
-	rd_kafka_anyconf_destroy(RD_GLOBAL, &rk->rk_conf);
+	rd_kafka_anyconf_destroy(_RK_GLOBAL, &rk->rk_conf);
         rd_list_destroy(&rk->rk_broker_by_id);
 
 	rd_kafkap_bytes_destroy((rd_kafkap_bytes_t *)rk->rk_null_bytes);
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -1214,9 +1214,10 @@ RD_EXPORT size_t rd_kafka_header_cnt (const rd_kafka_headers_t *hdrs);
  * @brief Configuration result type
  */
 typedef enum {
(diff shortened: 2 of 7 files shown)


accelerated:
I am still working on a PR for CppKafka. Librdkafka only validates the type and the existence of the option, there is no validation to enforce setting a producer option on a producer, etc...Anyone can set just about any option on any type of object and there's no error as long as the config name is valid. Librdkafka only checks scope with RD_GLOBAL everywhere or RD_TOPIC for topic based options. But today I can set a consumer topic option on a producer with no errors.

accelerated:
I have discussed this with @mfontanini [here](https://github.com/mfontanini/cppkafka/issues/48) and he seemed to agree that it would be nice to allow better config option control for users.

In particular see excerpt:
> **my comment**:
As for checking the Configuration::set(), one way of always being in sync is to validate against all settings at runtime and check for the global/consumer/producer flags. Right now it's not exported but this can be easily fixed with an extern declaration in the header file.
**mfontanini's reply**:
That is nice but if it's not currently being exported then we can't really use it. I do think this should be exported as it's useful in cases like this one.

>

mfontanini:
For the record, I'm okay with how things are now. This would only be useful for cppkafka if `Configuration` was changed to use a type specific to Consumer and Producer (e.g. `ConsumerConfiguration`) but this is not great because:

* It involves a breaking change unless you keep both the old and new classes but that's just confusing.
* There are still cases where you want a generic config to be used as the base for both the consumer and producer. e.g. in my code I always have a base config:

```c++
Configuration base_config = {
    { "metadata.broker.list", brokers },
    { "log.connection.close", false },
    { "client.id", getHostname() },
    { "api.version.request", true }
};
```

That I use to bootstrap the actual consumer and producer configs I'll use in the application. With a consumer/producer specific config, this wouldn't be possible unless you introduced yet another "common configuration" type but I'm not sure about this.

Anyway I don't want to divert the discussion towards cppkafka. I'm not 100% sure of how useful this would be but I do agree that this should be exported by rdkafka, mostly because I don't think there's a reason not to export it.

accelerated:
@mfontanini code will be 100% backwards compatible both compile and runtime. I am adding some traits to make things easier to use.

accelerated:
Please see cppkafka PR [here](https://github.com/mfontanini/cppkafka/pull/49)

edenhill:
This is a well crafted PR :+1:.

However, I think it is leaking too much of the internal librdkafka implementation into the public API, something that we try hard not to do since it makes maintenance harder.

If I understand the use-case correctly you basically wan't to error out if consumer config properties are used on the producer and vice versa.
I suggest a simpler approach:
 * add `rd_kafka_resp_err_t rd_kafka_conf_set_type (rd_kafka_conf_t *conf, rd_kafka_type_t type);` which sets the instance type of the configuration. Have it default to the current behavioiur (untyped) and only allow it to be changed once, otherwise return `ERR__CONFLICT`.
 * change the internal conf setters and getters, e.g., anyconf_set.., to check that scope matches the type, otherwise return `RD_KAFKA_CONF_UNKNOWN` (or perhaps INVALID).
 * errors will now propagate automatically for each .._conf_set() call where the type does not match the property's scope.
 * Existing applications and bindings only need to add a single line of code to enable this type checking.

accelerated:
@edenhill Thanks for looking into this. I have made a different set of changes to validate scope internally and which will be invisible to the user. If the changes are OK with you, I can remove the first commit which exposes the underlying configuration lists. Please check and let me know. 
Instead of adding the `rd_kafka_resp_err_t rd_kafka_conf_set_type (rd_kafka_conf_t *conf, rd_kafka_type_t type)`, I keep track of which configuration options have been set. If the user at some point in time adds an option which is a mismatch, then we just return a MISMATCH error.

edenhill:
@accelerated Great! Remove the previous commits (to clean up the history) and make sure to read the coding guidelines with regards to style:
https://github.com/edenhill/librdkafka/blob/master/CONTRIBUTING.md

accelerated:
Yeah, I am using spaces instead of tabs, but for some reason new line alignments are inserted with tabs. It's a bit annoying to say the least.

PR #1758, opened 2018-04-09
added `compresion.level' configuration option for fine-tuning gzip and LZ4

erkoln:
Hello,

librdkafka supports message sets compressed with either gzip, LZ4 or Snappy but currently provides no control over compression level used by the underlying algorithm when producing messages, meaning the default compression level is always used.

Default zlib level provides a good compression ratio, but does so at the cost of high CPU usage. When producing messages to a remote cluster, CPU can easily become a bottleneck for per-partition throughput.
On the other hand, Snappy and LZ4 use very little CPU but the compression ratio is not so good when compared to gzip. In this case, network bandwidth quickly becomes the bottleneck.

This patch allows compression level to be fine-tuned for LZ4 and gzip (this seemingly can't be controlled in the Snappy implementation used by librdkafka), thus making it possible to trade between CPU and bandwidth usage.

In my use-case, given the particular CPU and network constraints I have to do with, level-1 gzip typically provide the best overall message throughput, while providing about the same compression ratio as default-level gzip.


src/rdkafka_conf.c
```diff
@@ -779,6 +779,25 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 		} },
         { _RK_GLOBAL|_RK_PRODUCER, "compression.type", _RK_C_ALIAS,
           .sdef = "compression.codec" },
+	{ _RK_GLOBAL|_RK_PRODUCER, "compression.level", _RK_C_INT,
```
edenhill: librdkafka now supports default topic configuration to be set on the global object (it instanties a default topic config object itself), we don't need to add this to the global config but only to the topic one.

src/rdkafka_conf.c
```diff
@@ -779,6 +779,25 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
+	  _RK(compression_level),
+	  "compression level parameter for algorithm selected by configuration "
```
edenhill: "Compression .."

src/rdkafka_lz4.c
```diff
@@ -325,7 +325,7 @@ rd_kafka_lz4_decompress (rd_kafka_broker_t *rkb, int proper_hc, int64_t Offset,
  * @returns allocated buffer in \p *outbuf, length in \p *outlenp.
  */
 rd_kafka_resp_err_t
-rd_kafka_lz4_compress (rd_kafka_broker_t *rkb, int proper_hc,
+rd_kafka_lz4_compress (rd_kafka_broker_t *rkb, int proper_hc, int compLevel,
```
edenhill: no camel casing, use snake casing comp_level

src/rdkafka_msgset_writer.c
```diff
@@ -794,9 +794,14 @@ rd_kafka_msgset_writer_compress_gzip (rd_kafka_msgset_writer_t *msetw,
         const void *p;
         size_t rlen;
         int r;
+        int compLevel =
+                msetw->msetw_rktp->rktp_rkt->rkt_conf.compression_level;
+        if (compLevel>Z_BEST_COMPRESSION) {
```
edenhill: spaces: `if (comp_level > Z_BEST_COMPRESSION)`

src/rdkafka_msgset_writer.c
```diff
@@ -931,9 +936,18 @@ static int
 rd_kafka_msgset_writer_compress_lz4 (rd_kafka_msgset_writer_t *msetw,
                                      rd_slice_t *slice, struct iovec *ciov) {
         rd_kafka_resp_err_t err;
+        int compLevel =
+                msetw->msetw_rktp->rktp_rkt->rkt_conf.compression_level;
+        if (compLevel>RD_KAFKA_COMPLEVEL_LZ4_MAX) {
+                compLevel = RD_KAFKA_COMPLEVEL_LZ4_MAX;
+        }
```
edenhill: `} else {`

Please see the style guide: https://github.com/edenhill/librdkafka/blob/master/CONTRIBUTING.md#follow-code-style

src/rdkafka_msgset_writer.c
```diff
@@ -931,9 +936,18 @@ static int
 rd_kafka_msgset_writer_compress_lz4 (rd_kafka_msgset_writer_t *msetw,
                                      rd_slice_t *slice, struct iovec *ciov) {
         rd_kafka_resp_err_t err;
+        int compLevel =
+                msetw->msetw_rktp->rktp_rkt->rkt_conf.compression_level;
+        if (compLevel>RD_KAFKA_COMPLEVEL_LZ4_MAX) {
```
edenhill: Normalize these values in topic_new() instead so it doesn't need to be done for each messageset

src/rdkafka_conf.c
erkoln: OK, done in 81b010d
What about `compression.codec`. Shouldn't it be the same?

src/rdkafka_conf.c
erkoln: 5523bda :)

src/rdkafka_msgset_writer.c
erkoln: Right! done in f15ebaa

src/rdkafka_lz4.c
erkoln: a73440a

src/rdkafka_msgset_writer.c
erkoln: a73440a

src/rdkafka_msgset_writer.c
erkoln: a73440a

erkoln:
Hello,
I rebased my fork on current master to avoid polluting the commit log in case of merge.
This seems to have broken change requests tracking but I tried to reply with relevant commit hashes.
Sorry for that

src/rdkafka_conf.c
erkoln: Oups! Forgot to actually remove the field from config structure. Done in f7584f8

src/rdkafka_lz4.c
```diff
@@ -339,8 +339,11 @@ rd_kafka_lz4_compress (rd_kafka_broker_t *rkb, int proper_hc,
 
         /* Required by Kafka */
         const LZ4F_preferences_t prefs =
-                { .frameInfo = { .blockMode = LZ4F_blockIndependent } };
-
+                {
+												.frameInfo = { .blockMode = LZ4F_blockIndependent },
```
edenhill: indent looks weird here

src/rdkafka_topic.c
```diff
@@ -313,6 +313,25 @@ shptr_rd_kafka_itopic_t *rd_kafka_topic_new0 (rd_kafka_t *rk,
 	if (rkt->rkt_conf.compression_codec == RD_KAFKA_COMPRESSION_INHERIT)
 		rkt->rkt_conf.compression_codec = rk->rk_conf.compression_codec;
 
+				switch (rkt->rkt_conf.compression_codec) {
+				case RD_KAFKA_COMPRESSION_GZIP:
+								if(rkt->rkt_conf.compression_level > RD_KAFKA_COMPLEVEL_GZIP_MAX)
```
edenhill: indentation is off here.
Also check style guide, space between if and paren

src/rdkafka_topic.c
```diff
@@ -313,6 +313,25 @@ shptr_rd_kafka_itopic_t *rd_kafka_topic_new0 (rd_kafka_t *rk,
...
+												rkt->rkt_conf.compression_level =
+																RD_KAFKA_COMPLEVEL_GZIP_MAX;
+								break;
```
edenhill: There doesnt seem to be any code to set the default compression level for gzip

src/rdkafka_conf.c
```diff
@@ -923,6 +923,23 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 		} },
         { _RK_TOPIC | _RK_PRODUCER, "compression.type", _RK_C_ALIAS,
           .sdef = "compression.codec" },
+	{ _RK_TOPIC|_RK_PRODUCER, "compression.level", _RK_C_INT,
+	  _RKT(compression_level),
+	  "Compression level parameter for algorithm selected by configuration "
+	  "property `compression.codec`. Higher values will result in better "
+	  "compression at the cost of more CPU usage. Usable range is "
+	  "algorithm-dependent: "
+#if WITH_ZLIB
```
edenhill: You can remove the ifdefs here, I think it is fine to provide these ranges in the config doc even if corresponding support isn't included.

erkoln:
Squashed history, which will hopefully also fix PR status.

edenhill:
@andoma What's up with the mips builder?

src/rdkafka_topic.c
```diff
@@ -313,6 +313,30 @@ shptr_rd_kafka_itopic_t *rd_kafka_topic_new0 (rd_kafka_t *rk,
 	if (rkt->rkt_conf.compression_codec == RD_KAFKA_COMPRESSION_INHERIT)
 		rkt->rkt_conf.compression_codec = rk->rk_conf.compression_codec;
 
+        switch (rkt->rkt_conf.compression_codec) {
+        case RD_KAFKA_COMPRESSION_GZIP:
+                /* RD_KAFKA_COMPLEVEL_DEFAULT == Z_DEFAULT_COMPRESSION so there
```
edenhill: That's leaking abstraction, we should not rely on the value of Z_DEFAULT_COMPRESSION being equal our value, unless we explicitly it define it to be the same, which we don't (and shouldn't).

So I'd rather see us handling the default case explicitly here

src/rdkafka_topic.c
```diff
@@ -313,6 +313,30 @@ shptr_rd_kafka_itopic_t *rd_kafka_topic_new0 (rd_kafka_t *rk,
...
+                        rkt->rkt_conf.compression_level =
+                                RD_KAFKA_COMPLEVEL_GZIP_MAX;
+                break;
+        case RD_KAFKA_COMPRESSION_LZ4:
+                /* For LZ4, default compression level is 0 and needs mapping:
+                 * check upper and lower bounds. */
+                if (rkt->rkt_conf.compression_level > RD_KAFKA_COMPLEVEL_LZ4_MAX)
+                        rkt->rkt_conf.compression_level =
+                                RD_KAFKA_COMPLEVEL_LZ4_MAX;
+                else if (rkt->rkt_conf.compression_level < 0)
```
edenhill: `== RD_KAFKA_COMPLEVEL_DEFAULT`

src/rdkafka_topic.c
```diff
@@ -313,6 +313,30 @@ shptr_rd_kafka_itopic_t *rd_kafka_topic_new0 (rd_kafka_t *rk,
...
+                        rkt->rkt_conf.compression_level = 0;
+                break;
+        case RD_KAFKA_COMPRESSION_SNAPPY:
+        default:
+                /* Compression level has no effet in this case */
```
edenhill: effect

src/rdkafka_topic.c
erkoln: done, let me know if that's better :)

src/rdkafka_topic.c
erkoln: okay, done (5c90922)

src/rdkafka_topic.c
erkoln: fixed (5c90922)
oh the joy of coding while being sleep deprived...

edenhill:
Thank you for this PR!

erkoln:
You're welcome. :-)
Thank you for the great piece of software that librdkafka is!

PR #1760, opened 2018-04-09
provide scope validation for config options

accelerated:
### Description
* Provided scope validation for handle and topic configuration settings.
* Provided scope validation for callbacks.
* State is stored in the `scope` member of the configuration handles. As options are added, the configuration scope is also updated to reflect the scope of the option. If at some point an option is added which does not match the current scope state, `RD_KAFKA_CONF_SCOPE_MISMATCH` is returned. 
* Default topic configurations are made to reflect the scope of the handle configuration and vice-versa.
* Duplication of a configuration will preserve it's current scope value.
* By default the scope starts as `RD_GLOBAL` for handle configurations and `RD_TOPIC` for topic configurations.
* `RD_CGRP` will set the configuration scope to `RD_CONSUMER`.
* The feature can be enabled/disabled via `void rd_kafka_conf_enable_scope_validation(int enable)`. By default it is _disabled_.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -58,7 +58,7 @@
 #include <sys/timeb.h>
 #endif
 
-
+extern int validate_scope;
 
 static once_flag rd_kafka_global_init_once = ONCE_FLAG_INIT;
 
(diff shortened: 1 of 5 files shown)


accelerated:
@mfontanini The logic for the default topic is essentially the same. If you cannot intermix default topic configuration for both consumers and producers. Global configs are ok however. Since the default topic is attached to a handle config and thus can be used to set topic configs in other configurations, it will throw an error if applied to a config of the wrong scope.

mfontanini:
Does this mean that this (using cppkafka syntax and using it as it is today) doesn't work anymore?

```c++
Configuration config = {
    // this is a topic configuration attribute and will fall through
    // to the default topic configuration
    { "auto.offset.reset", "latest" }
};
```

accelerated:
The fall through should work as before. There are test cases which pass in librdkafka.

PR #1761, opened 2018-04-10
revise src/rdkafka_sasl_cyrus.c

TopHK:
When I use kerberos, I found a bug in src/rdkafka_sasl_cyrus.c.

When the keytab file is a wrong one, the data can still be transfered to kafka brokers. 

After I review the source of rdkafka for several times, I find that, the function rd_kafka_sasl_cyrus_client_new invokes the function rd_kafka_sasl_cyrus_kinit_refresh to kinit a keytab file. However, when rd_kafka_sasl_cyrus_kinit_refresh returns -1, it means kinit operation comes to a failure. But rd_kafka_sasl_cyrus_client_new can still execute the following code lines after invoking the function rd_kafka_sasl_cyrus_kinit_refresh.

### DIFF
--- a/src/rdkafka_sasl_cyrus.c
+++ b/src/rdkafka_sasl_cyrus.c
@@ -457,7 +457,9 @@ static int rd_kafka_sasl_cyrus_client_new (rd_kafka_transport_t *rktrans,
         memcpy(state->callbacks, callbacks, sizeof(callbacks));
 
         /* Acquire or refresh ticket if kinit is configured */ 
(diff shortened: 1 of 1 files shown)


edenhill:
The intention here was to be a bit lax with kinit, it is possible that we already have a ticket and kinit failing, in which case authentication work anyway.
And if kinit fails to refresh our ticket the broker-side authentication will fail anyway, so there is no need to fail early here, which your example also shows since you are able to authenticate even when kinit fails, probably due to cached credentials.

PR #1764, opened 2018-04-10
Added scope validation functionality for producers and consumers

accelerated:
### Description
This is an important feature which allows proper scope validation across all configuration options. It prevents a user from mistakenly setting the wrong configuration which is silently absorbed and never takes effect. Two new APIs have been added which allow a user to set strict scope guards on the all configuration options.

* Provided scope validation for handle and topic configuration settings by calling `rd_kafka_conf_set_type()` and `rd_kafka_topic_conf_set_type()`. By default this is disabled i.e. scope is unset. This scope setting is on a per-handle basis. 
* Provided scope validation for callbacks.
* State is stored in the `scope` member of the configuration handles. As options are added, the configuration scope is also updated to reflect the scope of the option. If at some point an option is added which does not match the current scope state, `RD_KAFKA_CONF_UNKNOWN` is returned and the internal errno is set. 
* Default topic configurations are made to reflect the scope of the handle configuration and vice-versa.
* Duplication of a configuration will preserve it's current scope value.
* `RD_CGRP` will set the configuration scope to `RD_CONSUMER`.


accelerated:
@edenhill Hi, I see one of the architecture tests has passed and all integration tests, not sure why the other architectures have a permission problem. Have you seen this before? In any case, if you can code review and comment on my changes that would be awesome.

src/rdkafka.c
```diff
@@ -1298,17 +1298,17 @@ static void rd_kafka_term_sig_handler (int sig) {
 
 
 rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *app_conf,
-			  char *errstr, size_t errstr_size) {
```
edenhill: No whitespace only diffs please.

src/rdkafka.c
```diff
@@ -58,7 +58,7 @@
 #include <sys/timeb.h>
 #endif
 
-
+extern int validate_scope;
```
edenhill: This is neither thread-safe or multi-instance safe.

src/rdkafka.c
```diff
@@ -1356,7 +1356,16 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *app_conf,
         }
 #endif
 
-        if (type == RD_KAFKA_CONSUMER) {
+        if (type == RD_KAFKA_PRODUCER) {
+                if (validate_scope && (conf->scope & _RK_CONSUMER)) {
+                    rd_snprintf(errstr, errstr_size,
+                                "Wrong scope: configuration provided is intended for consumers.");
```
edenhill: keep below 80 columns

src/rdkafka.h
```diff
@@ -1214,11 +1214,29 @@ RD_EXPORT size_t rd_kafka_header_cnt (const rd_kafka_headers_t *hdrs);
  * @brief Configuration result type
  */
 typedef enum {
-	RD_KAFKA_CONF_UNKNOWN = -2, /**< Unknown configuration name. */
-	RD_KAFKA_CONF_INVALID = -1, /**< Invalid configuration value. */
-	RD_KAFKA_CONF_OK = 0        /**< Configuration okay */
+	RD_KAFKA_CONF_SCOPE_MISMATCH = -3, /**< Invalid scope for consumer or
```
edenhill: I don't think this added enum is strictly required, using UNKNOWN and providing a descriptive error string is sufficient and avoids applications having to deal with another return value.

src/rdkafka.h
```diff
@@ -1214,11 +1214,29 @@ RD_EXPORT size_t rd_kafka_header_cnt (const rd_kafka_headers_t *hdrs);
...
+ *         together. For example one cannot mix consumer and producer settings
+ *         in the same configuration or topic configuration handle. Similar
+ *         checks are performed when setting callback methods which prevents
+ *         users from setting a consumer callback on a producer (and vice-versa)
+ *         as it would have no effect.
+ * @remark The type of scope of a configuration is determined by the first
+ *         non-global option which is set. Any subsequent options must be within
+ *         the same scope.
+ */
+void rd_kafka_conf_enable_scope_validation(int enable);
```
edenhill: This needs to be set per rd_kafka_conf_t and rd_kafka_topic_conf_t, not globally.

src/rdkafka.h
```diff
@@ -1310,19 +1328,19 @@ rd_kafka_conf_res_t rd_kafka_conf_set(rd_kafka_conf_t *conf,
  * \p events is a bitmask of \c RD_KAFKA_EVENT_* of events to enable
  * for consumption by `rd_kafka_queue_poll()`.
  */
-RD_EXPORT
-void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
```
edenhill: Existing APIs must not be changed since it breaks both the API and ABI guarantee.

The config scope validation is mainly useful for the generic _conf_set() APIs since they are typically exposed directly to the end-user.
The specific conf set functions, like this one and the ones for callbacks, are always added programmatically and it is sufficient to document what setting applies to which client type.

src/rdkafka_conf.c
```diff
@@ -585,28 +592,28 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           .copy = rd_kafka_conf_interceptor_copy },
 
         /* Global client group properties */
-        { _RK_GLOBAL|_RK_CGRP, "group.id", _RK_C_STR,
+        { _RK_GLOBAL|_RK_CONSUMER|_RK_CGRP, "group.id", _RK_C_STR,
```
edenhill: This is not necessarily true in the future, the client group mechanism in the Kafka protocol is not tied to the consumer but is generic.

src/rdkafka_conf.c
```diff
@@ -1121,10 +1128,29 @@ static int rd_kafka_conf_s2i_find (const struct rd_kafka_property *prop,
...
-			   const struct rd_kafka_property *prop,
-			   const char *value,
-			   char *errstr, size_t errstr_size) {
+                           const struct rd_kafka_property *prop,
+                           const char *value,
+                           char *errstr, size_t errstr_size) {
 	int ival;
+    
+	/* Check if configuration scopes match */
+	int* conf_scope = (scope & _RK_GLOBAL) ? &((rd_kafka_conf_t*)conf)->scope :
```
edenhill: Please read the coding guidelines:
 * pointer goes with the variable name, not type
 * <= 80 columns

src/rdkafka_conf.c
```diff
@@ -1451,24 +1485,32 @@ rd_kafka_conf_res_t rd_kafka_conf_set (rd_kafka_conf_t *conf,
...
+            conf->topic_conf = rd_kafka_topic_conf_new();
+            
+            /* Update the scope to match that of the handle configuration */
+            conf->topic_conf->scope |= (conf->scope & _RK_PC_MASK);
         }
-
-        return rd_kafka_topic_conf_set(conf->topic_conf, name, value,
-                                       errstr, errstr_size);
+        
+        rd_kafka_conf_res_t rc = rd_kafka_topic_conf_set(conf->topic_conf,
```
edenhill: coding guidelines:
 * no inline variable definitions, they must go at the beginning of the block

src/rdkafka_conf.c
```diff
@@ -1702,155 +1751,285 @@ rd_kafka_topic_conf_t *rd_kafka_topic_conf_dup (const rd_kafka_topic_conf_t
 }
 
 rd_kafka_topic_conf_t *rd_kafka_default_topic_conf_dup (rd_kafka_t *rk) {
-        if (rk->rk_conf.topic_conf)
-                return rd_kafka_topic_conf_dup(rk->rk_conf.topic_conf);
-        else
-                return rd_kafka_topic_conf_new();
+        if (rk->rk_conf.topic_conf) {
+            return rd_kafka_topic_conf_dup(rk->rk_conf.topic_conf);
+        }
```
edenhill: coding guidelines, brace placement

src/rdkafka.c
accelerated: Ok will shorten...although the coding style guide said not to exceed by too much the 80 col mark.

src/rdkafka.c
accelerated: Most of librdkafka APIs are not thread safe. Unless I make this into an atomic variable, this will eventually take effect when CPU cache is synced. I really don't think it's a big issue. Also this variable will most likely be set/unset once per program.

src/rdkafka.h
accelerated: I feel that UNKNOWN is too broad of an error and is misleading. If someone checks for it, and sees UNKNOWN, they will think the option does not exist which is even worse IMHO. Also most user facing APIs don't have an error string passed in. I do set errno however.

src/rdkafka.h
accelerated: Yes but why do it for each? It's a feature you want for all configs or none at all. Users will have to call this for every config object they create including the default ones, it's very redundant. This is supposed to be a universal check in place. And then if you miss one, you risk again setting the wrong config options.

edenhill: As a user of librdkafka, how can you know that your use of rdkafka is the only use in the current application runtime? Unless you know exactly what all of your dependencies, and their dependencies, and their dependencies do, you simply can't. This means that each rdkafka client instance must be completely separated from any other instance in the same process.

src/rdkafka.c
accelerated: One comment on this, the actual source code is riddled with a mix of leading tabs and whitespaces and often in the same function. I guess whomever made the changes a long time ago was not as careful. Some of the indentation i fixed was around my changes to match proper alignment with what i'm submitting. Otherwise it's impossible to make the code look good.

src/rdkafka.h
accelerated: Well the ABI is easily fixed by recompiling the code. I do think that anyone synch-ing with upstream code will all least recompile their stuff, especially if it's statically linked. Since C doesn't have a throw mechanism, unless you change the return type, it's impossible to validate. Also it's not just the config setters that are in question, it's also the callback setters which need this feature. This proposed change doens't really change the existing code, so I think it should be acceptable that at least a recompile should be done.

src/rdkafka_conf.c
accelerated: I'm not familiar with the internal protocol, but currently any of the settings above are for consumers only. IF (and this is a big IF) they will add some grouping to producers (seems weird conceptually), you can then have options for _RK_GLOBAL|_RK_PRODUCER|_RK_CGRP. I believe it makes the code clearer to leave these flags as-is.

src/rdkafka_conf.c
accelerated: ok

src/rdkafka_conf.c
accelerated: I will remove it but I would say that 99% of code that i've seen prefers braces. This is to prevent someone from adding a second line to the IF statement and forgetting to add braces. Also it makes the code much easier to read.

src/rdkafka.c
edenhill: Yes, this is unfortunate, but we're going to fix this throughout the code in one single commit when we have a style checker in place, before that I urge people not to commit whitespace diffs unless they change the line.

src/rdkafka.h
edenhill: ABI guarantee means that the application does **not** need to be recompiled when upgrading the library, this is how all shared libraries work on all the major platforms, Linux, osx, windows.
This is not up for discussion.

Regarding code-level setters, such as callback setters: yes, it would be good if they had a return value, but for historical reasons they don't and we can't change them.
My take on this is that it is not a critical concern since these are code-level setters, we can document our way around it to make sure the application developer will not use them for the wrong instance type.

accelerated:
>This is a great effort, but I think it is trying to do too much.
It should only perform checks on the user-facing conf_set() APIs, this should be sufficient to implement what you are after. The other checks will go in the application or bindings themselves.

I'm not sure what you're implying...you mean don't check when you create a new consumer or producer for example? Or set events or callbacks? These are just as important. If we limit the checks for config setters, there are still cases which can potentially mismatch. 
Your solution for clearing some flags is ok, but you also have to clear the topic and the global flags since you don't want to set a topic option on a handle config and vice-versa. And since this check is performed after the property is found (generic internal method) you are pretty much back to using the same logic in determining what config type it is.

src/rdkafka.c
accelerated: cool

src/rdkafka.h
accelerated: well in that case would you consider having a second set of API's which return an error? No ABI/API breakage.

src/rdkafka.h
accelerated: Typically applications would be multi threaded and they would link once the .a or .so of librdkafka. Settings will be done once and you control how APIs are called. I can't think of any use case where you would be, inside your own librdkafka application, linking some other 3rd party application which also happens to be a librdkafka client? or link your .so file? This is not a library like boost or STL where calls need to be completely re-entrant. The config settings array you have can be manipulated in the same way, so I don't see how it's any different. The extern declaration has only internal exposure, nothing leaks outside.

src/rdkafka.h
accelerated: Another alternative is to leave the API as-is and have the return type be changed via a compile flag...actually this entire feature could then be compile controlled.

accelerated: Example:
```
# ifdef __HAS_RETURN__
# define RET_TYPE int
# define RETURN(val) return ((int)val) 
# else
# define RET_TYPE void
# define RETURN(val) return
# endif
```
Then you can use 
```
RET_TYPE some_function()
{
int a;
RETURN(a);
}
```

src/rdkafka.h
mfontanini: I think @edenhill refers to the case where you use a third party library that also uses librdkafka, then you enable this globally and that library has e.g. a mix of producer/consumer configs in the same handle. Then that library will think something's wrong when setting those attributes and possibly fail to work because of that, when in reality it's just that you asked librdkafka to be more strict than it is by default and that as a side effect broke the third party library.

src/rdkafka.h
edenhill: That is not ABI or API safe, so we would need to provide two different libraries, and
we definitely don't want to go down that path.

Generally speaking, I see value in your proposed API changes in this PR, and if we were redesigning the API this would definitely go in, but we're not there, so each new API needs to be weighed careful, pros and cons, and if there is a viable workaround we typically reject API additions. Why? Because we're stuck with APIs "forever", and they thus need to be designed in a fashion to be future proof, which is hard. Add to that the API footprint size increasing and the maintenance cost, in KLOC, tests and support.

Specifically, if it stands between adding a second alternative function  for all conf setter functions, or adding a line to the documentation of each such function on what type of instance it should be used on, and the end-user is not directly affected, then we will choose the lesser approach of documenting it.
With documentation in place we can consider using set_dr_msg_cb() on a consumer config object a bug in the application, which it is.
Programmatic input does not need to be validated, only documented.
User input, on the other hand, needs proper validation and enforcement, and that's why I think the conf_set(name, value) interface should be changed for the new behaviour, but without affecting existing applications.

Adding a new conf_res_t enum will trigger a warning in any application that uses a switch-case on the conf_set() return value, we don't want that, so if we can solve that by using UNKNOWN (which a consumer property practically is on a producer, and vice versa) and providing a proper error string for the end-user, then we're fine.

accelerated: @edenhill what about my second proposal with the compile directives? (And I agree that having 2 sets of APIs is ugly.

src/rdkafka.h
accelerated: ok i'll remove the global setting

accelerated:
@edenhill Also for all the config settings which returned void (you specifically rejected any ABI change), I am setting errno and also I've added a _compile time_ setting for returning a proper config error type. By default, in all builds and when you release the code these will always be turned off so no ABI impact. However if anyone builds their own version, they can enable this flag and check proper return value instead of errno. Makes error checking much nicer IMHO. Please let me know what you think of this?
I have also updated the C++ version of the library to have the scope checks.

accelerated:
@edenhill Hi, any chance you could review this (pls see my last comments as well)? I will incorporate whichever changes you want. I would also like to roll this into cppkafka as soon as this is approved. Much appreciated, thanks!

edenhill:
Outdated.
We now warn when producer properties are set on consumers, and vice versa.

PR #1766, opened 2018-04-11
Topic Admin API

edenhill:


edenhill:
After the .NET SR and Avro PR reviews I'd like to return the favour.
I hereby present you with a 8.8 KLOC PR, congratulations!

I would focus on:
 * src/rdkafka_admin.c - the actual implementation of the admin request workers
 * src/rdkafka.h - the great (in size) public API
 * src/rdkafka_request.c - protocol request construction

mhowlett:
first comment: is it easy to make a librdkafka.redist nuget package from this PR? if so, that would probably help with the review of the API as i'd start implementing the .NET binding simultaneously and surely notice if i bumped into anything that seemed off.

edenhill:
Yeah, should be sufficient to create a tag on the adminapi branch, which will trigger the CI to upload artifacts to S3, which you then package with packaging/nuget/release.py and then upload to NuGet.
I'll push a tag and see what happens

src/rdkafka.c
```diff
@@ -551,7 +553,10 @@ void rd_kafka_get_err_descs (const struct rd_kafka_err_desc **errdescs,
 
 const char *rd_kafka_err2str (rd_kafka_resp_err_t err) {
 	static RD_TLS char ret[32];
-	int idx = err - RD_KAFKA_RESP_ERR__BEGIN;
```
mhowlett: i don't understand the rational for this change (same comment below). won't it just force extra type casting with no actual effect?

src/rdkafka.c
```diff
@@ -1427,6 +1442,12 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *app_conf,
 	if (rk->rk_conf.offset_commit_cb)
 		rk->rk_conf.enabled_events |= RD_KAFKA_EVENT_OFFSET_COMMIT;
 
+        rk->rk_controllerid = -1;
+
+        /* Admin client defaults */
+        rk->rk_conf.admin.request_timeout_ms =
+                RD_MIN(5*60*1000, rk->rk_conf.socket_timeout_ms);
```
mhowlett: why not just socket_timeout_ms? (and why the magic constant?)

src/rdkafka_admin.c
```diff
@@ -0,0 +1,2696 @@
...
+ *
+ * The internal async implementation relies on the following concepts:
+ *  - it uses a single rko (rd_kafka_op_t) to maintain state.
+ *  - the rko has a callback attached - called the worker callback.
+ *  - the worker callback is a small state machine that triggers
+ *    async operations (be it controller lookups, timeout timers,
+ *    protocol transmits, etc).
+ *  - the worker callback is only called on the rdkafka main thread.
+ *  - the callback is triggered by different events and sources by enqueuing
+ *    the rko on the rdkafka man ops queue.
```
mhowlett: 'main ops queue'?

src/rdkafka_admin.c
```diff
@@ -0,0 +1,2696 @@
...
+ *
+ * Let's illustrate this with a DeleteTopics example. This might look
+ * daunting, but it boils down to an asynchronous state machine being
+ * triggered by enqueuing the rko op.
+ *
+ *  1. [app thread] The user constructs the input arguments,
+ *     including a response rkqu queue and then calls DeleteTopics().
+ *
+ *  2. [app thread] DeleteTopics() creates a new internal op (rko) of type
+ *     RD_KAFKA_OP_CREATETOPICS, makes a **copy** on the rko of all the
```
mhowlett: `RD_KAFKA_OP_DELETETOPICS`?

src/rdkafka_admin.c
```diff
@@ -0,0 +1,2696 @@
...
+ *
+ * 19. [app thread] The application calls rd_kafka_queue_poll() to
+ *     receive the result of the operation. The result may have been
+ *     enqueued in step 18 thanks to succesful completion, or in any
+ *     of the earlier stages when an error was encountered.
+ *
+ * 20. [app thread] The application uses rd_kafka_event_DeleteTopics_result()
+ *     to retrieve the request-specific result type.
+ *
+ * 21. Done, so easy.
```
mhowlett: you might want to consider rewording # 21

src/rdkafka_aux.c
```diff
@@ -0,0 +1,99 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2018 Magnus Edenhill
```
mhowlett: You're either a good negotiator or this should be (c) Confluent. Assuming the former.

src/rdkafka.c
edenhill: :+1:
Other part of the code erroneously stored a int16_t in an int, resulting in invalid error codes, but that has been fixed so this is no longer neede. Reverting

src/rdkafka.c
edenhill: I don't remember! Reverting :)

src/rdkafka_admin.c
edenhill: :(

src/rdkafka_admin.h
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * @name CreateTopics
+ * @{
+ */
+
+/**
+ * @brief NewTopic type, used with CreateTopics.
+ */
+struct rd_kafka_NewTopic_s {
```
mhowlett: I'm going to call this `NewTopicSpecification` - which is what it actually is - in C# and hope no one notices i've deviated from the other languages.

edenhill: I strongly urge you to keep with the original Java API names unless there is a clear reason not to.
Inspiration:
https://docs.confluent.io/current/clients/javadocs/org/apache/kafka/clients/admin/AdminClient.html

src/rdkafka.h
```diff
@@ -4202,6 +4293,975 @@ rd_kafka_interceptor_add_on_request_sent (
...
+RD_EXPORT void
+rd_kafka_CreateTopics (rd_kafka_t *rk,
+                       rd_kafka_NewTopic_t **new_topics,
+                       size_t new_topic_cnt,
+                       const rd_kafka_AdminOptions_t *options,
+                       rd_kafka_queue_t *rkqu);
+
+
+/**
+ * @brief CreateTopics result type and methods
```
mhowlett: are you expecting binding libraries to use a common queue for all concurrent requests or a separate queue per request? If there are substantial advantages to using a single queue (that I'm not aware of), I think we really want an opaque pointer that can be passed into `rd_kafka_CreateTopics` which is then made available via `rd_kafka_CreateTopics_result_t`, otherwise we need to match up requests with results by looking at topic names.

I expect queue per request is fine, and I'm going to set things up like that in the C# binding unless otherwise informed... this approach also has the advantage of avoiding the need to pin memory in a garbage collected language.

src/rdkafka_admin.h
mhowlett: yeah, I've been following that.

src/rdkafka.h
edenhill: If you want to provide a blocking interface then you would typically use a temporary per API request, but if you want it async and poll the results from a single place you would use a shared queue, be it a specific one of for admin results or the main queue.

See rd_kafka_AdminOptions_set_opaque()

src/rdkafka_conf.c
```diff
@@ -2246,3 +2249,187 @@ void rd_kafka_conf_properties_show (FILE *fp) {
...
+        confval->valuetype = RD_KAFKA_CONFVAL_PTR;
+        confval->u.PTR = NULL;
+}
+
+
+/**
+ * @brief Set confval's value to \p valuep, verifying the passed
+ *        \p valuetype matches (or can be cast to) \p confval's type.
+ *
+ * @param dispname is the display name for the configuration value and is
```
mhowlett: `dispname`? i think you mean to document `confval`.

src/rdkafka.h
```diff
@@ -4205,6 +4296,987 @@ rd_kafka_interceptor_add_on_request_sent (
...
+ *          errors.
+ *
+ * @param errstrp is set to a human-readable error string, if there was an
+ *        error.
+ *
+ * @remark This is equivalent and interchangable to rd_kafka_event_error() and
+ *         rd_kafka_event_error_string() on a RD_KAFKA_EVENT_CREATETOPICS_RESULT event.
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_CreateTopics_result_error (
```
mhowlett: over-engineered? 
will these separate methods to get the error per type ever add value over just one?

mhowlett: oh, you say that in a remark.

I think it's not a good idea to provide these in addition to rd_kafka_event_error because you leave people wondering what they are for - their existence implies there is likely something more to it than a simple wrapper, and means I need to go investigate (spend extra time) before using.

edenhill: You are right, away they go!

CONFIGURATION.md
```diff
@@ -19,8 +19,8 @@ topic.metadata.refresh.fast.interval.ms  |  *  | 1 .. 60000      |           250
...
-socket.timeout.ms                        |  *  | 10 .. 300000    |         60000 | Default timeout for network requests. Producer: ProduceRequests will use the lesser value of socket.timeout.ms and remaining message.timeout.ms for the first message in the batch. Consumer: FetchRequests will use fetch.wait.max.ms + socket.timeout.ms.  <br>*Type: integer*
+debug                                    |  *  | generic, broker, topic, metadata, feature, queue, msg, protocol, cgrp, security, fetch, interceptor, plugin, consumer, admin, all |               | A comma-separated list of debug contexts to enable. Detailed Producer debugging: broker,topic,msg. Consumer: consumer,cgrp,topic,fetch <br>*Type: CSV flags*
+socket.timeout.ms                        |  *  | 10 .. 300000    |         60000 | Default timeout for network requests. Producer: ProduceRequests will use the lesser value of `socket.timeout.ms` and remaining `message.timeout.ms` for the first message in the batch. Consumer: FetchRequests will use `fetch.wait.max.ms` + `socket.timeout.ms`. Admin: Admin requests will use `socket.timeout.ms` or explicitly set `rd_kafka_AdminOptions_set_operation_timeout()` value. <br>*Type: integer*
```
mhowlett: '... or the explicitly set ...'

src/rdkafka.h
```diff
@@ -1314,6 +1318,39 @@ RD_EXPORT
...
+ *        librdkafka main thread.
+ *
+ * First set the event callback with this function, followed by creating an
+ * rd_kafka_t instance with rd_kafka_new().
+ * Get the instance's internal queue with rd_kafka_queue_get_internal() and
+ * pass it as the reply/response queue to an API that takes an event queue,
+ * such as rd_kafka_CreateTopics().
+ * As the response event is ready the event callback will be triggered
+ * from the internal librdkafka main thread.
+ * The application must destroyed the \c rkev passed to \p event cb using
```
mhowlett: ...must destroy...

src/rdkafka.h
```diff
@@ -1314,6 +1318,39 @@ RD_EXPORT
 void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
 
 
+/**
+ * @brief Generic event callback to be used with the event API to trigger
```
mhowlett: 'the event API' probably warrants more clarification.

jeffwidman:
Can you cleanup the git history on this a little? I was curious to review it commit-by-commit, but looks like they aren't really separated into logical commits, for example https://github.com/edenhill/librdkafka/pull/1766/commits/51fc40fad654e1739989450532fd64de1645b040 and https://github.com/edenhill/librdkafka/pull/1766/commits/c7d7f0e061fcf11427027c6e40982e7f29e7017f could be a single commit (at least based on a quick skim).

CONFIGURATION.md
```diff
@@ -40,6 +40,7 @@ log_level                                |  *  | 0 .. 7          |             6
 log.queue                                |  *  | true, false     |         false | Disable spontaneous log_cb from internal librdkafka threads, instead enqueue log messages on queue set with `rd_kafka_set_log_queue()` and serve log callbacks or events through the standard poll APIs. **NOTE**: Log messages will linger in a temporary queue until the log queue has been set. <br>*Type: boolean*
 log.thread.name                          |  *  | true, false     |          true | Print internal thread name in log messages (useful for debugging librdkafka internals) <br>*Type: boolean*
 log.connection.close                     |  *  | true, false     |          true | Log broker disconnects. It might be useful to turn this off when interacting with 0.9 brokers with an aggressive `connection.max.idle.ms` value. <br>*Type: boolean*
+background_event_cb                      |  *  |                 |               | Background queue event callback (set with rd_kafka_conf_set_background_event_cb()) <br>*Type: pointer*
```
mhowlett: it'd be good to beef this out a little bit. what are 'background queues'? what are the rules for using them?

edenhill: They're explained in the reference (set_background_event..), and this property is not settable with a standard string,string key-value API.
These C-type config properties (callbacks, opaques, etc) are included in CONFIGURATION.md as a reference, but they do have little value, we should probably take them all out at some point.

src/rdkafka.c
```diff
@@ -3007,13 +3133,58 @@ char *rd_kafka_clusterid (rd_kafka_t *rk, int timeout_ms) {
...
+}
+
+
 void *rd_kafka_opaque (const rd_kafka_t *rk) {
         return rk->rk_conf.opaque;
 }
 
 
 int rd_kafka_outq_len (rd_kafka_t *rk) {
-	return rd_kafka_curr_msgs_cnt(rk) + rd_kafka_q_len(rk->rk_rep);
```
mhowlett: not for this PR, but we should have a discussion (again) about what exactly it makes sense to expose to high level users. maybe it's just msgs_cnt, maybe it two separate numbers, or maybe documentation could just be improved. but this leaks too many internal details atm imo.

src/rdkafka.h
```diff
@@ -1314,6 +1318,46 @@ RD_EXPORT
...
+ *  1. First set the event callback on the configuration object with this
+ *     function, followed by creating an rd_kafka_t instance
+ *     with rd_kafka_new().
+ *  2. Get the instance's background queue with rd_kafka_queue_get_background()
+ *     and pass it as the reply/response queue to an API that takes an
+ *     event queue, such as rd_kafka_CreateTopics().
+ *  3. As the response event is ready and enqueued on the background queue the
+ *     event callback will be triggered from the background thread.
+ *
+ * The application must destroyed the \c rkev passed to \p event cb using
```
mhowlett: ... must destroy ...

src/rdkafka.h
```diff
@@ -1314,6 +1318,46 @@ RD_EXPORT
 void rd_kafka_conf_set_events(rd_kafka_conf_t *conf, int events);
 
 
+/**
+ * @brief Generic event callback to be used with the event API to trigger
+ *        callbacks for \c rd_kafka_event_t objects from a background
+ *        thread serving the background queue.
+ *
+ * How to use:
```
mhowlett: nice

src/rdkafka.h
```diff
@@ -3368,14 +3456,19 @@ void rd_kafka_log_syslog(const rd_kafka_t *rk, int level,
 /**
  * @brief Returns the current out queue length.
  *
- * The out queue contains messages waiting to be sent to, or acknowledged by,
- * the broker.
+ * The out queue length is the sum of:
+ *  - number of messages waiting to be sent to, or acknowledged by,
+ *    the broker.
+ *  - number of delivery reports waiting
```
mhowlett: ahh, i see this is better documented now. I think you could still add a little more verbosity to this point and the next.

src/rdkafka.h
```diff
@@ -4205,6 +4372,947 @@ rd_kafka_interceptor_add_on_request_sent (
...
+ *                Valid values are: CreateTopics, DeleteTopics,
+ *                                  CreatePartitions,
+ *                                  AlterConfigs, DescribeConfigs.
+ *
+ * @returns a new AdminOptions object (which must be freed with
+ *          rd_kafka_AdminOptions_destroy()), or NULL if \p for_api was set to
+ *          an unknown API name.
+ */
+RD_EXPORT rd_kafka_AdminOptions_t *
+rd_kafka_AdminOptions_new (rd_kafka_t *rk, const char *for_api);
```
mhowlett: why not use an enum for `for_api`?

src/rdkafka_admin.c
```diff
@@ -0,0 +1,2760 @@
...
+ * @brief Admin API implementation.
+ *
+ * The public Admin API in librdkafka exposes a completely asynchronous
+ * interface where the initial request API (e.g., ..CreateTopics())
+ * is non-blocking and returns immediately, and the application polls
+ * a ..queue_t for the result.
+ *
+ * The underlying handling of the request is also completely asynchronous
+ * inside librdkafka, for two reasons:
+ *  - everything is async in librdkafka so adding something new that isn't
```
mhowlett: v.minor: i don't think these 2 points are worth listing.

mhowlett:
a few outstanding comments, otherwise i'm happy.
 
I thought about the public API in a fair bit of detail, which i think is the most critical thing to get right. In the interests of time i only looked very briefly at the implementation (but will give it plenty of test time as i make the .NET binding).

src/rdkafka_admin.c
```diff
@@ -0,0 +1,2760 @@
...
+ *
+ *
+ */
+
+rd_kafka_DeleteTopic_t *rd_kafka_DeleteTopic_new (const char *topic) {
+        size_t tsize = strlen(topic) + 1;
+        rd_kafka_DeleteTopic_t *del_topic;
+
+        /* Single allocation */
+        del_topic = rd_malloc(sizeof(*del_topic) + tsize);
```
mhowlett: fancy

src/rdkafka.h
```diff
@@ -4205,6 +4372,947 @@ rd_kafka_interceptor_add_on_request_sent (
...
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success, or
+ *          RD_KAFKA_RESP_ERR__INVALID_ARG if timeout was out of range in which
+ *          case an error string will be written \p errstr.
+ *
+ * @remark This option is valid for CreateTopics, DeleteTopics and 
+ *         CreatePartitions.
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_AdminOptions_set_operation_timeout (rd_kafka_AdminOptions_t *options,
```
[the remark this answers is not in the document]
edenhill: I did not want to maintain an additional type for this specific use only, but I guess you are right.

src/rdkafka.h
edenhill: in rd_kafka_[consumer_]poll() queue

src/rdkafka.h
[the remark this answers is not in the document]
edenhill: Primarily that, but also because some of the Admin operations have broker-side timeouts ( e.g., createTopics and deleteTopics) which tells the broker how long it may block waiting for an operation to propagate throughout the cluster.
But from a user's perspective we also want a request-level timeout that includes broker lookup, connection, request time, and the operation time.
In a previous PR I tried to encode both these timeouts in a single value but it was ugly and did not go down well with the Mr Math.

We could call the operation timeout just `timeout` as in the Java API, but with two timeouts in play I believe this is too ambigious, thus the operation_timeout and request_timeout.

src/rdkafka_admin.c
edenhill: More docs can't be bad?! :)

Issue #1794, opened 2018-05-04
Producer message delivery fails when broker connectivity is down


krishanfrog:
How to reproduce
===========
I have a cluster of 4 brokers with a replication factor of 2 on all topics.

One of the broker loses network connectivity with all external services. The process is up but the instance just loses network connectivity for 10 minutes.

The group coordinator recognizes this and reassigns the leaders of the partitions assigned to this broker.

But none of the producer messages for these partitions are delivered. All messages throw an error

```
required feature not supported by broker
```

Once the connectivity is back up the leaders are reassigned back and now new messages are delivered.

Why did the messages fail to be sent to the new leader during the issue window? Is the flow different depending if the broker is down versus broker is up but not able to connect

Checklist
=========

Producer librdkafka version - 0.11.3
Apache Kafka version: 1.0.1

```
producer configuration: {
		"metadata.broker.list": "X:9092,Y:9092,Z:9092,U:9092",
		"max.in.flight.requests.per.connection": 5,
		"queue.buffering.max.messages": 50000,
		"queue.buffering.max.kbytes": 153600,
		"queue.buffering.max.ms": 100,
		"message.send.max.retries": 5,
		"retry.backoff.ms": 100,
		"batch.num.messages": 50,
		"dr_msg_cb": true,
		"log.connection.close": false,
		"socket.keepalive.enable": true,
		"statistics.interval.ms": 60000
	}
"topicConfig": {
		"request.required.acks": 1,
		"message.timeout.ms": 900000
	}
```
Operating system: Ubuntu 14.04

krishanfrog:
Some more information 

I have 40 producers. The produce message rate is not completely uniform across all producers. It ranges from 1000 messages per sec to 2500 messages per sec in these producers.

Issue 1 - Messages dropped
For 10 of these producers all messages attempted during the 1st 3 minutes of the issue window dropped and were not delivered (I got a failed delivery report for them after 3 minutes). After 3 minutes new messages were delivered to the new leader in the kafka cluster.

Issue 2 - Messages delivered out of order
For the other 30 producers no messages were delivered during the 1st 3 minutes of the issue window. After 3 minutes new messages were delivered to the new leader in the kafka cluster. After the original broker was back up the old messages in the 1st 3 minute window are now delivered 

@edenhill What are the reasons for these behaviours?

Is it expected to take upto 3 minutes to direct messages to the new leader of the cluster? I have verified that the leader rebalance happened immediately after the broker lost connectivity

[...]

PR #1799, opened 2018-05-09
Minor documentation updates from replies to #1794

briot:
From the very useful feedback you gave in #1794.
I am not sure of the best place for this kind of documentation, so I have duplicated part of it in the CONFIGURATION.md (which is a nice summary of all settings) and in rdkafka.h (which I think is a nice place where to describe the interaction between all the settings).


CONFIGURATION.md
```diff
@@ -100,7 +100,7 @@ queue.buffering.max.ms                   |  P  | 0 .. 900000     |             0
 linger.ms                                |  P  |                 |               | Alias for `queue.buffering.max.ms`
 message.send.max.retries                 |  P  | 0 .. 10000000   |             2 | How many times to retry sending a failing MessageSet. **Note:** retrying may cause reordering. <br>*Type: integer*
 retries                                  |  P  |                 |               | Alias for `message.send.max.retries`
-retry.backoff.ms                         |  P  | 1 .. 300000     |           100 | The backoff time in milliseconds before retrying a protocol request. <br>*Type: integer*
+retry.backoff.ms                         |  P  | 1 .. 300000     |           100 | The backoff time in milliseconds before retrying a protocol request.<br>*Type: integer*
```
edenhill: CONFIGURATION.md is automatically generated on build from rdkafka_conf.c, so you'll need to update that file, rerun 'make', and then commit both files.

src/rdkafka.h
```diff
@@ -3,24 +3,24 @@
  *
  * Copyright (c) 2012-2013 Magnus Edenhill
  * All rights reserved.
- * 
+ *
  * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met: 
- * 
+ * modification, are permitted provided that the following conditions are met:
+ *
```
edenhill: No whitespace diffs, please

briot:
Updated as per your comments. Thanks for taking the time to explain (I'll have further changes for the doc based on other email threads, so this was "testing the water"). The whitespace changes came from my default .vimrc that automatically remove end-of-line spaces. I am curious to know why you chose to preserve them, they are hard to see in diffs.

edenhill:
librdkafka sources has a mix of tab and whitespace indention and some trailing whitespacing, this all needs to be cleaned up, but instead of doing it incrementally the plan is to put a code style checker in place (astyle) and fix this across the board in one single dedicated commit.
So until that happens I ask people not to make whitespace diffs :)

Looking forward to more documentation updates, they're very valuable!
Meatier docs should go into INTRODUCTION.md (which should probably be renamed MANUAL.md at some point).

src/rdkafka_conf.c
```diff
@@ -829,7 +829,10 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  "Local message timeout. "
 	  "This value is only enforced locally and limits the time a "
 	  "produced message waits for successful delivery. "
-          "A time of 0 is infinite.",
+      "A time of 0 is infinite. "
```
edenhill: nit: indention seems off here (should be 8 whitespaces per tab), aligned with the lines above.

src/rdkafka.h
```diff
@@ -1332,8 +1332,9 @@ void rd_kafka_conf_set_dr_cb(rd_kafka_conf_t *conf,
  * the result of the produce request.
  * 
  * The callback is called when a message is succesfully produced or
- * if librdkafka encountered a permanent failure, or the retry counter for
- * temporary errors has been exhausted.
+ * if librdkafka encountered a permanent failure.
+ * Delivery errors occur when the retry count is exceeded, or when the
+ * message.timeout.ms timeout is exceeded.
```
edenhill: or there is a permanent error (e.g., UNKNOWN_TOPIC_OR_PARTITION)

INTRODUCTION.md
```diff
@@ -539,7 +539,9 @@ for a given partition by calling `rd_kafka_consume_start()`.
 
 After a topic+partition consumer has been started librdkafka will attempt
 to keep `queued.min.messages` messages in the local queue by repeatedly
-fetching batches of messages from the broker.
+fetching batches of messages from the broker. Librdkafka will fetch all
```
edenhill: lower case librdkafka
Regards,
The branding department

src/rdkafka.h
```diff
@@ -1332,8 +1332,9 @@ void rd_kafka_conf_set_dr_cb(rd_kafka_conf_t *conf,
  * the result of the produce request.
  * 
  * The callback is called when a message is succesfully produced or
- * if librdkafka encountered a permanent failure, or the retry counter for
- * temporary errors has been exhausted.
+ * if librdkafka encountered a permanent failure.
+ * Delivery errors occur when the retry count is exceeded, or when the
```
edenhill: or there is a permanent error (e.g., UNKNOWN_TOPIC_OR_PARTITION)

src/rdkafka.h
```diff
@@ -1377,13 +1378,19 @@ void rd_kafka_conf_set_consume_cb (rd_kafka_conf_t *conf,
...
- * in performing other operations along with the assinging/revocation,
+ * in performing other operations along with the assigning/revocation,
  * such as fetching offsets from an alternate location (on assign)
  * or manually committing offsets (on revoke).
  *
  * @remark The \p partitions list is destroyed by librdkafka on return
  *         return from the rebalance_cb and must not be freed or
  *         saved by the application.
+ *
+ * @remark The \p partitions list should not be modified for instance by
```
edenhill: It is actually okay to modify the partitions list to set initial offsets for each partition.
So maybe reword this as a precaution to only deliberately modifying the partitions list if the initial starting offsets should be changed. Otherwise pass a copy of the partitions list (using rd_kafka_topic_partition_list_copy() to functions which may modify it.

nit: position() is the current consume position, it is typically outdated in ASSIGN_PARTITIONS.

src/rdkafka.h
```diff
@@ -1516,6 +1523,7 @@ void rd_kafka_conf_set_log_cb(rd_kafka_conf_t *conf,
  * Function arguments:
  *   - \p rk - Kafka handle
  *   - \p json - String containing the statistics data in JSON format
+ *     See https://github.com/edenhill/librdkafka/wiki/Statistics
```
edenhill: Move this below the argument list.

src/rdkafka.h
```diff
@@ -2907,6 +2919,10 @@ rd_kafka_committed (rd_kafka_t *rk,
  * of the last consumed message + 1, or RD_KAFKA_OFFSET_INVALID in case there was
  * no previous message.
  *
+ * @remark  In this context the last consumed message is the offset consumed
+ *          by the current librdkafka instance and, in case of rebalancing, not
+ *          necessarily the last message consumed from the kafka queue.
```
edenhill: change last line to
".. the last message fetched from the partition."

src/rdkafka.h
```diff
@@ -2957,6 +2973,22 @@ rd_kafka_position (rd_kafka_t *rk,
  * `rd_kafka_topic_new()`.
  *
  * `rd_kafka_produce()` is an asynch non-blocking API.
+ * See `rd_kafka_conf_set_dr_msg_cb` on how to setup a callback to be called
+ * once the delivery status (success or failure) is known.
+ *
+ * Since producing is asynchronous, you should call `rd_kafka_flush()` before
+ * you destroy the producer. Otherwise, any outstanding messages will be
+ * discarded and you won't know. An alternative is to check for delivery
```
edenhill: change to "silently discarded."

src/rdkafka.h
```diff
@@ -2957,6 +2973,22 @@ rd_kafka_position (rd_kafka_t *rk,
...
+ * reports via `rd_kafka_conf_set_dr_msg_cb()`.
```
edenhill: Remove the alternative line, it is already mentioned above.

src/rdkafka.h
```diff
@@ -2957,6 +2973,22 @@ rd_kafka_position (rd_kafka_t *rk,
  * `rd_kafka_topic_new()`.
  *
  * `rd_kafka_produce()` is an asynch non-blocking API.
+ * See `rd_kafka_conf_set_dr_msg_cb` on how to setup a callback to be called
+ * once the delivery status (success or failure) is known.
```
edenhill: add: "The delivery report callback is triggered by the application calling rd_kafka_poll() (at regular intervals) or rd_kafka_flush() (at termination)."

src/rdkafka.h
```diff
@@ -2957,6 +2973,22 @@ rd_kafka_position (rd_kafka_t *rk,
...
+ *
+ * Since producing is asynchronous, you should call `rd_kafka_flush()` before
+ * you destroy the producer. Otherwise, any outstanding messages will be
+ * discarded and you won't know. An alternative is to check for delivery
+ * reports via `rd_kafka_conf_set_dr_msg_cb()`.
+ *
+ * When errors occur, librdkafka automatically retries to produce the
```
edenhill: "When temporary errors occur"

src/rdkafka.h
```diff
@@ -2957,6 +2973,22 @@ rd_kafka_position (rd_kafka_t *rk,
...
+ * messages. Retries are triggered after retry.backoff.ms but only if the
```
edenhill: change "but only if" to "and when"

src/rdkafka.h
```diff
@@ -2957,6 +2973,22 @@ rd_kafka_position (rd_kafka_t *rk,
...
+ * leader broker for the given partition is available. Otherwise, librdkafka
+ * falls back to polling the topic metadata to monitor when a new leader is
+ * elected (see the topic.metadata.refresh.fast.interval.ms and
+ * topic.metadata.refresh.interval.ms configurations) and then performs a
+ * retry. A delivery error will always occur if overall it took more than
```
edenhill: change last line to "A delivery error will occur if the message could not be produced within `message.timeout.ms`."

src/rdkafka.h
```diff
@@ -2957,6 +2973,22 @@ rd_kafka_position (rd_kafka_t *rk,
...
+ * message.timeout.ms.
```
edenhill: Add: "See the Message reliability chapter in INTRODUCTION.md for more information"

src/rdkafka_conf.c
```diff
@@ -829,7 +829,10 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	  "Local message timeout. "
 	  "This value is only enforced locally and limits the time a "
 	  "produced message waits for successful delivery. "
-          "A time of 0 is infinite.",
+	  "A time of 0 is infinite. "
```
edenhill: mix of tabs and spaces. Use 8 whitespaces as tab indention.

briot: That's a thorough review, thanks !
I think and hope I have properly taken into account all the changes while preserving the individual commits (since they refer to past issues).
Only your last comment for the whitespace issue is mysterious to me. As it is now in the patch, the line "A time of 0 is infinite" seems to be indented exactly as all other lines in this file (i.e. vim tells me it starts at column 4-7), which I think was not the case before and had to be corrected.

edenhill: Right, this is one of those cases where the existing code uses the old tab-for-indent.
What I typically do is change the existing related statement to the new space-for-indent style.
But leave it as it is, it is fine :)

edenhill:
Great stuff, thank you!

PR #1816, opened 2018-05-23
Fixed murmur2 partitioner to make it compatible with java version

lins05:
* In [kafka java producer](https://github.com/apache/kafka/blob/1.1.0/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L69), the partition is `abs(hash(key)) % num_partition`
* while in librdkafka it's`hash(key) % num_partition`. 

This incompatibility would cause trouble when the python producer and java producer are used to write to the same topic.


edenhill:
This is great, thank you!

PR #1823, opened 2018-05-25
Some small enhancements to gen-ssl-certs.sh

jcaesar:
I promised a pull request in https://github.com/edenhill/librdkafka/issues/1765 and forgot about it. Not necessarily bad, because I found a few more things. In detail, this changes:

 - specify the keyalg to be RSA as it may default to DSA, which Java/Kafka(?) does not cope well with
 - pass the validity period when signing a non-java client key
 - change the key length to the current "recommended" minimum

(The last one might not be the best idea, if you are also using this script in some kind of CI setting. Please tell me what you think.)

### DIFF
--- a/tests/gen-ssl-certs.sh
+++ b/tests/gen-ssl-certs.sh
@@ -60,7 +60,7 @@ elif [[ $OP == "server" && ! -z "$CA_CERT" && ! -z "$PFX" && ! -z "$CN" ]]; then
 
     #Step 1
     echo "############ Generating key"
-    keytool -storepass "$PASS" -keypass "$PASS" -keystore ${PFX}server.keystore.jks -alias localhost -validity $VALIDITY -genkey <<EOF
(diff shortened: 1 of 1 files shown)


tests/gen-ssl-certs.sh
```diff
@@ -131,7 +131,7 @@ EOF
     else
 	# Standard OpenSSL keys
 	echo "############ Generating key"
-	openssl genrsa -des3 -passout "pass:$PASS" -out ${PFX}client.key 1024 
+	openssl genrsa -des3 -passout "pass:$PASS" -out ${PFX}client.key 2048
```
edenhill: It is not used in any tests so this change is fine

edenhill:
Great stuff, thanks!

Is this ready to be merged?

jcaesar:
Checked it again, should be ok.

edenhill:
Thanks for this!

PR #1826, opened 2018-05-28
Proper log message on SSL connection close

edenhill:
Includes honouring log.connection.close=true for SSL disconnects.


src/rdkafka_int.h
```diff
@@ -413,17 +413,14 @@ static RD_UNUSED RD_INLINE
 rd_kafka_resp_err_t rd_kafka_set_last_error (rd_kafka_resp_err_t err,
 					     int errnox) {
         if (errnox) {
-#ifdef _MSC_VER
-                /* This is the correct way to set errno on Windows,
+                /* MSVC:
```
mhowlett: i don't think you want this comment here anymore.

PR #1850, opened 2018-06-15
Add support for set_events(EVENT_ERROR)

edenhill:
.. which allows errors to be emitted as events.

Needed by Golang client issue 200.


rnpridgeon:
LGTM

Issue #1859, opened 2018-06-24
Callback-based IO event notifier


fede1024:
I'm working on improving the way librdkafka interacts with the rust futures for [rust-rdkafka](https://github.com/fede1024/rust-rdkafka). The main challenge is to notify the event loop whenever there are new events to be processed. I believe issue #749 asked for basically the same feature, but the solution provided there uses file-descriptor, while in the rust case (and I suppose in other cases as well) it would be more efficient to just have a callback that is triggered whenever a queue goes from 0 to a non-0 number of events.

I made a small change to the library [here](https://github.com/fede1024/librdkafka/commit/87898c5ca8137c370809ecb26f0e12507eb47946) and I've successfully tested this approach from the rust library, but before opening a PR (with proper tests and documentation), I'd like to know if there is any downside to this approach, if there is any reason why this woud be a bad idea, etc.

Thanks! :)

edenhill:
Thanks for the background, the use-case makes sense and should be catered for by librdkafka.

Your change looks good, go ahead with creating a PR, I left some comments on your initial commits.

PR #1860, opened 2018-06-24
Add callback-based IO event notifications

fede1024:
See issue #1859.

I've added a test based on 0040. I'm not super happy with the sleep-based waiting loop, but even with a better approach I don't think we'd be able to reduce the execution time of the test by more than 2 seconds. Let me know if you have any suggestion.

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -2484,12 +2484,33 @@ size_t rd_kafka_queue_length (rd_kafka_queue_t *rkqu);
  *
  * librdkafka will maintain a copy of the \p payload.
  *
+ * @remark IO and callback event triggering are mutually exclusive.
  * @remark When using forwarded queues the IO event must only be enabled
  *         on the final forwarded-to (destination) queue.
  */
 RD_EXPORT
 void rd_kafka_queue_io_event_enable (rd_kafka_queue_t *rkqu, int fd,
 				     const void *payload, size_t size);
+
+/**
+ * @brief Enable callback event triggering for queue.
+ *
+ * The callback will be called from an internal librdkafka thread
+ * when a new element is enqueued on a previously empty queue.
+ *
+ * To remove event triggering call with \p event_cb = NULL.
+ *
+ * @remark IO and callback event triggering are mutually exclusive.
+ * @remark Since the callback may be triggered from internal librdkafka
+ *         threads, the application must not perform any pro-longed work in
+ *         the callback, or call any librdkafka APIs (for the same rd_kafka_t
+ *         handle).
+ */
+RD_EXPORT
+void rd_kafka_queue_cb_event_enable (rd_kafka_queue_t *rkqu,
+                                     void (*event_cb) (rd_kafka_t *rk,
+                                                       void *opaque),
+                                     void *opaque);
 
 /**@}*/
 
--- a/src/rdkafka_queue.c
+++ b/src/rdkafka_queue.c
@@ -719,6 +719,8 @@ void rd_kafka_q_io_event_enable (rd_kafka_q_t *rkq, int fd,
                 qio->fd = fd;
                 qio->size = size;
                 qio->payload = (void *)(qio+1);
+                qio->event_cb = NULL;
+                qio->event_cb_opaque = NULL;
                 memcpy(qio->payload, payload, size);
         }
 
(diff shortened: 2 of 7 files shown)


src/rdkafka.h
```diff
@@ -2370,13 +2370,34 @@ size_t rd_kafka_queue_length (rd_kafka_queue_t *rkqu);
  *
  * librdkafka will maintain a copy of the \p payload.
  *
+ * @remark IO and callback event triggering are mutually exclusive.
```
fede1024: I think this should actually be "@remark FD and callback event..." since they are both IO events.

src/rdkafka.h
```diff
@@ -2370,13 +2370,34 @@ size_t rd_kafka_queue_length (rd_kafka_queue_t *rkqu);
  *
  * librdkafka will maintain a copy of the \p payload.
  *
+ * @remark FD and callback IO event triggering are mutually exclusive.
```
edenhill: IO in this context refers to the file descriptor triggering.
I don't think the callback based triggering should be named IO since it doesn't really do any IO or can't be directly integrated with an IO scheduler in the way the fd-based one can.

So please revert your IO -> FD changes, and go with the following nomenclature:
IO event = FD
callback event = callback

src/rdkafka.h
```diff
@@ -2370,13 +2370,34 @@ size_t rd_kafka_queue_length (rd_kafka_queue_t *rkqu);
...
  * @remark When using forwarded queues the IO event must only be enabled
  *         on the final forwarded-to (destination) queue.
  */
 RD_EXPORT
 void rd_kafka_queue_io_event_enable (rd_kafka_queue_t *rkqu, int fd,
 				     const void *payload, size_t size);
 
+/**
+ * @brief Enable IO event triggering for queue using a callback.
```
edenhill: drop "IO"

src/rdkafka.h
```diff
@@ -2370,13 +2370,34 @@ size_t rd_kafka_queue_length (rd_kafka_queue_t *rkqu);
...
+ *
+ * The callback will be called from an internal librdkafka thread
+ * when a new element is enqueued on a previously empty queue.
+ *
+ * To remove event triggering call with \p event_cb = NULL.
+ *
+ * @remark FD and callback IO event triggering are mutually exclusive.
```
edenhill: FD -> IO

src/rdkafka.h
```diff
@@ -2370,13 +2370,34 @@ size_t rd_kafka_queue_length (rd_kafka_queue_t *rkqu);
...
+ * @remark Since the callback may be triggered from internal librdkafka
+ *         threads, the application must not perform any pro-longed work in
+ *         the callback, or call any librdkafka APIs (for the same rd_kafka_t
+ *         handle).
+ */
+RD_EXPORT
+void rd_kafka_queue_io_event_cb_enable (rd_kafka_queue_t *rkqu,
```
edenhill: drop _io_

src/rdkafka_queue.c
```diff
@@ -722,6 +724,46 @@ void rd_kafka_queue_io_event_enable (rd_kafka_queue_t *rkqu, int fd,
 }
 
 
+/**
+ * @brief Enable or disable(event_cb==NULL) callback-based wake-ups for queue
+ */
+void rd_kafka_q_io_event_cb_enable (rd_kafka_q_t *rkq,
```
edenhill: drop _io_

src/rdkafka_queue.c
```diff
@@ -722,6 +724,46 @@ void rd_kafka_queue_io_event_enable (rd_kafka_queue_t *rkqu, int fd,
...
+
+        if (event_cb) {
+                rkq->rkq_qio = qio;
+        }
+
+        mtx_unlock(&rkq->rkq_lock);
+
+}
+
+void rd_kafka_queue_io_event_cb_enable (rd_kafka_queue_t *rkqu,
```
edenhill: dito

tests/0083-io_event_cb.c
```diff
@@ -0,0 +1,210 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2012-2013, Magnus Edenhill
```
edenhill: Set copyright to 2018

tests/0083-io_event_cb.c
```diff
@@ -0,0 +1,210 @@
...
+         * 4) Expect CB
+         * 5) Consume the available messages
+         * 6) Wait 1 interval expecting no CB.
+         * 7) Produce remaing half
+         * 8) Expect CB
+         * 9) Done.
+         */
+        while (recvd < msgcnt) {
+                TEST_SAY("Waiting for event\n");
+                int event_count = wait_event_cb(1 * wait_multiplier);
```
edenhill: variable definitions must go at the top of the scope, before any function calls.

tests/0083-io_event_cb.c
```diff
@@ -0,0 +1,210 @@
...
+                TEST_ASSERT(event_count <= 1, "Event cb called %d times", event_count);
+
+                if (event_count == 1) {
+                        TEST_SAY("Events received: %d\n", event_count);
```
edenhill: dito

src/rdkafka.h
fede1024: I see, makes sense. Thank you.

src/rdkafka.h
fede1024: I'm going to replace `io` with `cb`: `rd_kafka_queue_cb_event_enable` (similarly to `rd_kafka_queue_io_event_enable`). Please let me know if you would rather call it `rd_kafka_queue_event_cb_enable`.

src/rdkafka.h
fede1024: Dropped.

src/rdkafka.h
```diff
@@ -2359,7 +2359,7 @@ size_t rd_kafka_queue_length (rd_kafka_queue_t *rkqu);
 
 
 /**
- * @brief Enable IO event triggering for queue.
+ * @brief Enable FD event triggering for queue.
```
edenhill: This should still be IO

src/rdkafka_queue.h
```diff
@@ -281,6 +285,11 @@ void rd_kafka_q_io_event (rd_kafka_q_t *rkq) {
 	if (likely(!rkq->rkq_qio))
 		return;
 
+        if (rkq->rkq_qio->event_cb) {
+                rkq->rkq_qio->event_cb(rkq->rkq_rk, rkq->rkq_qio->event_cb_opaque);
+                return;
+	}
```
edenhill: tab -> 8 whitespaces

tests/CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ set(
     0078-c_from_cpp.cpp
     0079-fork.c
     0081-fetch_max_bytes.cpp
+    0083-io_event_cb.c
```
edenhill: this file will need to be added to win32/tests/tests.vcxproj too

edenhill:
The branch is conflicting, rebase on latest master.

src/rdkafka_queue.h
fede1024: Sorry I missed this.

tests/CMakeLists.txt
fede1024: Will do. I'll also rename the test to cb_event, according to the naming convention.

fede1024:
I've rebased, squashed and pushed again. Let me know if you have any additional comment.

Is there any chance this could make it into 0.11.5 or is it too close to release?

edenhill:
Thank you!

Issue #1858, opened 2018-06-22
High CPU Usage caused by librdkafka


wuqingjun:
Description - updated
===========
 
The dump file shows 15 threads, 4 of them shows very long CPU time.  I am posting the call stacks of the 4 threads here, if you like, you can also open the attached file. 

The symptom is similar to this issue but it should have been fixed: https://github.com/edenhill/librdkafka/issues/1569 

And I really see the Kafka brokers are down and up at the time this issue happened, but I am unable to reproduce this. 

Dump file info, I got this by running ~*k command in windbg:
...
```
  10  Id: 2780.2cac Suspend: 0 Teb: 0000003d`ad520000 Unfrozen
 # Child-SP          RetAddr           Call Site
00 (Inline Function) --------`-------- librdkafka!rd_clock+0x46 [e:\src\one\oss\librdkafka\src\rdtime.h @ 82] 
01 0000003d`ae7ffa80 00007ffe`ba56ba00 librdkafka!rd_kafka_broker_serve+0x13f [e:\src\one\oss\librdkafka\src\rdkafka_broker.c @ 2493] 
02 0000003d`ae7ffac0 00007ffe`ba56b021 librdkafka!rd_kafka_broker_ua_idle+0xe0 [e:\src\one\oss\librdkafka\src\rdkafka_broker.c @ 2552] 
03 0000003d`ae7ffb10 00007ffe`ba5c3f5b librdkafka!rd_kafka_broker_thread_main+0x2e1 [e:\src\one\oss\librdkafka\src\rdkafka_broker.c @ 3459] 
04 0000003d`ae7ffd70 00007ffe`d3708364 librdkafka!_thrd_wrapper_function+0x1b [e:\src\one\oss\librdkafka\src\tinycthread.c @ 627] 
05 0000003d`ae7ffda0 00007ffe`d40d70b1 kernel32!BaseThreadInitThunk+0x14
06 0000003d`ae7ffdd0 00000000`00000000 ntdll!RtlUserThreadStart+0x21

  11  Id: 2780.3be4 Suspend: 0 Teb: 0000003d`ad522000 Unfrozen
 # Child-SP          RetAddr           Call Site
00 0000003d`ae8ff178 00007ffe`d085795f ntdll!NtWaitForMultipleObjects+0x14
01 0000003d`ae8ff180 00007ffe`d085785e KERNELBASE!WaitForMultipleObjectsEx+0xef
[...]

PR #1870, opened 2018-07-02
Block broker thread on event queue instead of IO if IO is irrelevant in current state

edenhill:
This fixes the case where idle (typically bootstrap) broker threads
are consuming CPU on Windows (with a low socket.blocking.max.ms) due
to aggressively short timeouts on IO.

The drawback is that it might take up to 1000ms to detect disconnects
on idle connections, which should not be a problem in practice.

Related issue:
 * #1858
 * https://github.com/confluentinc/confluent-kafka-dotnet/issues/468

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -2505,40 +2505,84 @@ static int rd_kafka_broker_ops_serve (rd_kafka_broker_t *rkb, int timeout_ms) {
 static void rd_kafka_broker_serve (rd_kafka_broker_t *rkb,
                                    rd_ts_t abs_timeout) {
         rd_ts_t now;
-        int initial_state = rkb->rkb_state;
         int remains_ms = rd_timeout_remains(abs_timeout);
+        int ops_timeout = RD_POLL_NOWAIT;
+        int io_timeout;
+
+        /* As a defensive measure, don't block forever.
+         * When blocking on the queue we will want to detect IO disconnects
+         * eventually.
+         * 1 wakeup/second is reasonable. */
+        if (unlikely(remains_ms == RD_POLL_INFINITE))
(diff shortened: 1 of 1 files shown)


wuqingjun:
Is this PR under review?

src/rdkafka_broker.c
```diff
@@ -2505,40 +2505,84 @@ static int rd_kafka_broker_ops_serve (rd_kafka_broker_t *rkb, int timeout_ms) {
...
-        /* Serve broker ops */
-        if (rd_kafka_broker_ops_serve(rkb,
-                                      !rkb->rkb_transport ?
-                                      remains_ms : RD_POLL_NOWAIT))
-                remains_ms = RD_POLL_NOWAIT;
+        /* As a defensive measure, don't block forever.
+         * When blocking on the queue we will want to detect IO disconnects
+         * eventually.
+         * 1 wakeup/second is reasonable. */
+        if (unlikely(remains_ms == RD_POLL_INFINITE))
```
rnpridgeon: Does it make sense to defend against all 'excessive' values for `abs_timeout` as opposed to just `RD_POLL_INFINITE`?  A value of 300000ms still seems fairly intrusive given we would wakeup infinite blocking calls every 1 second.

edenhill: Right, so a MIN(remains_ms, 1000) makes sense!

edenhill:
I'm not sure this is the best approach.
We should considering using forwardable condvars, rather than the entire queue, but that's a bigger change.

PR #1930 enables low-latency wakeups on Windows by using TCP sockets, so that might a better way forward now.

edenhill:
I'm keeping this open as a reminder to check if it still applies

Issue #815, opened 2016-09-30
OpenSSL global destructor fix


edenhill:
Issue seen in https://github.com/arnaud-lb/php-rdkafka/issues/59 where both librdkafka and curl use OpenSSL. When libraries are unloaded from memory on PHP termination the global callbacks will reference unmapped memory causing a crash.

PostgreSQL's fix for the same is here:
https://www.postgresql.org/message-id/20081203200427.0A23B7545A4@cvs.postgresql.org

dzuelke:
This was my dumb attempt, but it doesn't seem to be working: https://github.com/edenhill/librdkafka/compare/master...dzuelke:unset_ssl_callbacks

edenhill:
rd_kafka_transport_term() is never called :)
I just left it in there as documentation as why it shouldnt be used..

edenhill:
I have a fix on the way though

dzuelke:
Oh I see now. My bad =)

dzuelke:
I can test right away once you got a fix ready!

edenhill:
Awesome, it will be on the partition_changes branch in a few minutes, just need it to pass the SSL and SASL tests.

edenhill:
There, go crazy!

dzuelke:
No more crash! Good stuff @edenhill :)

edenhill:
wohoo! Thanks for all the fantastic work on this!
The partition_changes branch will end up in master in a week or two.

edenhill:
Do you need it cherry-picked in master sooner than that?

dzuelke:
Nah I think I'll wait for 0.9.2. If I want to ship PHP+Kafka support on Heroku earlier I can always use that branch, but I won't be able to do that next week anyway, so the week after sounds mighty fine :)

dzuelke:
Also general 👊🏼 for your A++ responsiveness to issues. Remind me to buy you a drink if we ever run into each other; I got the memory of a goldfish so I'll certainly forget.

edenhill:
I clearly remember you saying 2 drinks

edenhill:
This issue is now fixed on master, please try to verify the fix in your environment.
Thank you.

dzuelke:
Still works with master! Thanks!

PR #1882, opened 2018-07-14
Provide SSL callbacks to handle retrieving and processing certificates

noahdav:
We have a custom security provider. In our case we need to be able to allow this security provider to auth the certificate from the broker as well as retrieve the client certificate. Since this is custom logic providing a callback to the client seems to handle this scenario.


noahdav:
@edenhill I am seeing this failure in doozer/target/stretch-mips.  I am confused on why this is failing to compile since it is working on other arch.  I do not see any arch specific code here.  I do not have access to this arch to test this on so I am having troubles to fix this break.  Is there any way to test this offline?

edenhill:
The MIPS build is now fixed on master, so try rebasing your branch.

noahdav:
@edenhill rebasing did not seem to help.  I tried to asking on OpenSSL but the only help that I received is that there are some issues with the headers for MIPS in 1.0.2.  At this point I am not sure the best way to proceed.  Do you have any suggestions?  If not how do you feel about adding #ifdef to exclude MIPS since this does not seem to work for MIPS?

Since in MIPS apparently the X509_STORE_CTX does not have the cert field we cannot get the peer cert to validate.  X509_store_ctx_get_current_cert will only return the error and in the normal case when no errors occur it returns NULL.

edenhill:
Okay, let's do this:
 * Add a configure check to configure.librdkafka that checks for the ctx field
 * #ifdef out the relevant (smallest) section of rdkafka_transport.c using the define set by the configure check
 * Add an error return to rd_kafka_conf_set_ssl_verify_cb() that checks the ifdef too, return ERR__NOT_IMPLEMENTED if not set.
 * Document the `@return ..` for  ..set_ssl..() and mention it is not supported on MIPS platforms.

noahdav:
@edenhill After digging into this further it looks like struct x509_store_ctx_st is not defined.  

rdkafka_transport.c:682:23: error: dereferencing pointer to incomplete type 'X509_STORE_CTX {aka struct x509_store_ctx_st}'
             cert = ctx->cert;
                       ^~
In openssl/ssl.h:

# ifndef OPENSSL_NO_DEPRECATED
#  ifndef OPENSSL_NO_X509
#   include <openssl/x509.h>
#  endif

However I do not see anywhere that we are defining these which would cause x509.h to not be included and therefore x509_vfy.h to not get included either.  I can certainly add a check to see if we are building on MIPS but not sure what other checks we should use since the struct is not defined.

noahdav:
@edenhill What do we need to do in order to get this merged into master?

src/rdkafka_transport.c
```diff
@@ -694,7 +822,16 @@ static int rd_kafka_transport_ssl_connect (rd_kafka_broker_t *rkb,
 		goto fail;
 #endif
 
-        rd_kafka_transport_ssl_clear_error(rktrans);
+    if (rktrans->rktrans_rkb->rkb_rk->rk_conf.ssl.handshake_info) {
```
edenhill: This should be made into standard debug.
If it isn't too noisy (like up to 10 lines per SSL handshake) then it can use the `security` context, otherwise we should add a `sslhandshake` debug context - if there is value to it.
I believe one can use the `openssl s_client ..` to debug SSL issues just as well?

src/rdkafka_transport.c
```diff
@@ -877,6 +1014,147 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
 		}
 	}
 
+    /* Register the peer verify callback on the context */
+    if (rk->rk_conf.ssl.cert_verify_cb) {
+
+        unsigned int mode = SSL_VERIFY_PEER | SSL_VERIFY_CLIENT_ONCE | SSL_VERIFY_FAIL_IF_NO_PEER_CERT;
+
+        rd_kafka_dbg(rk, SECURITY, "SSL",
+            "Setting verification mode %u", mode);
```
edenhill: This doesnt seem very useful since the modes are hardcoded, remove it.

src/rdkafka_transport.c
```diff
@@ -877,6 +1014,147 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+    }
+
+    if (rk->rk_conf.ssl.cert_retrieve_cb) {
+
+        unsigned char* buffer = NULL;
+        long len = 0;
+
+        rd_kafka_dbg(rk, SECURITY, "SSL",
+            "Calling callback to retrieve client's public key certificate");
+        len = rk->rk_conf.ssl.cert_retrieve_cb(RD_KAFKA_CERTIFICATE_PUBLIC_KEY, &buffer, rk->rk_conf.opaque);
```
edenhill: pass an errstr + size so that we can have the user give us a meaningful error message when this fails.
Should return -1 on error instead of 0.

src/rdkafka_transport.c
```diff
@@ -877,6 +1014,147 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+        SSL_CTX_set_cert_verify_callback(
+            ctx,
+            verify_broker_callback,
+            rk);
+    }
+
+    if (rk->rk_conf.ssl.cert_retrieve_cb) {
+
+        unsigned char* buffer = NULL;
+        long len = 0;
```
edenhill: use ssize_t

src/rdkafka_transport.c
```diff
@@ -877,6 +1014,147 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+
+        SSL_CTX_set_cert_verify_callback(
+            ctx,
+            verify_broker_callback,
+            rk);
+    }
+
+    if (rk->rk_conf.ssl.cert_retrieve_cb) {
+
+        unsigned char* buffer = NULL;
```
edenhill: drop `unsigned`, not needed, we're not using it in any other apis

src/rdkafka_transport.c
```diff
@@ -877,6 +1014,147 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+        long len = 0;
+
+        rd_kafka_dbg(rk, SECURITY, "SSL",
+            "Calling callback to retrieve client's public key certificate");
+        len = rk->rk_conf.ssl.cert_retrieve_cb(RD_KAFKA_CERTIFICATE_PUBLIC_KEY, &buffer, rk->rk_conf.opaque);
+        if (len) {
```
edenhill: this is such a big block of code it should go into its own function that is called from here.

src/rdkafka_transport.c
```diff
@@ -877,6 +1014,147 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+
+                                PKCS12_free(p12);
+                                BIO_free(bio);
+                                EVP_PKEY_free(pkey);
+                                X509_free(cert);
+                                if (ca != NULL)
+                                    sk_X509_pop_free(ca, X509_free);
+
+                                goto fail;
+                            }
```
edenhill: `} else {`

src/rdkafka_transport.c
```diff
@@ -877,6 +1014,147 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+                    
+                    BIO_free(bio);
+                }
+                else
+                    rd_snprintf(errstr, errstr_size,
+                        "Failed to initialize BIO");
+        }
+        else
+            rd_kafka_dbg(rk, SECURITY, "SSL",
+                "Callback failed to return valid private key certificate");
```
edenhill: If this fails and sets errstr it also needs to `goto fail` to bug out, otherwise the errstr will be ignored.

src-cpp/HandleImpl.cpp
```diff
@@ -123,6 +122,34 @@ int RdKafka::open_cb_trampoline (const char *pathname, int flags, mode_t mode,
...
 }
 
+int RdKafka::cert_verify_cb_trampoline(unsigned char* cert, long len, void *opaque)
+{
+    RdKafka::HandleImpl *handle = static_cast<RdKafka::HandleImpl *>(opaque);
+
+#if WITH_SSL
+    return handle->cert_verify_cb_->cert_verify_cb(cert, len);
+#else
+    RD_UNUSED(cert);
```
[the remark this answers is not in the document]
noahdav: I can certainly remove these.  I was trying to be consistent with other places in the code which ifdef WITH_SSL such the config which sets these in rdkafka_conf.h/c.  On a side note I cannot seem to compile code these days without WITH_SSL being defined so this may not be necessary anymore.

src/rdkafka_conf.c
```diff
@@ -533,6 +533,20 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	_RK(ssl.keystore_password),
 	"Client's keystore (PKCS#12) password."
 	},
+    { _RK_GLOBAL, "ssl.handshake.info.enable", _RK_C_BOOL,
+    _RK(ssl.handshake_info),
```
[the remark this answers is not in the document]
noahdav: Sure I will change this to security.

noahdav:
@edenhill I believe that I addressed all of your comments.  Can you please take another look and let me know if there are still any outstanding issues.  Thanks:)

src-cpp/HandleImpl.cpp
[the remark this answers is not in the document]
edenhill: I mean the RD_UNUSED, I haven't seen any compilers warn about that.

As for !WITH_SSL: I'll investigate, thanks for letting me know!

src-cpp/rdkafkacpp.h
```diff
@@ -715,7 +715,61 @@ class RD_EXPORT OffsetCommitCb {
...
+    /**
+    * @brief Set certificate retrieve callback.
+    *
+    * The retrieval of the cetificate specified by the type will be returned by this callback.
+    * This is used to get the certificates required to create the SSL connection.
+    *
+    * @remark upon compleation the buffer must point to a valid buffer.
+    *
+    * @param buffer must point to valid data upon compleation.  It is the responsibility of the
+    *   implementation to ensure that the data is valid until the class is destructed.
```
edenhill: add that the CertRetrieveCb instance must outlive the RdKafka client instance.

src-cpp/rdkafkacpp.h
```diff
@@ -715,7 +715,61 @@ class RD_EXPORT OffsetCommitCb {
...
+    * @param type the type of buffer which will be retrieved by the callback.
+    *   CERTIFICATE_PUBLIC_KEY
+    *   CERTIFICATE_PRIVATE_KEY
+    *   CERTIFICATE_PRIVATE_KEY_PASS
+    *
+    * @remark When type is CERTIFICATE_PUBLIC_KEY or CERTIFICATE_PRIVATE_KEY then upon return the buffer
+    * must point to the certificate.  When type is For CERTIFICATE_PRIVATE_KEY_PASS then buffer
+    * must point to a narrow string containing the password of the private key.
+    *
+    * @returns the number of bytes in the returned buffer.
```
edenhill: add: or -1 on error, in which case the callback implemention must provide a human readable error string in `std::string &errstr`.

change return type to ssize_t
add errstr parameter

src-cpp/rdkafkacpp.h
```diff
@@ -715,7 +715,61 @@ class RD_EXPORT OffsetCommitCb {
...
+    */
+    virtual size_t cert_retrieve_cb(Type type, char** buffer) = 0;
```
edenhill: there's still plenty of `*` on the type side, they must go on the variable side:
`char **buffer`

src-cpp/rdkafkacpp_int.h
```diff
@@ -498,11 +554,22 @@ class ConfImpl : public Conf {
       return Conf::CONF_OK;
     }
 
+  Conf::ConfResult get(CertVerifyCb *&ssl_cert_verify_cb) const {
+      if (!rk_conf_)
+      return Conf::CONF_INVALID;
```
edenhill: fix indent

src-cpp/rdkafkacpp_int.h
```diff
@@ -498,11 +554,22 @@ class ConfImpl : public Conf {
...
+      ssl_cert_verify_cb = this->ssl_cert_verify_cb_;
+      return Conf::CONF_OK;
+  }
 
+  Conf::ConfResult get(CertRetrieveCb *&ssl_cert_retrieve_cb) const {
+      if (!rk_conf_)
+      return Conf::CONF_INVALID;
```
edenhill: fix indent

src-cpp/rdkafkacpp_int.h
```diff
@@ -498,11 +554,22 @@ class ConfImpl : public Conf {
       return Conf::CONF_OK;
     }
 
+  Conf::ConfResult get(CertVerifyCb *&ssl_cert_verify_cb) const {
```
edenhill: there is a mix of cert_verify_ and ssl_cert_verify_.., I would prefer the latter both for the callbacks as for the classes to make it absolutely clear that this is Ssl Certificates and nothing else.

src/rdkafka.h
```diff
@@ -1677,6 +1677,37 @@ void rd_kafka_conf_set_open_cb (rd_kafka_conf_t *conf,
                                                 void *opaque));
 #endif
 
+/**
+* @enum rd_kafka_certificate_type_t
+*
+* @brief Type of certificates
+*
+* @sa rd_kafka_conf_set_get_cert_cb()
```
edenhill: it is called ssl_cert_retrieve_cb(), no?

src/rdkafka.h
```diff
@@ -1677,6 +1677,37 @@ void rd_kafka_conf_set_open_cb (rd_kafka_conf_t *conf,
...
+*/
+typedef enum rd_kafka_certificate_type_t {
+    RD_KAFKA_CERTIFICATE_PUBLIC_KEY,
```
edenhill: add something like this for each entry:
```
    RD_KAFKA_CERTIFICATE_PUBLIC_KEY,  /**< Broker's public key */
```

src/rdkafka.h
```diff
@@ -1677,6 +1677,37 @@ void rd_kafka_conf_set_open_cb (rd_kafka_conf_t *conf,
...
+rd_kafka_conf_res_t rd_kafka_conf_set_ssl_cert_verify_cb(rd_kafka_conf_t *conf,
+    int (*ssl_cert_verify_cb) (char *cert, size_t len, void *opaque));
+
+/**
+* @brief Sets the callback to recieve the client certificate
+*
+*/
+RD_EXPORT
+void rd_kafka_conf_set_ssl_cert_retrieve_cb(rd_kafka_conf_t *conf,
+    size_t (*ssl_cert_retrieve_cb) (rd_kafka_certificate_type_t type, char **buffer, void *opaque));
```
edenhill: ssize_t and add `char *errstr, size_t errstr_size` arguments to callback.

src/rdkafka_conf.c
```diff
@@ -1890,6 +1899,32 @@ void rd_kafka_conf_set_open_cb (rd_kafka_conf_t *conf,
 }
 #endif
 
+rd_kafka_conf_res_t
+rd_kafka_conf_set_ssl_cert_verify_cb(rd_kafka_conf_t *conf,
+    int (*ssl_cert_verify_cb) (char *cert, size_t len, void *opaque)) {
+#ifdef __mips__
```
edenhill: I think this should read: `if defined(__mips__) || !WITH_SSL`, we want to return UNKNOWN in both those cases

src/rdkafka_conf.c
```diff
@@ -1890,6 +1899,32 @@ void rd_kafka_conf_set_open_cb (rd_kafka_conf_t *conf,
...
+    return RD_KAFKA_RESP_ERR__NOT_IMPLEMENTED;
+#endif
+#if WITH_SSL
+    conf->ssl.cert_verify_cb = ssl_cert_verify_cb;
+#else
+    RD_UNUSED(conf);
```
edenhill: remove

src/rdkafka_conf.c
```diff
@@ -1890,6 +1899,32 @@ void rd_kafka_conf_set_open_cb (rd_kafka_conf_t *conf,
 }
 #endif
 
+rd_kafka_conf_res_t
+rd_kafka_conf_set_ssl_cert_verify_cb(rd_kafka_conf_t *conf,
+    int (*ssl_cert_verify_cb) (char *cert, size_t len, void *opaque)) {
+#ifdef __mips__
+    return RD_KAFKA_RESP_ERR__NOT_IMPLEMENTED;
```
edenhill: this is an rd_kafka_resp_err_t error code, but the method returns an rd_akfka_conf_res_t. change it to return RD_KAFKA_CONF_UNKNOWN

src/rdkafka_conf.c
```diff
@@ -1890,6 +1899,32 @@ void rd_kafka_conf_set_open_cb (rd_kafka_conf_t *conf,
...
+    conf->ssl.cert_verify_cb = ssl_cert_verify_cb;
+#else
+    RD_UNUSED(conf);
+    RD_UNUSED(ssl_cert_verify_cb);
+#endif
+
+    return RD_KAFKA_RESP_ERR_NO_ERROR;
+}
+
+void rd_kafka_conf_set_ssl_cert_retrieve_cb(rd_kafka_conf_t *conf,
```
edenhill: add rd_kafka_conf_res_t return type and return UNKNOWN if !WITH_SSL.

also remove UNUSED here

src/rdkafka_conf.h
```diff
@@ -159,6 +159,8 @@ struct rd_kafka_conf_s {
 		char *crl_location;
 		char *keystore_location;
 		char *keystore_password;
+        int (*cert_verify_cb) (char* cert, size_t len, void *opaque);
```
edenhill: pointer '*' goes on the variable, not type.

src/rdkafka_transport.c
```diff
@@ -475,13 +476,16 @@ void rd_kafka_transport_ssl_term (void) {
 
 	CRYPTO_set_id_callback(NULL);
 	CRYPTO_set_locking_callback(NULL);
-        CRYPTO_cleanup_all_ex_data();
+    EVP_cleanup();
+    CRYPTO_cleanup_all_ex_data();
+    ERR_free_strings();
```
edenhill: remove this call, it is not safe to do when SSL is used by multiple libs in the same process, see #815

src/rdkafka_transport.c
```diff
@@ -475,13 +476,16 @@ void rd_kafka_transport_ssl_term (void) {
...
 
 	for (i = 0 ; i < rd_kafka_ssl_locks_cnt ; i++)
 		mtx_destroy(&rd_kafka_ssl_locks[i]);
 
 	rd_free(rd_kafka_ssl_locks);
 
+    rd_kafka_ssl_verify_context_index = -1;
```
edenhill: initialize to -1 in the definition instead

src/rdkafka_transport.c
```diff
@@ -663,6 +669,128 @@ static int rd_kafka_transport_ssl_passwd_cb (char *buf, int size, int rwflag,
 	return pwlen;
 }
 
+int verify_broker_callback(X509_STORE_CTX* ctx, void* arg)
```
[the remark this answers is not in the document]
edenhill: these are not fixed

src/rdkafka_transport.c
```diff
@@ -694,7 +824,16 @@ static int rd_kafka_transport_ssl_connect (rd_kafka_broker_t *rkb,
 		goto fail;
 #endif
 
-        rd_kafka_transport_ssl_clear_error(rktrans);
+    if (rktrans->rktrans_rkb->rkb_rk->rk_conf.debug | RD_KAFKA_DBG_SECURITY) {
+
+        rd_kafka_dbg(rktrans->rktrans_rkb->rkb_rk, SECURITY, "SSL",
+            "Registering handshake debug info");
+
+        SSL_set_ex_data(rktrans->rktrans_ssl, rd_kafka_ssl_verify_context_index, rktrans->rktrans_rkb->rkb_rk);
```
edenhill: verify_context_index is set up once per runtime, but is set multiple times, once for each rk instance here, that is not safe since an application may instantiate multiple rk clients and destroy them whenever.

Do we really need this handshake debug? What exactly does it provide? Do you have any example output?

src/rdkafka_transport.c
```diff
@@ -806,6 +945,138 @@ static int rd_kafka_transport_ssl_handshake (rd_kafka_transport_t *rktrans) {
...
 }
 
+/**
+* Registers the SSL client certificates on the SSL_CTX
+*
+* Returns -1 on error, 0 if client certificates are successfully registered.
+*/
+int rd_kafka_transport_set_client_certificates(rd_kafka_t *rk,
+    SSL_CTX *ctx,
+    char **errstr, size_t *errstr_size) {
```
edenhill: these two should be passed by the caller, not returned from this function:
change it to `char *errstr, size_t errstr_size`

edenhill:
Can you make a test-case that uses this functionality on Windows?

src/rdkafka_transport.c
```diff
@@ -663,6 +669,128 @@ static int rd_kafka_transport_ssl_passwd_cb (char *buf, int size, int rwflag,
...
 }
 
+int verify_broker_callback(X509_STORE_CTX* ctx, void* arg)
+{
+    int validate = 0;
+    char errstr[512];
+    rd_kafka_t* rk = NULL;
+
+    errstr[0] = '\0';
+    if (arg) {
```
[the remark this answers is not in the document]
noahdav: not that I am aware of

src/rdkafka_transport.c
```diff
@@ -663,6 +669,128 @@ static int rd_kafka_transport_ssl_passwd_cb (char *buf, int size, int rwflag,
...
+        rk = arg;
+
+        X509* cert = NULL;
+        rd_kafka_dbg(rk, SECURITY, "SSL",
+            "Verification of peer certificate.");
```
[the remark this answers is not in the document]
noahdav: Ok removed.

src/rdkafka_transport.c
```diff
@@ -663,6 +669,128 @@ static int rd_kafka_transport_ssl_passwd_cb (char *buf, int size, int rwflag,
...
+#ifndef __mips__
+        if(ctx)
+            cert = ctx->cert;
+#endif
+        if (cert) {
```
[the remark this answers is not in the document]
noahdav: removed this anyhow since it is always passed to the callback

src/rdkafka_conf.c
```diff
@@ -533,6 +533,20 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
 	"Client's keystore (PKCS#12) password."
 	},
+    { _RK_GLOBAL, "ssl.handshake.info.enable", _RK_C_BOOL,
+    _RK(ssl.handshake_info),
+    "Report diagnostic information during SSL handshake",
+    0, 1, 0
+    },
+    { _RK_GLOBAL, "ssl.certificate.verify_cb", _RK_C_PTR,
+    _RK(ssl.cert_verify_cb),
+    "Client certificate verification callback."
```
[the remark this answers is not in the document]
noahdav: yes broker cert but client is doing the verification.  I can change this if you like.

src/rdkafka_transport.c
noahdav: OK in this case i will remove it.  I found it helpful during debugging but but using openssl s_client would probably be just as good.

src-cpp/HandleImpl.cpp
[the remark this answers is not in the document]
noahdav: I will remove for now.  If the compiler barks about this I may need to add it back.

src/rdkafka_transport.c
noahdav: Its probably ok to not have this.  I will remove it.  There is not much value over openssl s_client.  If it is needed we can add it back at a later time.

src/rdkafka_transport.c
```diff
@@ -663,6 +669,128 @@ static int rd_kafka_transport_ssl_passwd_cb (char *buf, int size, int rwflag,
...
+            if (subj)
+                OPENSSL_free(subj);
+            if (issuer)
+                OPENSSL_free(issuer);
+
+            len = i2d_X509(cert, &buf);
+            rd_kafka_dbg(rk, SECURITY, "SSL",
+                "Calling client callback to verify certificate");
+
+            if (!rk->rk_conf.ssl.cert_verify_cb(buf, len, rk->rk_conf.opaque)) {
```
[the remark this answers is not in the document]
noahdav: If we do this who is responsible for allocating/freeing the memory?  If the callback just returns us a pointer then it needs to stick around long enough for us to return/log the error message.  In this case how would they know that is is safe to free that memory.  Possibility another callback but this is kind of clunky for just the reason why the validation failed.  The other option is for us to allocate some memory before we call the callback but then the callback is stuck to a specific size.  Do you have any other ideas on how this would work?

src/rdkafka_transport.c
```diff
@@ -663,6 +669,128 @@ static int rd_kafka_transport_ssl_passwd_cb (char *buf, int size, int rwflag,
...
+    else
+        rd_snprintf(str, sizeof(errstr),
+            "undefined ");
+
+    len = (int)strlen(errstr);
+    str += len;
+
+    if (type & SSL_CB_LOOP)
+        rd_snprintf(str, sizeof(errstr) - len,
+            "%s:%s\n", str, SSL_state_string_long(ssl));
```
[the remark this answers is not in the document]
noahdav: We removed this anyhow.

src/rdkafka_transport.c
```diff
@@ -663,6 +669,128 @@ static int rd_kafka_transport_ssl_passwd_cb (char *buf, int size, int rwflag,
...
+        else if (ret < 0)
+        {
+            rd_snprintf(str, sizeof(errstr) - len,
+                "%s:error in %s\n",
+                str, SSL_state_string_long(ssl));
+        }
+    }
+
+    if (strlen(errstr) > 0)
+        rd_kafka_log(rk, LOG_INFO, "SSL", "%s", errstr);
```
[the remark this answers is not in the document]
noahdav: removed handshake callback.

src-cpp/rdkafkacpp_int.h
noahdav: This is indented like the other ones above it.  Do you really want me to make this on different?

src-cpp/rdkafkacpp_int.h
noahdav: Same as above should we make this one different.

src-cpp/rdkafkacpp.h
noahdav: Again who should allocate this in the C layer?  In C++ we have a std::string which the callback can set but in the C layer this needs to be allocated.  

As for the ssize_t it is not defined so not sure that we want to define it in C++,  It is defined in rdkafka.h but do we want to duplicate that?  What is wrong with size_t in this case?

src/rdkafka_transport.c
noahdav: I am confused here?  Do you want me to remove all of the changes here to this function?  If not can you explain what is not safe here?

noahdav: @edenhill  can you please comment on this so I can address this too.

src/rdkafka_transport.c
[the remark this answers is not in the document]
noahdav: @edenhill please comment if this is sufficient.

noahdav:
@edenhill Thank you for your time reviewing this.  I went through to again to fix coding styles.  please respond to my comments as I believe that the others are addressed.  Sorry if I missed some of your comments in last iteration.  I believe that they should now be addressed.

noahdav:
@edenhill Thank you for your time reviewing this. I went through to again to fix coding styles.  Please respond to my comments as I believe that the others are addressed. Sorry if I missed some of your comments in last iteration. I may of lost a few with how many that there were.  I believe that they should now be addressed.

noahdav:
@edenhill 
I am finally getting back to this.  For the test I am looking at how to do this but I am not seeing any tests which configure kafka.  In order to do this I will need to set up kafka with correct SSL certs.  I assume we will need to generate self signed ones for both client and broker.  Is there a place which does something like this?  If not I can create the tests assuming that the certs exist.

edenhill:
Could you provide an example of how to use this with the Windows key store?

edenhill:
Let's schedule this feature for the 1.0.0 release.

noahdav:
Yes I will provide an example code to use this with the windows store.  We have been using it for a few months now internally.  I will write up a sample.

@edenhill When is the 1.0.0 release scheduled?

edenhill:
Perfect!

1.0.0 is scheduled for November

src/rdkafka_transport.c
dzuelke: What @edenhill means, @noahdav, is that this may cause conflicts with other programs or libraries that are also using OpenSSL; #815 or https://www.postgresql.org/message-id/20081203200427.0A23B7545A4@cvs.postgresql.org are examples of this, although I am not sure how `ERR_free_strings` specifically is affected.

noahdav: Thanks @dzuelke for the clarification.  The article that you liked talks about having callbacks set to unloaded modules.  In this case I do not see how this will happen.  I am not understanding how ERR_free_strings is different from CRYPTO_cleanup_all_ex_data or the other cleanup routines here.  If you feel strongly I can remove this but I am not clear to the issue by calling ERR_free_strings.

noahdav:
@edenhill I updated examples/rdkafka_example.cpp with example of using these callbacks as we discussed on windows.  Please take a look and let me know what else needs to be addressed before this feature can be checked in.

noahdav:
@edenhill as we discusses to use a certificate in the windows I generated the cert using OpenSSL then converted it to a pfx file which can easily be imported to the Windows cert store

openssl.exe pkcs12 -export -out cert.pfx -inkey client.key -in client.pem

noahdav:
@edenhill any update when we can merge this change?  Or if there are still any outstanding issues please let me know as I think I have addressed the previous comments.

LighthouseJ:
Hello.  I've been using this branch (noahdav/librdkafka:ssl) for about a week or so but I've had a small issue.  When I define the retrieve_cb in the config, after I create the high-level consumer, the receive_cb function pointer is never copied and never executed.  I had to modify a few lines to not skip copying the callback pointer.  Attached is my change.  
[0001-Not-skipping-copy-of-the-retrieve_cb.patch.txt](https://github.com/edenhill/librdkafka/files/2600965/0001-Not-skipping-copy-of-the-retrieve_cb.patch.txt)

I don't believe that this is the best fix for this issue, but it ensured that the receive_cb is still available when it is used at runtime.
The verify_cb may similarly be affected.

Thoughts or any feedback in case I did something wrong?

noahdav:
@LighthouseJ 

Thanks for the feedback.  I am not sure how this callback is being handled any differently than any of the other callbacks.  It should be copied in a similar way.  Can you share your sample code.  Did you look at any other callback?

LighthouseJ:
@noahdav I'm not sure either.  I hoped @edenhill would comment if my patch is appropriate or not.  I've been using the SSL soft certs in the high-level consumer without issue so I was under the (possibly false) impression that when the high-level consumer is created and the provided config is duplicated, that these two new callback function pointers are not copied since the dupe code (what calls that function) does not correct discern between fields that should be copied and rdkafka system fields that shouldn't.

I will try and reproduce on the branch and see if I can demonstrate it.

edenhill:
I'm rebasing and making some final adjustments to this PR, I'll have a look at the callback issue when that's done.

noahdav:
@edenhill is there any update to when this will merge?  Which release are we targeting?

noahdav:
@edenhill Sounds great.  Thanks for the update.  This will be great for at least us to get in the official branch.

src/rdkafka_transport.c
```diff
@@ -901,6 +1115,27 @@ int rd_kafka_transport_ssl_ctx_init (rd_kafka_t *rk,
...
+    if (rk->rk_conf.ssl.ssl_cert_verify_cb) {
+
+        SSL_CTX_set_verify(
+            ctx, SSL_VERIFY_PEER | SSL_VERIFY_CLIENT_ONCE | SSL_VERIFY_FAIL_IF_NO_PEER_CERT,
+            NULL);
+
+        rd_kafka_dbg(rk, SECURITY, "SSL",
+            "Registering verification callback");
+
+        SSL_CTX_set_cert_verify_callback(
```
edenhill: Reading the SSL docs on [SSL_CTX_set_verify](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_verify.html) and [SSL_CTX_set_cert_verify](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_cert_verify_callback.html) makes me believe we should not do the latter, but the former.
The latter replaces OpenSSL's verification mechanism, which I don't think we want, while the former allows additional checks after the OpenSSL verification has been done.

Do you have a use-case for replicating the OpenSSL verification, or are the simpler additional verification sufficient?

Also see answers on SO: https://stackoverflow.com/questions/2729254/ssl-ctx-set-cert-verify-callback-vs-ssl-ctx-set-verify

**Do not push any commits to this PR**, I'm working on it locally with a bit of merge fixes, refactoring, etc.

noahdav: My understanding of this is that we do not replace OpenSSL's verification since we pass NULL to the callback in SSL_CTX_set_verify.  We are only specifying the verification flags.  Since we specify the flag SSL_VERIFY_PEER we need to use SSL_CTX_set_cert_verify_callback to validate the server to perform mutual auth.  

Looking at this PR we may want to add a property to set the verification flags so that somone can specify the flags they want.  In some cases SSL_VERIFY_FAIL_IF_NO_PEER_CERT may not be appropriate.

noahdav: My understanding of this is that we do not replace OpenSSL's verification since we pass NULL to the callback in SSL_CTX_set_verify.  We are only specifying the verification flags.  Since we specify the flag SSL_VERIFY_PEER we need to use SSL_CTX_set_cert_verify_callback to validate the server to perform mutual auth.  

Looking at this PR we may want to add a property to set the verification flags so that someone can specify the flags they want.  In some cases SSL_VERIFY_FAIL_IF_NO_PEER_CERT may not be appropriate.

edenhill: SSL_CTX_set_cert_verify() will replace OpenSSL's verification code according to [docs](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_cert_verify_callback.html): 

> . If a verification callback callback is specified via SSL_CTX_set_cert_verify_callback(), the supplied callback function is called instead. 

I don't think we should do this, we should use SSL_CTX_set_verify() instead that allows an application to provide *additional* checks.

Also, we should have the verification enabled by default, regardless if a custom SSL verify callback is provided, i.e., we always have SSL_VERIFY_PEER set. 
The other flags are server-side only.

LighthouseJ: I agree that SSL_CTX_set_cert_verify() will override the default cert checking mechanism.
I couldn't find confirmation, but is something like OCSP or another custom CA downloading and checking system may be ligitimately able to be used?

@edenhill Is your concern leaving it implemented technical or security in nature?

noahdav: @edenhill 

I agree that we definitely do not want to override the default certificate verification process.  We do however want to provide a way for the client to perform mutual auth based on the server/broker certificate.  If I remember correctly I could not get the cert in the SSL_CTX_set_verify callback.  When I called X509_STORE_CTX_get_current_cert this would only return the 0 dept certificate if there was an error in the validation process.  Otherwise NULL was returned.

I am ok with using this as long as there is a reliable way to retrieve the server certificate.

LighthouseJ: Why conditionally perform mutual auth depending on the peer cert?
If you have an expected CA file/path configured, then openssl will already look for and verify the cert chain by default.

I believe that you'd use SSL_CTX_set_verify_callback to verify a peer cert against a custom store, e.g. a custom Windows CAPI store.

noahdav: @LighthouseJ

We need to verify a cert in the windows cert store.  In our case we have custom logic to verify the certificate in addition to it existing in the windows cert store.  Do you have any suggestions on how to handle this scenario?

LighthouseJ: @noahdav If you need custom logic aside from the cert belonging in the default places for Windows, then you would likely need a custom verify callback function then.  IMHO I think your use case then necessitates the openssl callback function being exposed through librdkafka.

noahdav: @LighthouseJ That is why I created this PR in the first place.  This is being used internally in Microsoft Azure.

edenhill: @noahdav I'll see if we can get the full cert chain using the simpler callback form.

noahdav: @edenhill That would be great.  Let me know what you figure out.  I do think it would be best to not override the default verification but just plug the callback logic when we try to allow additional checks on top of what OpenSSL already does.

noahdav: @edenhill is there any update on this?  Were you able to get the entire cert chain in a different way?

src/rdkafka_conf.c
```diff
@@ -625,6 +625,15 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	_RK(ssl.keystore_password),
 	"Client's keystore (PKCS#12) password."
 	},
+    { _RK_GLOBAL, "ssl.certificate.verify_cb", _RK_C_PTR,
+    _RK(ssl.ssl_cert_verify_cb),
+    "Verify the broker certificate callback."
+    },
+    { _RK_GLOBAL, "ssl.certificate.retrieve_cb", _RK_C_PTR,
```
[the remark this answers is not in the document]
noahdav: @edenhill 

The only problem I see with that approach is that the memory management is a little more complicated from the callers point of view.  They would essentially need to keep the memory around during the entire lifetime of the configuration instance.  Currently once it is sent to OpenSSL it can safely be freed.  However I am ok with either approach since it is only a single buffer that needs to be maintained.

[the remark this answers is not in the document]
noahdav: Or if we copy it into the config as a buffer not the pointer that would be ok too.  Since it is a public key there is no need to clear the buffer when freeing the memory.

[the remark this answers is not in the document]
edenhill: Yeah, all config values are copied, so should be fine.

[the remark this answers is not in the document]
noahdav: @edenhill Are you going to make this change or do you want me to do this and push new changes?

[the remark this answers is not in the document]
edenhill: @noahdav Thanks, but I'm on it already!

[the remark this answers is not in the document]
noahdav: @edenhill I looked at this again and tried to see why I did not do just make it a property originally.  The retrieve callback is actually called 3 times.  Once for public key, once for private key, and once for private key password.  Are you planning on making all of those properties?  If so my only objection is that the private key and private key password need to be securely cleared from memory after use since it is sensitive data.  On windows we call SecureZeroMemory to make sure the memory is cleared.  I am assuming some equivalent API on Linux/Unix exists or you could use memset_s with a memory barrier to make sure these are securely wiped before we free the memory.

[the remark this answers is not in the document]
edenhill: Good and timely input! I'm just now adding functionality to do exactly this for sensitive configuration properties.
Thanks for the hint on SecureZeroMemory :+1:

By "after use", I'm assuming you mean after the rd_kafka_t instance is destroyed, right?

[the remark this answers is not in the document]
noahdav: Currently I was managing this in my consumer/producer wrapper object after the connection was established.  If you want to move this to before we destroy the instance that is fine too.

Are you planning on having all 3 properties (public key, private key, and private key password) instead of the single callback?  I thought the callback was simpler to have this all in one place since the caller is already implementing the validate callback anyhow.  Also it allowed the caller to implement all three of these since it is not valid to have some of them implemented.  Either way is fine with me as long as we have this same functionality.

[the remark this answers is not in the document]
edenhill: I'm adding `ssl.certificate.pem` (client's public key) and `ssl.key.pem` (client's private key), taking a PEM-formatted string.
`ssl.key.password` (private key PEM's password, optional) remains unchanged.

edenhill:
A summary of changes in the upcoming force-push:

 * The retrieve_cb() has been removed in favour of using standard configuration properties. The reasoning behind this ís simplification, reduced code size, cross-language maintenance (some high-level languages, such as Go, can't use callbacks in a good way), and memory ownership issues (having the callback return cert memory to librdkafka in an effective and allocator-agnostic way is tricky).
 * Refactored the code and made it more in line with the librdkafka style guide.
 * Values of security-sensitive configuration properties are now cleared from memory when the value is no longer used, or on rd_kafka_t destruction. This is used for ssl.certificate.pem, ssl.key.pem and ssl.key.password.
 * The ssl_cert_verify_cb was changed to provide more information to the callback, including the full X509_STORE_CTX.

Future commits:
 * Rework the Windows SSL cert store example to its own file.
 * Add integration tests.

edenhill:
@noahdav Can you please have a look at PR https://github.com/edenhill/librdkafka/pull/2309, which is based on your PR with the above changes?

PR #1906, opened 2018-07-30
Remove callback in KafkaConsumer's example since it is actually not implemeted

chaodhib:
remove the callback from the high level consumer example since it is not implemented

The callback is not implemented. Yet, it is featured in the example. This is highly confusing as can be shown by these tickets: [HERE](https://github.com/edenhill/librdkafka/issues/1905), [HERE ](https://github.com/edenhill/librdkafka/issues/1137) and [HERE](https://github.com/edenhill/librdkafka/issues/1620).

### DIFF
--- a/examples/rdkafka_consumer_example.cpp
+++ b/examples/rdkafka_consumer_example.cpp
@@ -210,16 +210,6 @@ void msg_consume(RdKafka::Message* message, void* opaque) {
   }
 }
 
-
-class ExampleConsumeCb : public RdKafka::ConsumeCb {
- public:
-  void consume_cb (RdKafka::Message &msg, void *opaque) {
-    msg_consume(&msg, opaque);
-  }
-};
-
-
-
 int main (int argc, char **argv) {
   std::string brokers = "localhost";
   std::string errstr;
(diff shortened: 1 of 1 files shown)


chaodhib:
Maybe the entire commit https://github.com/edenhill/librdkafka/commit/c09efc160949cd3a1c0c8480ff22f6f7943c09e1 should be rolled back?

examples/rdkafka_consumer_example.cpp
```diff
@@ -452,10 +446,12 @@ int main (int argc, char **argv) {
    * Consume messages
    */
   while (run) {
-    RdKafka::Message *msg = consumer->consume(1000);
-    if (!use_ccb) {
-      msg_consume(msg, NULL);
+    if (use_ccb) {
```
edenhill: suggest you remove all occurrences of use_ccb, including the flag.

edenhill:
ping

chaodhib:
Done. What about the rest of the changes in https://github.com/edenhill/librdkafka/commit/c09efc160949cd3a1c0c8480ff22f6f7943c09e1? Should it be kept?

examples/rdkafka_consumer_example.cpp
```diff
@@ -306,15 +295,6 @@ int main (int argc, char **argv) {
       }
       break;
 
-      case 'f':
```
edenhill: remove `f:` from getopt above

chaodhib: Done

edenhill:
Thank you!

PR #1921, opened 2018-08-06
rdhdrhistogram: Fix incorrect float -> int cast causing havoc on MIPS.

andoma:
In rd_hdr_histogram_new() when minValue is 0, log2() will return -inf.
Casting -inf to int is undefined, and on MIPS this results in 2^31
but on other architectures it becomes zero instead (which is desired
in this function). Fix this by changing so the RD_MAX() compare
is done on floating point (0 is > -inf) and thus unitMagnitude
becomes 0 (as expected).

This bug caused the while() loop in said function to never terminate
which in turn caused the unit tests to hang.


edenhill:
Good find, thank you mr Smas! ❤️

Issue #1913, opened 2018-08-02
Message stalled in queue when broker (partition leader) die


chronidev:
Description
===========
When a broker, which is the leader of the partition, goes down while a producer produce message at continuous rate, some messages timeout at the end of the message.timeout.ms and result in messages loss.
```
[2018-08-02 04:34:03,735] kafka producer delivery_cb: _MSG_TIMED_OUT: Local: Message timed out 1533176942362;360001
``` 
It seems the messages stay in a queue somewhere and are not re-send to the new leader. 

I have Repeated log message like this:
```
[2018-08-02 04:34:00,734] CONNECT [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: broker3.testy.fr:9092/3: broker in state DOWN connecting
[2018-08-02 04:34:00,734] CONNECT [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: broker3.testy.fr:9092/3: Connecting to ipv4#10.67.146.9:9092 (plaintext) with socket 15
[2018-08-02 04:34:00,734] STATE [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: broker3.testy.fr:9092/3: Broker changed state DOWN -> CONNECT
[2018-08-02 04:34:00,734] BROADCAST [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: Broadcasting state change
[2018-08-02 04:34:00,734] BROKERFAIL [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: broker3.testy.fr:9092/3: failed: err: Local: Broker transport failure: (errno: Connection refused)
[2018-08-02 04:34:00,734] FAIL [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: broker3.testy.fr:9092/3: Connect to ipv4#10.67.146.9:9092 failed: Connection refused
[2018-08-02 04:34:00,734] STATE [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: broker3.testy.fr:9092/3: Broker changed state CONNECT -> DOWN
[2018-08-02 04:34:00,734] BROADCAST [rdkafka#producer-1] [thrd:broker3.testy.fr:9092/bootstrap]: Broadcasting state change
[...]

PR #1923, opened 2018-08-06
Purge and retry buffers in outbuf queue on connection fail (#1913)

edenhill:
There are multiple parts to this fix:
 * A request/response handler can now check if a failing request
   was sent out on the wire (or did not make it past the output
   queue).
 * The retry code now only increments the retry count for actually
   sent requests.
 * When the broker connection goes down, requests in the output queue
   are now purged to have their handler callbacks called which in turn
   will trigger a (now free) retry.
 * ProduceRequests are not retried, but their messages are put back
   on the partition queue (existing behaviour).


chronidev:
I was able to test the code, it seems to correct the issue.

PR #1924, opened 2018-08-06
consumer: use RD_KAFKA_NO_REPLYQ when calling rd_kafka_seek async

JonECG:
Fixes an error where a null mtx would attempted to be accessed within
RD_KAFKA_REPLYQ as tmpq would not be created if no timeout is provided.


edenhill:
Awesome! Thank you!

PR #1932, opened 2018-08-09
Add rd_kafka_destroy_flags()

edenhill:
From Dispose() you would call `rd_kafka_destroy_flags(RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)`.

This should probably be changed in the Python destructor as well.

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -2043,9 +2043,52 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type, rd_kafka_conf_t *conf,
  * @brief Destroy Kafka handle.
  *
  * @remark This is a blocking operation.
+ * @remark rd_kafka_consumer_close() will be called from this function
+ *         if the instance type is RD_KAFKA_CONSUMER, a \c group.id was
+ *         configured, and the rd_kafka_consumer_close() was not
+ *         explicitly called by the application. This in turn may
+ *         trigger consumer callbacks, such as rebalance_cb.
+ *         Use rd_kafka_destroy_flags() with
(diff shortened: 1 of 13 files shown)


src/rdkafka.h
```diff
@@ -2043,11 +2043,54 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type, rd_kafka_conf_t *conf,
  * @brief Destroy Kafka handle.
  *
  * @remark This is a blocking operation.
+ * @remark rd_kafka_consumer_close() will be called from this function
+ *         if the instance type is RD_KAFKA_CONSUMER, a \c group.id was
+ *         configured, and the rd_kafka_consumer_close() was not
+ *         explicitly called by the application. This in turn may
```
mhowlett: more important to note is it may trigger broker requests (callbacks may be triggered regardless).

src/rdkafka.h
```diff
@@ -2043,11 +2043,54 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type, rd_kafka_conf_t *conf,
...
 RD_EXPORT
 void        rd_kafka_destroy(rd_kafka_t *rk);
 
 
+/**
+ * @brief Destroy Kafka handle according to specified destroy flags
+ *
+ */
+RD_EXPORT
+void rd_kafka_destroy_flags (rd_kafka_t *rk, int flags);
```
mhowlett: i guess this name is ok. the `Ex` extension used by MS has the advantage that most people looking at it will know it's a new improved version of an older method by convention whereas the word flags is specific to this case. it also potentially reads as 'destroying the flags'. On the other hand, i'm not confused.

mhowlett:
lgtm with the disclaimer that the logic in this PR is difficult to review properly given limited familiarity with the code.

PR #1945, opened 2018-08-13
Early feedback on headers support for librdkafka

davidtrihy:
Looking for feedback on some changes I've made to support Headers in line with this issue here https://github.com/edenhill/librdkafka/issues/1861

As of now there are no tests, documentation isn't up to date and the coding standards are not conforming but this builds fine.

I'm mainly looking for early feedback on the implementation and interface.

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -49,6 +49,7 @@
 /**@cond NO_DOC*/
 #include <string>
 #include <list>
+#include <map>
 #include <vector>
 #include <stdint.h>
 
@@ -78,7 +79,6 @@ extern "C" {
 };
 
 namespace RdKafka {
-
 
 /**
  * @name Miscellaneous APIs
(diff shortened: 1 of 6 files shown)


src-cpp/rdkafkacpp.h
```diff
@@ -49,6 +49,7 @@
 /**@cond NO_DOC*/
 #include <string>
 #include <list>
+#include <map>
```
davidtrihy: Need to remove this

src-cpp/rdkafkacpp.h
```diff
@@ -396,6 +396,8 @@ std::string  err2str(RdKafka::ErrorCode err);
 /* Forward declarations */
 class Producer;
 class Message;
+class Header;
```
davidtrihy: Don't think this is needed anymore

davidtrihy:
Closing as I'm opening up a new one with a lot more changes

PR #1949, opened 2018-08-14
Use astyle as a C style checker

edenhill:
...using a modified version of astyle that allows
space pad after the function name in the function definition.

This currently only checks the C code since the --style=google
check in astyle is not fully compatible with Google C++ style guide.

Use `make checkstyle` to report all C style violations,
and `make fixstyle` to automatically fix them.


phrocker:
I'm curious why this never moved forward? Seems like a great approach

edenhill:
@phrocker Need to find a good time (no big or critical code changes) for when to fix all the formatting/style issues in the current-tree.

phrocker:
@edenhill I'll try to carve some time to help out if that's cool with you

PR #1951, opened 2018-08-14
highlighting legacy properties

mhowlett:
When reading the config property docs quickly, it's easy to miss that these are legacy properties. I did, as did the reporter of https://github.com/confluentinc/confluent-kafka-dotnet/issues/362


edenhill:
good stuff!

PR #1954, opened 2018-08-15
Improved handling of removed partitions

edenhill:
..and some minor other fixes


edenhill:
It is probably easiest to review this per commit.

Issue #1955, opened 2018-08-16
【producer】broker got invalid request when broker network thead got big pressure


hulaxiaodai:
Description
===========
In my product env, when i produce messages about 500mb/s, the borker got InvalidMessageException . and got an  invalid request size. 
It looks like the request was truncation, then network thread got next 4 bytes request size is wrong;
/*
if(sizeBuffer.remaining > 0)
      read += Utils.read(channel, sizeBuffer)
    // have we allocated the request buffer yet?
    if(contentBuffer == null && !sizeBuffer.hasRemaining) {
      sizeBuffer.rewind()
      val size = sizeBuffer.getInt()
      if(size <= 0)
        throw new InvalidRequestException("%d is not a valid request size.".format(size))
      if(size > maxSize)
        throw new InvalidRequestException("Request of length %d is not valid, it is larger than the maximum size of %d bytes.".format(size, maxSize))
      contentBuffer = byteBufferAllocate(size)
    }
*/

It is reproduce when broker got big flow only, and  allocate huge memory, then cause gull gc some time. It makes broker  got performance decline.

How to reproduce
================
I debug kafka server on idea, and user librdkafka  produce messages about 80000/s.
then I break  server ,and let it run after a few minutes. broker get exception.

broker log, below:
[2018-08-16 20:25:44,053] ERROR [KafkaApi-0] Error processing ProducerRequest with correlation id 126 from client log_collector-2d6f96a6b009 on partition [org.qa_test_pt_1KB_2018,0] (kafka.server.KafkaApis)
kafka.message.InvalidMessageException: Message is corrupt (stored crc = 978447105, computed crc = 4105360175)
	at kafka.message.Message.ensureValid(Message.scala:166)
	at kafka.log.Log$$anonfun$analyzeAndValidateMessageSet$1.apply(Log.scala:383)
	at kafka.log.Log$$anonfun$analyzeAndValidateMessageSet$1.apply(Log.scala:361)
	at scala.collection.Iterator$class.foreach(Iterator.scala:727)
	at kafka.utils.IteratorTemplate.foreach(IteratorTemplate.scala:32)
	at kafka.log.Log.analyzeAndValidateMessageSet(Log.scala:361)
	at kafka.log.Log.append(Log.scala:257)
	at kafka.cluster.Partition$$anonfun$appendMessagesToLeader$1.apply(Partition.scala:380)
[...]

PR #1960, opened 2018-08-21
Disconnect regardless of socket.max.fails when partial request times out (#1955)

edenhill:
We can't restore the connection to a clean state when a partial
request times out, instead fail the connection and reconnect,
regardless of the socket.max.fails setting.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -26,7 +26,7 @@ socket.send.buffer.bytes                 |  *  | 0 .. 100000000  |             0
 socket.receive.buffer.bytes              |  *  | 0 .. 100000000  |             0 | Broker socket receive buffer size. System default is used if 0. <br>*Type: integer*
 socket.keepalive.enable                  |  *  | true, false     |         false | Enable TCP keep-alives (SO_KEEPALIVE) on broker sockets <br>*Type: boolean*
 socket.nagle.disable                     |  *  | true, false     |         false | Disable the Nagle algorithm (TCP_NODELAY) on broker sockets. <br>*Type: boolean*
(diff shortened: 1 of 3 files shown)


CONFIGURATION.md
```diff
@@ -26,7 +26,7 @@ socket.send.buffer.bytes                 |  *  | 0 .. 100000000  |             0
 socket.receive.buffer.bytes              |  *  | 0 .. 100000000  |             0 | Broker socket receive buffer size. System default is used if 0. <br>*Type: integer*
 socket.keepalive.enable                  |  *  | true, false     |         false | Enable TCP keep-alives (SO_KEEPALIVE) on broker sockets <br>*Type: boolean*
 socket.nagle.disable                     |  *  | true, false     |         false | Disable the Nagle algorithm (TCP_NODELAY) on broker sockets. <br>*Type: boolean*
-socket.max.fails                         |  *  | 0 .. 1000000    |             1 | Disconnect from broker when this number of send failures (e.g., timed out requests) is reached. Disable with 0. NOTE: The connection is automatically re-established. <br>*Type: integer*
+socket.max.fails                         |  *  | 0 .. 1000000    |             1 | Disconnect from broker when this number of send failures (e.g., timed out requests) is reached. Disable with 0. WARNING: It is highly recommended to leave this setting at its default value of 1 to avoid the client and broker to become desynchronized in case of request timeouts. NOTE: The connection is automatically re-established. <br>*Type: integer*
```
mhowlett: is it worth considering making this a binary property i.e. "disconnect from broker on send failure"?

edenhill: Yeah, that would probably make more sense, but I'd rather see this property removed altogether. I don't really see a use-case where becoming desynchronized with the broker is something you'd want.   This is a legacy thing..

PR #1970, opened 2018-08-25
Update batch.num.messages description

pmbuko:
Makes the `batch.num.messages` description under the Performance header of INTRODUCTION.md agree with the description in CONFIGURATION.md


edenhill:
Thank you!

Issue #1962, opened 2018-08-21
0.11.5 FreeBSD/clang compilation warnings


mutability:
Compiling librdkafka 0.11.5 on FreeBSD 11.1 with clang 4.0.0 produces warnings about gettimeofday & alloca:

```
$ CC=clang CXX=clang ./configure
[...]
$ gmake
[...]
clang -MD -MP -g -O2 -fPIC -Wall -Wsign-compare -Wfloat-equal -Wpointer-arith -Wcast-align -DLIBRDKAFKA_GIT_VERSION="\"v0.11.5\""  -c rdkafka_topic.c -o rdkafka_topic.o
In file included from rdkafka_topic.c:30:
In file included from ./rdkafka_int.h:46:
./rdtime.h:98:2: warning: implicit declaration of function 'gettimeofday' is invalid in C99 [-Wimplicit-function-declaration]
        rd_gettimeofday(&tv, NULL);
        ^
./rdposix.h:147:33: note: expanded from macro 'rd_gettimeofday'
#define rd_gettimeofday(tv,tz)  gettimeofday(tv,tz)
                                ^
rdkafka_topic.c:838:23: warning: implicitly declaring library function 'alloca' with type 'void *(unsigned long)' [-Wimplicit-function-declaration]
        partbrokers = rd_alloca(mdt->partition_cnt * sizeof(*partbrokers));
                      ^
./rdposix.h:73:23: note: expanded from macro 'rd_alloca'
#define rd_alloca(N)  alloca(N)
                      ^
rdkafka_topic.c:838:23: note: include the header <stdlib.h> or explicitly provide a declaration for 'alloca'
./rdposix.h:73:23: note: expanded from macro 'rd_alloca'
#define rd_alloca(N)  alloca(N)
                      ^
2 warnings generated.
```

(the same warnings are repeated when compiling other parts of librdkafka)

 - [x] librdkafka version (release number or git tag): `v0.11.5`
 - [x] Operating system:
```
$ freebsd-version 
11.1-RELEASE-p9
$ clang --version
FreeBSD clang version 4.0.0 (tags/RELEASE_400/final 297347) (based on LLVM 4.0.0)
Target: x86_64-unknown-freebsd11.1
Thread model: posix
InstalledDir: /usr/bin
```

edenhill:
That looks like a missing include or two.
PRs to fix this are welcome since FreeBSD is not a platform we actively maintain.

PR #1978, opened 2018-08-30
Fix warnings on FreeBSD; it is very picky about _POSIX_C_SOURCE et al

mutability:
This fixes the warnings in #1962 on FreeBSD

This turned out to be a bit of a rabbit hole: FreeBSD disables all the usual default visibility if you set _POSIX_C_SOURCE _at all_, so the warnings were coming from the implicit _XOPEN_SOURCE=700 being disabled.

Also, there doesn't seem to be any way to get the standard headers to set __BSD_VISIBLE (needed for alloca) when the default visibility is disabled, so we have to do that directly, which is ugly.

### DIFF
--- a/src/rd.h
+++ b/src/rd.h
@@ -38,6 +38,16 @@
 #ifndef _POSIX_C_SOURCE
 #define _POSIX_C_SOURCE 200809L  /* for timespec on solaris */
 #endif
+/*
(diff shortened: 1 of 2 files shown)


edenhill:
Please address the CI build failures

PR #1984, opened 2018-09-04
locking callbacks as required, dont call CRYPTO_cleanup_all_ex_data

ameihm0912:
Updates global SSL initialization to only configure callbacks if they
have not already been set. This makes the library behave better in
processes where OpenSSL is being used by other shared libraries, and
will not arbitrarily clobber callback functions set by the main
application using rdkafka.

Also remove call to CRYPTO_cleanup_all_ex_data, which makes an
assumption nothing will utilize OpenSSL in the process anymore and can
cause problems in the scenarios described above.

### DIFF
--- a/src/rdkafka_transport.c
+++ b/src/rdkafka_transport.c
@@ -480,19 +480,23 @@ static void rd_kafka_transport_libcrypto_THREADID_callback(CRYPTO_THREADID *id)
  * Global OpenSSL cleanup.
  */
 void rd_kafka_transport_ssl_term (void) {
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 	int i;
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
-	CRYPTO_set_id_callback(NULL);
-	CRYPTO_set_locking_callback(NULL);
-        CRYPTO_cleanup_all_ex_data();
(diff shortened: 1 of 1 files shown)


src/rdkafka_transport.c
```diff
@@ -491,14 +491,16 @@ void rd_kafka_transport_ssl_term (void) {
...
 	
-	rd_kafka_ssl_locks_cnt = CRYPTO_num_locks();
-	rd_kafka_ssl_locks = rd_malloc(rd_kafka_ssl_locks_cnt *
-				       sizeof(*rd_kafka_ssl_locks));
-	for (i = 0 ; i < rd_kafka_ssl_locks_cnt ; i++)
-		mtx_init(&rd_kafka_ssl_locks[i], mtx_plain);
-
-	CRYPTO_set_id_callback(rd_kafka_transport_ssl_threadid_cb);
-	CRYPTO_set_locking_callback(rd_kafka_transport_ssl_lock_cb);
+	if (!CRYPTO_get_locking_callback()) {
```
edenhill: There's a race condition here, but I guess there isn't much we can do about it.

edenhill:
Can you rebase this on latest master?
Also pay attention that OpenSSL v1.1.0 support was added

ameihm0912:
@edenhill this has been rebased against master and adjusted for the OpenSSL version ifdef.

PR #1997, opened 2018-09-12
CMake: Build shared C++ library

Oxymoron79:
Allow the C++ library to be built as a shared library.

Also, the shared libraries are versioned by setting the SOVERSION property, similar to the mklove build.


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.2)
-project(RdKafka)
+project(RdKafka VERSION 0.11.6)
```
edenhill: Where does this version eventually end up? SONAME/SOVERSION?
I'm not sure that's completely correct since a version bump of librdkafka does not necessarily mean there is an incompatibility in the C++ API, existing apps should not need relinking.

Having said that; ABI compat for C++ is a mess...

Oxymoron79: Only the major version number (0) is used as the SOVERSION. This causes CMake to generate symbolic links (similar to mklove):
```
librdkafka.so -> librdkafka.so.0
librdkafka++.so -> librdkafka++.so.0
```

I also have a pull-request in queue to generate pkg-config files with CMake, where the full version is needed.

edenhill: But librdkafka's SONAME is currently at 1 (https://github.com/edenhill/librdkafka/blob/master/src-cpp/Makefile#L3)
Additionally, our bump to v1.0.0 will not be an ABI-breaking change (but v2.0.0 probably will be).
The SONAME and librdkafka version are thus loosely coupled, so I don't think it is useful here

Oxymoron79: Ah OK, I see.
I'm going to remove the project version from this pull-request and decouple the SOVERSION from the project version using CMake variables.

edenhill:
Will merge after the v0.11.6 maintenance release

PR #2000, opened 2018-09-13
OpenSSL 1.1.0 compatibility

nouzun:
Changes to be able to build librdkafka with OpenSSL 1.1.0

### DIFF
--- a/src/rdkafka_transport.c
+++ b/src/rdkafka_transport.c
@@ -446,7 +446,7 @@ static char *rd_kafka_ssl_error (rd_kafka_t *rk, rd_kafka_broker_t *rkb,
     return errstr;
 }
 
-
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 static RD_UNUSED void
 rd_kafka_transport_ssl_lock_cb (int mode, int i, const char *file, int line) {
 	if (mode & CRYPTO_LOCK)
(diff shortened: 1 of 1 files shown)


src/rdkafka_transport.c
```diff
@@ -503,6 +507,13 @@ void rd_kafka_transport_ssl_init (void) {
 	SSL_load_error_strings();
 	SSL_library_init();
 	OpenSSL_add_all_algorithms();
+#else
+#ifdef HAVE_OPENSSL_CRYPTO_THREADID_SET_CALLBACK
+	CRYPTO_THREADID_set_callback(rd_kafka_transport_ssl_threadid_cb);
```
edenhill: Can't find if this is actually needed on OpenSSL >=1.1.0, but other projects seem to not use it on that version, so maybe it should be moved to the < 1.1.0 block?

https://github.com/pycurl/pycurl/issues/491

src/rdkafka_transport.c
```diff
@@ -503,6 +507,13 @@ void rd_kafka_transport_ssl_init (void) {
...
+#endif
+	OPENSSL_init_ssl(0, NULL);
```
edenhill: This isn't needed according to docs:
https://www.openssl.org/docs/man1.1.0/ssl/OPENSSL_init_ssl.html

nouzun: I removed it.

src/rdkafka_transport.c
nouzun: True. I moved it to the < 1.1.0 block.

PR #2001, opened 2018-09-13
Idempotent Producer support

edenhill:
And some additional goodies:
 * stricter config checking (for 1.0)
 * Added purge() API
 * fatal errors
 * Added message_status() API


CONFIGURATION.md
```diff
@@ -97,6 +97,8 @@ rebalance_cb                             |  C  |                 |
 offset_commit_cb                         |  C  |                 |               | Offset commit result propagation callback. (set with rd_kafka_conf_set_offset_commit_cb()) <br>*Type: pointer*
 enable.partition.eof                     |  C  | true, false     |          true | Emit RD_KAFKA_RESP_ERR__PARTITION_EOF event whenever the consumer reaches the end of a partition. <br>*Type: boolean*
 check.crcs                               |  C  | true, false     |         false | Verify CRC32 of consumed messages, ensuring no on-the-wire or on-disk corruption to the messages occurred. This check comes at slightly increased CPU usage. <br>*Type: boolean*
+enable.idempotence                       |  P  | true, false     |         false | When set to `true`, the producer will ensure that messages are successfully produced exactly once and in the original produce order. The following configuration properties are adjusted automatically (if necessary) when idempotence is enabled: `max.inflight.requests.per.connection` is capped at 5, `retries` must be greater than 0, `acks` is  set to `all`, `queuing.strategy` is set to `fifo`. <br>*Type: boolean*
```
mhowlett: it would be good to clarify here whether these are automatically adjusted only in the case they haven't been explicitly set or whether setting this overwrites explicit settings as well (which is how it reads to me). in the latter case, I think the config should be considered invalid and call to `rd_kafka_new` fail (haven't looked at what is actually done yet).

edenhill: It works the way you describe it, kafka_new() will fail if a user has set incompatible properties. But yeah, it needs to be made clear here.

Unmodified properties are automatically adjusted.

INTRODUCTION.md
```diff
@@ -388,6 +392,244 @@ may be produced out of order, since a sub-sequent message in a sub-sequent
...
+more information.
+
+
+#### Guarantees
+
+There are three types of guarantees that the idempotent producer can satisfy:
+
+ * Exactly-once - a message is only written to the log once.
+                  Does NOT cover the exactly-once consumer case.
+ * Ordering - a series of messages are written to the log in the
```
mhowlett: would be good to clarify whether this does or does not also imply exactly once. The way i read it is that it does not, but i'm left with some confusion.

INTRODUCTION.md
```diff
@@ -388,6 +392,244 @@ may be produced out of order, since a sub-sequent message in a sub-sequent
...
+Using the Idempotent Producer prevents reordering even with `max.in.flight` > 1,
+see [Idempotent Producer](#idempotent-producer) below for more information.
+
+
+### Idempotent Producer
+
+librdkafka supports the idempotent producer which provides strict ordering and
+and exactly-once producer guarantees.
+The idempotent producer is enabled by setting the `enable.idempotence`
+configuration property to `true`, this will automatically adjust a number of
```
mhowlett: the end of https://github.com/confluentinc/confluent-kafka-dotnet/pull/623#issuecomment-421098810 describes some flow on problems that this will cause.

CONFIGURATION.md
mhowlett: sounds good

INTRODUCTION.md
```diff
@@ -388,6 +392,244 @@ may be produced out of order, since a sub-sequent message in a sub-sequent
...
+   it was not written to the log.
+   Application retry will risk ordering, but not duplication.
+ * `RD_KAFKA_MSG_STATUS_POSSIBLY_PERSISTED` - the message was transmitted
+   to the broker, but no acknowledgement was received.
+   Application retry will risk ordering and duplication.
+ * `RD_KAFKA_MSG_STATUS_PERSISTED` - the message was written to the log by
+   the broker and fully acknowledged.
+   No reason for application to retry.
+
+This method should be called by the application on delivery report error.
```
mhowlett: In the bindings I think this means we will now have a new error type `MessageError` which is a subclass of `Error` that adds an additional enum property `Status`.

INTRODUCTION.md
```diff
@@ -388,6 +392,244 @@ may be produced out of order, since a sub-sequent message in a sub-sequent
...
+The follow sections describe librdkafka's handling of the
+Idempotent Producer specific errors that may be returned by the broker.
+
+
+##### RD_KAFKA_RESP_ERR_OUT_OF_ORDER_SEQUENCE_NUMBER
+
+This error is returned by the broker when the sequence number in the
+ProduceRequest is larger than the expected next sequence
+for the given PID+Epoch+Partition (last BaseSeq + msgcount + 1).
+Note: sequence 0 is always accepted.
```
mhowlett: even after successful produce requests to a PID/Epoch/Partition?

INTRODUCTION.md
```diff
@@ -461,29 +703,51 @@ A poll-based API is used to provide signaling back to the application,
...
+    An application typically does not have to perform any action when
+    an error is raised through the error callback, the client will
+    automatically try to recover from all errors, given that the
+    client and cluster is correctly configured.
+    In some specific cases a fatal error may occur which will render
+    the client more or less inoperable for further use:
+    if the error code in the error callback is set to
+    `RD_KAFKA_RESP_ERR__FATAL` the application should retrieve the
+    underlying fatal error and reason using the `rd_kafka_fatal_error()` call,
+    and then begin terminating the instance.
```
mhowlett: In the bindings, I think this means we want a new error type `ErrorEvent` which is a subclass of `Error` and provides an additional `bool` property `IsFatal`. It doesn't seem like we want a more fine-grained property than this to indicate severity level.

INTRODUCTION.md
```diff
@@ -747,6 +1016,49 @@ is returned.
...
+the application.
+`rd_kafka_purge(RD_KAFKA_PURGE_F_QUEUE)` is automatically called by the client
+when a producer fatal error has occurred, messages in-flight are not purged
+automatically to allow waiting for the proper acknowledgement from the broker.
+The purged messages in queue will fail with error code set to
+`RD_KAFKA_RESP_ERR__PURGE_QUEUE`.
+
+
+
+
```
mhowlett: good documentation 👍

examples/rdkafka_consumer_example.cpp
```diff
@@ -96,11 +96,14 @@ class ExampleEventCb : public RdKafka::EventCb {
 
     switch (event.type())
     {
-      case RdKafka::Event::EVENT_ERROR:
+      case RdKafka::Event::
+        EVENT_ERROR:
```
mhowlett: is this on a new line by mistake?

examples/rdkafka_performance.c
```diff
@@ -131,8 +130,15 @@ uint64_t wall_clock (void) {
 }
 
 static void err_cb (rd_kafka_t *rk, int err, const char *reason, void *opaque) {
-	printf("%% ERROR CALLBACK: %s: %s: %s\n",
-	       rd_kafka_name(rk), rd_kafka_err2str(err), reason);
+        if (err == RD_KAFKA_RESP_ERR__FATAL) {
```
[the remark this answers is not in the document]
mhowlett: oh ok, i see, it makes a lot of sense because it's a global thing you want to get access to from anywhere, not just the context of where the error was first known. very good.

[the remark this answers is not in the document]
mhowlett: It would still have been nice to use a bit to indicate an error is fatal rather than have a separate error code, but less important given the above.

INTRODUCTION.md
edenhill: This lists what guarantees the idemp producer satisfies; not necessarily which guarantees you can select to use.
Ordering and exactly-once are always enabled with idemp, gapless was added to cover corner cases which the Java producer handles in a different way.

I'll clarify

INTRODUCTION.md
edenhill: Maybe for .NET, which is the only client using different classes for succeeded or failed messages.

While the status will always be PERSISTED for successfully delivered messages, I don't think it hurts to provide a Status() method on the base producer message class so it is usable both for succeeded and failed messages. I could see a future enhancement which sets POSSIBLY_PERSISTED if producing with acks=1.

INTRODUCTION.md
```diff
@@ -388,6 +392,244 @@ may be produced out of order, since a sub-sequent message in a sub-sequent
...
+The librdkafka msgseq is used, along with a base msgseq value stored
+at the time the PID/Epoch was bumped, to calculate the Kafka protocol's
+message sequence number.
+
+With Idempotent Producer enabled there is no risk of reordering despite
+`max.in.flight` > 1 (capped at 5).
+
+**Note**: "msgseq" in log messages refer to the librdkafka msgseq, while "seq"
+          refers to the protocol message sequence, "baseseq" is the msgseq of
+          the first message in a batch.
```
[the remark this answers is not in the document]
edenhill: Thanks for the thumbs-up, made me find an error in that sentence :)
"baseseq" is the *seq* of .., not *msgseq*

INTRODUCTION.md
edenhill: Yes, I believe so. Seq 0 resets the sequence on the broker.
Right @hachikuji ?

INTRODUCTION.md
edenhill: Yep, that's pretty much what the C++ API looks like: it combines log_cb, stats_cb, throttle_cb and error_cb into an Event object.
The Error Event has a fatal() a method.

Fatal errors are basically exceptions: ERR__FATAL indicates that the a fatal error has been raised, you then use rd_kafka_fatal_error() to extract the actual fatal error.
We should abstract in the most idiomatic way in the high-level clients, probably through exceptions raised by poll()/flush()/consume()/etc for .NET and Python.
For Go we'll keep using KafkaError but expose a fatal() method.

edenhill:
Naming-wise I'm not fully convinced that "POSSIBLY_PERSISTED" and "gapless" are 100%, comments, suggestions?

edenhill:
WIP Python branch https://github.com/confluentinc/confluent-kafka-python/tree/idempotence which adds support for fatal errors.

INTRODUCTION.md
mhowlett: i was very good with the complaining, but failed to come up with something superior.

INTRODUCTION.md
mhowlett: yep, status properly belongs a level above error, you're right.

INTRODUCTION.md
```diff
@@ -388,6 +392,250 @@ may be produced out of order, since a sub-sequent message in a sub-sequent
...
+
+To help the application decide what to do in these error cases, a new
+per-message API is introduced, `rd_kafka_message_persistance_status()`,
+which returns one of the following values:
+
+ * `RD_KAFKA_MSG_STATUS_NOT_PERSISTED` - the message has never
+   been transmitted to the broker, or failed with an error indicating
+   it was not written to the log.
+   Application retry will risk ordering, but not duplication.
+ * `RD_KAFKA_MSG_STATUS_POSSIBLY_PERSISTED` - the message was transmitted
```
mhowlett: Responding to another comment by you about this name: when I read this through the first time, my reaction to this name wasn't negative at all. thinking about it more, I think I prefer `RD_KAFKA_MSG_STATUS_MAYBE_PERSISTED`, but my preference is marginal.

STATISTICS.md
```diff
@@ -189,6 +193,17 @@ rebalance_cnt | int | | Total number of rebalances (assign or revoke)
 assignment_size | int gauge | | Current assignment's partition count
 
 
+## eos
+
+Field | Type | Example | Description
+----- | ---- | ------- | -----------
```
rnpridgeon: 👍

src-cpp/rdkafkacpp.h
```diff
@@ -1380,6 +1460,25 @@ class RD_EXPORT MessageTimestamp {
...
+
+    /**< Message was transmitted to broker, but no acknowledgement was
+     *   received.
+     *   Application retry risks ordering and duplication. */
+    MSG_STATUS_POSSIBLY_PERSISTED = 1,
+
+    /**< Message was written to the log and fully acknowledged.
+     *   No reason for application to retry.
+     *   Note: this value should only be trusted with \c acks=all. */
+    MSG_STATUS_PERSISTED =  2
```
mhowlett: it's questionable, but I think doing as you are an calling successful acks=1 produce requests `PERSISTED` is ok.

src/rdkafka.h
```diff
@@ -1385,6 +1496,14 @@ void rd_kafka_conf_set_dr_cb(rd_kafka_conf_t *conf,
  *
  * An application must call rd_kafka_poll() at regular intervals to
  * serve queued delivery report callbacks.
+ *
+ * The broker-assigned offset can be retrieved with \c rkmessage->offset
+ * ((if `producer.offset.report` is true), and the timestamp can be
```
mhowlett: the property was `produce.offset.report` and it no longer exists.

mhowlett: `produce.offset.report` is also referred to in a comment in `rd_kafka_message_s`

src/rdkafka.h
```diff
@@ -1385,6 +1496,14 @@ void rd_kafka_conf_set_dr_cb(rd_kafka_conf_t *conf,
...
+ * retrieved using rd_kafka_message_timestamp().
+ *
+ * @remark The Idempotent Producer may return invalid timestamp
+ *         and offset for retried messages that were previously
+ *         successfully delivered but not properly acknowledged.
```
mhowlett: after reading this i'm wondering:
1. can there potentially be more than one DR for a given message (assume no)?
2. could the DR returned for a successfully delivered message have invalid properties - timestamp and offset?
3. could the DR for a successfully produced message have an error code != `NoError`?

if yes to any of the above, doesn't seem ideal.

src/rdkafka.h
```diff
@@ -1511,9 +1630,19 @@ void rd_kafka_conf_set_offset_commit_cb (
 /**
  * @brief Set error callback in provided conf object.
  *
- * The error callback is used by librdkafka to signal critical errors
+ * The error callback is used by librdkafka to signal warnings and errors
  * back to the application.
  *
+ * These errors are generally to be considered informational and non-permanent,
```
mhowlett: 'These errors should generally be considered ...'

src/rdkafka_conf.c
```diff
@@ -1024,6 +1111,67 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
+
+        if (!(prop = rd_kafka_conf_prop_find(_RK_GLOBAL, name)))
+            return 0;
+
+        return rd_kafka_anyconf_is_modified(conf, prop);
+}
+
+
+/**
+ * @returns true if property has been set/modified, else 0.
```
mhowlett: nit: probably should use `rd_false`, `rd_true`, or at least not switch between int and bool mid sentence.

src/rdkafka_idempotence.c
```diff
@@ -0,0 +1,431 @@
...
+
+
+/**
+ * @brief Acquire Pid by looking up a suitable broker and then
+ *        sending an InitProducerIdRequest to it.
+ *
+ * @param rkb may be set to specify a broker to use, otherwise a suitable
+ *            one is looked up.
+ *
+ * @returns 1 if a request was enqueued, or 0 if no broker was available.
```
mhowlett: nit: slightly incorrect

src/rdkafka_idempotence.c
```diff
@@ -0,0 +1,431 @@
...
+
+        rd_rkb_dbg(rkb, EOS, "GETPID",
+                   "Failed to acquire PID: %s", rd_kafka_err2str(err));
+
+        if (err == RD_KAFKA_RESP_ERR__DESTROY)
+                return; /* Ignore */
+
+        rd_assert(thrd_is_current(rk->rk_thread));
+
+        /* FIXME: Handle special errors, maybe raise certain errors
```
mhowlett: did you mean to fixme?

src/rdkafka_idempotence.c
```diff
@@ -0,0 +1,431 @@
...
+                rd_kafka_wrunlock(rkb->rkb_rk);
+                rd_kafka_broker_destroy(rkb);
+                return 1;
+        }
+
+        rd_kafka_wrunlock(rkb->rkb_rk);
+
+        rd_rkb_dbg(rkb, EOS, "GETPID",
+                   "Can't acquire ProducerId from this broker: %s", errstr);
+        rd_kafka_idemp_restart_request_pid_tmr(rk, 500);
```
mhowlett: nit: the 500's a bit magical, used in a few places, could be a constant somewhere.

tests/test.c
```diff
@@ -401,12 +416,30 @@ void test_socket_enable (rd_kafka_conf_t *conf) {
 
 static void test_error_cb (rd_kafka_t *rk, int err,
 			   const char *reason, void *opaque) {
-        if (test_curr->is_fatal_cb && !test_curr->is_fatal_cb(rk, err, reason))
-                TEST_SAY(_C_YEL "rdkafka error (non-fatal): %s: %s\n",
+        if (test_curr->is_fatal_cb && !test_curr->is_fatal_cb(rk, err, reason)) {
+                TEST_SAY(_C_YEL "rdkafka error (non-testfatal): %s: %s\n",
```
mhowlett: maybe 'but not test fatal'

tests/0086-purge.c
```diff
@@ -0,0 +1,308 @@
...
+ *  - purge(INFLIGHT) => no change in len()
+ *  - purge(QUEUE) => len() should drop to 0, dr errs should be ERR__PURGE_QUEUE
+ *
+ * Remote test (WITH_SOCKEM):
+ *  - Limit in-flight messages to 10
+ *  - Produce 20 messages to the same partition, in batches of 10.
+ *  - Make sure only first batch is sent.
+ *  - purge(QUEUE) => len should drop to 10, dr err ERR__PURGE_QUEUE
+ *  - purge(INFLIGHT|QUEUE) => len should drop to 0, ERR__PURGE_INFLIGHT
+ */
```
mhowlett: I guess there's scope for additional variants, but I think this is good enough.

mhowlett:
This is LGTM except I'm a bit un-easy about the level of test coverage.

src/rdkafka.h
edenhill: 1. No
 2. Yes, if the message was retried and got the ERR__DUPLICATE error back from the broker, which indicates that the message was persisted on the first produce but the ack was lost. But there is nothing we can do about it but document the behaviour.
 3. No

src/rdkafka_conf.c
edenhill: haha

src-cpp/rdkafkacpp.h
rnpridgeon: I disagree, `acks=1` should be `MSG_STATUS_POSSIBLY_PERSISTED`. Broker failures and log divergence is actually quite common in practice.

Issue #1998, opened 2018-09-12
Failed to create thread: No error information (0)


nick-zh:
I am in a alpine 3.8 container and installing (make and make install) librdkafka 0.11.5.
When i am instantiating a new Producer (in php), i am getting:
```Failed to create thread: No error information (0)```
Any insight on why this might happen? With 0.11.4 it is working fine

**Comment by nick-zh** [comment]
So i was able to narrow it down, it seems that installing it this way:
```
git clone --depth 1 --branch v0.11.5 https://github.com/edenhill/librdkafka.git && \
cd librdkafka
./configure
make
make install
```
causes this to happen, when i install the package from edge:
```
apk add librdkafka librdkafka-dev --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/community
```
it seems to work. I leave it open for now @edenhill maybe you have some insight in why this might happen. If not, you can close it.

**Comment by edenhill** [comment]
I don't see how the error information is lost.
pthread_create() fails here and sets errno:
https://github.com/edenhill/librdkafka/blob/master/src/tinycthread.c#L665

errno is printed from the caller, now with a value of 0, which shouldnt be possible:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.c#L1734

Unless the free() at tinycthread.c line 676 resets the errno, which it shouldnt, and looking at the musl source it doesn't seem to do that unless syscall() does.

**Comment by nick-zh** [comment]
Hey @edenhill 
So i was able to reproduce this, you can unpack my zip file and do a
```
docker-compose up
docker exec -it <fpm container id> sh
./setup.sh
php kafka_producer.php
```
[debug.zip](https://github.com/edenhill/librdkafka/files/2387538/debug.zip)

So in my php file, when i remove this line:
```
$topicConfig->set('message.timeout.ms', 5000);
```
I get an error with information:
```
Uncaught RdKafka\Exception: Failed to create thread: Invalid seek (29)
```
If i leave it in, i still get, no information on the error though:
```
Uncaught RdKafka\Exception: Failed to create thread: No error information (0)
```

edenhill:
:100: for reproducible environment!

edenhill:
Found the issue: librdkafka's internal C11 threads (tinycthread) is mixed with the musl-provided C11 threads, causing weird errors and crashes.

We'll need to disable our own C11 threads implementation when a system one is available.

PR #2007, opened 2018-09-17
Use C11 threads when available

edenhill:
When libc provides C11 threads, which is the case with musl on alpine, the internal librdkafka c11 thread implementation (tinycthread), interferes with the musl symbols, leading to weird errors and crashes.

There are no code changes, only refactoring.

This PR makes librdkafka use the system C11 threads if available.

Reported in #1998

### DIFF
--- a/src/tinycthread_extra.c
+++ b/src/tinycthread_extra.c
@@ -1,0 +1,153 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2018 Magnus Edenhill
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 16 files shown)


src/tinycthread_extra.c
```diff
@@ -0,0 +1,153 @@
...
+#if HAVE_PTHREAD_SETNAME_GNU
+        if (!pthread_setname_np(pthread_self(), name))
+                return thrd_success;
+#endif
+        return thrd_error;
+}
+
+int thrd_is_current(thrd_t thr) {
+#if defined(_TTHREAD_WIN32_)
+        return GetThreadId(thr) == GetCurrentThreadId();
```
rnpridgeon: Use C11 thrd_current() ?

src/tinycthread_extra.c
```diff
@@ -0,0 +1,153 @@
...
+#endif
+}
+
+
+
+
+int cnd_timedwait_ms(cnd_t *cnd, mtx_t *mtx, int timeout_ms) {
+        if (timeout_ms == -1 /* INFINITE*/)
+                return cnd_wait(cnd, mtx);
+#if defined(_TTHREAD_WIN32_)
```
rnpridgeon: Is it appropriate to check for C11 support as opposed to tiny thread here?

src/tinycthread_extra.c
edenhill: Windows has this really weird two-level thread ids.
When you create a thread you get a thread handle, which is not comparable to other thread handles.

src/tinycthread_extra.c
edenhill: Windows' timed cond wait already takes a relative timeout, so this is an optimization to avoid converting to absolute and then back to relative again.
pthread cond wait always take abs time.

Issue #1806, opened 2018-05-15
Missing license and/or copyright for scripts


jvymazal:
Description
===========
Some script and "pseudo-source" files in repo are missing license and/or attribution notice. This can lead to problems for packagers in some distributions. Please consider adding them.

How to reproduce
================
open thinkgs like  [lds-gen](https://github.com/edenhill/librdkafka/blob/master/lds-gen.py), and [various](https://github.com/edenhill/librdkafka/blob/master/configure) [configure](https://github.com/edenhill/librdkafka/blob/master/configure.librdkafka) [files](https://github.com/edenhill/librdkafka/blob/master/dev-conf.sh) (possibly others, have not checked every file in project)
Thanks :-)

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `3fd5e8b4514cb51da23ec452fc7987e26adb026e`
 - [x] Apache Kafka version: `N/A`
 - [x] librdkafka client configuration: `N/A`
 - [x] Operating system: `Fedora 28 x64`
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka - N/A
 - [x] Provide broker log excerpts - N/A
 - [x] Critical issue - no

edenhill:
From a packager's perspective the licenses are only needed for packaged files, and I don't believe any of the scripts are supposed to be included in a prebuilt package of librdkafka.

jvymazal:
Not for the pre-built one, but [source-packages](http://mirrors.isu.net.sa/pub/fedora/fedora-epel/7/SRPMS/l/librdkafka-0.11.0-1.el7.src.rpm) exist as well :-) .

PR #2009, opened 2018-09-18
Added copyright check (#1806)

edenhill:
This is the right thing to do and helps packagers (e.g. debian) to avoid lintian warnings.

No code changes.


rnpridgeon:
Please elaborate on why this "is the right thing to do" to the uninitiated

edenhill:
When possible all files should have a license header, and I think this surfaced for the debian package maintainers as some of the packaged files did not.

PR #2017, opened 2018-09-21
Fix 64-bit IBM i build error

ThePrez:
64-bit builds on IBM i fail, as follows:
```
ld: 0711-317 ERROR: Undefined symbol: .le64toh
ld: 0711-345 Use the -bloadmap or -bnoquiet option to obtain more information.
```
(IBM i is using parts of the AIX runtime, and this fix is also applicable to AIX, hence why it is included in the AIX path)


YUbuntu0109:
HI😊

PR #2027, opened 2018-09-27
Makefile: fix install rule

pacovn:
Added the 'all' dependency to the 'install' rule so it builds and installs when calling "make install" after the "./configure"


pacovn:
@edenhill The actual issue is that "make install" returns exit code 0 when failing because no previous "make" was done.

pacovn:
I've created issue #2048 for this.

Issue #1542, opened 2017-11-22
Counting error in rdkafka_performance


LionelCons:
I ran `rdkafka_performance` with:
```
$ rdkafka_performance -P -t performance-tests -c 60000000 -s 1000 -S 1 -a 0
```

It worked fine and at some point it has produced all the messages requested:
```
% 59895281 messages produced (59895281000 bytes), 59395313 delivered (offset 0, 0 failed) in 1032819ms: 57507 msgs/s and 57.51 MB/s, 1583935 produce failures, 499969 in queue, no compression
% 59943352 messages produced (59943352000 bytes), 59443359 delivered (offset 0, 0 failed) in 1033819ms: 57498 msgs/s and 57.50 MB/s, 1586516 produce failures, 499995 in queue, no compression
% 59999218 messages produced (59999218000 bytes), 59499222 delivered (offset 0, 0 failed) in 1034820ms: 57497 msgs/s and 57.50 MB/s, 1588434 produce failures, 499996 in queue, no compression
% 60000000 messages produced (60000000000 bytes), 59521253 delivered (offset 0, 0 failed) in 1035821ms: 57462 msgs/s and 57.46 MB/s, 1588447 produce failures, 478747 in queue, no compression
% 60000000 messages produced (60000000000 bytes), 59541214 delivered (offset 0, 0 failed) in 1036827ms: 57426 msgs/s and 57.43 MB/s, 1588447 produce failures, 458786 in queue, no compression
% 60000000 messages produced (60000000000 bytes), 59561058 delivered (offset 0, 0 failed) in 1037853ms: 57388 msgs/s and 57.39 MB/s, 1588447 produce failures, 438943 in queue, no compression
```

So far, so good.

However, near the end of the run, it reported:
```
% 60000000 messages produced (60000000000 bytes), 59960874 delivered (offset 0, 0 failed) in 1058937ms: 56623 msgs/s and 56.62 MB/s, 1588447 produce failures, 39128 in queue, no compression
% 60000000 messages produced (60000000000 bytes), 59979812 delivered (offset 0, 0 failed) in 1059938ms: 56587 msgs/s and 56.59 MB/s, 1588447 produce failures, 20198 in queue, no compression
% 60000000 messages produced (60000000000 bytes), 59997103 delivered (offset 0, 0 failed) in 1060944ms: 56550 msgs/s and 56.55 MB/s, 1588447 produce failures, 2901 in queue, no compression
[...]

PR #2028, opened 2018-09-27
Fixing Counting error in rdkafka_performance #1542

gnanasekarl:
Removing the deduction of outq len from the total message produced count, which creates unnecessary mismatch between delivered and produced counters if in case there is some messages in outq.
Reference: issue#1542

### DIFF
--- a/examples/rdkafka_performance.c
+++ b/examples/rdkafka_performance.c
@@ -1432,7 +1432,6 @@ int main (int argc, char **argv) {
 		outq = rd_kafka_outq_len(rk);
                 if (verbosity >= 2)
                         printf("%% %i messages in outq\n", outq);
-
 		cnt.t_end = t_end;
 
 		if (cnt.tx_err > 0)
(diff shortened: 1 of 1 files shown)


gnanasekarl:
Hi, 
AppVeyor build failed due to the below reason
_"
# Install CoApp module
#Install-Module CoApp -Force
Installing OpenSSL v1.0 32-bit ...
Downloading...
Exception calling "DownloadFile" with "2" argument(s): "The request was aborted: Could not create SSL/TLS secure channel."
At line:10 char:1
+ (New-Object Net.WebClient).DownloadFile('https://slproweb.com/downloa ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
 
Installing...
"_

What is this?

edenhill:
You can ignore that error, CI environment related

PR #2032, opened 2018-10-01
Fix high cpu sasl loop after disconnect

stefanseufert:
Should fix https://github.com/edenhill/librdkafka/issues/1937

### DIFF
--- a/src/rdkafka_transport.c
+++ b/src/rdkafka_transport.c
@@ -1412,6 +1412,16 @@ static void rd_kafka_transport_io_event (rd_kafka_transport_t *rktrans,
 					     errstr);
 			return;
 		}
+
+		if (events & POLLHUP) {
+			errno = EINVAL;
+			rd_kafka_broker_fail(rkb, LOG_ERR,
+					     RD_KAFKA_RESP_ERR__AUTHENTICATION,
+					     "Disconnected");
+
+			return;
+		}
+
 		break;
 
 	case RD_KAFKA_BROKER_STATE_APIVERSION_QUERY:


src/rdkafka_transport.c
```diff
@@ -1403,6 +1403,15 @@ static void rd_kafka_transport_io_event (rd_kafka_transport_t *rktrans,
 
 	case RD_KAFKA_BROKER_STATE_AUTH:
 		/* SASL handshake */
+
+        if (events & POLLHUP) {
```
edenhill: We must read remaining socket buffer data if events&POLLIN, to make sure we receive any auth errors sent by the broker before it disconnects the socket.
So move this check down below the sasl_io_event() call.

src/rdkafka_transport.c
```diff
@@ -1403,6 +1403,15 @@ static void rd_kafka_transport_io_event (rd_kafka_transport_t *rktrans,
...
+            rd_kafka_broker_fail(rkb, LOG_ERR,
```
edenhill: set `errno = EINVAL;` prior to calling broker_fail(). It logs the errno in debug messages and we want to have it set to a known value.

src/rdkafka_transport.c
edenhill: Set tab indent to 8 whitespaces.

stefanseufert: Done as requested.

src/rdkafka_transport.c
stefanseufert: Done as requested.

src/rdkafka_transport.c
stefanseufert: Done as requested.

stefanseufert:
Rebased and squashed fixes into original commit as per contributing guidelines

stefanseufert:
@edenhill anything else you want me to so here? According to your comments in various threads, 11.6 seems to be close and   I really hope to see this included.

edenhill:
0.11.6 would be a stretch, it is way past its deadline and needs to get out the door.
But we're doing 1.0.0 release candidates very soon, which would be a better fit for this.

stefanseufert:
This is a really nasty bug for everyone using SASL auth. A restart of any broker node will immediatedly make the clients go 100% CPU per thread. I've seen boxes logging CPU loads in excess of 350 due to this issue, mailing them completely unaccessible. Since this also affects producers it will result in message loss. There are no known workarounds and there ist no way to detect the problem in the calling code. As far as I can see all platforms are affected. Not sure how much worse a bug could be?

The only reason, why this hasn't cought much more attention seems to be the fact that auth isn't used that commonly.

The fix on the other hand seems trivial ans will only affect those people using auth.

... but it certainly is your call.

edenhill:
Thanks alot for this!
(yes, it will go into 0.11.6 ;) )

PR #2054, opened 2018-10-14
KIP-62: add support for max processing time (max.poll.interval.ms)

edenhill:
Changed defaults:
 * session.timeout.ms = 10000


ijuma:
Nice! Do we track these config default changes somewhere so that we can tell users in upgrade notes and such?

edenhill:
@ijuma Yep, something like "Default changes" here: https://github.com/edenhill/librdkafka/releases/tag/v0.11.4

ijuma:
Sounds good.

edenhill:
Added `rebalance_reason`.
A `max_poll_interval` metric would definitely be useful, but requires a lock rather than an atomic, so will leave this until it is really needed.

edenhill:
Added `rebalance_reason`.
A `max_poll_interval` metric would definitely be useful, but requires a lock rather than an atomic, so will leave this until it is really needed.

edenhill:
Added `rebalance_reason`.
A `max_poll_interval` metric would definitely be useful, but requires a lock rather than an atomic, so will leave this until it is really needed.

PR #2071, opened 2018-10-24
Remove Conflict Markers in rdlist.h

motorspin:


motorspin:
Let me know if you would like me to add anything to the commit as a description. This seemed like a small oversight so wasn't sure if I should put anything!

edenhill:
Good find! Thank you!

PR #2075, opened 2018-10-26
CMake: Generate pk-gconfig files

Oxymoron79:
Generate the pkg-config files with CMake.

It uses the configure_file function from CMake to generate the pkg-config files for the C and C++ libraries (shared and optionally static variant) from the same template file.

Since the pkg-config files contain the version, the CMake project version is set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.2)
-project(RdKafka)
+project(RdKafka VERSION 0.11.6)
 
 # Options. No 'RDKAFKA_' prefix to match old C++ code. {
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 3.2)
-project(RdKafka VERSION 0.11.6)
+
+include("packaging/cmake/parseversion.cmake")
+parseversion("src/rdkafka.h")
+
+project(RdKafka VERSION ${RDKAFKA_VERSION})
 
 # Options. No 'RDKAFKA_' prefix to match old C++ code. {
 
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.2)
-project(RdKafka)
+project(RdKafka VERSION 0.11.6)
```
edenhill: No hardcoding of versions. It should be based on the version in src/rdkafka.h

src/CMakeLists.txt
```diff
@@ -194,6 +194,66 @@ endif()
...
+set(PKG_CONFIG_REQUIRES "")
+if(WITH_ZLIB)
+  string(APPEND PKG_CONFIG_REQUIRES "zlib ")
+endif()
+if(WITH_SSL)
+    string(APPEND PKG_CONFIG_REQUIRES "libssl ")
+endif()
+if(WITH_SASL_CYRUS)
+  string(APPEND PKG_CONFIG_REQUIRES "libsasl2 ")
+endif()
```
edenhill: +WITH_ZSTD and +WITH_LZ4_EXT

src/CMakeLists.txt
```diff
@@ -194,6 +194,66 @@ endif()
...
+set(PKG_CONFIG_CFLAGS
+    "-I\${includedir}"
+)
+set(PKG_CONFIG_LIBS
+    "-L\${libdir} -lrdkafka"
+)
+set(PKG_CONFIG_LIBS_PRIVATE
+    "-lpthread -lrt"
```
edenhill: -lrt is linux specific

src/CMakeLists.txt
Oxymoron79: Linking with external LZ4 library (`WITH_LZ4_EXT`) is not yet supported by the CMake build files.
Adding support for that should be its own issue IMHO.

edenhill:
This is great, thanks!

Oxymoron79:
You're welcome.
Thanks for merging!

PR #2104, opened 2018-11-14
OpenSSL initialization compatibility for v1.1.0 and later

wiml:
OpenSSL v1.1.0 made all of the init stuff optional and deprecated the old functions in favor of `OPENSSL_init_ssl` and `OPENSSL_init_crypto`. There are compatibility macros in place, but only if you compile with openssl's api compatibility switch set.

This change lets librdkafka compile against OpenSSL with the API compatibility level set as late as 1.1.0.

I've only done moderate testing ("it works for me").


edenhill:
Verified with openssl v1.1.0f and ASAN on osx. ✅

edenhill:
Thank you for this!

PR #2116, opened 2018-11-22
Minor doc update for queue.buffering.max.*

briot:
The producer queue is shared for all topics and partitions (we do
 not have one queue per toppar).


edenhill:
The count and size are per producer, not per partition:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_int.h#L296

So your doc change is fine, but it needs to go in rdkafka_conf.c, then rebuild and commit both rdkafka_conf.c and CONFIGURATION.md, the latter being auto generated from the former.
Make sure to only commit lines to CONFIGURATION.md that you actually intended since its output depends on build options.

briot:
I had forgotten this was auto-generated, will amend the commit thanks.

After some more thoughts on the topic, it had seemed like the producer queue limits (count and size) might be per broker. If a producer sends to two brokers (because messages are sent to different topics/partition), are we still sharing the same queue ?

On the one hand, it makes it easier to limit memory usage by the producer, but it also makes it somewhat harder to control the batches in particular to get better compression

edenhill:
These limits are global per rd_kafka_t instance, not per partition nor broker.

src/rdkafka_conf.c
```diff
@@ -917,12 +917,15 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
         { _RK_GLOBAL|_RK_PRODUCER|_RK_HIGH, "queue.buffering.max.messages",
           _RK_C_INT,
 	  _RK(queue_buffering_max_msgs),
-	  "Maximum number of messages allowed on the producer queue.",
+	  "Maximum number of messages allowed on the producer queue. "
+	  "This queue is shared by all topics and partitions.",
+
```
edenhill: remove blank line

edenhill:
Thank you!

PR #2129, opened 2018-12-01
v1.0.0 deprecations

edenhill:
See #2020

PR #2132, opened 2018-12-05
Change producer default to acks=all

edenhill:


CONFIGURATION.md
```diff
@@ -123,7 +123,7 @@ dr_msg_cb                                |  P  |                 |
...
 -----------------------------------------|-----|-----------------|--------------:|------------| --------------------------
-request.required.acks                    |  P  | -1 .. 1000      |             1 | high       | This field indicates how many acknowledgements the leader broker must receive from ISR brokers before responding to the request: *0*=Broker does not send any response/ack to client, *1*=Only the leader broker will need to ack the message, *-1* or *all*=broker will block until message is committed by all in sync replicas (ISRs) or broker's `min.insync.replicas` setting before sending response.  <br>*Type: integer*
+request.required.acks                    |  P  | -1 .. 1000      |            -1 | high       | This field indicates the number of acknowledgements the leader broker must receive from ISR brokers before responding to the request: *0*=Broker does not send any response/ack to client, *1*=Only the leader broker will need to ack the message, *-1* or *all*=broker will block until message is committed by all in sync replicas (ISRs) or broker's `min.insync.replicas` setting before sending response. <br>*Type: integer*
```
rnpridgeon: >*-1* or *all*=broker will block until message is committed by all in sync replicas (ISRs) or broker's `min.insync.replicas` setting before sending response.

Is this correct? If there are 4 replicas in the ISR and `min.insync.replicas=2` 2 acks won't satisfy my request. So it's not really an `or` so much that `min.insync.replicas` sets the bare minimum number of in-sync replicas needed in order to deem a partition healthy enough to accept new messages from my producer. 

It might be helpful to break these two concepts up to make this more clear. 

> *-1* or *all*=broker will block until message is committed by all in sync replicas (ISRs). If there are less than  `min.insync.replicas` in the ISR the produce request will fail.

src/rdkafka_conf.c
```diff
@@ -1011,19 +1011,20 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
-	  "in sync replicas (ISRs) or broker's `min.insync.replicas` "
-	  "setting before sending response. ",
-	  -1, 1000, 1,
-	  .s2i = {
-			{ -1, "all" },
-		}
-	},
+          _RKT(required_acks),
+          "This field indicates the number of acknowledgements the leader "
+          "broker must receive from ISR brokers before responding to the "
```
rnpridgeon: >ISR brokers

The broker isn't really a member of the ISR the replica is. This is an important distinction because one healthy partition does not necessarily mean you have an unhealthy broker.

edenhill: "This is an important distinction because one healthy partition does not necessarily mean you have an unhealthy broker."
Not really following that

rnpridgeon: All brokers: 1,2,3,4,5 
All replicas: 1,2,3,4,5
Leader 2 ISR list for TopicPartition `tp-1`: [1,2,4,5]

This does not necessarily mean that all replicas held by broker 3 are out of sync. It means that replica 3 for Topic-Partition `tp-1` is out of sync. Calling the ISR member `ISR Brokers` implies that the broker has fallen out of sync rather than the replica that lives on that broker.

edenhill: According to TopicMetadata the leader considers itself part of the ISR:
```
  topic "test" with 4 partitions:
    partition 0, leader 4, replicas: 4,2,3, isrs: 4,2,3
```

rnpridgeon: As it should be but again this is reporting the state of the replica not necessarily the broker.

rnpridgeon: From a producer perspective which thinks only at the TopicPartition level  I guess this is not true.

CONFIGURATION.md
```diff
@@ -53,8 +53,8 @@ default_topic_conf                       |  *  |                 |
...
-broker.version.fallback                  |  *  |                 |        0.10.0 | medium     | Older broker versions (before 0.10.0) provide no way for a client to query for supported protocol features (ApiVersionRequest, see `api.version.request`) making it impossible for the client to know what features it may use. As a workaround a user may set this property to the expected broker version and the client will automatically adjust its feature set accordingly if the ApiVersionRequest fails (or is disabled). The fallback broker version will be used for `api.version.fallback.ms`. Valid values are: 0.9.0, 0.8.2, 0.8.1, 0.8.0. Any other value, such as 0.10.2.1, enables ApiVersionRequests. <br>*Type: string*
+api.version.fallback.ms                  |  *  | 0 .. 604800000  |             0 | medium     | Dictates how long the `broker.version.fallback` fallback is used in the case the ApiVersionRequest fails. **NOTE**: The ApiVersionRequest is only issued when a new connection to the broker is made (such as after an upgrade). <br>*Type: integer*
```
rnpridgeon: :+1:

CONFIGURATION.md
```diff
@@ -53,8 +53,8 @@ default_topic_conf                       |  *  |                 |
...
+broker.version.fallback                  |  *  |                 |        0.10.0 | medium     | Older broker versions (before 0.10.0) provide no way for a client to query for supported protocol features (ApiVersionRequest, see `api.version.request`) making it impossible for the client to know what features it may use. As a workaround a user may set this property to the expected broker version and the client will automatically adjust its feature set accordingly if the ApiVersionRequest fails (or is disabled). The fallback broker version will be used for `api.version.fallback.ms`. Valid values are: 0.9.0, 0.8.2, 0.8.1, 0.8.0. Any other value >= 0.10, such as 0.10.2.1, enables ApiVersionRequests. <br>*Type: string*
```
rnpridgeon: This is already in #2129 is it not?

edenhill: Added "Any other value >= 0.10" for clarification

PR #2139, opened 2018-12-10
Change connection close log levels, depending on in-flight requests

edenhill:


edenhill:
Ping @jcarey03

mhowlett:
LGTM. I don't have context for this change, but in case it's relevant, want to note that if the word 'error' appears in a log message, people will assume it's an error even if the level is info or debug.

PR #2141, opened 2018-12-11
Build: avoid interactive 'mv' prompt

pacovn:
Signed-off-by: F. Aragon <paco@voltanet.io>


edenhill:
Thank you for this, will merge after the next release.

edenhill:
It took some time, but thank you!

Issue #2138, opened 2018-12-10
calling unsubscribe immediately before destroy causes infinite loop


mhowlett:
Description
===========

there's a regression in one of the .net integration tests (librdkafka 1.0-RC4 vs 0.11.6) in which consumer unsubscribe is called immediately before the consumer is disposed (effectively calling `Librdkafka.destroy_flags(handle, (IntPtr)Librdkafka.DestroyFlags.RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE)`). The issue is the call to dispose blocks indefinitely. 

https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Exiting.cs#L76

librdkafka is emitting the following log message many thousands of times:

`127.0.0.1:9092/0: Handle is terminating in state APIVERSION_QUERY: 2 refcnts (0x7fefb0061b28), 0 toppar(s), 0 active toppar(s), 0 outbufs, 0 waitresps, 0 retrybufs: failed 0 request(s) in retry+outbuf`

PR #2142, opened 2018-12-11
Consumer: don't send Leave when terminating with NO_CONSUMER_CLOSE (#2138)

edenhill:
@mhowlett Please verify this fix with the .NET client


mhowlett:
LGTM & verified that with this branch of librdkafka, the .NET Consumer_Exiting integration test no longer hangs here: https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Exiting.cs#L76

PR #2146, opened 2018-12-12
Add dup config to C++ interface

noahdav:
Added the ability to duplicate a configuration in the C++ interface.  This is exposed in the C interface but we need this in C++ interface too.


edenhill:
Outdated. Please update and reopen if still relevant.
Thanks

PR #2152, opened 2018-12-12
Use separate connection to the group coordinator

edenhill:
Create separate connection to consumer Group Coordinator

    Since JoinGroupRequests may block for up to max.poll.interval.ms,
    which may be set very high (hours..), any sub-sequent requests
    on the same connection, such as Metadata refreshes, would time out
    and tear down the connection, triggering another rebalance.

    This commit adds support for logical brokers, in this case
    a GroupCoordinator broker, which initially doesn't have an address
    but is updated by the cgrp to connect to the group coordinator.
    This connection is then used for all group-related protocol requests,
    Join, Sync, Leave, OffsetCommit, leaving the standard broker
    connections for other protocol use.

    As a side-effect this commit also deletes legacy cgrp code from when
    it used to run on the broker thread.

    Also improves HOLB detection and handling.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -136,9 +136,11 @@ static void rd_kafka_mk_brokername (char *dest, size_t dsize,
 				    const char *nodename, int32_t nodeid,
 				    rd_kafka_confsource_t source) {
 
-	/* Prepend protocol name to brokername, unless it is a
-	 * standard plaintext broker in which case we omit the protocol part. */
-	if (proto != RD_KAFKA_PROTO_PLAINTEXT) {
+        /* Prepend protocol name to brokername, unless it is a
+         * standard plaintext or logical broker in which case we
+         * omit the protocol part. */
+        if (proto != RD_KAFKA_PROTO_PLAINTEXT &&
+            source != RD_KAFKA_LOGICAL) {
 		int r = rd_snprintf(dest, dsize, "%s://",
 				    rd_kafka_secproto_names[proto]);
 		if (r >= (int)dsize) /* Skip proto name if it wont fit.. */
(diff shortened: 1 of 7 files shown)


src/rdkafka_broker.c
```diff
@@ -640,14 +644,43 @@ static int rd_kafka_broker_bufq_timeout_scan (rd_kafka_broker_t *rkb,
...
+                        /* Head of line blocking:
+                         * If this is not the first request in queue, but the
+                         * initial first request did not time out,
+                         * it typically means the first request is a
+                         * long-running blocking one, holding up the
+                         * sub-sequent requests.
+                         * In this case log what is likely holding up the
+                         * requests and what caused this request to time out. */
+                        if (holb && holb == TAILQ_FIRST(&rkbq->rkbq_bufs)) {
+                                rd_snprintf(holbstr, sizeof(holbstr),
```
mhowlett: nice log message. is the `holb == TAILQ_FIRST(&rkbq->rkbq_bufs` condition superfluous? (always true if `hobl != NULL`)?

src/rdkafka_broker.c
```diff
@@ -2479,6 +2529,20 @@ static int rd_kafka_broker_cmp_by_id (const void *_a, const void *_b) {
 }
 
 
+/**
+ * @brief Set the broker logname (used in logs) to a copy of \p logname.
+ *
+ * @locality any
+ * @locks none
+ */
```
mhowlett: what does locks none mean, given you're locking `rkb->rkb_logname_lock`

mhowlett:
LGTM, probably want a test or two?

src/rdkafka_broker.c
edenhill: holb points to the first element in the queue before anything has timed out.
At this point the first element, holb, might have timed out and removed from the queue, so it might no longer be the first element.

src/rdkafka_broker.c
edenhill: `@locks` instructs the caller what locks it needs to hold to call this function.

rnpridgeon: TIL

PR #2155, opened 2018-12-13
Include error message in CreatePartitions result

### LINKED ISSUE #2154: Missing broker message upon create partitions error
Description
===========
If an error occurs while handling a create partitions request, the broker returns an error message.  See the `error_message` field in CreatePartitionsResponse. Currently librdkafka does not expose this message but instead always returns the default message for the error code.

How to reproduce
================
- Call `rd_kafka_CreatePartitions()` with invalid arguments. For example specifying a new partition count less or equals to current count.

- Call `rd_kafka_topic_result_error_string()` and get the default message instead of the custom message from the broker.

Checklist
=========

 - [x] librdkafka version (release number or git tag): ` fa71cce`
 - [X] Apache Kafka version: `>1.0.0`
 - [X] librdkafka client configuration: `default`
 - [X] Operating system: `all`

mimaison:
fixes #2154

Co-authored-by: Mickael Maison <mickael.maison@gmail.com>
Co-authored-by: Edoardo Comar <ecomar@uk.ibm.com>

### DIFF
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -1750,15 +1750,14 @@ rd_kafka_CreatePartitionsResponse_parse (rd_kafka_op_t *rko_req,
                                 errstr = (char *)rd_kafka_err2str(error_code);
                         else
                                 RD_KAFKAP_STR_DUPA(&errstr, &error_msg);
+                } else {
+                        errstr = NULL;
                 }
 
 
                 terr = rd_kafka_topic_result_new(ktopic.str,
                                                  RD_KAFKAP_STR_LEN(&ktopic),
-                                                 error_code,
-                                                 error_code ?
-                                                 rd_kafka_err2str(error_code) :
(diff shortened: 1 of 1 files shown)


src/rdkafka_admin.c
```diff
@@ -1750,15 +1750,14 @@ rd_kafka_CreatePartitionsResponse_parse (rd_kafka_op_t *rko_req,
                                 errstr = (char *)rd_kafka_err2str(error_code);
                         else
                                 RD_KAFKAP_STR_DUPA(&errstr, &error_msg);
+                } else {
+                        errstr = NULL;
```
edenhill: errstr is the parent's buffer, which might be needed later in this function, so suggest not setting it to NULL here.
Instead change the result_new(..) call below to `error_code ? errstr : NULL`

edenhill:
Thank you for this!

PR #2157, opened 2018-12-17
Update librdkafka.vcxproj

yedong113:


edenhill:
What's the reason for this change?

Issue #1946, opened 2018-08-13
Add batched API to rd_kafka_query_watermark_offsets


JonECG:
Description
===========
Currently rd_kafka_query_watermark_offsets can only be called for an individual partition at a time and is a blocking call. It would be great if we could use batch multiple partitions into fewer requests. I tried my hand at it with the following:

rdkafka.h
```
/**
 * @brief Query broker for low (oldest/beginning) or high (newest/end) offsets
 *        for multiple partitions
 *
 * \p logical_offset must be RD_KAFKA_OFFSET_BEGINNING or RD_KAFKA_OFFSET_END.
 *
 * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success or an error code on failure.
 */
RD_EXPORT rd_kafka_resp_err_t
rd_kafka_query_watermark_offsets_batch (rd_kafka_t *rk,
                                        rd_kafka_topic_partition_list_t *partitions,
                                        int logical_offset, int timeout_ms);
```

rdkafka.c
```
struct _query_wmark_offsets_batch_state {
        rd_kafka_resp_err_t err;
        rd_kafka_topic_partition_list_t *partitions;
        rd_ts_t ts_end;
        int     requestidx;
        int     state_version;
};

static void rd_kafka_query_wmark_offsets_batch_resp_cb (rd_kafka_t *rk,
                                                        rd_kafka_broker_t *rkb,
                                                        rd_kafka_resp_err_t err,
                                                        rd_kafka_buf_t *rkbuf,
                                                        rd_kafka_buf_t *request,
                                                        void *opaque) {
        struct _query_wmark_offsets_batch_state *state;
        rd_kafka_topic_partition_list_t *offsets;
        rd_kafka_topic_partition_t *rktpar;

        if (err == RD_KAFKA_RESP_ERR__DESTROY) {
                /* 'state' has gone out of scope when query_watermark..()
                 * timed out and returned to the caller. */
                return;
        }

        state = opaque;

        offsets = rd_kafka_topic_partition_list_new(1);
        err = rd_kafka_handle_Offset(rk, rkb, err, rkbuf, request, offsets);
        if (err == RD_KAFKA_RESP_ERR__IN_PROGRESS) {
                rd_kafka_topic_partition_list_destroy(offsets);
                return; /* Retrying */
        }

	    /* Retry if no broker connection is available yet. */
	    if ((err == RD_KAFKA_RESP_ERR__WAIT_COORD ||
[...]

PR #2160, opened 2018-12-18
[WIP] Add rd_kafka_query_watermark_offsets_list api (#1946)

JonECG:
Allows multiple watermarks in a single request

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2937,6 +2937,126 @@ rd_kafka_query_watermark_offsets (rd_kafka_t *rk, const char *topic,
         return RD_KAFKA_RESP_ERR_NO_ERROR;
 }
 
+struct _query_wmark_offsets_list_state {
+        rd_kafka_resp_err_t             err;
+        rd_kafka_topic_partition_list_t *partitions;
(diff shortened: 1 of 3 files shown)


QuentinPerez:
hi @edenhill, any plan to merge it ? 

I need to fetch the high watermark for many partitions and with the current API that can take a lot of time to fetch them. 

Let me know if I can help :)

mxk1235:
@JonECG @QuentinPerez this is cool, we use something similar, although we grab both earliest and latest offset in one function call.  

my fork isn't quite ready for a PR, but i would welcome any input https://github.com/mxk1235/librdkafka/commit/4e08ca4fefb7c3c393218100821b99f701b5dbeb 

fwiw, I am trying to track down an UNKNOWN_PARTITION error that pops up relatively often, yet appears transient.

mxk1235:
just wanted to add, UKNOWN_PARTITION errors appear to be retry-able on the consumer side.  it's hard to catch in progress, possibly due to OffsetRequest responses being in-flight.   consumer->poll(100) makes sure all responses are processed, and then rd_kafka_query_watermark_offsets_list() can be retried.

PR #2177, opened 2019-01-08
If the fetch list has changed, start fetching from a random position.

LavaSpider:
This is to address the case where a program is too busy, so it has paused all consuming, then resumes when it is no longer busy. On resume, the fetch order will always be in the order of the topic/partition assignment. This means that if the busy state happens frequently, partitions earlier in the assignment will be preferred for consuming messages.


edenhill:
Thank you for this. This will not make the v1.0.0 release so will let it sit until the release is out.

LavaSpider:
Sounds good, glad I could help

Issue #2171, opened 2019-01-01
Error: Local: Host resolution failure


Vijendra07Kulhade:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
I was working with node-rdkafka and observed this error. After further debugging the issue I found that error seems to appear from native library which node-rdkafka use i.e librdkafka.
After spending sometime on it we figured out that there is a limit on hostname size.
Which is set `rdkafka_proto.h`
`#define RD_KAFKA_NODENAME_SIZE  128`
This is trimming the host name and resulting the error.

How to reproduce
================
This error will come if broker host is big enough. Also this error still come if we try to give IPs as IPs will get resolved to the hostname. After changing NODENAME_SIZE. It worked. 
`#define RD_KAFKA_NODENAME_SIZE  256`

Let me know if this fix can be incorporated in any of the release. Thanks

**IMPORTANT**: Always try to reproduce the issue on the latest released version (see https://github.com/edenhill/librdkafka/releases), if it can't be reproduced on the latest version the issue has been fixed.

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version 2.5.1
 - [x] Apache Kafka version:  0.10.2.0
 - [x] librdkafka client configuration: `var producer = new Kafka.Producer({
    // 'debug' : 'all',
    'client.id':'noderd-kafka',
    'bootstrap.servers': 'lognlognhostname:9092',
    'broker.version.fallback':'0.10.2.0',
    'heartbeat.interval.ms':3000,
    'compression.codec': 'none',
    'request.required.acks':1,
    'api.version.request':false,
    'retry.backoff.ms': 200,
    'message.send.max.retries': 10,
    'socket.keepalive.enable': true,
    'queue.buffering.max.messages': 10000,
    'queue.buffering.max.ms': 1000,
    'batch.num.messages': 10000,
    'dr_cb': true
},{
    'message.timeout.ms':300000,
    'produce.offset.report':true,
    'acks':1,
    'offset.store.sync.interval.ms':10
});`
 - [x] Operating system: `Mac Os Mojave`
 - [ ] Provide logs ``

edenhill:
You're absolutely right, the max nodename/hostname size should be 256.
Do you care to submit a PR?

Vijendra07Kulhade:
@edenhill Although its very small change, But I can submit the PR.

[...]

PR #2178, opened 2019-01-08
Error: Local: Host resolution failure #2171

Vijendra07Kulhade:
@edenhill Please review 
1. Updated the  NODE Name size to 256


edenhill:
Thank you!

Issue #2163, opened 2018-12-20
Idempotence: Out of order error on timeout


edenhill:
Originally reported in https://github.com/confluentinc/confluent-kafka-dotnet/issues/704

PR #2188, opened 2019-01-16
Proper handling of message timeouts with the idempotent producer

edenhill:
Issue #2163 

.. and a recent regression fix for consumer hanging when group coordinator broker goes down and comes up, thank you soak tests.


INTRODUCTION.md
```diff
@@ -485,6 +486,45 @@ a topic's partition count is known, which would insert these messages
...
+broker's point of view.
+The message status for timed out queued messages will be
+`RD_KAFKA_MSG_STATUS_NOT_PERSISTED`.
+
+If messages time out (also due to `message.timeout.ms`) while in-flight
+to the broker the protocol request will fail, the broker connection will
+be closed by the client, and the timed out messages will be removed
+from the producer queue. In this case the in-flight messages may be
+written to the topic log by the broker, even though a delivery report with
+error `ERR__MSG_TIMED_OUT` will be raised.
```
ijuma: It's worth explaining why this is OK.

INTRODUCTION.md
```diff
@@ -426,7 +426,8 @@ There are three types of guarantees that the idempotent producer can satisfy:
               This guarantee is enabled by default, but may be disabled
               by setting `enable.gapless.guarantee` if individual message
               failure is not a concern.
-
+              Message timeouts (`message.timeout.ms`) will cause gaps which
+              are permitted by the gap-less guarantee.
```
mhowlett: I find this section confusing to grok, and suggest framing it 
a bit differently as below (you may want to improve on my 
choice of words, which is not perfect..).

```
The idempotent producer provides two guarantees:

 * Exactly-once - a message is only written to the log once.
 * Ordering - a series of messages are written to the log in the
              order they were produced.

Additionally, the behavior in the event of message timeouts 
(`message.timeout.ms`) is configurable via the 
`enable.gapless.guarantee` configuration parameter. By 
default, this is set to `true`, which guarantees that a 
series of messages are written once and in order
without risk of skipping messages. The sequence of
message may be cut short and fail before all messages
are written, but may not fail individual messages in the
series. It's important to note that with the gapless 
guarantee enabled, a message delivery failure will result
in a fatal producer error. If individual message failure is
not a concern, `enable.gapless.guarantee` can be set
to `false`.

Note: The idempotent producer feature only applies to
the producer - it does NOT cover the exactly once 
consume case.
```

INTRODUCTION.md
```diff
@@ -485,6 +486,45 @@ a topic's partition count is known, which would insert these messages
...
 
+#### Message timeout considerations
+
+If messages time out (due to `message.timeout.ms`) while in the producer queue
+there will be gaps the in the series of produced messages:
+
+E.g., Messages 1,2,3,4,5 are produced by the application.
+      While messages 2,3,4 are transmitted to the broker the connection to
+      the broker goes down.
+      While the broking is down the message timeout expires for message 2 and 3.
```
mhowlett: broker

INTRODUCTION.md
```diff
@@ -485,6 +486,45 @@ a topic's partition count is known, which would insert these messages
...
+for a given partition when one or more of its queued (not transmitted)
+messages are timed out. When all requests are drained the Epoch is bumped and
+the base sequence number is reset to the first message in the queue, effectively
+skipping the timed out messages as if they had never existed from the
+broker's point of view.
+The message status for timed out queued messages will be
+`RD_KAFKA_MSG_STATUS_NOT_PERSISTED`.
+
+If messages time out (also due to `message.timeout.ms`) while in-flight
+to the broker the protocol request will fail, the broker connection will
```
mhowlett: comma after broker

INTRODUCTION.md
```diff
@@ -485,6 +486,45 @@ a topic's partition count is known, which would insert these messages
 after the partition-explicit messages regardless of produce order.
 
 
+#### Message timeout considerations
```
mhowlett: `enable.gapless.guarantee` isn't mentioned at all in this section, i think it would be good if it was.

INTRODUCTION.md
ijuma: I don't understand the reason for this config?

edenhill: @ijuma There are certain error cases (any non-idempotent permanent error) which may cause messages to permanently fail producing, which will lead to "gaps" in the message series.
E.g.: messages 1,2,3,4,5 are produced by the application, 1,2 are produced, 3,4 fail due to invalid message (size, or whatever), 5 is produced, leading to 1,2,5 being persisted in the log.
With the gap-less guarantee we provide the application with an additional guarantee (over exactly once and ordering) to stop producing if it would result in gaps in the message series, and instead raise a fatal error.

Message timeouts are not covered by the gap-less guarantee.

src/rdkafka_topic.c
```diff
@@ -1166,30 +1166,24 @@ static const char *rd_kafka_toppar_needs_query (rd_kafka_t *rk,
 
 /**
  * @brief Scan all topics and partitions for:
- *  - timed out messages.
```
rnpridgeon: :+1:

src/rdkafka_broker.c
```diff
@@ -3051,20 +3059,49 @@ static void rd_kafka_broker_internal_serve (rd_kafka_broker_t *rkb,
...
- * @brief Scan toppar's xmit queue for message timeouts.
- * @locality broker thread
- * @locks none
+ * @brief Scan toppar's xmit and producer queue for message timeouts and
+ *        enqueue delivery reports for timed out messages.
+ *
+ * @returns the number of messages timed out.
+ *
+ * @locality toppar's broker handler thread
+ * @locks toppar_lock MUST be held
```
rnpridgeon: I'm a bit confused by this annotation. It's unclear to me if this intended to inform the developer to acquire this lock prior to calling this function or this function acquires the lock. I ask because the former appears to be the case for this function however `rd_kafka_toppar_producer_serve` has the same annotation yet it locks the toppar shortly after entering the function. Sure you wouldn't want to acquire the lock prior to calling `rd_kafka_toppar_producer_serve` otherwise you'd end up in a deadlock. 

annotation
https://github.com/edenhill/librdkafka/pull/2188/files#diff-214080cd229adac5b1596529b6eecfe4R3129

grab lock:
https://github.com/edenhill/librdkafka/pull/2188/files#diff-214080cd229adac5b1596529b6eecfe4R3154

There is a subtle difference in the annotations, perhaps that's intentional.

INTRODUCTION.md
ijuma: Interesting. That could be useful. The naming is a bit unfortunate though because of the timeouts edge case.

edenhill: I believe it is useful until we add transactions, which is a better and more resilient way to solve this.
We could leave it out, or mark it as experimental, and make the default gapless.guarantee=false (which is the Java producer behaviour), until we know what to do with this.

edenhill: @mhowlett 

Thanks for the improved verbiage.

This bit is not true though, since message timeouts are okay even with gapless.guarantee enabled:

> Additionally, the behavior in the event of message timeouts 
> (`message.timeout.ms`) is configurable via the 
> `enable.gapless.guarantee`

ijuma: If this config hasn't been in any release and we are not sure about it, I'd vote for not including it in 1.0.

src/rdkafka_broker.c
```diff
@@ -3051,20 +3059,49 @@ static void rd_kafka_broker_internal_serve (rd_kafka_broker_t *rkb,
...
+                                             rd_kafka_toppar_t *rktp,
+                                             rd_ts_t now) {
+        rd_kafka_msgq_t xtimedout = RD_KAFKA_MSGQ_INITIALIZER(xtimedout);
+        rd_kafka_msgq_t qtimedout = RD_KAFKA_MSGQ_INITIALIZER(qtimedout);
+        int xcnt, qcnt, cnt;
+        uint64_t first, last;
+
+        xcnt = rd_kafka_msgq_age_scan(rktp, &rktp->rktp_xmit_msgq,
+                                      &xtimedout, now);
+        qcnt = rd_kafka_msgq_age_scan(rktp, &rktp->rktp_msgq,
```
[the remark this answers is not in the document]
edenhill: The message queues are sorted and msgq_age_scan() will only scan the queue until the first message with an expire team in the future is encountered (https://github.com/edenhill/librdkafka/pull/2188/files#diff-7f64c9289f107afb7d5ab47e90485926R611).

This is thus pretty much a no-op if no messages are to be timed out (typical case).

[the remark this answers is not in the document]
edenhill: And for the record: `@locks ` tells the caller what locks it must or must not hold when calling this function. It does not say what locks the function itself acquires.

INTRODUCTION.md
edenhill: For reference, this is the configuration property doc string:
```
        { _RK_GLOBAL|_RK_PRODUCER|_RK_HIGH, "enable.gapless.guarantee",
          _RK_C_BOOL,
          _RK(eos.gapless),
          "When set to `true`, any error that could result in a gap "
          "in the produced message series when a batch of messages fails, "
          "will raise a fatal error (ERR__GAPLESS_GUARANTEE) and stop "
          "the producer. "
          "Requires `enable.idempotence=true`.",
          0, 1, 1 },
```

edenhill:
Changes since last round of reviews:
 * Disabled gapless guarantee by default, and mark the property as experimental.
 * Fix idempo state hang when draining partitions and there is nothing to drain.
 * Fix race condition in produce_batch() which could cause messages to get stuck in the unassigned partition queue.
 * Log warning when experimental properties are used
 * Fix some minor issues (stats, output, tests, etc).

Issue #1704, opened 2018-02-23
RdKafka::Conf::get(propName, propVal) returns an empty value for topic configuration properties


ramakrishnavaradarajan:
Description
===========
RdKafka::Conf::get(const std::string &propName, std::string &propValue) method returns an empty property value for all topic configuration properties

How to reproduce
================
This issue can be reproduced in v0.11.3 : https://github.com/edenhill/librdkafka/releases/tag/v0.11.3

Here's a simple reproducer (below). This issue can be reproduced with any topic configuration property, but for illustration purposes, I chose "message.timeout.ms" below.

`RdKafka::Conf* topicConf = RdKafka::Conf::create(RdKafka::Conf::CONF_TOPIC);`
`std::string errStr("");`
`topicConf->set("message.timeout.ms", "10", errStr);`
`std::string propVal;`
`topicConf->get("message.timeout.ms", propVal);`
`/* propVal will be empty */`

Checklist
=========
 - [x] librdkafka version (release number or git tag): `v0.11.3`
 - [x] Apache Kafka version: `Kafka not required for bug reproduction`
 - [x] librdkafka client configuration: `default client configuration is sufficient for reproduction`
 - [x] Operating system: `No specific OS versions required for bug reproduction`

PR #2195, opened 2019-01-22
Fix issue #1704

masahiro-kondo-quick:
I challenged a small fix. Please confirm it.
Fix issue #1704 (RdKafka::Conf::get(propName, propVal) returns an empty value for topic configuration properties)


edenhill:
Thank you!

PR #2203, opened 2019-01-25
Minor improvements to documentation.

SkyToGround:
I clarified the documentation slightly as we had a bug caused by delete being called on a `TopicMetadataVector*`.


edenhill:
Thank you!

PR #2206, opened 2019-01-29
Pass member_id to rkas_get_metadata_cb()

ignitenet-martynas:
I am currently working on adding support for the sticky assignment strategy to librdkafka in my fork, and that implies being able to persist the current assignment of partitions to a consumer as userData. I'm using the `rkas_on_assignment_cb()` callback to store the assignment for a particular group member and then look it up and serialize and return it from `rkas_get_metadata_cb()`.

However, to be able to look up the current assignment for a particular consumer I need to identify it and currently no argument is passed to `rkas_get_metadata_cb()` that can be used to identify the consumer that the metadata is being assembled for.

Therefore I added `member_id` as the last argument of `rkas_get_metadata_cb()` and changed `rd_kafka_JoinGroupRequest()` to pass it.

This should not be introducing any binary or source code incompatibilities because `rkas_get_metadata_cb()` is only used internally within the library.

### DIFF
--- a/src/rdkafka_assignor.c
+++ b/src/rdkafka_assignor.c
@@ -138,7 +138,8 @@ rd_kafka_consumer_protocol_member_metadata_new (
 
 rd_kafkap_bytes_t *
 rd_kafka_assignor_get_metadata (rd_kafka_assignor_t *rkas,
-				const rd_list_t *topics) {
+				const rd_list_t *topics,
+				const char *member_id RD_UNUSED) {
         return rd_kafka_consumer_protocol_member_metadata_new(
                 topics, rkas->rkas_userdata,
                 rkas->rkas_userdata_size);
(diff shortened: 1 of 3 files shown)


edenhill:
Can you elaborate on how you are using the member id in your custom assignor?

ignitenet-martynas:
I am generally trying to copy the Java implementation of sticky assignor at https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java

Basically:

1. In `rd_kafka_sticky_assignor_on_assignment_cb()` I store a copy of a member's current assignment into a hash table based on member_id
2. In `rd_kafka_sticky_assignor_get_metadata_cb()` I look the stored copy up and then serialize it into the userData part of the metadata.

However, I do need the `member_id` passed into `rd_kafka_sticky_assignor_get_metadata_cb()` in order to do the lookup in step 2.

edenhill:
Would you be up for implementing and contributing the Sticky Assignor in librdkafka?

edenhill:
With the sticky assignor (et.al) now merged to master, would you like to revisit this?

PR #2209, opened 2019-01-31
Handle additional retvals from cnd_timedwait_abs()

rogerclermont:
Handle additional retvals from `cnd_timedwait_abs() ` to avoid infinite loop.


rogerclermont:
Apologies; I had intended to first create this on the fork.

PR #2211, opened 2019-01-31
Handle additional retvals from cnd_timedwait_abs()

rogerclermont:
Handle additional retvals from cnd_timedwait_abs() to avoid infinite loop.


edenhill:
Thank you!

PR #2216, opened 2019-02-05
Batch produce requests to multiple topics and partitions.

LavaSpider:
I know this is a large diff, the intent is to reduce the overall number of produce requests since it seems that there is a high overhead per request on the broker.

This code allows for producing to multiple topics and partitions within a single produce request. It does not currently address producing multiple message sets for the same toppar within the same request.

The batch producing has been thoroughly vetted, however it was originally written on top of rdkafka 0.11.4. I have rebased to pick up the sparse connection and idempotent producer changes, and re-run the latency measurements as shown below.

I added the config var "produce.request.max.partitions" to limit the number of toppars contained within one request. Setting this to 1 will mimic existing behavior.

I am attaching latency measurements from before and after the change producing 1,000,000 messages to a topic which has 200 partitions, and one which has 1,000 partitions. The message rate is 20,000 messages per second. The only change is switching out the code used to produce, all config values remain the same.

On the charts, the X axis is message number and the Y axis is time in ms.
For our test, we are running against a 4 node kafka cluster. The producer and consumer are on the same machine, so there should be no deviation of timestamps.

The charts are divided into seven sub-charts, which show the following:
1.  Time between call to rd_kafka_produce until rdkafka message is created.
2. Time between 1 and message set being created.
3. Time between 2 and produce request being sent by producer in rd_kafka_send.
4. Time between 3 to fetch request received by consumer in rd_kafka_recv.
5. Time between 4 to call to rd_kafka_consume0.
6. Time between 5 and message handled by app code.
7. Total time between rd_kafka_produce to message consumed by consumer app.

**Latency charts: 1 topic 200 partitions**
**BASELINE**
consumer_baseline_200
**BATCH**
consumer_batch_200

**Latency charts: 1 topic 1000 partitions**
**BASELINE**
consumer_baseline_1000
**BATCH**
consumer_batch_1000

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -235,6 +235,8 @@ typedef struct rd_kafka_topic_conf_s rd_kafka_topic_conf_t;
 typedef struct rd_kafka_queue_s rd_kafka_queue_t;
 typedef struct rd_kafka_op_s rd_kafka_event_t;
 typedef struct rd_kafka_topic_result_s rd_kafka_topic_result_t;
+typedef struct rd_kafka_produce_ctx_s rd_kafka_produce_ctx_t;
+typedef struct rd_kafka_produce_calculator_s rd_kafka_produce_calculator_t;
 /* @endcond */
 
 
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -94,7 +94,78 @@ const char *rd_kafka_secproto_names[] = {
 	NULL
 };
 
+typedef struct produce_batch_s {
+        int batch_initialized;
+        TAILQ_HEAD(,rd_kafka_toppar_s) batch_toppars;
+        rd_kafka_produce_calculator_t batch_calc;
+} produce_batch_t;
 
+
+/* These rd_kafka_broker_produce_batch_* functions are helpers
+ * for delaying the construction of a message batch until the
+ * counts of messages, partitions, and topics are known.
+ */
+static void
+rd_kafka_broker_produce_batch_init(produce_batch_t *batch,
+                                   rd_kafka_broker_t *rkb) {
+        TAILQ_INIT(&batch->batch_toppars);
+        rd_kafka_produce_calculator_init(&batch->batch_calc, rkb);
+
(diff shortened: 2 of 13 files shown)


write2jaydeep:
Hi 
is this PR merged with the master branch? how does it effective to use in production?

LavaSpider:
No it has not yet been merged yet. It looks like I need to rebase it off of master.

This change mainly affects users who have a large number of topics/partitions they are producing to. In that case, this will significantly reduce the number of produce requests sent to a broker. There seems to be a per-request overhead on the kafka broker, so limiting the overall number of the requests improves individual message latency.

cjf2xn:
Has there been any movement on this or a similar PR? My use case would benefit from this change.

LavaSpider:
Sorry, this PR has been in limbo for a while. It was originally made
against an older version of rdkafka.

Since then support for idempotent produce requests was added. That addition
modified code in similar areas as this change and I had not had time to
rebase. The rebase is happening currently as we need the idempotent
producer behavior, so I should be able to update this PR soon.

kphelps:
Is there any plan to get this change across the finish line? If not, I could try to take a shot at rebasing this in the near future. I've been using rdkafka heavily on the consumer end, but this issue has been a blocker for migrating some high partition count use cases to use rdkafka's producer.

wmorgan6796:
echoing what @kphelps said in saying I would also be interested in getting this across the finish line and helping where need be. This relates to my discussion here: https://github.com/confluentinc/librdkafka/discussions/3998 

We currently have a set up with high partition count ( >2000 Partitions) that we're producing to and this would greatly help us cut down on resource contention from excessive threading since to get around this we need to create multiple producers to keep up with the message load (40+ Million msg/s which each node doing 500k+  msg/s)

dopuskh3:
@emasab @milindl  @anchitj what is blocking from rebasing this change on top of master and getting it through the finish line. We are very interested in that feature and can help if needed. Just trying to gauge how hard it would be to rebase this change and if there are particular design choices that were preventing this PR from being merged in the first place.

PR #2218, opened 2019-02-06
Fix recent regression: messages in UA partition does not time out.

edenhill:


src/rdkafka_topic.c
```diff
@@ -1226,8 +1229,11 @@ void rd_kafka_topic_scan_all (rd_kafka_t *rk, rd_ts_t now) {
 		for (p = RD_KAFKA_PARTITION_UA ;
 		     p < rkt->rkt_partition_cnt ; p++) {
 
-			if (!(s_rktp = rd_kafka_toppar_get(rkt, p, 0)))
-				continue;
+                        if (!(s_rktp = rd_kafka_toppar_get(
+                                      rkt, p,
+                                      p == RD_KAFKA_PARTITION_UA ?
```
rnpridgeon: Much nicer than 0 none 0

PR #2221, opened 2019-02-12
CMake: Add check for C11 thread support.

Oxymoron79:
This PR adds a check in the CMake files to detect C11 thread support on the build host to enable the conditional compilation of tinycthread.
Glibc version 2.28 introduced support for C11 threads, which leads to similar problems as reported in https://github.com/edenhill/librdkafka/issues/1998 on system that already use that Glibc version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,7 @@ endif(WIN32)
 # * HAVE_ATOMICS_64_SYNC
 # * HAVE_REGEX
 # * HAVE_STRNDUP
+# * WITH_C11THREADS
 # * LINK_ATOMIC
 include("packaging/cmake/try_compile/rdkafka_setup.cmake")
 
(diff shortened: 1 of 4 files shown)


edenhill:
LGTM.
Is this good to go, @Oxymoron79 ?

Oxymoron79:
@edenhill Yes, it's good to go. The fix works on the systems that I tested.

edenhill:
Thank you!

edenhill:
@Oxymoron79 This is unrelated to this issue, but the CMake builds are breaking because of missing BUILT_WITH define:
this was recently added to mklove as a string containing the WITH_x build options used for the build (without the WITH_ prefix).

Do you have any idea how to do this with CMake and would you like to take a stab at it?
If it turns out to be complicated we could just set it to "CMAKE", it is only used in an initial debug message to show the build options.

Oxymoron79:
@edenhill Sure, I will try to find a solution in the next days. From a first quick look at the buildlog, I think it won't be hard to fix.

PR #2230, opened 2019-02-19
Hide sparse connections, document debug contexts, misc warnings

edenhill:
We've had zero issues with sparse connections during the last 5 months, so we should remove the dual stack mode, but that's too big of a change this close to release.
To avoid breaking changes post v1.0.0 we hide the configuration property so it can be removed later.


INTRODUCTION.md
```diff
@@ -915,6 +910,45 @@ the logging level will be LOG_WARNING (4), else LOG_INFO (6).
...
+eos           | producer | Idempotent Producer debugging.
+all           | *        | All of the above.
+
+
+Suggested debugging settings for troubleshooting:
+
+Problem space          | Type     | Debug setting
+-----------------------|----------|-------------------
+Producer not delivering messages to broker | producer | `broker,topic,msg`
+Consumer not fetching messages | consumer | Start with `consumer`, or use `cgrp,fetch` for detailed information.
```
mhowlett: do you mean / is it better to say: 'start with consumer, add cgrp,fetch for detailed information'

INTRODUCTION.md
```diff
@@ -915,6 +910,45 @@ the logging level will be LOG_WARNING (4), else LOG_INFO (6).
...
+Consumer starts reading at unexpected offset | consumer | `consumer` or `cgrp,fetch`
```
mhowlett: ditto

INTRODUCTION.md
edenhill: `consumer` is a high-level consumer debugging, while `cgrp,fetch` is much more detailed.
They are typically not to be used together.

PR #2240, opened 2019-02-27
Use __sun for detecting Solaris

adeason:
Without this, I need to run ./configure with CPPFLAGS='-Dsun' for compilation to succeed on Solaris.

### DIFF
--- a/src/rdendian.h
+++ b/src/rdendian.h
@@ -66,7 +66,7 @@
  #include <endian.h>
 #elif defined __BSD__
   #include <sys/endian.h>
-#elif defined sun
+#elif defined __sun
   #include <sys/byteorder.h>
   #include <sys/isa_defs.h>
 #define __LITTLE_ENDIAN 1234
@@ -89,7 +89,7 @@
 #define le64toh(x) (x)
 #define htole16(x) (x)
 #define htole64(x) (x)
-#endif /* sun */
+#endif /* __sun */
 
 #elif defined __APPLE__
   #include <machine/endian.h>
--- a/src/rdkafka_transport.c
+++ b/src/rdkafka_transport.c
@@ -144,7 +144,7 @@ ssize_t rd_kafka_transport_socket_sendmsg (rd_kafka_transport_t *rktrans,
                          rktrans->rktrans_sndbuf_size);
         msg.msg_iovlen = (int)iovlen;
 
-#ifdef sun
+#ifdef __sun
         /* See recvmsg() comment. Setting it here to be safe. */
         socket_errno = EAGAIN;
 #endif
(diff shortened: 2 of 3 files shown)


adeason:
The appveyor failure looks unrelated as far as I can tell; it's failing on a prereq download for building on windows, I think?

edenhill:
I see other projects using both, `#if defined(sun) || defined(__sun)`, is there an authoritative source which to use?

adeason:
On Mon, 04 Mar 2019 02:07:29 -0800
Magnus Edenhill <notifications@github.com> wrote:

> I see other projects using both, `#if defined(sun) || defined(__sun)`,
> is there an authoritative source which to use?

I don't think so. Some hints, though:

In the man page for the Solaris native compiler ("Oracle Solaris
Studio"):
<https://docs.oracle.com/cd/E37069_01/html/E54439/uc-cc-1.html>, search
for "-Dname[=def]" to find the list of predefined symbols. 'sun' and
'__sun' are both in there, but near the end it mentions:

> If +p is used, sun, unix, sparc and i386 are not defined.

The +p option says it turns off "non-standard preprocessor asserts". So
from that, 'sun' is considered non-standard.

But I was using gcc (which I assumed was required for librdkafka). I
don't see any mention of these symbols specifically for gcc, but this
page has something relevant to say:
<https://gcc.gnu.org/onlinedocs/cpp/System-specific-Predefined-Macros.html>

> The C standard requires that all system-specific macros be part of the
> reserved namespace. All names which begin with two underscores, or an
> underscore and a capital letter, are reserved for the compiler and
> library to use as they wish. However, historically system-specific
> macros have had names with no special prefix; for instance, it is
> common to find unix defined on Unix systems. For all such macros, GCC
> provides a parallel macro with two underscores added at the beginning
> and the end. If unix is defined, __unix__ will be defined too. There
> will never be more than two underscores; the parallel of _mips is
> __mips__.
[...]
> We are slowly phasing out all predefined macros which are outside the
> reserved namespace. You should never use them in new programs, and we
> encourage you to correct older code to use the parallel macros
> whenever you find it.

So that page indicates 'sun' is nonstandard since it doesn't begin with
any underscores, and you should avoid nonstandard platform macros. But
it also says:

> We don’t recommend you use the system-specific macros that are in the
> reserved namespace, either. It is better in the long run to check
> specifically for features you need, using a tool such as autoconf.

My guess is that 'sun' is the older symbol, and at some point Sun
switched to using '__sun' to try to be more standard and avoid naming
conflicts. But "older" here is quite old, like pre-Solaris SunOS era,
which would have already been obsolete in the 90s. So checking for 'sun'
feels like one of those ridiculous checks for compatibility with
30-year-old systems that are unlikely to ever encounter a modern release
of librdkafka; but I'm not sure, since it's hard to find detailed info.
And maybe 'sun' will go away at some point; right now I only see it
output by the Solaris compiler, and not by gcc.

I agree checking for both '__sun' and 'sun' is common, though. If you
prefer to check for both, I don't mind resubmitting doing that instead.

-- 
Andrew Deason
adeason@dson.org

edenhill:
Fantastic research, thank you!
You have me convinced that we should be fine with just `__sun`.

edenhill:
Thank you for this!

PR #2243, opened 2019-03-06
Fix tinycthread const_seg/data_seg pragma mismatch

JonECG:
Fixes a pragma leak in tincythread on windows that could poison a dll

### DIFF
--- a/src/tinycthread.c
+++ b/src/tinycthread.c
@@ -537,9 +537,9 @@ static void NTAPI _tinycthread_tss_callback(PVOID h, DWORD dwReason, PVOID pv)
   #endif
   PIMAGE_TLS_CALLBACK p_thread_callback = _tinycthread_tss_callback;
   #ifdef _M_X64
-    #pragma data_seg()
-  #else
     #pragma const_seg()
+  #else
+    #pragma data_seg()
   #endif
 #else
   PIMAGE_TLS_CALLBACK p_thread_callback __attribute__((section(".CRT$XLB"))) = _tinycthread_tss_callback;


src/tinycthread.c
```diff
@@ -537,9 +537,9 @@ static void NTAPI _tinycthread_tss_callback(PVOID h, DWORD dwReason, PVOID pv)
   #endif
   PIMAGE_TLS_CALLBACK p_thread_callback = _tinycthread_tss_callback;
   #ifdef _M_X64
-    #pragma data_seg()
-  #else
     #pragma const_seg()
+  #else
+    #pragma data_seg()
```
edenhill: I'm not sure I fully understand what is going on here, care to elaborate?

JonECG: Sure thing, the diff reads a bit weird, but before the change, the code is as follows:

```
  #ifdef _M_X64
    #pragma const_seg(".CRT$XLB")
  #else
    #pragma data_seg(".CRT$XLB")
  #endif
  PIMAGE_TLS_CALLBACK p_thread_callback = _tinycthread_tss_callback;
  #ifdef _M_X64
    #pragma data_seg()
  #else
    #pragma const_seg()
  #endif
```

Afterwards, it would be:

```
  #ifdef _M_X64
    #pragma const_seg(".CRT$XLB")
  #else
    #pragma data_seg(".CRT$XLB")
  #endif
  PIMAGE_TLS_CALLBACK p_thread_callback = _tinycthread_tss_callback;
  #ifdef _M_X64
    #pragma const_seg()
  #else
    #pragma data_seg()
  #endif
```

Using `#pragma XXXX_seg( ... )` on windows will explicitly set which segment/section a symbol will be written to in the object file. Calling it parameter-less will reset it back to its default value.
See: https://docs.microsoft.com/en-us/cpp/preprocessor/const-seg?view=vs-2017

As for the significance of ".CRT$XLB", it's fairly undocumented, but it looks like that's a section reserved for entrypoint functions, specifically for thread local storage, on windows. See: https://stackoverflow.com/questions/14538159/about-tls-callback-in-windows

However, the usage of ".CRT$XLB" only caused the side effect in my case due to another library being statically linked that was also trying to do some TLS shenanigans. The real issue is that the const/data segment is not reset on windows after declaring the callback functor. This causes any static memory declared after this point to also be shoved in that segment, this can include static cstrings from the assert macro.

It is apparent from the code that the intention was for the pragma that was changed before to be reset after declaring the callback. Otherwise, the resets that were there previously do nothing, since that const/data segment had not been overwritten.

Hope that clears some things up.

edenhill: Superb explanation, thank you!

PR #2244, opened 2019-03-07
Allow ./configure to auto-bootstrap dependencies

edenhill:
.. and a bunch of build/packaging fixes to produce reliable builds, optionally statically linked.

Daughter projects should be adjusted to use `./configure --install-deps --source-deps-only ..` rather than any `apt-get install libssl-dev..` in their bootstrap scripts, since the latter is error prone.


README.md
```diff
@@ -109,9 +109,42 @@ See the [wiki](https://github.com/edenhill/librdkafka/wiki) for a FAQ.
...
+On Windows, reference [librdkafka.redist](https://www.nuget.org/packages/librdkafka.redist/) NuGet package in your Visual Studio project.
+
+
+For other platforms, follow the source building instructions below.
+
+
 ### Building
 
       ./configure
+      # Or, to have configure automatically install dependencies:
```
ybyzek: "to have configure automatically install" reads strangely (at first I thought configure was a verb).  Perhaps:

```
# Or, to have `configure` automatically install dependencies:
```

or

```
# Or, to have it automatically install dependencies:
```

or

```
# Or, to automatically install dependencies:
```

README.md
```diff
@@ -109,9 +109,42 @@ See the [wiki](https://github.com/edenhill/librdkafka/wiki) for a FAQ.
...
+      # ./configure --install-deps
+      # Or, let configure automatically build dependencies from source:
```
ybyzek: Same comment as above

README.md
edenhill: 3rd option is good, thanks!

mklove/modules/configure.base
```diff
@@ -304,6 +308,301 @@ function mkl_allvar_set {
...
+            cmd="yum install -y $pkgs"
+            ;;
+
+        alpine-0)
+            method=apk
+            pkgs=$(mkl_meta_get $name apk)
+            cmd="apk add $pkgs"
+            ;;
+
+        osx-*)
```
rnpridgeon: nit: It seems like sort of a stretch to cal brew MacOsx's `native` package installer.

mklove/modules/configure.base
```diff
@@ -304,6 +308,301 @@ function mkl_allvar_set {
...
+
+# Returns 0 (yes) if this dependency has a source builder, else 1 (no)
+function mkl_dep_has_builder {
+    local name=$1
+    local func="${name}_install_source"
+    mkl_func_exists $func
+}
+
+
+# Returns 0 (yes) if this dependency has a source builder, else 1 (no)
```
rnpridgeon: has an installer*

mklove/modules/configure.base
```diff
@@ -342,13 +641,27 @@ $(mkl_var_get MKL_FAIL__${n}__reason)
...
+            centos|rhel|redhat|fedora)
+                local rpms=$(mkl_meta_get $conf "rpm")
+                pkg_cmd="sudo yum install -y"
+                if [[ ${#rpms} > 0 ]]; then
+                    install_pkgs="$install_pkgs $rpms"
+                fi
+                ;;
+            alpine)
+                local apks=$(mkl_meta_get $conf "apk")
+                pkg_cmd="apk install "
```
rnpridgeon: Is there a `-y` equivalent for apk ?

mklove/modules/configure.cc
```diff
@@ -49,7 +49,7 @@ function checks {
 	    mkl_mkvar_append CPPFLAGS CPPFLAGS "-m$MBITS"
 	    mkl_mkvar_append LDFLAGS LDFLAGS "-m$MBITS"
 	fi
-	if [[ -z "$ARFLAGS" && $MBITS == 64 && $MKL_DISTRO == "SunOS" ]]; then
+	if [[ -z "$ARFLAGS" && $MBITS == 64 && $MKL_DISTRO == "sunos" ]]; then
```
rnpridgeon: Use `TR` to make case insensitive?

mklove/modules/configure.host
```diff
@@ -51,12 +62,23 @@ function checks {
         esac
     fi
 
-    if [[ -z $sys ]]; then
+    # Convert to lower case
+    sys=$(echo $sys | tr '[:upper:]' '[:lower:]')
```
rnpridgeon: 🥇

mklove/modules/configure.base
edenhill: I'm not sure I agree. Homebrew is the defacto package manager for developers on osx.
And it is native to the platform.

mklove/modules/configure.base
edenhill: No, it is non-interactive it seems.
More importantly, there is no "install", it is called "add"

PR #2248, opened 2019-03-12
Character Length for Broker

AkshayDubey29:
Character Length for Broker to support more than 128 Characters lengths


src/rdkafka_int.h
```diff
@@ -185,7 +185,7 @@ struct rd_kafka_s {
 
         rd_kafka_conf_t  rk_conf;
         rd_kafka_q_t    *rk_logq;          /* Log queue if `log.queue` set */
-        char             rk_name[128];
+        char             rk_name[256];
```
edenhill: this is the client instance name (for logging), not the broker name.

PR #2252, opened 2019-03-13
Idempotent producer example 

edenhill:
And sneaked in a release checksum script.

### DIFF
--- a/examples/rdkafka_idempotent_producer.c
+++ b/examples/rdkafka_idempotent_producer.c
@@ -1,0 +1,342 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2019, Magnus Edenhill
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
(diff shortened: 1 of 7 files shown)


src/rdkafka.h
```diff
@@ -673,6 +673,27 @@ rd_kafka_resp_err_t rd_kafka_fatal_error (rd_kafka_t *rk,
                                           char *errstr, size_t errstr_size);
 
 
+/**
+ * @brief Trigger a fatal error for testing purposes.
```
mhowlett: ideally I think this functionality should be tested via integration tests using a mock broker, or intercepting and modifying calls to a real broker - I don't like the idea of having methods in a public API for testing purposes. I admit though that is difficult to set up and that the testing in this PR brings value. In the event we start doing integration testing as described above though, I think we should get rid of this method. so could maybe mark as such, but just my opinion.

edenhill: As stated elsewhere, it is practically impossible to trigger idempotent producer fatal errors without a full mock broker.

I think this API serves an important purpose for our users to allow them to easily test their (error handling) code, we wouldn't want to require them setting up a eos-breaking mock broker just to verify their code logic.

examples/rdkafka_idempotent_producer.c
```diff
@@ -0,0 +1,342 @@
...
+ * This callback is triggered by rd_kafka_poll() or rd_kafka_flush()
+ * for client instance-level errors, such as broker connection failures,
+ * authentication issues, etc.
+ *
+ * These errors should generally be considered informational as
+ * the underlying client will automatically try to recover from
+ * any errors encountered, the application does not need to take
+ * action on them.
+ *
+ * But with idempotence truly fatal errors can be raised when
```
mhowlett: comma after idempotence [you can rely on my to pick up on the important stuff]

examples/rdkafka_idempotent_producer.c
```diff
@@ -0,0 +1,342 @@
...
+ * the idempotence guarantees can't be satisfied, these errors
```
mhowlett: new sentence after satisfied, not comma.

examples/rdkafka_idempotent_producer.c
```diff
@@ -0,0 +1,342 @@
...
+                                 * The internal queue represents both
+                                 * messages to be sent and messages that have
+                                 * been sent or failed, awaiting their
+                                 * delivery report callback to be called.
+                                 *
+                                 * The internal queue is limited by the
+                                 * configuration property
+                                 * queue.buffering.max.messages */
+                                rd_kafka_poll(rk, 1000/*block for max 1000ms*/);
+                                goto retry;
```
mhowlett: Interesting that you prefer goto. seems justifiable - one level less nesting, less loc, not spaghetti here.

examples/rdkafka_idempotent_producer.c
```diff
@@ -0,0 +1,342 @@
...
+                 * you register). */
+                rd_kafka_poll(rk, 0/*non-blocking*/);
+
+                msgcnt++;
+
+                /* Since fatal errors can't be triggered in practice,
+                 * use the test API to trigger a fabricated error after
+                 * some time. */
+                if (msgcnt == 13)
+                        rd_kafka_test_fatal_error(
```
mhowlett: I don't think an example should contain a test method.

examples/rdkafka_idempotent_producer.c
edenhill: Yep, those are the reasons. :+1:

examples/rdkafka_idempotent_producer.c
edenhill: I believe it is important to show our users how to test their code.

src/rdkafka.h
mhowlett: very good point that users would find this useful for testing. ok, sounds good to me.

examples/rdkafka_idempotent_producer.c
edenhill: It is also very unlikely that people will leave this as-is in the code as they move the example over to their real code.

Issue #2259, opened 2019-03-19
ERR__ALL_BROKERS_DOWN not generated for high-level consumer


edenhill:
If the high-level consumer has all its bootstrap brokers down, there will be no ERR__ALL_BROKERS_DOWN event emitted due to the logical GroupCoordinator broker never transitioning from INIT to DOWN state.

 - [x] librdkafka version (release number or git tag): master / v1.0.0-RC8

PR #2260, opened 2019-03-19
Selectively include logical brokers in ERR__ALL_BROKERS_DOWN calculation (#2259)

edenhill:
..depending on if they have an address set or not.


edenhill:
You're absolutely right. Added a test.

PR #2262, opened 2019-03-21
CMake: Generate BUILT_WITH define

Oxymoron79:
As mentioned in #2221 I have been working on generating the BUILT_WITH define directive in CMake.

I was able to replicate the generation of the BUILT_WITH list for most entries except GNULD and LDS, since I haven't found a proper way to do the necessary checks in CMake.

Also, the entries for the C and C++ compilers (GCC, GXX) are created from the CMake compiler ID strings for GNU, Clang and MSVC only and are not exported to config.h.

Are these restrictions acceptable for now?

Along the way, I also added the missing checks to enable the hdr histogram compilation and for the support of the CRC32C instruction.


CMakeLists.txt
```diff
@@ -18,6 +18,49 @@ option(ENABLE_DEVEL "Enable development asserts, checks, etc" OFF)
 option(ENABLE_REFCNT_DEBUG "Enable refcnt debugging" OFF)
 option(ENABLE_SHAREDPTR_DEBUG "Enable sharedptr debugging" OFF)
 set(TRYCOMPILE_SRC_DIR "${CMAKE_CURRENT_LIST_DIR}/packaging/cmake/try_compile")
+set(BUILT_WITH "")
```
edenhill: Let's include "CMAKE" here to differentiate from the mklove builds.

CMakeLists.txt
```diff
@@ -18,6 +18,49 @@ option(ENABLE_DEVEL "Enable development asserts, checks, etc" OFF)
...
+  list(APPEND BUILT_WITH "GXX")
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  list(APPEND BUILT_WITH "CLANGXX")
+elseif(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
+  list(APPEND BUILT_WITH "MSVCXX")
+endif()
+# }
+
+# GNU install {
+find_program(GnuInstall "install")
```
edenhill: Not sure if this is needed with cmake, it already does the install thing, right?
And as for build option, INSTALL isn't relevant, but everything is included in the mklove string.

CMakeLists.txt
```diff
@@ -18,6 +18,49 @@ option(ENABLE_DEVEL "Enable development asserts, checks, etc" OFF)
...
+if(NOT GnuInstall-NOTFOUND)
+  set(WITH_INSTALL ON)
+  list(APPEND BUILT_WITH "INSTALL")
+endif()
+# }
+
+# PkgConfig {
+find_package(PkgConfig QUIET)
```
edenhill: dito

CMakeLists.txt
```diff
@@ -18,6 +18,49 @@ option(ENABLE_DEVEL "Enable development asserts, checks, etc" OFF)
...
+if(PkgConfig_FOUND)
+  set(WITH_PKGCONFIG ON)
+  list(APPEND BUILT_WITH "PKGCONFIG")
+endif()
+# }
+
+# LIBM {
+include(CheckLibraryExists)
+check_library_exists(m pow "" WITH_HDRHISTOGRAM)
```
edenhill: Stumbled upon this too, thanks 👍

CMakeLists.txt
```diff
@@ -41,6 +87,9 @@ else ()
   endif()
```
edenhill: nit above: we're not bundling zstd (dont need to fix it in this PR)

CMakeLists.txt
Oxymoron79: PkgConfig is used (if available on the host system) to configure SASL. So I would leave that in.

CMakeLists.txt
Oxymoron79: I think you're right that CMake is not using GNU install for the install procedure. I will remove it.

CMakeLists.txt
Oxymoron79: Agreed.
In that case, I would also like to put the unmodified compiler ID strings from CMake in the list, instead of parsing the ID strings and generate different ID strings. I think that would be cleaner and easier to maintain...

CMakeLists.txt
Oxymoron79: If I don't fix it now, I will eventually forget about it :grinning:

CMakeLists.txt
edenhill: What does the compiler string look like?

Oxymoron79: They are listed here: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html
E.g. for the GNU compiler collection it's GNU for both compilers C and C++.

CMakeLists.txt
edenhill: Go for it!

CMakeLists.txt
edenhill: Go for it!

Oxymoron79: It's already in https://github.com/edenhill/librdkafka/pull/2262/commits/b773fa76629c0ac6142a854720904a5a1be8556d ....

CMakeLists.txt
Oxymoron79: Fixed in https://github.com/edenhill/librdkafka/pull/2262/commits/1d0fadb1a9e9f42fc4b0674decb86d9b9704252e.

CMakeLists.txt
Oxymoron79: Removed in https://github.com/edenhill/librdkafka/commit/b773fa76629c0ac6142a854720904a5a1be8556d.

Oxymoron79:
> Can you rebase on latest master (has a conflicting BUILT_WITH commit)?
Rebased and squashed review commits on master.

edenhill:
Thank you!

PR #2287, opened 2019-04-18
Update error response codes

benesch:
Update to include error codes from v2.2 of Kafka.

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -428,8 +428,22 @@ enum ErrorCode {
         ERR_LISTENER_NOT_FOUND = 72,
         /** Topic deletion is disabled */
         ERR_TOPIC_DELETION_DISABLED = 73,
+        /** Leader epoch is older than broker epoch */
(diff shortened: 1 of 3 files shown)


benesch:
Good point! Done.

src-cpp/rdkafkacpp.h
```diff
@@ -428,8 +428,22 @@ enum ErrorCode {
         ERR_LISTENER_NOT_FOUND = 72,
         /** Topic deletion is disabled */
         ERR_TOPIC_DELETION_DISABLED = 73,
+        /** The leader epoch is older than the broker epoch */
```
edenhill: To be consistent with existing lingua, please remove the "The " prefixes of all these descriptions, and any the's embedded in the text too, so:
"Leader epoch is older than broker epoch"

benesch: Ok, done.

edenhill:
Thank you for this!

benesch:
Thanks for reviewing! Looking forward to the next release.

PR #2298, opened 2019-04-25
trust the headers

kermitbu:


edenhill:
I have no idea what this is for, can you please explain the reasoning behind this change?

kermitbu:
> I have no idea what this is for, can you please explain the reasoning behind this change?

I don't want to malloc headers every time I produce data, but my modification method is not suitable and I am improving the method

edenhill:
So just pass NULL as headers.

kermitbu:
But I want to pass the header for extra info, What should I do？

edenhill:
You either pass the headers that should be added to the message, or you dont.
If your headers are identical for each message you will still need to duplicate them for each time you call produce().

kermitbu:
Well, I have to make a version for business needs.
malloc for each production data request is a time-consuming activity.

edenhill:
What rate are you producing messages at?
In your measurements, what is the performance drop or CPU cost of creating a new Headers object for each produce() call?

kermitbu:
Message generation rates are around 1.4 million per second, and reducing malloc times is a point of optimization based on past experience

edenhill:
And your headers are identical for each message?

kermitbu:
Each message has a different header, which I want to use to log breakpoint information when an exception occurs. The breakpoint information is not the same as offset, and the breakpoint information is a two-dimensional data.
My business is to collect 1.4 million data per second from upstream, and then write these data into multiple partitions under multiple topics. When there is an exception, I need to get the maximum number of multiple partitions in Kafka. Is there any good way?

kermitbu:
I have considered recording breakpoint information into another topic separately before, but I think it is equivalent to sending data twice and involves transaction processing, so I don't want to consider such a scheme.

edenhill:
So if each message has unique headers they would need to be separately allocated anyway, since there is no point in reusing them.

edenhill:
You can use metadata() to retrieve the number of partitions for a topic.
But it is better to rely on the partitioner to distribute messages among the existing partitions.

kermitbu:
Each message has a unique header, but it takes time for each header to reapply for memory. My idea is that the memory of this header is reused and used again after sending. This is controlled by the user.

In addition, if I want to reuse the memory of value and control the application and release of memory by myself, and I don't want to make a copy of rdkafka, can I set the msgflags parameter of rd_kafka_produce to 0?

I learned that the metadata information included all partition information for each topic, but I could not find what I wanted

edenhill:
The msgflags only applies to the produced value, not key or headers.

kermitbu:
Thanks for your reply. I plan to use tcmalloc to replace the malloc of glibc

PR #2301, opened 2019-04-25
Added rd_kafka_conf() to retrieve the client's configuration object

edenhill:


mhowlett:
can this be used to get default values? (without knowing, that's what i'd assume the main use of this would be?) if so test for that would be good. regardless, LGTM.

edenhill:
To get default values you would use instantiate a fresh new config object with rd_kafka_conf_new().
This is for retrieving the values that are actually used by a client instance, including such values that have been automatically adjusted (e.g., for idempotence).

The need for this came up with the SASL OAUTHBEARER token refresh callback that may need to use other configuration parameters for its authentication.

PR #2305, opened 2019-04-29
configure: Add option to disable automagic dependency on zstd

Whissi:
This commit will add an option which will allow you to explicit disable
zstd usage.

### DIFF
--- a/mklove/modules/configure.libzstd
+++ b/mklove/modules/configure.libzstd
@@ -9,8 +9,12 @@
 #   mkl_check libzstd [<action>]
 #
 
+mkl_toggle_option "Feature" ENABLE_ZSTD "--enable-zstd" "Enable support for ZSTD compression" "y"
+
 function manual_checks {
-    local action=$1
+    local action=${1:-disable}
+
+    [[ $ENABLE_ZSTD == y ]] || return 0
 
     mkl_meta_set "libzstd" "brew" "zstd"
     mkl_meta_set "libzstd" "apk" "zstd-dev zstd-static"


mklove/modules/configure.zstd
```diff
@@ -9,8 +9,12 @@
 #   mkl_check zstd [<action>]
 #
 
+mkl_toggle_option "Feature" ENABLE_ZSTD "--enable-zstd" "Enable support for ZSTD compression" "y"
+
 function manual_checks {
-    local action=$1
+    local action=${1:-disable}
```
edenhill: Why is this needed?

mklove/modules/configure.libzstd
```diff
@@ -9,8 +9,12 @@
 #   mkl_check libzstd [<action>]
 #
 
+mkl_toggle_option "Feature" ENABLE_ZSTD "--enable-zstd" "Enable support for ZSTD compression" "y"
+
 function manual_checks {
-    local action=$1
+    local action=${1:-disable}
```
edenhill: why is a default value needed? it is always passed an action

benesch: Not the PR author, but I assume @Whissi is simply following the precedent set by libssl and libsasl2:

  * https://github.com/edenhill/librdkafka/blob/b3223ecea638bfacc6d9ec981aa4df1d22ee9542/mklove/modules/configure.libssl#L22
  * https://github.com/edenhill/librdkafka/blob/b3223ecea638bfacc6d9ec981aa4df1d22ee9542/mklove/modules/configure.libsasl2#L17

If this is indeed unnecessary, perhaps it should be removed from those other modules as well.

benesch:
Not the PR author, but I'm interested in seeing this merged! Please see my response above. I actually think it would make sense to merge this as is for consistency with the other mklove lib modules and then if you'd like I can submit a follow-on PR to remove the default expansion across all of the mklove lib modules.

mklove/modules/configure.libzstd
Whissi: Yes, that was my intention: Keep current style.

edenhill:
Thank you!

PR #2307, opened 2019-05-01
CMake: Take care of zstd static lib on Win32

myd7349:
When built as a static library on Win32, zstd has output names like zstd_static.lib (Release) and zstd_staticd.lib (Debug):
https://github.com/facebook/zstd/blob/69baaee3e42f90dedea2c946bc19bfeac4e782ee/build/cmake/lib/CMakeLists.txt#L141
and `find_library(ZSTD zstd)` will not work in such cases.

### DIFF
--- a/packaging/cmake/Modules/FindZstd.cmake
+++ b/packaging/cmake/Modules/FindZstd.cmake
@@ -1,4 +1,3 @@
-# https://github.com/facebook/folly/blob/master/CMake/FindZstd.cmake
 #
 # - Try to find Facebook zstd library
 # This will define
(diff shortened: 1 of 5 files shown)


packaging/cmake/Modules/FindZstd.cmake
```diff
@@ -0,0 +1,28 @@
+# https://github.com/facebook/folly/blob/master/CMake/FindZstd.cmake
+#
```
edenhill: Need to include the folly license in this directory as LICENSE.FindZstd

myd7349: Hi! @edenhill Thanks for your review! I have added a license file for `FindZstd.cmake`: https://github.com/myd7349/librdkafka/blob/win32-zstd-static-lib/LICENSE.FindZstd
BTW, I cut these lines off: https://github.com/facebook/folly/blob/master/LICENSE#L180-L200

edenhill:
thank you!

myd7349:
Hi! @edenhill Thanks!
BTW, the PR id in the commit message seems to be wrong.

PR #2309, opened 2019-05-02
Continuing on PR #1882: SSL cert verification and in-memory keys

edenhill:
This PR is a continuation on @noahdav's work on PR #1882.

Summary of changes compared to #1882:
 * The retrieve_cb() has been removed in favour of using standard configuration properties. The reasoning behind this ís simplification, reduced code size, cross-language maintenance (some high-level languages, such as Go, can't use callbacks in a good way), and memory ownership issues (having the callback return cert memory to librdkafka in an effective and allocator-agnostic way is tricky).
 * Refactored the code and made it more in line with the librdkafka style guide.
 * Values of security-sensitive configuration properties are now cleared from memory when the value is no longer used, or on rd_kafka_t destruction. This is used for ssl.certificate.pem, ssl.key.pem and ssl.key.password.
 * The ssl_cert_verify_cb was changed to provide more information to the callback, including the full X509_STORE_CTX.
 * Add integration tests.

To be done:
 * Rework the Windows SSL cert store example to its own file.


noahdav:
@edenhill Thanks for the work on this PR.  I really appreciate it.  Due to the size of the change I would like to test this a little bit before I review.  I should be able to test by end of week.  Hope that works for your schedule.

edenhill:
Re PEM format:
I too stumbled upon the PEM issue as I was diving into the example code.
What's the format of the certificates in the cert store? DER?
If so we could provide rd_kafka_conf_set_ssl_key() and set_certificate(), set_ca(), etc.

Another question:
OpenSSL does not make use of the Windows cert store for finding CA/root certs.
But from what I understand this should be fairly simple to add to librdkafka, populating the OpenSSL cert store with all the Root certificates from the windows store.
Do you think this is a good idea? And if so, should we do it by default if no ssl.ca.xxxx is specified? (as it works on Linux/osx which sets up the default ca path if no ca is specified).

src/rdkafka_conf.h
```diff
@@ -203,23 +205,34 @@ struct rd_kafka_conf_s {
...
-        int (*ssl_cert_verify_cb) (char *cert, size_t len, char *errstr, size_t errstr_size, void *opaque);
-        ssize_t (*ssl_cert_retrieve_cb) (rd_kafka_certificate_type_t type, char **buffer, char *errstr, size_t errstr_size, void *opaque);
-	} ssl;
+                char *key_location;
+                char *key_pem;
+                char *key_password;
+                char *cert_location;
+                char *cert_pem;
+                char *ca_location;
+                char *ca_pem;
```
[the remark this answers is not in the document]
edenhill: Good catch!
I was thinking of adding them, but opted out and forgot to remove these.
This is sort of related to my question on Root CAs.

noahdav:
@edenhill thanks for your work on this.

Good point with the root store.  I think it would make sense to do this by default when no ssl.ca is specified.  In our case this seems to work without this but I think it would be a good idea to do.

When I did the original PR on this I was using the PFX format as that is what the windows cert store exports.  This has the public and private key in the buffer.

To convert to a PFX file from the public and private keys I do the following where client.key is the private key and client.pem is the public key

openssl.exe pkcs12 -export -out cert.pfx -inkey client.key -in client.pem

To verify that the pfx file has the private key I can use openssl to do this too.  
openssl pkcs12 -in cert.pfx -nocerts -out priv.pem -nodes

The original PR I did assumed that the private key was in PFX format which looking back at this is probably not a good assumption for all cases.  On windows I can get the PFX as a buffer which OpenSSL can parse form PFXExportCertStoreEx.  I think if there was a way to specify the PFX to librdkafka that would be the best option here.  Then we can set the private key as we did before.

EVP_PKEY* pkey;
X509 *cert;

BIO* bio = BIO_new_mem_buf(pPfx, cbPfx);
PKCS12 *pkcs12 = d2i_PKCS12_bio(bio, nullptr);
PKCS12_parse(pkcs12, get_password(), &pkey, &cert, NULL))

SSL_CTX_use_PrivateKey(ctx, pkey);
SSL_CTX_check_private_key(ctx);

Once there is a way to specify the PFX to librdkafka I can test the verify callback.  If it has the full peer certificate I should be validate it as I did previously.

edenhill:
So it sounds like we should support PEM, DER and PKCS12 for at least the client's public and private keys, yeah?

noahdav:
Yes PKCS12 will work for our scenarios.  Although it contains the public and private keys we only need it to parse the private key.  The public key can still be in PEM format if that simplifies anything.

LighthouseJ:
> OpenSSL does not make use of the Windows cert store for finding CA/root certs.

Core OpenSSL yes, but this is not technically true.  OpenSSL has an implementation for it using the OpenSSL dynamic engine interface.  The engine is called "capi", which stands for Windows Crypto API, and it provides the glue (on Windows) to make Windows Crypto calls to access Windows cert stores, show Windows prompt for choosing certificates, access hardware devices, etc...

edenhill:
@LighthouseJ Interesting, have you had any luck using it?

LighthouseJ:
@edenhill Yes, actually I've needed a similar change to what @noahdav did in the original PR specifically to do w/ Windows Crypto API but @noahdav did a good job at capturing a lot of what I needed, mainly the plumbing in librdkafka that I needed, I just needed to add a few things in, but unfortunately needed a little hack. :)

I was anticipating making a PR on github for folks w/ changes I need once this change is settled and merged in.

One notable difference and reason for the hack is that in circumstance where Windows will not let you have the private key because it's stored on a hardware device (so not read from a PEM file), Windows gives you a realistic OpenSSL-looking thing but you cannot serialize it to a BIO for example or do parsing on it, and instead you can only use it.  The calling code in rdkafka_transport.c that was there attempted to parse it and it effectively taint the blackbox private key object, so I had to short-circuit some of the code to just use the object w/o a parse.

Once this is merged and I can find the time, I can make a PR with this change in possibly in a less hacky way.

noahdav:
I originally looked at capi but did not think it was a good idea to require the implementation of the callbacks to need to deal with both OpenSSL and CryptAPI.  I only wanted to have each part of the code to deal with a single dependency and normalize it to OpenSSL since that is what librdkafka already does.

I am all for someone implementing the capi glue to librdkafka in a way that abstracts it and can handle the keys.  In our scenario the certificates are in the windows cert store but we do not interact with the windows cert store at all.  They are manged, rotated, etc by another service and we have an API which will eventually give us the PCCERT_CONTEXT.  Since we only have the PCCERT_CONTEXT I still need a way to convert this back to OpenSSL.  I think that the PKCS12 is the best way to handle this since I can convert back to the private key for OpenSSL.  If someone has a better option I am all for exploring another way.

@edenhill are you still planning on adding back the PKCS12 support or is this something I can add back? 
 I cannot use this PR as it currently stands for our scenario without this support.

edenhill:
@noahdav I'm working on adding specific config setters for the different format types, e.g.:
`rd_kafka_conf_set_ssl_privatekey_pkcs12(..)`, but also keep the PEM-based configuration property since that will be useful for the high-level client bindings.

noahdav:
@edenhill That will be great.  I will test it once you have the changes.  Thank you for all of your support on getting the PR completed.

src-cpp/rdkafkacpp.h
```diff
@@ -993,6 +1070,48 @@ class RD_EXPORT Conf {
...
 
+  /** @brief Use with \p name = \c \"ssl_cert_verify_cb\"
+      @remark this is not supported on the MIPS platform
+  */
+  virtual Conf::ConfResult set(const std::string &name,
+                               SslCertificateVerifyCb *ssl_cert_verify_cb,
+                               std::string &errstr) = 0;
+
+  /**
+   * @brief Set certificate/key \p cert_type from the \p cert_enc encoded
```
edenhill: @noahdav Does this look okay?

edenhill:
FYI: I gave OpenSSL's CAPI engine (Windows Crypto engine) a try and read thru its source; it does not provide a root CA certificate store, but it does seem to provide client certificate lookups.
I'm not sure the added dependency of capi.dll adds enough value for it to be included, so I'll skip it for now.
As for the CA root certs, as part of this PR I'll build that into librdkafka when on Windows and no ssl.ca.location has been specified.

noahdav:
@edenhill The API looks good to me however it is failing to parse the private key.  PKCS12_parse is failing.  I am trying to debug what is going on here to understand why this is now failing.  I will let you know what I find out.

edenhill:
@LighthouseJ 
> One notable difference and reason for the hack is that in circumstance where Windows will not let you have the private key because it's stored on a hardware device (so not read from a PEM file), Windows gives you a realistic OpenSSL-looking thing but you cannot serialize it to a BIO for example or do parsing on it, ..

I think we should provide a method to retrieve the SSL_CTX object from a librdkafka rd_kafka_t handle to allow deeper application integration with OpenSSL. That'd work for your use-case, right?

edenhill:
@noahdav I (force) pushed a couple of changes (including CA certs are now automatically read from the Root store), see examples/win_ssl_cert_store.cpp which is your original example extracted to its own file and reworked for the new API.
I can't get it to parse the private key from a PKCS#12 store, the OpenSSL error stack points to this line:
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/pkcs12/p12_mutl.c#L90

But maybe I'm just using a lousy key (that I found in my key store..:) )

edenhill:
Extra bonus, added `ssl.endpoint.identification.algorithm=https`

noahdav:
Hmm. That is the same error I am seeing in OpenSSL as well. I am not sure what the difference is here. My original code worked with the callbacks but for some reason this is not working. I do not see any differences in how you are parsing the pkcs12 vs what I originally did. I will dig into it more today to see if I can find anything out.

edenhill:
@noahdav Thank you!

noahdav:
@edenhill I finally figured out the issue here.  It looks like OpenSSL is not initialized yet since rd_kafka_ssl_init is not yet called.  Therefore we did not call SSL_library_init() so when we are trying to parse the pkcs12 it fails to find the digest algorithm.  When I originally made this change using the callbacks they were only called after OpenSSL was initialized so we would not see this issue.  I would assume since we are not properly initializing OpenSSL that many other thing can fail unexpectedly as well.

I will spend some time to familiarize myself with when rd_kafka_ssl_init gets called to see if there is a way to initialize OpenSSL prior to calling rd_kafka_conf_set_ssl_cert or for that matter any of the OpenSSL config methods.  Please let me know if you have thoughts to how this can be done.

edenhill:
@noahdav Aah, good find! Makes total sense. I'll push a fix for this soon.

noahdav:
After working around this initialization issue I am trying to test my verification callback.  So far I see a couple of issues.  First off in my test I am using a self signed cert.  So the first call to the rd_kafka_transport_ssl_cert_verify_cb function will call my callback but it does not have the context to understand why preverify_ok is 0.  In this case it is due to X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN.

Once I masked the X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN in the callback and return 1 in rd_kafka_transport_ssl_verify is SSL_get_verify_result is still returning the original error (X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN) so the verification still fails.  Do you know how to have the callback mask the error in this case?


I also think we should change the callback to pass a function table so that the callback can get information about the cert/validation process.  We can always extend it as needed but as long as the x509_ctx is passed to the callback they can parse it as currently implemented.  I assume each of the function pointers in the table would require the x509_ctx to be passed in anyhow so this would be needed.  I think this approach is better than offloading this to the callback as it make the callback not coupled to OpenSSL (librdkafka version dependent).  In our case we have no OpenSSL as we only use windows CryptAPI.  I do not think it is reasonable to ask implementer to take on an OpenSSL dependency in their application code if not necessary.  Also if for some reason we do not provide everything needed in rd_kafka_cert_verify_ft the context is still provided to the callback.

struct rd_kafka_cert_verify_ft {
error_code;
depth;
verification_result;
context;
...
}

noahdav:
@edenhill Thanks for the push to init OpenSSL.  That did the trick.  Back to my other issue.  We need a way to handle additional checks.  Since in this case the X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN error is failing for me I need a good way to call X509_STORE_CTX_set_error to success in this case.  I think this too could be added to the rd_kafka_cert_verify_ft that I mentioned above.  This would allow callbacks to be abstracted from OpenSSL.  Also the callback needs to be able to set X509_STORE_CTX_set_error in the case of a failure too.

noahdav:
@edenhill I went ahead and implemented the callbacks I was talking about.  It needs documentation comments and style fixes per the style guidelines.  However I wanted to get your thoughts on this type of change.  Please take a look at noahdav:ssl_cert_cb to see the changes and you can incorporate them in this PR.  I think for the callback to have the context to truly provide this validation check it needs this info at a minimum but will probably need additional callbacks as we discover them.  Please let me know your thoughts?

LighthouseJ:
> @LighthouseJ
> 
> > One notable difference and reason for the hack is that in circumstance where Windows will not let you have the private key because it's stored on a hardware device (so not read from a PEM file), Windows gives you a realistic OpenSSL-looking thing but you cannot serialize it to a BIO for example or do parsing on it, ..
> 
> I think we should provide a method to retrieve the SSL_CTX object from a librdkafka rd_kafka_t handle to allow deeper application integration with OpenSSL. That'd work for your use-case, right?

Not sure, but you might have a better idea because I have not had a chance to dig into this PR's worth of evolved work.

So the initial attempt by @noahdav was designed to put the entire cert into the buffer that's returning, and then use a sequence of calls to parse out and use certificates:

BIO_new_mem_buf (put bytes into BIO)
d2i_PKCS12_bio (convert BIO to PKCS12)
PKCS12_parse (extract private key)
and finally:
SSL_CTX_use_PrivateKey (use the private key)
to read and use the key in the SSL context.

This wasn't suitable to my use case because I could not possess the bytes to the private key, but got a pointer to them in memory.

I used CAPI along these lines:

```
    ENGINE *    ptr_engine;
    X509    *   ptr_client_cert;
    EVP_PKEY *  ptr_private_key;
...
    ENGINE_load_builtin_engines();
    ENGINE_load_dynamic();
    ERR_clear_error();
    ptr_engine = ENGINE_by_id("dynamic");
...
    ENGINE_ctrl_cmd_string(ptr_engine, "SO_PATH", "./capi.dll", 0);
    ENGINE_ctrl_cmd_string(ptr_engine, "ID", "capi", 0);
    ENGINE_ctrl_cmd_string(ptr_engine, "LIST_ADD", "1", 0);
    ENGINE_ctrl_cmd_string(ptr_engine, "LOAD", NULL, 0);

    ENGINE_init(ptr_engine);
...

ssize_t 
GetCAPICertificatesCb::ssl_cert_retrieve_cb(Type type, char **buffer, std::string &errstr)
{
    ...
    // only want to do this once, or a limited time if the load fails or needs to be re-read
    int load_result =
        ENGINE_load_ssl_client_cert(ptr_engine, NULL, NULL,
                &ptr_client_cert, &ptr_private_key,
                NULL, NULL, NULL);
    ...
    switch (type)
    {
    case CERTIFICATE_PRIVATE_KEY:
    {

        if (ptr_private_key)
        {

            // copies the pointer to the private key to the first character
            *buffer = (char *)ptr_private_key;
            return 1;

        break;
    } // case CERTIFICATE_PRIVATE_KEY:
    ...
    } // switch (type)
```

See the relevant call:
https://github.com/openssl/openssl/blob/master/engines/e_capi.c#L140
https://github.com/openssl/openssl/blob/master/engines/e_capi.c#L1691


I added a patch to @noahdav's original PR to look for buffer size of exactly 1 and then run this code:

```
if (len == 1) // *buffer points to a prepared EVP_PKEY pointer loses efficacy when serialized to a buffer
{
    rd_kafka_dbg(rk, SECURITY, "SSL",
        "Using an opaque EVP_PKEY pointer as the private key");
    EVP_PKEY * pkey = (EVP_PKEY *)buffer;

    r = SSL_CTX_use_PrivateKey(ctx, pkey);

    if (r != 1) {
        return -1;
    }

    rd_kafka_dbg(rk, SECURITY, "SSL",
        "Checking client private key");

    r = SSL_CTX_check_private_key(ctx);

    if (r != 1) {
        return -1;
    }
}
else
// do the original use case...
```

I just needed to bind the EVP_PKEY (which is prepared by Windows) to the SSL_CTX, in other words, only the last step above.

I realize that you're offering to give the SSL_CTX to users to use, which would smooth this out, but this is a little bit of what I did to skirt around the interface a little.  I did not look into this PR to see if this, or maybe making an EVP_PKEY an optional return, is a better choice.

edenhill:
@noahdav
I'm not sure it is a great idea to override OpenSSL's verification, I was under the impression that our verify callback was for additional verification.
For the self signed certs itself, you could set the cert as the ssl.ca.location to circumvent that, right?

As for the function table, I see what you are getting at, and the only thing that seems to be missing from today's signature is the per-depth error, the depth itself and the x509 are already provided.
Adding it as a function table in a public structure is unfortunately not extensible, since any change to that struct would break the ABI compat guarantees librdkafka provides.
We would have to make it an opaque/privately-defined object type with a lot of boiler plate for a feature that will not be that widely used. I'd like to avoid that if possible.
I feel that it might be easier to add the per-depth error code as an argument to the verify callback (possibly replacing the preverify_ok? since it makes it redundant).

@LighthouseJ 
The interface you are using, retrieve_cb, has been removed in favour of conf setters.
While it would be easy to provide a CERT_ENC_EVP_PKEY type  for conf->set_ssl_cert() it would not really work in practice since OpenSSL does not provide any ABI guarantees: if the application uses a slightly different version, or possibly even build, of OpenSSL, the passed EVP_PKEY may not be compatible with the OpenSSL version librdkafka is using.

So I suggest you convert the PKEY to PEM or PKCS#12 and pass that set_ssl_cert()

LighthouseJ:
> @LighthouseJ
> The interface you are using, retrieve_cb, has been removed in favour of conf setters.
Good to know!

> While it would be easy to provide a CERT_ENC_EVP_PKEY type for conf->set_ssl_cert() it would not really work in practice since OpenSSL does not provide any ABI guarantees: if the application uses a slightly different version, or possibly even build, of OpenSSL, the passed EVP_PKEY may not be compatible with the OpenSSL version librdkafka is using.

While I agree with this statement in general, I think for the purposes that librdkafka uses OpenSSL, it is quite stable, at least as far as librdkafka uses OpenSSL.
Part of the stability that CAPI inherits comes from the OpenSSL dynamic engine which is a C-function pointer system that keeps the OpenSSL engine interface stable.

I'm able to build librdkafka w/ OpenSSL 1.1.0g (previously 1.0.1 line as well) and use it without issue, so I believe 1.1.0 branch works fine w/ librdkafka from a librdkafka pre-1.0.0 release tag.

Could you please include the CERT_ENC_EVP_PKEY as a valid type of cert to return anyway, and I can try and use it to see if it fits my purpose?

> 
> So I suggest you convert the PKEY to PEM or PKCS#12 and pass that set_ssl_cert()

I did look into this before I made the ugly patch. I think when I found the right sequence of commands, I got no errors, but I got a zero-length buffer where the private key was supposed to be, so I thought that either Windows or OpenSSL was not allowing me to extract a private key as a method of protecting the integrity of the hardware device.

edenhill:
>  I think for the purposes that librdkafka uses OpenSSL, it is quite stable, at least as far as librdkafka uses OpenSSL.

As long as the OpenSSL project does not guarantee ABI stability we can't, and shouldn't, make any such assumptions. This will lead to bugs that are extremely difficult to track down.

Convert the ENV_PKEY to PEM or PKCS should be fairly straight forward with libssl.

noahdav:
@edenhill The problem with using the x509 that the callback passes today is that we are now requiring an application to take a hard dependency on OpenSSL.  Many windows applications do are using the Win32 Crypt API(s) instead.  Actually to make matters worse it is the same version that librdkafka uses so any application using a different version of OpenSSL is also unable to use this.  I think we really should not leak OpenSSL constructs to the callbacks in users applications.  Using ssl.ca.location will not work with certs in the windows store.  Most of our certs are trusted but even when they are self signed in some cases there is additional validation which is performed.  The application also needs the ability to set any error context when the validation fails.  This is currently missing from the API without using OpenSSL in the callback.  

Why is it breaking to add a function to the end of the function table.  A struct can be extended as long as you add it to the end.  Also we can #ifdef depending on the librdkafka version if necessary.  If this approach will not work then can we add a control function to perform these operations?

int ssl_verify_control(void* context, int32_t op, void** buffer, size_t bufferSize) {
switch(op) {
  case GET_CURRENT_ERROR:
     ...
    break;

    case SET_CURRENT_ERROR:
    ...
    break;

    case GET_DEPTH:
     ...
    break;

    ...
  }
  return 1;
}

edenhill:
@noahdav Yeah, I agree, I'll take out the x509 parameter from the verify callback.

As for public structs, an application needs to be recompiled if the struct layout changes, thus breaking the ABI compatibility.

And I don't think a function table is really necessary, seeing how the only missing information is the per-cert error code which we can add to the verify_cb in some form.

> The application also needs the ability to set any error context when the validation fails.

Why does the certificate verification need to be more fine grained than success or failure?

> Using ssl.ca.location will not work with certs in the windows store.

Right, but if we allowed CA certs to be specified with set_ssl_cert  it would work, right?
Also, if the CA cert is in the Root store librdkafka will pick it up automatically now.

noahdav:
@edenhill  After the callback the call to SSL_get_verify_result is what librdkafka is using to determine success/failure.  Therefore we need to set the error so that this function will propagate the failure form the callback.

The callback needs the ability to retrieve any error and depth so that it can perform additional checks based upon the error which occurred.  

Adding the root CA may help in some of the cases but in many cases there is additional cert checks we want to apply.  This still needs to be done somehow.  I think it is simplest to delegate this to the callback to determine the logic of what success/failure is.

edenhill:
> The callback needs the ability to retrieve any error and depth so that it can perform additional checks based upon the error which occurred.

Since the callback is called once for each depth, does it really need to be able to extract errors for other depths than it is currently being called for?

As for error reporting, the user shall set an error string if the callback returns false, and that error string will be logged by librdkafka.

noahdav:
@edenhill I agree.  The callback only needs the current depth/error.  As for error reporting to OpenSSL I think we should have a way to set the OpenSSL error.  What is you objection to allowing the callback to set this error?

edenhill:
> What is you objection to allowing the callback to set this error?

There are two issues with it:
 1. `X509_STORE_CTX_set_error()` expects an SSL / libcrypto error code - this leaks the abstraction that we're trying to provide, this means an application will need to include openssl/err.h, et.al, to set this to a meaningful value.
 2. the ctx error will just show up as a generic and weirdly formatted openssl error (because they are all weirdly formatted :) ) - while the errstr could provide some actual information on the verification error, not just the kind of error.

So I don't think it adds much value.

noahdav:
@edenhill fair enough.  I now understand your concern.  I think this is valid.  So for the callback to handle a certificate that is self signed on the server what is your suggestion to be able to successfully validate it?  Currently it will fail saying it is self signed.  Other than this I think this PR looks great.

I will test today the SSL_get_verify_result how it responds to failure in the cert validation callback.  I was under the impression that we had to set the error context in addition to the true/false return code.

noahdav:
@edenhill Your concerns are valid.  I understand and after thinking more about it I agree with what you are saying.  I think it would be simplest to allow a self signed cert if the client asks for it to be ignored.  Maybe we can add a flags to the config?  Otherwise we can add the root ca if we can support adding it to set_ssl_cert.

I tested this and other than the self signed cert it seems to be working.  When the verify callback returns false we keep retrying which causes the callback to get called over and over in the connect loop.  This does not seem like a good idea but it does eventually times out.  I think we should try to understand if the fail to connect is due to verify failure and in this case do not try to reconnect.

edenhill:
> I think it would be simplest to allow a self signed cert if the client asks for it to be ignored.

I looked at SSL_CTX_set_verify() again and it actually allows us to override the preverify_ok failure that you are seeing, so if we just provide the error to the verify callback  you can check if it is the self-signed-cert error and then return ok rather than fail (from preverify_ok). But we might need to clear the error on the context, I'll look into it.

> Maybe we can add a flags to the config? Otherwise we can add the root ca if we can support adding it to set_ssl_cert.

But this is probably a better approach, I'll add support for setting the CA certificates using set_ssl_cert().  I'll add DER and PKCS#12/PFX support for CAs, that should be enough, right?

edenhill:
>  I think we should try to understand if the fail to connect is due to verify failure and in this case do not try to reconnect.

Yeah, librdkafka currently treats all errors as temporary. We're looking into enriching the error handling with indication if an error is temporary or permanent to allow the application to take proper action.

edenhill:
One open question is how we pass the OpenSSL error code to the verify_cb.
 * Unchanged, people will have to look up the int define in openssl/err.h (et.al)?
 * Add librdkafka defines for a few "well known", I dont even know where to begin
 * Convert them to a text string and let the application `strcmp/strstr()`?

noahdav:
@edenhill Thanks so much for your work on this.  I really appreciate it.

> Yeah, librdkafka currently treats all errors as temporary. We're looking into enriching the error handling with indication if an error is temporary or permanent to allow the application to take proper action.

As long as we have the same behavior as rest of librdkafka I guess that is good enough for now.

> One open question is how we pass the OpenSSL error code to the verify_cb.

As much as I hate to admit it I think that using the OpenSSL error codes is the way to go.  This will allow an application to either define the ones they care about or use OpenSSL defines in err.h/x509_vfy.h/etc.  Although it is not pretty I think it is much better then string parsing to figure out the errors.

> I looked at SSL_CTX_set_verify() again and it actually allows us to override the preverify_ok failure that you are seeing, so if we just provide the error to the verify callback you can check if it is the self-signed-cert error and then return ok rather than fail (from preverify_ok). But we might need to clear the error on the context, I'll look into it.

I looked at this too and we will need a way to clear the error in the context otherwise SSL_CTX_set_verify() will return the original error.  I was using X509_STORE_CTX_set_error to do this in the function table.  If you have another way to do this so that the application does not have a OpenSSL link time dependency that would be great.

>  I'll add support for setting the CA certificates using set_ssl_cert(). I'll add DER and PKCS#12/PFX support for CAs, that should be enough, right?

Yes as long as you support PKCS#12/PFX we can convert all of our certs into this.

edenhill:
@noahdav Can you pull and try out the latest version?

The verify_cb changed: x509 and preverify_ok are gone, instead we have an in/out x509_error that you can use to change or clear the X509 CTX error.

CA loading from PKCS, DER and PEM was also added.

noahdav:
@edenhill Thanks for your work on this.  I am really excited to get this all complete.  I will take another look at this today.

noahdav:
@edenhill This will work for my scenario.  Thank you for all of your help on this feature.  This will be great to get it in.

I am seeing is that when I am just testing on my localmachine the callback gets called multiple times.  I am setting the metadata.broker.list to localhost:9093.  Then in rd_kafka_transport_io_event the rd_kafka_transport_t already has two entries in it.  One for the localhost (broker_id = -1) and one for <fqdn>:9093 (broker_id = 0).  Since there are two entries this causes the callback to get called two times (for each cert in the chain).  I do not see where the dns lookup is occurring but it seems that this should not create the extra connection in this case.  I can work around this by ignoring when broker_id < 0 but I think this should not create the extra connection.

On a simpler note you only allow the callback to clear the cert error.  Should we allow them to set the error too?  If not I am ok with this but we should have another flag that if set clears error.  Then have the current error field only and input not in out.  I do not think that this is clear that on input the x509_error param can have any value but on output only 0 has any effect.

edenhill:
librdkafka will map a bootstrap broker to a broker reported through cluster metadata if the hostname and port matches exactly, but it will not do so based on what the hostname resolves to since DNS entries may change or resolve to multiple addresses, so the behaviour you are describing is expected.

You can set the context error by updating the error code in x509_error and returning 0/false from the verify_cb.

noahdav:
@edenhill thanks you are right.  I missed that part of returning false that will set error.  This seems good to me.

noahdav:
@edenhill It looks like I spoke too soon.  In rd_kafka_conf_set_ssl_cert the map needs to initialize the ca cert.

rd_kafka_cert_t **cert_map[RD_KAFKA_CERT__CNT] = {
                [RD_KAFKA_CERT_PUBLIC_KEY] = &conf->ssl.cert,
                [RD_KAFKA_CERT_PRIVATE_KEY] = &conf->ssl.key,
                [RD_KAFKA_CERT_CA] = &conf->ssl.ca
        };

After making this change I am still getting error  X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN in the callback.  Any idea what I need to do in order to get OpenSSL to recognize the self signed ca?

edenhill:
Good catch!

What's your config?

noahdav:
I just tried to use set_ssl_cert to with CERT_CA CERT_ENC_PEM which returned success.  The pem file is the ca cert that I used to generate the certificate public and private keys.  Is there something else that I am missing here?

security.protocol=ssl
ssl.certificate.pem=...
ssl.key.password=...
set_ssl_cert with CERT_PRIVATE_KEY CERT_ENC_PKCS12
enable.ssl.certificate.verification=true
ssl_cert_verify_cb = verify_cb

edenhill:
So the thing is that CERT_CA, ssl.ca.pem AND PUBLIC_KEY and ssl.certificate.* all map to SSL_CTX_use_certificate(), which seems to be a singletonish thing in the context, i.e., sub-sequent calls will replace the previous one.

I'm not really sure if I should be using another API, possibly SSL_CTX_add_extra_chain_cert(), but that should only be needed if there is indeed a chain of certs that need to be added.
In your case it sounds like you only need to add one certificate.

Can you try commenting out your `ssl.certificate.pem` config and see if it helps, since that overwrites the CA?

noahdav:
I did that but then the public key cannot be found which causes create of the procurer to fail.  I can certainly work around this in the callback by setting the error back to 0 but not the most elegant solution.  Looking at the docs I think you are right that if we want to support this so a cert can be added with its CA we need to add it using  SSL_CTX_add_extra_chain_cer or SSL_CTX_add1_chain_cert.

edenhill:
Okay @noahdav, this should now be fixed, can you give it another spin?

noahdav:
@edenhill  What should the expected behavior be here.  I set the ca cert using set_ssl_cert.  However I am still seeing the validate callback being called with error code X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN.

edenhill:
@noahdav Can you export your private key and the root CA to files and then use openssl command line to verify correctness?
`openssl verify -CAfile yourca.pem yourkey.pem`

noahdav:
@edenhill OpenSSL verify works fine.  OpenSSL returns client.pem: OK when I use the public key of the ca and the cert public key.  

In my testing I do the following:

Set ca public key in PEM format via set_ssl_cert()
ssl.certificate.pem=<public key pem>
ssl.key.password=<pass>
Set private key in PFX/PKCS#12 format via set_ssl_cert()
enable.ssl.certificate.verification=true
ssl_cert_verify_cb=<callback

When the callback gets invoked the error is still 19(X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN).

edenhill:
If you specify the client's public key as both certificate.pem and CA, then it is effectively a self-signed certificate.
You should have a root-CA cert (by your own authority or a proper one) signing your priv/pub key, right?

LighthouseJ:
> If you specify the client's public key as both certificate.pem and CA, then it is effectively a self-signed certificate.

IMHO, this isn't technically true. A self-signed CA has the same subject and issuer DN, and is of course a CA.

Help for X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN says:

> X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN: self signed certificate in certificate chain
>    the certificate chain could be built up using the untrusted certificates but the root could not be found  locally.

I think it's okay but not ideal to send the self-signed cert in the chain, but the primary cause for the verification error is if the remote end did not trust the CA.  The certs you send over the wire is untrusted, and the CA(s) need to be trusted via -CAfile or -CAdir (or JSSE CA truststore) first.

I typically have a self-signed root CA, which signs intermediate CA, which signs server certs and client certs per application.  I then create cert chains (which are ordered!) with intermediate and leaf cert, which avoids that verify failure as long as the peer trusts my CA files.

I create my own CA's and certs for various purposes using this guide:
https://jamielinux.com/docs/openssl-certificate-authority/

edenhill:
Oh, sorry. The CA cert you pass to librdkafka is for verifying the broker certificate, not the client's

noahdav:
> If you specify the client's public key as both certificate.pem and CA, then it is effectively a self-signed certificate.

I can try this but I thouhght I should be using the root-CA.

> You should have a root-CA cert (by your own authority or a proper one) signing your priv/pub key, right?

Yes I have a ca-cert which I used to sign both client and broker certs.

edenhill:
@noahdav 

Okay, so if you have a CA-cert that was used to sign the client's key and broker's certificate then you should be fine with setting:
 * `ssl.key.*` or `CERT_PRIVATE_KEY` = the client's private key
 * `ssl.certificate.*` or CERT_PUBILIC_KEY` = the client's public key / certificate
 * `ssl.ca.*` or `CERT_CA` = the ca-cert you used to sign the client and broker certificates

I'm guessing you are only using `set_ssl_cert()`, and not mixing these properties, right?
What encodings are you using for these three?

Can you verify the above?

edenhill:
Okay @noahdav, I think I've got this sorted out now.
OpenSSL has a bunch of different, but similarily named, certificate stores, and I was using the wrong one.
Please give the latest version of this branch a try and report back please!

noahdav:
@edenhill That fixed my issue.  I am now able to validate the self signed cert.  Thanks again for all of your help with this.  I am excited to get this merge to master soon.  I will review one more time early next week if that works for you.  I took a look and did not see anything related to this in the latest commits.  Can you also point me to the commit which fixed this so I can see what actual changed?

edenhill:
@noahdav That's fantastic news, thank you!
We're now passing a X509_STORE to SSL_CTX_set_cert_store() (which is used for  verifying the peer cert):

https://github.com/edenhill/librdkafka/pull/2309/files#diff-c5104996425e1a3e383d42a77a8124ccR725

noahdav:
@edenhill Thank  you for all your help with this.  I will do a full of this early next week and let you know if I see anything.

wesnerm:
From issue #2319.

>Sorry, for the lack of response. I haven't had to opportunity to test it.
>
>Since I get my packages from NuGet, I need to figure out the interaction between package management, wrapper libraries (Confluent.NET) and private versions of packages from pull requests.
>
>The PR does address other features that I have been interested in such as in-memory keys and the use of Windows roots.
>
>When I verify this works or doesn't work, I will send a message to both this issue and the PR.
>
>Most likely, I will only able to get to this after the PR is merged.
>
>Thanks for your work.

CONFIGURATION.md
```diff
@@ -59,12 +59,20 @@ ssl.cipher.suites                        |  *  |                 |
...
 ssl.sigalgs.list                         |  *  |                 |               | low        | The client uses the TLS ClientHello signature_algorithms extension to indicate to the server which signature/hash algorithm pairs may be used in digital signatures. See manual page for `SSL_CTX_set1_sigalgs_list(3)`. OpenSSL >= 1.0.2 required. <br>*Type: string*
 ssl.key.location                         |  *  |                 |               | low        | Path to client's private key (PEM) used for authentication. <br>*Type: string*
-ssl.key.password                         |  *  |                 |               | low        | Private key passphrase <br>*Type: string*
+ssl.key.password                         |  *  |                 |               | low        | Private key passphrase (for use with ssl.key.location) <br>*Type: string*
```
mhowlett: also for use with rd_kafka_conf_set_ssl_cert according to it's docs.

CONFIGURATION.md
```diff
@@ -59,12 +59,20 @@ ssl.cipher.suites                        |  *  |                 |
...
 ssl.certificate.location                 |  *  |                 |               | low        | Path to client's public key (PEM) used for authentication. <br>*Type: string*
-ssl.ca.location                          |  *  |                 |               | medium     | File or directory path to CA certificate(s) for verifying the broker's key. <br>*Type: string*
+ssl.certificate.pem                      |  *  |                 |               | low        | Client's public key string (PEM format) used for authentication. <br>*Type: string*
+ssl_certificate                          |  *  |                 |               | low        | Client's public key as set by rd_kafka_conf_set_ssl_cert() <br>*Type: *
+ssl.ca.location                          |  *  |                 |               | low        | File or directory path to CA certificate(s) for verifying the broker's key. <br>*Type: string*
+ssl_ca                                   |  *  |                 |               | low        | CA certificate as set by rd_kafka_conf_set_ssl_cert() <br>*Type: *
```
mhowlett: I guess not having a string config property for CA certs was deliberate? Seems to be the only thing missing to complete the set. I don't know why it would be useful though.

examples/rdkafka_performance.c
```diff
@@ -3,24 +3,24 @@
  *
  * Copyright (c) 2012, Magnus Edenhill
  * All rights reserved.
- * 
+ *
```
mhowlett: perhaps there is less reason to be scared by the line count of this PR than I thought! white space changes ideally shouldn't be here.

CONFIGURATION.md
wesnerm: Actually, there are several new settings that offer in-memory certificates; hence the title for this PR. I believe for example that the string config property for CA certs is ssl_ca.

examples/win_ssl_cert_store.cpp
```diff
@@ -0,0 +1,424 @@
...
+   * Typically include path in a real application would be
+   * #include <librdkafka/rdkafkacpp.h>
+   */
+#include "rdkafkacpp.h"
+
+
+
+class ExampleStoreRetriever {
+public:
+        ExampleStoreRetriever (std::string const &subject, std::string const &pass)
```
mhowlett: minor point, but different indentation/formatting would be better in the below few lines to aid understanding at a glance... will refrain from similar comments elsewhere. indentation + i'd re-order the initializer list to logically group the passed in paramters together, and the NULL initializers together.

src-cpp/rdkafkacpp.h
```diff
@@ -830,6 +860,60 @@ class RD_EXPORT OffsetCommitCb {
...
+ */
+class RD_EXPORT SslCertificateVerifyCb {
+public:
+  /**
+   * @brief SSL broker certificate verification callback.
+   *
+   * The verification callback is triggered from internal librdkafka threads
+   * upon connecting to a broker. On each connection attempt the callback
+   * will be called for each certificate in the broker's certificate chain,
+   * starting at the root certification, as long as the application callback.
```
mhowlett: incorrect full stop

src-cpp/rdkafkacpp.h
```diff
@@ -830,6 +860,60 @@ class RD_EXPORT OffsetCommitCb {
...
+   * returns 1 (valid certificate).
+   *
+   * \p broker_name and \p broker_id correlate to the broker the connection
```
mhowlett: correlate -> correspond?

src/rdkafka_cert.c
```diff
@@ -0,0 +1,556 @@
...
+        return pwlen;
+}
+
+
+
+/**
+ * @name Public API
+ * @brief These public methods must be available regardless if
+ *        librdkafka was built with OpenSSL or not.
+ * @{
```
mhowlett: are they? this is within the `#if WITH_SSL` block.

CONFIGURATION.md
edenhill: @mhowlett Right, I don't believe it is useful. We can add it later if necessary.

@wesnerm ssl_ca is the backing store for set_ssl_cert(), it can't be set directly with a string setter.

examples/rdkafka_performance.c
edenhill: I generally agree, but this example is so rarely visited that I took the opportunity to fix its style when there was other business in this area. 🤷‍♂

examples/win_ssl_cert_store.cpp
edenhill: Fixing initializer order, but keeping indentation:
this file is styled by Visual Studio, while other C++ code uses the Google C++ style, the reason I didn't use the Google style here is that this is a Windows-only file and it will most likely be edited in Visual Studio only.

src/rdkafka_cert.c
edenhill: good find!

edenhill:
Thanks for the review @mhowlett !

@noahdav As soon as you give the thumps up we'll merge this.

src-cpp/rdkafkacpp.h
noahdav: looks good to me

noahdav:
@edenhill Thanks for all your help getting this completed.  This will be great to get into next release.

edenhill:
@noahdav Did you get a chance to review?

noahdav:
@edenhill yes it looks good.  I did already approve the PR.

edenhill:
@noahdav Oh, sorry, thanks!

edenhill:
@noahdav 
Thanks a lot for all the time and effort you've spent on this PR, from the code contribution, to reviews discussions, testing, and verification, it has been hugely valuable! :100:


We're looking to make a release including this changes before end of June.

noahdav:
@edenhill My pleasure.  I am excited to have this in the community release.  Thank you too for all your help with this.

larsbrekken:
Hi guys, this functionality has worked great for us when running our code from a Windows Service with extensive permissions, however when running it from a web site with a default App Pool Identity user, we get the following permissions-related error:
```
DEBUG|[thrd:app]: librdkafka built with OpenSSL version 0x1000212f 
DEBUG|[thrd:app]: Failed to open Windows certificate Root store: Access is denied...: falling back to OpenSSL default CA paths
```

I am not familiar with the Windows system calls for accessing the store, but in the docs I see a mention of a store called "Ca" in addition to the "Root" store that librdkafka uses.

https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certopenstore

That page also mentions a simpler call, and the example on that page uses the "Ca" store.

Again, not an expert on this and the "Root" store might be the right one for librdkafka. 

Unfortunately the permissions issue is blocking us from rolling out with the new functionality and have to stick with referencing a CA file on disk.

edenhill:
@larsbrekken Try changing this line to "Ca" and see if it helps:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_ssl.c#L658

If so we can make the store name configurable.

noahdav:
@LighthouseJ Can you share what you did to get the capi working?  We have a new requirement which does not allow us to export the private key to feed into OpenSSL.  Therefore we need to use capi to access the cert store directly.  I have not seen many examples of using this so I was hoping to see what you did in order to get this working.  We will also need to wire this up into librdkafka.  I am hoping to build on top of what you have already done.

Thanks

LighthouseJ:
@noahdav I was sure that I shared details on this... but I can't find it in this PR or your original.  I'm not at work now so I can't provide details now, but I will soon.

I vaguely outlined what you're looking for in this comment:
https://github.com/edenhill/librdkafka/pull/2309#issuecomment-489642251

In your PR, you allowed a user who has access to a soft cert to read and provide it to librdkafka.  With a smartcard or softcerts in Windows cert manager, you don't have access to read the private key, but with OpenSSL (and Windows CAPI), you can get an opaque object that points to the private key, and then use it within librdkafka.  My hack was to your original changeset, where the retrieve_cb could return a special payload result, instead of a PEM file, is a pointer to the OpenSSL private key opaque object.  The library was slightly modified to also look for a special (short) payload, and if found, cast as a pointer, do a check for validity, then call openssl to use it as the private key.

The essential change needed to librdkafka would be to modify the current function where retrieve_cb() used to be to return different types of private keys, so not just a PEM cert in a byte buffer, but also a private key object (a native OpenSSL object?)

Give me a few days and I'll reply with info on how to query capi w/ capi ui chooser dialog (happens in my implementation of your retrieve_cb()) and what I did to wire it into rdkafka to use it when authenticating with openssl.

I know retrieve_cb() was not kept in the final changes by Magnus, butI haven't looked into how difficult it would be to port the change that we probably both need into current rdkafka, but I hope to give you enough info to try and make it in.

LighthouseJ:
@noahdav

Here's the critical part.  In your branch, in the rd_kafka_transport_set_client_certificates() function, right after you call the user callback and you added:
```
BIO* bio = BIO_new_mem_buf(buffer, (long)len);
if (bio) {
    PKCS12 *p12 = d2i_PKCS12_bio(bio, NULL);
    ...
    < read file into BIO, use as private key if parse OK >
} else
        "Failed to initialize BIO");
```

Before that part, I patched it to do this:
```
if (len == 1) // *buffer points to a prepared EVP_PKEY pointer loses efficacy when serialized to a buffer
{
    rd_kafka_dbg(rk, SECURITY, "SSL",
        "Using an opaque EVP_PKEY pointer as the private key");

    EVP_PKEY * pkey = (EVP_PKEY *)buffer;

    r = SSL_CTX_use_PrivateKey(ctx, pkey);

    if (r != 1) { return -1; }

    rd_kafka_dbg(rk, SECURITY, "SSL",
        "Checking client private key");

    r = SSL_CTX_check_private_key(ctx);
    if (r != 1) { return -1; }
				
    rd_kafka_dbg(rk, SECURITY, "SSL", "Private key OK");
} // if (len == 1)
else // Assume the user passed a larger buffer and read normally
{ // do the rest of your code
    if (bio) {
        PKCS12 *p12 = d2i_PKCS12_bio(bio, NULL);
        ...
        < read file into BIO, use as private key if parse OK >
    } else
            "Failed to initialize BIO");
```

Here's how I initialize OpenSSL engine w/ capi:
```
// types in header implementation of your RdKafka::SslCertificateRetrieveCb class
    ENGINE *    ptr_engine;

    X509    *   ptr_client_cert;

    EVP_PKEY *  ptr_private_key;
// code to initialize in .cpp file
    ENGINE_load_builtin_engines();
    ENGINE_load_dynamic();
    ERR_clear_error();
    ptr_engine = ENGINE_by_id("dynamic");

    if (ptr_engine)
    {
        ENGINE_ctrl_cmd_string(ptr_engine, "SO_PATH", "./capi.dll", 0);
        ENGINE_ctrl_cmd_string(ptr_engine, "ID", "capi", 0);
        ENGINE_ctrl_cmd_string(ptr_engine, "LIST_ADD", "1", 0);
        ENGINE_ctrl_cmd_string(ptr_engine, "LOAD", NULL, 0);

        ENGINE_init(ptr_engine);

        // now loaded
    }
    else
    { // failed to load the dynamic engine
    }
```

I do a one-time lazy init at the top of your ssl_cert_retrieve_cb:
```
    if ((ptr_client_cert == NULL) || (ptr_private_key == NULL))
    {
        int load_result =
            ENGINE_load_ssl_client_cert(ptr_engine, NULL, NULL,
                &ptr_client_cert, &ptr_private_key,
                NULL, NULL, NULL);
        // log load_result to help diagnose

        if (load_result != 0) // 0 == user canceled, 1 == user selected a cert
        {
            // Loading user certificate info OK
        }
    }
```

And in the body of ssl_cert_retrieve_cb, I do this:
```
    switch (type)
    {
    case CERTIFICATE_PUBLIC_KEY:
    {
        if (ptr_client_cert)
        {
            // Loading public key...

            // need to convert the provided public key "internal" openssl format
            // to 'encoded' DER-formatted certificate
            int size_encoded =
                i2d_X509(ptr_client_cert, (unsigned char **)buffer);

            // Encoded <size_encoded> bytes into the outgoing buffer

            return (size_t)size_encoded;
        }
        else
        {
            // Unable to load public key because the client certificate was invalid
        }

        break;
    } // case CERTIFICATE_PUBLIC_KEY:
    case CERTIFICATE_PRIVATE_KEY:
    {
        if (ptr_private_key)
        {
            // Pushing private key pointer to kafka library

            // copies the pointer to the private key to the first character
            *buffer = (char *)ptr_private_key;
            return 1;
        }
        else
        {
            // Unable to find the private key
        }

        break;
    } // case CERTIFICATE_PRIVATE_KEY:

    case CERTIFICATE_PRIVATE_KEY_PASS:
        // do nothing
        break;
        
    } // switch (type)
    // default case -- return blanks!
    *buffer = NULL;
    return 0;
}
```

And there's appropriate cleanup...

ENGINE_load_ssl_client_cert(...) was the site where the user was asked for their cert, and saves what they chose to the member variables, and pointers to client cert and key are passed to the patch of librdkafka to use as-is *without* passing through BIO or any conversion.

To get the same level of function, you'll have to compile openssl with capi engine, and put capi.dll in the same directory as the executable running per the SO_PATH used.  Also, the certificate chooser dialog is **not** shown by default, you have to compile openssl w/ OPENSSL_CAPIENG_DIALOG preprocessor defined else capi will just choose the first cert/key pair found which may not be what you want.

noahdav:
@LighthouseJ Thank you for this.  It looks very helpful.  I will go through this and let you know what I come up with.

PR #2274, opened 2019-04-08
CMake: Remove redundant LIBRDKAFKACPP_EXPORTS definition

myd7349:
Hi! I am creating a portfile for `librdkafka` in `vcpkg` these days. And I encountered several `unresolved external symbol` errors related to those static const data members defined  in `RdKafka::Topic` when I built `librdkafka` as a shared library on Win32: https://github.com/Microsoft/vcpkg/pull/5921#issuecomment-480585430

After some research, I found that this issue seems caused by a redundant `LIBRDKAFKACPP_EXPORTS` definition in this file: https://github.com/edenhill/librdkafka/blob/master/src-cpp/CMakeLists.txt#L37-L39
This redundant definition causes `RD_EXPORT` expanded as `__declspec(dllexport)` in a CMake-based project.

A small CMake project to help reproduce this problem: https://github.com/myd7349/Ongoing-Study/tree/master/cpp/CMake/vcpkg/librdkafka_pr_2274

PR #2310, opened 2019-05-07
Minor fixes for Win32

myd7349:
- Remove redundant LIBRDKAFKACPP_EXPORTS definition in tests missed in #2274
  I want to apologize for it since I should have fixed it in #2274. But I didn't run the tests before: https://github.com/myd7349/vcpkg/blob/8aeafa192dc621f17be56108cabcc556fe6522b9/ports/librdkafka/portfile.cmake#L32, so I just missed it at that time.
  We can reproduce the link error caused by this definition on Win32 by:
  ```
  mkdir build
  cd build
  cmake ..
  cmake --build .
  ```
- Fix C4996 warning related to strdup
  When I was trying to build librdkafka on UWP, a C4996 error is reported by VS: 
  >e:\vcpkg\buildtrees\librdkafka\src\a34ca9b746-a24f4e04b2\src\rdkafka_queue.h(652): error C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _strdup. See online help for details. (compiling source file

  Though the UWP build is still failed for some other reasons([install-x86-uwp-dbg-out.log](https://github.com/edenhill/librdkafka/files/3152143/install-x86-uwp-dbg-out.log)), I think this fix is reasonable since all other places in the code use `rd_strdup` instead of `strdup`.

### DIFF
--- a/src/rdkafka_queue.h
+++ b/src/rdkafka_queue.h
@@ -649,7 +649,7 @@ rd_kafka_set_replyq (rd_kafka_replyq_t *replyq,
 	replyq->q = rkq ? rd_kafka_q_keep(rkq) : NULL;
 	replyq->version = version;
 #if ENABLE_DEVEL
-	replyq->_id = strdup(__FUNCTION__);
+	replyq->_id = rd_strdup(__FUNCTION__);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


edenhill:
Thanks!

PR #2323, opened 2019-05-11
CMake: Fix configuration on Win32

myd7349:
On Windows with MSVC, this line:
```cmake
check_library_exists(m pow "" WITH_HDRHISTOGRAM)
```
causes `WITH_HDRHISTOGRAM` set to `OFF`.
As a result:
- When specifying `WITHOUT_WIN32_CONFIG=ON`, the histogram source will not be built.
- When specifying `WITHOUT_WIN32_CONFIG=OFF`, there will be link erros since `WITH_HISTOGRAM` is defined as 1 in `win32_config.h`

References:
https://stackoverflow.com/questions/32816646/can-cmake-detect-if-i-need-to-link-to-libm-when-using-pow-in-c


myd7349:
The configuration step on Win32 through CMake still have room for improvement.
1.

CMakeLists.txt
```diff
@@ -35,7 +35,12 @@ endif()
 
 # LIBM {
 include(CheckLibraryExists)
-check_library_exists(m pow "" WITH_HDRHISTOGRAM)
+check_library_exists(m pow "" HAVE_LIBM)
```
edenhill: On MSVC we don't need to, and probably shouldnt, check for libm at all.
What do you think?

edenhill:
I think I like the more aggressive approach, better to get this all right in one go.

CMakeLists.txt
myd7349: @edenhill Thanks for your review! I will update this PR later.

myd7349:
Here is a related discussion about `libm`: https://github.com/mesonbuild/meson/issues/5390#issuecomment-491640529

myd7349:
`snappy` support doesn't work out-of-box, reported at:
https://github.com/microsoft/vcpkg/pull/7456

edenhill:
@myd7349 WITH_SNAPPY should always be set to 1 in CMakeLists.txt

PR #2337, opened 2019-05-24
fix off by one issue in rd_kafka_offsets_store

meridional:
There is a discrepancy between `rd_kafka_offsets_store` and `rd_kafka_offset_store` where only the latter automatically increment the offset by 1 and it's probably the desired the behavior in most scenarios.

This PR basically lets `rd_kafka_offsets_store` to do the same.

Please let me know if there is rationale behind the discrepancy or if this is indeed a bug.

More context:
I encountered this issue when running a go consumer, and used `consumer.StoreOffsets` which calls `rd_kafka_offsets_store`. And it came as a surprise to me that restarting an up-to-date consumer would replay the latest message.

Will add tests if necessary.

### DIFF
--- a/src/rdkafka_offset.c
+++ b/src/rdkafka_offset.c
@@ -714,7 +714,7 @@ rd_kafka_offsets_store (rd_kafka_t *rk,
                 }
 
                 rd_kafka_offset_store0(rd_kafka_toppar_s2i(s_rktp),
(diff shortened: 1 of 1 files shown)


edenhill:
Yeah this is a lapse in the implementation, but we can't change it since it would break existing applications.

We should update the documentation to point out the difference though, and deprecate rd_kafka_offset_store().

PR #2349, opened 2019-06-03
Fix wrong legacy batch absolute offset calculation

martinivanov:
Currently, the number of messages in a legacy (v0, v1) message set is used to derive the base offset of the messages. Messages in a message set are not guaranteed to have a consecutive relative offset, which leads to message offsets being overridden by librdkafka and eventually lost to the consumer. The PR changes the usage of the number of messages in the message set to the relative offset of the last message in the set for the calculation of the base offset, which removes the possible inconsistency.

### DIFF
--- a/src/rdkafka_msgset_reader.c
+++ b/src/rdkafka_msgset_reader.c
@@ -1060,7 +1060,13 @@ static void rd_kafka_msgset_reader_postproc (rd_kafka_msgset_reader_t *msetr,
                                              int64_t *last_offsetp) {
         rd_kafka_op_t *rko;
 
-        if (msetr->msetr_relative_offsets) {
+        rko = rd_kafka_q_last(&msetr->msetr_rkq,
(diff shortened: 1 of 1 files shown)


martinivanov:
Here is the equivalent Java client code:
https://github.com/apache/kafka/blob/c09e25fac2aaea61af892ae3e5273679a4bdbc7d/clients/src/main/java/org/apache/kafka/common/record/AbstractLegacyRecordBatch.java#L379
https://github.com/apache/kafka/blob/c09e25fac2aaea61af892ae3e5273679a4bdbc7d/clients/src/main/java/org/apache/kafka/common/record/AbstractLegacyRecordBatch.java#L400

dkholod:
Hi guys.
On topics with a lot of changes and compaction we often see the same issues. Messages are lost.
Is it any plans for a fix?

edenhill:
Thank you!

PR #2358, opened 2019-06-12
CMake: Add check for GNU thread name setting support.

billygout:
Tested by building with cmake with and without successful compile of [by introducing a typo into]  `packaging/cmake/try_compile/pthread_setname_gnu_test.c`, and checking by running `gdb` on one of the `examples/` apps and checking the difference in the output of `disassemble thrd_setname`.


edenhill:
Think you forgot the compile test file itself:
```
CMake Error at CMakeLists.txt:15 (add_executable):
  Cannot find source file:

    /project/repo/checkout/packaging/cmake/try_compile/pthread_setname_gnu_test.c
```

billygout:
Doh! Thanks

billygout:
fixed,rebased,force-pushed

edenhill:
Thank you!

PR #2364, opened 2019-06-14
Bump message.timeout.ms limit to INT32_MAX

sarkanyi:
Because 0 is broken and some people need more than just 15 minutes.


edenhill:
Thank you!

PR #2365, opened 2019-06-17
max.poll.interval.ms: properly handle blocking calls

edenhill:
Previous to this fix the app_polled time was only updated upon
entry to a (possibly) blocking function, such as consumer_poll().
If the call had a timeout > max.poll.interval.ms and there were no messages
for at least max.poll.interval.ms, the max poll check would kick in and
evict the consumer from the group due to inactivity.

We now signal to app_polled that we're in a blocking call
which prevents the max poll check to fire.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -1953,7 +1953,7 @@ rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *app_conf,
         rd_interval_init(&rk->rk_suppress.sparse_connect_random);
         mtx_init(&rk->rk_suppress.sparse_connect_lock, mtx_plain);
 
-        rd_atomic64_init(&rk->rk_ts_last_poll, rd_clock());
+        rd_atomic64_init(&rk->rk_ts_last_poll, INT64_MAX);
 
 	rk->rk_rep = rd_kafka_q_new(rk);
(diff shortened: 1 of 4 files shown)


src/rdkafka.c
```diff
@@ -2609,13 +2609,18 @@ rd_kafka_consume_callback0 (rd_kafka_q_t *rkq, int timeout_ms, int max_cnt,
...
+                rd_kafka_app_poll_blocking(rkq->rkq_rk);
+
+        res = rd_kafka_q_serve(rkq, timeout_ms, max_cnt,
+                               RD_KAFKA_Q_CB_RETURN,
+                               rd_kafka_consume_cb, &ctx);
 
-        return rd_kafka_q_serve(rkq, timeout_ms, max_cnt,
-                                RD_KAFKA_Q_CB_RETURN,
-                                rd_kafka_consume_cb, &ctx);
+        rd_kafka_app_polled(rkq->rkq_rk);
```
rnpridgeon: Wouldn't this call to rd_kafka_app_polled be redundant now that `rd_kafka_q_serve` calls `rd_kafka_app_polled`()?

edenhill: s/q_serve/q_serve_rkmessages/

and yep, you are correct, fixed it :+1:

PR #2370, opened 2019-06-18
Reverse openssl requirement on v1.0.2

edenhill:
Ubuntu 14.04 is still on v1.0.1


CONFIGURATION.md
```diff
@@ -71,7 +71,7 @@ ssl.crl.location                         |  *  |                 |
...
 ssl.keystore.password                    |  *  |                 |               | low        | Client's keystore (PKCS#12) password. <br>*Type: string*
 enable.ssl.certificate.verification      |  *  | true, false     |          true | low        | Enable OpenSSL's builtin broker (server) certificate verification. This verification can be extended by the application by implementing a certificate_verify_cb. <br>*Type: boolean*
-ssl.endpoint.identification.algorithm    |  *  | none, https     |          none | low        | Endpoint identification algorithm to validate broker hostname using broker certificate. https - Server (broker) hostname verification as specified in RFC2818. none - No endpoint verification. <br>*Type: enum value*
+ssl.endpoint.identification.algorithm    |  *  | none, https     |          none | low        | Endpoint identification algorithm to validate broker hostname using broker certificate. https - Server (broker) hostname verification as specified in RFC2818. none - No endpoint verification. OpenSSL >= 1.0.2 required. <br>*Type: enum value*
```
rnpridgeon: NIT: Does it make sense to reference the specific section of RFC2818 , 3.1, rather than the RFC as a whole since the only applicable part of that RFC is the verification process?

edenhill: Good point, but nah, noone is going to read it anyway.

PR #2372, opened 2019-06-19
tests: added quick mode (-Q) to allow running integration tests on CI

edenhill:


rnpridgeon:
Is it worth making note of the admin client chages in the commit message? Other than that lgtm

PR #2377, opened 2019-06-23
Consumer transaction support

mhowlett:
same as #2352 - opening again as github isn't updating the PR to reflect rebase to master / conflict resolution / commit squash.


src/rdkafka_broker.c
```diff
@@ -3614,20 +3650,125 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                         */
+
+                                        for (k = 0 ; k < AbortedTxnCnt; k++) {
+                                                int64_t pid;
+                                                int64_t first_offset;
+                                                int64_t *v;
+                                                rd_kafka_aborted_txn_start_offsets_t node, *node_ptr;
+
+                                                rd_kafka_buf_read_i64(rkbuf, &pid);
+                                                rd_kafka_buf_read_i64(rkbuf, &first_offset);
```
edenhill: the other code in this function uses CamelCase for protocol field names, suggest doing the same here: FirstOffset

src/rdkafka_broker.c
```diff
@@ -3614,20 +3650,125 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                        /* 
+                                         * sort each PID txn list
+                                         *
+                                         * note: aborted_txn_list is used as a queue to traverse the avl tree.
+                                         * it's also retained to facilitate destroying it.
+                                         */
+                                        aborted_txn_list_idx = 0;
+                                        rd_list_init(&aborted_txn_list, 0, NULL);
+                                        rd_list_add(&aborted_txn_list, aborted_txn_offsets->ravl_root);
+                                        while (rd_list_cnt(&aborted_txn_list) > (int)aborted_txn_list_idx) {
```
edenhill: Why are you traversing the aborted transactions again here?
Can't this txn_list be built in the parse loop above? In `if (!node_ptr) {...`

src/rdkafka_msgset.h
```diff
@@ -29,6 +29,13 @@
 #ifndef _RDKAFKA_MSGSET_H_
 #define _RDKAFKA_MSGSET_H_
 
+typedef struct rd_kafka_aborted_txn_start_offsets_s {
+        rd_avl_node_t avl_node;
+        int64_t pid;
+        int32_t offsets_idx;
```
edenhill: should be int as it points to an rd_list index.

src/rdkafka_msgset.h
```diff
@@ -29,6 +29,13 @@
 #ifndef _RDKAFKA_MSGSET_H_
 #define _RDKAFKA_MSGSET_H_
 
+typedef struct rd_kafka_aborted_txn_start_offsets_s {
```
edenhill: Add `@struct doc-string`

src/rdkafka_msgset_reader.c
```diff
@@ -403,6 +411,7 @@ rd_kafka_msgset_reader_decompress (rd_kafka_msgset_reader_t *msetr,
                                             rkbufz,
                                             msetr->msetr_rktp,
                                             msetr->msetr_tver,
+                                            NULL,
```
edenhill: Add a comment saying that there are no aborted transaction support for MsgVersion < 2

src/rdkafka_msgset_reader.c
```diff
@@ -692,8 +701,129 @@ rd_kafka_msgset_reader_msg_v2 (rd_kafka_msgset_reader_t *msetr) {
...
 
-        rd_kafka_buf_read_bytes_varint(rkbuf, &hdr.Key);
+        /* Handle control messages */
+        if (msetr->msetr_v2_hdr->Attributes & RD_KAFKA_MSGSET_V2_ATTR_CONTROL) {
+                struct {
+                        int64_t KeySize;
+                        int16_t Version;
+                        int16_t Type;
+                } ctrl_data;
+                rd_kafka_aborted_txn_start_offsets_t node, *node_ptr;
```
edenhill: `node` is the name of the avl element, which is an implementation detail, this variable should be named something pertaining to txn_start_offsets_t  (same comment for broker.c)

src/rdkafka_msgset_reader.c
```diff
@@ -692,8 +701,129 @@ rd_kafka_msgset_reader_msg_v2 (rd_kafka_msgset_reader_t *msetr) {
...
+                                "unsupported version %"PRId16
+                                " at offset %"PRId64,
+                                rktp->rktp_rkt->rkt_topic->str,
+                                rktp->rktp_partition,
+                                ctrl_data.Version, hdr.Offset);
+                        rd_kafka_buf_skip_to(rkbuf, message_end);
+                        return RD_KAFKA_RESP_ERR_NO_ERROR; /* Continue with next msg */
+                }
+
+                if (unlikely(ctrl_data.KeySize != 4)) {
```
edenhill: drop braces for single-statement blocks.

src/rdkafka_msgset_reader.c
```diff
@@ -692,8 +701,129 @@ rd_kafka_msgset_reader_msg_v2 (rd_kafka_msgset_reader_t *msetr) {
...
 
+                case RD_KAFKA_CTRL_MSG_ABORT:
+                        if (msetr->msetr_rkb->rkb_rk->rk_conf.isolation_level !=
+                                        RD_KAFKA_READ_COMMITTED)
+                                break;
+
+                        if (unlikely(!msetr->msetr_aborted_txn_offsets))
+                                goto unexpected_abort_txn;
+
+                        node.pid = msetr->msetr_v2_hdr->PID;
```
edenhill: This logic should be abstracted to `rd_kafka_something_abortd_txn.._find(.., pid);`

src/rdkafka_msgset_reader.c
```diff
@@ -692,8 +701,129 @@ rd_kafka_msgset_reader_msg_v2 (rd_kafka_msgset_reader_t *msetr) {
...
+                                        " before or at %"PRId64,
+                                        rktp->rktp_rkt->rkt_topic->str,
+                                        rktp->rktp_partition,
+                                        hdr.Offset, abort_start_offset);
+                                break;
+                        }
+
+                        /* This marks the end of this (aborted) transaction,
+                         * advance to next aborted transaction in list */
+                        node_ptr->offsets_idx += 1;
```
edenhill: `foo++;`

src/rdkafka_broker.c
```diff
@@ -3605,6 +3639,8 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
 			struct rd_kafka_toppar_ver *tver, tver_skel;
                         rd_kafka_toppar_t *rktp;
                         shptr_rd_kafka_toppar_t *s_rktp = NULL;
+                        rd_avl_t *aborted_txn_offsets = NULL;
+                        rd_list_t aborted_txn_list;
```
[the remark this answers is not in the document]
mhowlett: keeping the list around is only used as an optimization in the destroy method, which is relatively local to here. it's not passed into rd_kafka_msgset_parse  (it shouldn't be either). i'm not sure what functionality you want comparmentalized and unit tested. you're right, it's slightly messy, but I don't see a big enough advantage to encapsulating it up. my original implementation here avoided this complexity but was less efficient because it also created the list in the destroy method as well. that's actually my (slight) preference because it makes the code simpler to read, and the perf. downside is hit only infrequently and wouldn't be of practical consequence, but either or...

src/rdkafka.h
```diff
@@ -2904,13 +2904,13 @@ void rd_kafka_queue_cb_event_enable (rd_kafka_queue_t *rkqu,
...
-#define RD_KAFKA_OFFSET_STORED -1000  /**< Start consuming from offset retrieved
-				       *   from offset store */
-#define RD_KAFKA_OFFSET_INVALID -1001 /**< Invalid offset */
+#define RD_KAFKA_OFFSET_BEGINNING -2   /**< Start consuming from beginning of
+					*   kafka partition queue: oldest msg */
+#define RD_KAFKA_OFFSET_END       -1   /**< Start consuming from end of kafka
+					*   partition queue: next msg */
+#define RD_KAFKA_OFFSET_STORED -1000   /**< Start consuming from offset retrieved
+					*   from offset store */
+#define RD_KAFKA_OFFSET_INVALID -1001  /**< Invalid offset */
```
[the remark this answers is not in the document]
mhowlett: whoops, hang over from when there was something else here.

src/rdkafka_broker.c
mhowlett: ok, I didn't fully appreciate that convention. Will also change pid to Pid.

src/rdkafka_broker.c
```diff
@@ -3614,20 +3650,125 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                                rd_kafka_buf_read_i64(rkbuf, &pid);
+                                                rd_kafka_buf_read_i64(rkbuf, &first_offset);
+
+                                                node.pid = pid;
+                                                node_ptr = RD_AVL_FIND(aborted_txn_offsets, &node);
+                                                if (!node_ptr) {
+                                                        node_ptr = rd_malloc(sizeof(*node_ptr));
+                                                        node_ptr->pid = pid;
+                                                        node_ptr->offsets_idx = 0;
+                                                        rd_list_init(&node_ptr->offsets, 0, NULL);
```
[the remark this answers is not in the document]
mhowlett: that is a very good point, adds much convenience.

src/rdkafka_broker.c
```diff
@@ -3614,20 +3650,125 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                                }
+
+                                                v = rd_list_add(&node_ptr->offsets, NULL);
+                                                *v = first_offset;
+                                        }
+
+                                        /* 
+                                         * sort each PID txn list
+                                         *
+                                         * note: aborted_txn_list is used as a queue to traverse the avl tree.
```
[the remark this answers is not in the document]
mhowlett: put it in `rd_kafka_msgset_reader_s`

src/rdkafka_broker.c
mhowlett: each list in the PID tree needs sorting. elements could alternatively be placed in the individual lists in order, but that seems less efficient / ideal AFAICT.

the txn_list was originally used only as a utility to enable the breadth first traversal of the tree, being disposed immediately after. I keep it around now on your previous suggestion that this could be reused in the AVL tree destroy, where visiting all the nodes is again needed in order to destroy the lists. if the AVL tree implementation had a free_cb like list this list wouldn't be required at all. but it doesn't.

mhowlett: oh and yes, this list can be built in the above loop (great idea! much better!), but the traversal is still needed for the sorting.

src/rdkafka_msgset.h
mhowlett: now you point it out, I'm thinking `size_t` is probably most ideal here, but `rd_list_elem` takes an `int` so that is most appropriate, yes.

src/rdkafka_msgset_reader.c
mhowlett: i've changed this (and in broker.c), though I feel it is harder to follow now with the extra characters and also that the word node is more informative/useful, because it matches a key aspect of the algorithm, which the reader will be trying to follow when reading this. now this information is obfuscated in favor of something that provides less value.

mhowlett: sorry I've changed this back to node again because i think it's making things worse (away from txn_start_offsets / _ptr). If you want this changed, tell me precisely what name you want and I'll change it.

src/rdkafka_msgset_reader.c
mhowlett: sounds good, this is messy enough that that would add value.

src/rdkafka_broker.c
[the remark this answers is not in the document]
mhowlett: yes, great idea, this cleans things up a lot.

src/rdkafka_broker.c
```diff
@@ -3614,20 +3650,125 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                if (AbortedTxnCnt > 0) {
+                                        size_t aborted_txn_list_idx;
+
+                                        if (unlikely(rkb->rkb_rk->rk_conf.isolation_level ==
+                                            RD_KAFKA_READ_UNCOMMITTED))
+                                                rd_rkb_log(rkb, LOG_WARNING, "FETCH",
+                                                        "%.*s [%"PRId32"]: "
+                                                        "%"PRId32" aborted transaction(s) "
+                                                        "encountered in READ_UNCOMMITTED "
+                                                        "fetch response: ignoring",
```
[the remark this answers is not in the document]
mhowlett: i've now put this check after the aborted transactions are parsed, changed the warning to an error and destroy the aborted transactions data in the if block. reasoning: this case should never be able to happen in practice. if it does (for some reason that should never be able to happen), the user wants uncommitted values, so the aborted transactions should be ignored. since this should never happen, I the inefficiency of reading/parsing aborted transactions when they aren't required is a non issue.

src/rdkafka_broker.c
mhowlett: update: this is now all encapsulated up.

src/rdkafka_msgset_reader.c
```diff
@@ -692,8 +741,123 @@ rd_kafka_msgset_reader_msg_v2 (rd_kafka_msgset_reader_t *msetr) {
...
+
+                case RD_KAFKA_CTRL_MSG_ABORT:
+                        if (msetr->msetr_rkb->rkb_rk->rk_conf.isolation_level !=
+                                        RD_KAFKA_READ_COMMITTED)
+                                break;
+
+                        if (unlikely(!msetr->msetr_aborted_txns))
+                                goto unexpected_abort_txn;
+
+                        aborted_txn_start_offset = rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid(
```
edenhill: this naming scheme does not make it easy to fit within 80 columns ;)
I dont think it is necessary to repeated "aborted_txn", that's already in there.
So maybe:
`rd_kafka_aborted_txns_pop_pid(msetr->..., ..PID);`
That tells me it is getting and removing something that has something to do with a PID, I can read the docstring for the function to understand more, or add a comment here.
C you know.

src/rdkafka_msgset_reader.c
```diff
@@ -692,8 +741,123 @@ rd_kafka_msgset_reader_msg_v2 (rd_kafka_msgset_reader_t *msetr) {
...
+
+                        if (unlikely(aborted_txn_start_offset == -1))
+                                goto unexpected_abort_txn;
+
+                        if (unlikely(aborted_txn_start_offset >= hdr.Offset)) {
+                                rd_rkb_log(msetr->msetr_rkb, LOG_ERR, "TXN",
+                                        "%s [%"PRId32"]: "
+                                        "Abort txn ctrl msg bad order "
+                                        "at offset %"PRId64". Expected "
+                                        " before or at %"PRId64,
```
edenhill: Needs to say what we do: do we ignore these messages? do we stop?

src/rdkafka_msgset_reader.c
```diff
@@ -692,8 +741,123 @@ rd_kafka_msgset_reader_msg_v2 (rd_kafka_msgset_reader_t *msetr) {
...
+                                "Received abort txn ctrl msg for "
+                                "unknown txn PID %"PRId64" at "
+                                "offset %"PRId64": ignoring",
+                                rktp->rktp_rkt->rkt_topic->str,
+                                rktp->rktp_partition,
+                                msetr->msetr_v2_hdr->PID, hdr.Offset);
+                        break;
+
+                default:
+                        rd_kafka_buf_parse_fail(rkbuf,
```
edenhill: We should simply ignore unknown control types (since they are not for us), not fail.
Should still do a debug log on it though.

src/rdkafka_msgset_reader.c
```diff
@@ -868,12 +1032,8 @@ rd_kafka_msgset_reader_v2 (rd_kafka_msgset_reader_t *msetr) {
                 goto done;
         }
 
-        /* Ignore control messages */
-        if (unlikely((hdr.Attributes & RD_KAFKA_MSGSET_V2_ATTR_CONTROL))) {
+        if (hdr.Attributes & RD_KAFKA_MSGSET_V2_ATTR_CONTROL)
```
edenhill: To be safe, move this increment as far up as possible so we don't miss the increment in the case we ignore some ctrl type or whatever.

This counter is used to differentiate between an empty/parse-failed data msgset  or a ctrl msgset, which is then used by the backoff code.

src/rdkafka_msgset_reader.c
```diff
@@ -902,6 +1062,29 @@ rd_kafka_msgset_reader_v2 (rd_kafka_msgset_reader_t *msetr) {
...
+                }
+
+                if (msetr->msetr_aborted_txns != NULL &&
+                    msetr->msetr_v2_hdr->Attributes & RD_KAFKA_MSGSET_V2_ATTR_TRANSACTIONAL &&
+                    !(msetr->msetr_v2_hdr->Attributes & RD_KAFKA_MSGSET_V2_ATTR_CONTROL)) {
+
+                        int64_t txn_start_offset = rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid(
+                                msetr->msetr_aborted_txns, msetr->msetr_v2_hdr->PID);
+                        
+                        if (txn_start_offset >= 0 &&
```
edenhill: the contract with for_pid is -1 on error, so use that here:
`if (txn_start_offset != -1 && `

src/rdkafka_msgset_reader.c
```diff
@@ -902,6 +1062,29 @@ rd_kafka_msgset_reader_v2 (rd_kafka_msgset_reader_t *msetr) {
...
+                            msetr->msetr_v2_hdr->BaseOffset >= txn_start_offset) {
+                                /* MessageSet is part of an aborted transaction */
+                                rd_slice_widen(&rkbuf->rkbuf_reader, &save_slice);
```
edenhill: It'd be nice with a debug message here:
"..: skipping %d message(s) in aborted transaction"

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        const rd_kafka_aborted_txn_start_offsets_t *a = _a, *b = _b;
+        return a->pid - b->pid;
+}
+
+
+/**
+ * @brief Free resources associated with an AVL tree node.
+ */
+void rd_kafka_aborted_txn_node_destroy (void *offsets) {
+        rd_list_destroy(&((rd_kafka_aborted_txn_start_offsets_t *)offsets)->offsets);
```
edenhill: This cast is hard to read, assign to a temporary variable like we do in cmp

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+}
+
+
+rd_kafka_aborted_txns_t *
+rd_kafka_aborted_txns_new (int32_t txn_cnt) {
```
edenhill: docstring

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        rd_kafka_aborted_txns_t *aborted_txns;
+        aborted_txns = rd_malloc(sizeof(aborted_txns));
+        aborted_txns->avl = rd_malloc(sizeof(*aborted_txns->avl));
+        rd_avl_init(aborted_txns->avl, rd_kafka_aborted_txn_cmp_by_pid, 0);
+        aborted_txns->list = rd_malloc(sizeof(*aborted_txns->list));
```
edenhill: avl and list should not be allocated but defined as their non-pointer types in the struct and just init()ed here.

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        rd_list_init(aborted_txns->list, 0,
```
edenhill: can we initialize this list to txn_cnt?

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+                rd_kafka_aborted_txn_node_destroy);
+        aborted_txns->cnt = txn_cnt;
+        return aborted_txns;
+}
+
+
+void
+rd_kafka_aborted_txns_destroy (rd_kafka_aborted_txns_t *aborted_txns) {
```
edenhill: docstring

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        rd_list_destroy(aborted_txns->list);
+        rd_avl_destroy(aborted_txns->avl);
+        rd_free(aborted_txns->avl);
+        rd_free(aborted_txns->list);
+        rd_free(aborted_txns);
+}
+
+
+rd_kafka_aborted_txn_start_offsets_t*
```
edenhill: `*` should not stick to the type but the variable/function, in this case, just add a space here.

compare:
`obj_t* a, b;` <-- looks like a and b are same type
to
`obj_t *a, b;` <-- but they're not!

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+rd_kafka_aborted_txns_offsets_for_pid (rd_kafka_aborted_txns_t *aborted_txns,
```
edenhill: Change to _txns_pop_pid() and combine with move_to_next below to avoid two lookups.

src/rdkafka_msgset_reader.c
edenhill: docstring

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid (rd_kafka_aborted_txns_t *aborted_txns,
+                                                       int64_t pid) {
+        rd_kafka_aborted_txn_start_offsets_t node, *node_ptr;
+        node.pid = pid;
+        node_ptr = RD_AVL_FIND(aborted_txns->avl, &node);
+
+        if (node_ptr == NULL)
+                return -1;
+        
+        if (unlikely(node_ptr->offsets_idx >= rd_list_cnt(&node_ptr->offsets)))
```
edenhill: Add a comment for why this happens.

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+                return -1;
+
+        int64_t abort_start_offset =
```
edenhill: all variables needs to be declared at the head of the scope/block.

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        }
+}
+
+
+/**
+ * @brief Unit tests for all functions that operate on
+ * rd_kafka_aborted_txns_t
+ */
+int unittest_aborted_txns (void)
+{
```
edenhill: open brace goes on the funciton line.

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        rd_kafka_aborted_txns_add(aborted_txns, 2, 11);
+        rd_kafka_aborted_txns_add(aborted_txns, 2, 7);
+        rd_kafka_aborted_txns_add(aborted_txns, 1, 3);
+        rd_kafka_aborted_txns_sort(aborted_txns);
+
+        start_offset = rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid(
+                aborted_txns, 1);
+        RD_UT_ASSERT(3 == start_offset,
+                     "queried start offset was %"PRId64", "
+                     "exected 3", start_offset);
```
edenhill: expected

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        rd_kafka_aborted_txns_move_to_next(aborted_txns, 1);
+        rd_kafka_aborted_txns_move_to_next(aborted_txns, 1);
+        rd_kafka_aborted_txns_move_to_next(aborted_txns, 2);
+
+        start_offset = rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid(
+                aborted_txns, 1);
+        RD_UT_ASSERT(-1 == start_offset,
+                     "queried start offset was %"PRId64", "
+                     "exected -1", start_offset);
+        start_offset = rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid(
```
edenhill: add empty line to be consistent with code above.

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1401,230 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+                rd_list_sort(&el->offsets, rd_kafka_offset_cmp);
+        }
+}
+
+
+/**
+ * @brief Unit tests for all functions that operate on
+ * rd_kafka_aborted_txns_t
+ */
+int unittest_aborted_txns (void)
```
edenhill: 👏

tests/testcpp.h
```diff
@@ -96,6 +96,27 @@ namespace Test {
                               (int)randomized);
   }
 
+  /**
+  * @brief Create a topic using the Admin API
```
edenhill: indent

tests/testcpp.h
```diff
@@ -96,6 +96,27 @@ namespace Test {
...
+  static RD_UNUSED void create_topic (RdKafka::Handle *use_handle, const char *topicname,
+                                      int partition_cnt, int replication_factor) {
+    rd_kafka_t *use_rk = NULL;
+    if (use_handle != NULL)
+      use_rk = use_handle->c_ptr();
+    test_create_topic(use_rk, topicname, partition_cnt, replication_factor);
+  }
+
+  /**
+  * @brief Delete a topic using the Admin API
```
edenhill: indent

edenhill: "using the Admin API" is not true, it will fall back to kafka-topics if not available.

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+
+static void test_assert(bool cond, std::string msg) {
+  if (!cond)
+    Test::Say(msg);
+  assert(cond);
+}
+
+
+static void execute_java_produce_cli(std::string &bootstrapServers, std::string &topic, std::string cmd) {
```
edenhill: &gt; 80 cols

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,553 @@
...
+  std::string errstr;
+  std::string topic_name;
+  RdKafka::KafkaConsumer *c;
+  std::vector<RdKafka::Message *> msgs;
+
+  std::string bootstrap_servers = get_bootstrap_servers();
+  Test::Say("bootstrap.servers: " + bootstrap_servers);
+
+
+  // Test 0
```
edenhill: These numbers are not that helpful, it'd be good if there was a short comment saying what was being tested for each one.

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+  topic_name = Test::mk_topic_name("0098-consumer_txn-0", 1);
+  c = create_consumer(topic_name, "READ_COMMITTED");
+  Test::create_topic(c, topic_name.c_str(), 1, 3);
+
+  execute_java_produce_cli(bootstrap_servers, topic_name, "0");
+
+  msgs = consume_messages(c, topic_name, 0);
+  test_assert(msgs.size() == 5, tostr() << "Consumed unexpected number of messages. Expected 5, got: " << msgs.size());
+  test_assert(0 == msgs[0]->key()->c_str()[0], "Unexpected key");
```
edenhill: Check key_length >= 1 before dereferencing.

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,553 @@
...
+  test_assert(4 == msgs[4]->key()->c_str()[0], "Unexpected key");
+  delete_messages(msgs);
+  c->close();
+  delete c;
+
+  c = create_consumer(topic_name, "READ_UNCOMMITTED");
+  msgs = consume_messages(c, topic_name, 0);
+  test_assert(msgs.size() == 10, tostr() << "Consumed unexpected number of messages. Expected 10, got: " << msgs.size());
+  test_assert(0 == msgs[0]->key()->c_str()[0], "Unexpected key");
```
edenhill: check key length

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,553 @@
...
+  Test::Say("Consuming from topic " + topic + "\n");  
+  std::vector<RdKafka::Message *> result = std::vector<RdKafka::Message *>();
+
+  while (true) {
+    RdKafka::Message *msg = c->consume(tmout_multip(1000));
+    switch (msg->err())
+    {
+      case RdKafka::ERR__TIMED_OUT:
+        continue;
+      case RdKafka::ERR__PARTITION_EOF:
```
edenhill: I'd like to see a test for PARTITION_EOF in two modes where the LSO is < HWMARK, one for committed and one for uncommitted, to verify that it behaves according to the isolation.level.

edenhill:
There's a couple of compilation warnings on the Appveyor builders that needs to be addressed.

src/rdkafka_broker.c
```diff
@@ -3612,20 +3614,62 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                        for (k = 0 ; k < AbortedTxnCnt; k++) {
+                                                int64_t Pid;
+                                                int64_t FirstOffset;
+                                                rd_kafka_buf_read_i64(rkbuf, &Pid);
+                                                rd_kafka_buf_read_i64(rkbuf, &FirstOffset);
+                                                rd_kafka_aborted_txns_add(aborted_txns, Pid, FirstOffset);
+                                        }
+                                        rd_kafka_aborted_txns_sort(aborted_txns);
+
+                                        if (unlikely(rkb->rkb_rk->rk_conf.isolation_level ==
```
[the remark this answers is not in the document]
mhowlett: makes sense. tbh the only reason I have it as is over your suggestion is I was too lazy to research how to do the skip.

src/rdkafka_conf.c
```diff
@@ -982,6 +982,19 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
                         { RD_KAFKA_OFFSET_METHOD_BROKER, "broker" }
                 }
         },
+        { _RK_GLOBAL|_RK_CONSUMER|_RK_MED, "isolation.level",
+          _RK_C_S2I,
+          _RK(isolation_level),
+          "Controls how to read messages written transactionally: "
```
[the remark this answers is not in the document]
mhowlett: During testing, I observed that messages produced transactionally (and aborted) do not necessarily make it to the log. Assuming i'm not delusional, that makes 'written' more correct than 'produced', as sometimes messages produced transactionally won't make it through.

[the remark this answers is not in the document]
mhowlett: the apache docs use the word written, though are ambiguous about my observation above.

src/rdkafka_conf.c
```diff
@@ -982,6 +982,19 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
+          "`read_committed` - only return transactional messages which have "
+          "been committed. `read_uncommitted` - return all messages, even "
+          "messages for aborted and non-committed transactions.",
```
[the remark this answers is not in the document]
mhowlett: changed to match the wording of the apache docs for this config parameter.

src/rdkafka_msgset.h
```diff
@@ -30,6 +30,74 @@
...
 
 
+
+/** 
+ * @struct rd_kafka_aborted_txns_t
+ * 
+ * @brief A collection of aborted transactions.
+ */
+typedef struct rd_kafka_aborted_txns_s {
+    rd_avl_t *avl;
```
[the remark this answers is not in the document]
mhowlett: i do find the combination of an 8 ws indent and 80 char line limit coding convention a bit comical

src/rdkafka_msgset.h
```diff
@@ -30,6 +30,74 @@
...
+rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid (rd_kafka_aborted_txns_t *aborted_txns,
+                                                         int64_t pid);
+
+
+/** 
+ * @brief Move to the next aborted transaction start
+ * offset for the specified pid.
+ */
+void
+rd_kafka_aborted_txns_move_to_next(rd_kafka_aborted_txns_t *aborted_txns,
```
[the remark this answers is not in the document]
mhowlett: this is fighting muscle memory. did a pass through changing all of these, but missed one.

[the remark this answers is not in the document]
mhowlett: or three

src/rdkafka_msgset.h
```diff
@@ -30,6 +30,74 @@
...
+rd_kafka_aborted_txns_destroy (rd_kafka_aborted_txns_t *aborted_txns);
+
+
+/** 
+ * @brief Get the next aborted transaction start
+ * offset for the specified pid. Returns -1 if
+ * there is none.
+ */
+int64_t
+rd_kafka_aborted_txns_next_aborted_txn_offset_for_pid (rd_kafka_aborted_txns_t *aborted_txns,
```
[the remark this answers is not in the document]
mhowlett: cool. i did not realize that resulted in a compiler warning if the pointed to value changes.

src/rdkafka_msgset.h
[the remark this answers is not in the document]
mhowlett: the lookup without move_next is used in rd_kafka_msgset_reader_v2, so we need both. i don't think there's a good argument for efficiency here in spite of that as aborted transactions are not the normal case.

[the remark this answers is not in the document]
mhowlett: what about `rd_kafka_aborted_txns_get_offset`?

src/rdkafka_msgset_reader.c
mhowlett: ok. except i think we want get and move still.

src/rdkafka_msgset_reader.c
mhowlett: ignore.
ok.

src/rdkafka_msgset_reader.c
mhowlett: yep, good point

src/rdkafka_msgset_reader.c
mhowlett: don't necessarily agree it's better but ok.

src/rdkafka_broker.c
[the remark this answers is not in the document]
edenhill: The original code did a skip, look at the diff.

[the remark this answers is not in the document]
edenhill: right there in the diff:
https://github.com/edenhill/librdkafka/pull/2377/files#diff-214080cd229adac5b1596529b6eecfe4L3627

src/rdkafka_msgset.h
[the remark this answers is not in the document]
edenhill: It is basically the Linux kernel style, the idea is to keep down the level of nesting.

src/rdkafka_msgset.h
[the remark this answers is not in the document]
edenhill: > . i don't think there's a good argument for efficiency here in spite of that as aborted transactions are not the normal case.

We did the worst-case napkin calculations, and there can be a large number of aborted transactions in a MessageSet - while it might not be normal we still need to handle it properly.

If we can easily avoid two lookups, we should.
And in this case there is just a single error check (`if (start_offset > hdr.offset)`) between the get and the pop, and that error check  may very well be done after the pop.

Every cycle counts

src/rdkafka_msgset_reader.c
edenhill: it is a nit, do as you please :)

src/rdkafka_broker.c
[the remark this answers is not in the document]
mhowlett: yes - more precisely a combination of laziness and poor memory.

src/rdkafka_msgset.h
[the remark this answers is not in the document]
mhowlett: yes, you are completely right, i was being delusional.

src/rdkafka_msgset_reader.c
mhowlett: also changed the offsets parameter name to `_node_ptr`

src/rdkafka_msgset_reader.c
mhowlett: this function is documented in the header. i think bad to duplicate that, and better this be in the header. the function above is an implementation detail, so documented in the .c.

src/rdkafka_msgset_reader.c
mhowlett: why would I have done this? need a vacation.

src/rdkafka_msgset_reader.c
mhowlett: yep. it would also then be possible to do away with storing the count separately in rd_kafka_aborted_txns_t, though i've left as is as i think that's semantically better (and I think getting at the list capacity isn't part of the public api).

src/rdkafka_msgset_reader.c
mhowlett: header

src/rdkafka_msgset_reader.c
mhowlett: ahh, i see now that's going against the convention. will change. seems counter intuitive to put documentation in the bit that's supposed to be private and not put it in the bit that's supposed to be public though.

src/rdkafka_msgset_reader.c
mhowlett: moving to implementation

src/rdkafka_msgset_reader.c
mhowlett: i've made `rd_kafka_aborted_txns_get_offset` and `rd_kafka_aborted_txns_pop_offset` which call directly through to a common implementation called `rd_kafka_aborted_txns_next_offset`.

src/rdkafka_msgset_reader.c
mhowlett: i feel that is over-commenting.

tests/0098-consumer-txn.cpp
mhowlett: ouch. LoC of this PR is now at a much more impressive level.

tests/0098-consumer-txn.cpp
mhowlett: these numbers are very helpful in linking code between the java generator (where there is a short comment saying what is being tested in each one), and here. that is outlined in the comment at the top of the file. I will duplicate the comments from the java file here.

tests/0098-consumer-txn.cpp
mhowlett: yep, good plan.

src/rdkafka_msgset_reader.c
mhowlett: changed to get and pop

src/rdkafka_msgset_reader.c
mhowlett: the behavior is kind of undefined. I've changed it to say that messages in aborted transactions may be delivered to the application (which seems likely).

src/rdkafka_msgset_reader.c
edenhill: So for the internal API what I do is document the function at its definition (in the .c file), close to the implementation to make sure the docs stay up to date,
while in the public API the functions are documented in the public rdkafka.h file (because that's what people see).

src/rdkafka_msgset_reader.c
edenhill: Make static if only used in this file.

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1407,284 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+/** 
+ * @brief Get the next aborted transaction start
+ * offset for the specified pid.
+ * 
+ * @param increment_idx it true, the offset index will be incremented.
+ * 
+ * @returns the start offset or -1 if there is none.
+ */
+int64_t
+rd_kafka_aborted_txns_next_offset (rd_kafka_aborted_txns_t *aborted_txns,
```
edenhill: make static

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1407,284 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+        node.pid = pid;
+        return RD_AVL_FIND(&aborted_txns->avl, &node);
+}
+
+
+/** 
+ * @brief Get the next aborted transaction start
+ * offset for the specified pid.
+ * 
+ * @param increment_idx it true, the offset index will be incremented.
```
edenhill: if true

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,725 @@
...
+  Test::delete_topic(c, topic_name.c_str());
+
+  c->close();
+  delete c;
+}
+
+extern "C" {
+  int main_0098_consumer_txn (int argc, char **argv) {
+    if (test_quick)
+      return 0;
```
edenhill: `Test::Skip("Test skipped due to quick mode\n");`

tests/0098-consumer-txn.cpp
mhowlett: note that every test uses the `consume_messages` function which relies on partition eof events to know when to finish and every test is run for each isolation mode. I believe the scenario you suggest should already be covered, with the caveat that i observed producers don't necessarily send aborted transactions, though i recall making a scenario where they do. will check.

mhowlett: good call, this is broken.

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1407,284 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
 }
+
+
+/**
+ * @brief Offset comparator
+ */
+static int rd_kafka_offset_cmp (const void *_a, const void *_b) {
+        const int64_t *a = _a, *b = _b;
+        const int64_t diff = *a - *b;
+        return diff < 0 ? -1 : (diff > 0 ? 1 : 0);
```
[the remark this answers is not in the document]
mhowlett: integer overflow - can you think of a better way?

src/rdkafka_msgset_reader.c
```diff
@@ -1216,3 +1407,284 @@ rd_kafka_msgset_parse (rd_kafka_buf_t *rkbuf,
...
+}
+
+
+/**
+ * @brief Pid comparator for rd_kafka_aborted_txn_start_offsets_t
+ */
+static int rd_kafka_aborted_txn_cmp_by_pid (const void *_a, const void *_b) {
+        const rd_kafka_aborted_txn_start_offsets_t *a = _a, *b = _b;
+        const int64_t diff = a->pid - b->pid;
+        return diff < 0 ? -1 : (diff > 0 ? 1 : 0);
```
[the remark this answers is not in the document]
mhowlett: dito

tests/0098-consumer-txn.cpp
mhowlett: also added a check for ls_offset and hi_offset in the stats. the code is a bit less than ideal, but you can feel free to not care about that.

src/rdkafka_msgset_reader.c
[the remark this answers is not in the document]
edenhill: compare2 [here](https://stackoverflow.com/a/10997428/1821055) is more efficient.

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,851 @@
...
+ */
+
+
+/**
+ * @brief extract a single value from a json file immediately following the
+ * specified nested field sequence.
+ *
+ * @returns -1 if no such value exists.
+ */
+static int64_t extract_json_value(std::string json, std::vector<std::string> &fields) {
```
edenhill: use rapidjson (if available, else skip this test), see https://github.com/edenhill/librdkafka/blob/master/tests/0053-stats_cb.cpp#L46

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,851 @@
...
+  try {
+    return std::stol(json.substr(pos1, pos2-pos1));
+  }
+  catch (...) {
+    return -1;
+  }
+}
+
+
+static bool _should_capture_stats;
```
edenhill: keep these in your TestEventCb class

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,851 @@
...
+        break;
+    }
+    break;
+  }
+
+  Test::Say("Read all messages from topic: " + topic + "\n");
+
+  _should_capture_stats = true;
+  limit_count = 0;
+  while (limit_count++ < 20) {
```
edenhill: Where's the value 20 coming from?

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,851 @@
...
+}
+
+extern "C" {
+  int main_0098_consumer_txn (int argc, char **argv) {
+    if (test_quick) {
+      Test::Skip("Test skipped due to quick mode\n");
+      return 0;
+    }
+    
+    do_test_consumer_txn_test();
```
edenhill: Instead of skipping this entire test suite, it is better if you cut down on the number of tests but still run the most important ones.

mhowlett: yes, but that means i need to work out how to get the java application built prior to the tests being run, and I was avoiding that effort.

tests/0098-consumer-txn.cpp
mhowlett: yes, this is stupid, i just need to block the consume longer than the stats interval.

tests/0098-consumer-txn.cpp
mhowlett: makes sense

tests/0098-consumer-txn.cpp
edenhill: Isn't that just a `make -C tests/java`?

tests/0098-consumer-txn.cpp
edenhill: Then check time and not iterations, if consume() wakes up from other events it makes the test unreliable.

You can even rely on the test timeout to take care of it for you, just wait indefinitely for the break event, the test timeout will kill the test if it runs for longer than the timeout you set in test_conf_init

tests/0098-consumer-txn.cpp
edenhill: I'll add rapidjson to the travis installer so it is available for the integration test runner (but not the other ones).

tests/0098-consumer-txn.cpp
mhowlett: A kafka distribution is needed for the java client dependency - that doesn't seemed to be something that's installed already, so I'm still skipping these tests in test_quick mode still

tests/0098-consumer-txn.cpp
mhowlett: using # iterations is dead simple and didn't involve me doing an additional 2 minutes of irksome investigation into how to check time in C, and/or if there is some other idiomatic way you do it in librdkafka (did i mention i need a vacation?).

anyway, relying on test timeout is a great idea, going with that.

tests/0098-consumer-txn.cpp
mhowlett: done - took me longer to do that btw and is about the same number of lines code than the existing solution. but point taken.

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,892 @@
...
+  }
+
+  Test::Say("Read all messages from topic: " + topic + "\n");
+
+  TestEventCb::should_capture_stats = true;
+
+  /* rely on the test timeout to prevent an infinite loop in
+   * the (unlikely) event that the statistics callback isn't
+   * called. */
+  while (true) {
```
edenhill: nit: `while (!TestEventCb::has_captured_stats)`

tests/0098-consumer-txn.cpp
```diff
@@ -0,0 +1,892 @@
...
+    c->consume(tmout_multip(500));
```
edenhill: this is leaking memory if a message or consumer error (which is a Message with err set) is returned.

mhowlett: there are also memory leaks above.

fmd:
Is there any likely movement on this @mhowlett, @edenhill? It seems very close

tests/java/TransactionProducerCli.java
```diff
@@ -216,6 +216,7 @@ public static void main (String[] args) throws Exception {
             case "6":
                 makeTestMessages(producer3, topic, 0, 0x10, 1, TransactionType.None, FlushType.Yes);
                 makeTestMessages(producer1, topic, 0, 0x20, 3, TransactionType.BeginOpen, FlushType.Yes);
+                System.exit(0);
```
mhowlett: for versions > 2.3.0, `producer.close` apparently auto-aborts the transactions, whereas in earlier versions this is not the case.

edenhill: Maybe add a comment saying we dont want the transactions to be aborted

edenhill:
Merged!  Great work on this @mhowlett !

maks-rafalko:
Do you have an idea when it can be released? Was waiting exactly for this PR ;)

Great work guys!

fmd:
Amazing!

PR #2378, opened 2019-06-24
Fixed socket recv error handling for MSVC build

binary85:
Added socket recv error handling for MSVC build.
There's no returns when recv returns SOCKET_ERROR and error code is not WSAEWOULDBLOCK.
It causes crashes(calls rd_buf_write with size -1).

### DIFF
--- a/src/rdkafka_transport.c
+++ b/src/rdkafka_transport.c
@@ -307,26 +307,21 @@ rd_kafka_transport_socket_recv0 (rd_kafka_transport_t *rktrans,
                          0);
 
                 if (unlikely(r == SOCKET_ERROR)) {
+                        int errno_save = socket_errno;
+                        if (errno_save == EAGAIN
 #ifdef _MSC_VER
-                        int errno_save = WSAGetLastError();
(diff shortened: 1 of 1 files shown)


src/rdkafka_transport.c
```diff
@@ -308,10 +308,14 @@ rd_kafka_transport_socket_recv0 (rd_kafka_transport_t *rktrans,
 
                 if (unlikely(r == SOCKET_ERROR)) {
 #ifdef _MSC_VER
-                        if (WSAGetLastError() == WSAEWOULDBLOCK)
+                        int errno_save = WSAGetLastError();
```
edenhill: How about consolidating the win and non-win code?
Something like this:
```
     int errno_save = socket_errno;
     if (errno_save == EAGAIN
#ifdef _MSC_VER
        || errno_save == WSAEWOULDBLOCK
#endif 
        )
          return sum;

      rd_snprintf(errstr, errstr_size, "%s", socket_strerror(errno_save));
      return -1;
```

binary85: That's nice.

edenhill:
Thank you!

Issue #2373, opened 2019-06-20
Destroy toppar even when retry OffsetRequest


binary85:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
Hello. 
I'm tracing crashes due to accessing freed rktp, and found a suspicious routine.

At `rd_kafak_partition.c::rd_kafka_toppar_handle_Offset`,
when retries request in [`rd_kafka_handle_Offset`](https://github.com/edenhill/librdkafka/blob/8681f884020e880a4c6cda3cfc672f0669e1f38e/src/rdkafka_partition.c#L1281), returns `RD_KAFKA_RESP_ERR__IN_PROGRESS` error. and returns without calling `rd_kafka_toppar_destroy`[(line 1331)](https://github.com/edenhill/librdkafka/blob/8681f884020e880a4c6cda3cfc672f0669e1f38e/src/rdkafka_partition.c#L1331).
But after `rd_kafka_handle_Offset` call, err can be updated to `RD_KAFKA_RESP_ERR__OUTDATED` when op version is outdated[(line 1292)](https://github.com/edenhill/librdkafka/blob/8681f884020e880a4c6cda3cfc672f0669e1f38e/src/rdkafka_partition.c#L1292). In this case, calls `rd_kafka_toppar_destroy`.
And when response handling for retried req, will call `rd_kafka_toppar_destroy` again.
At that moment, rktp can be deleted unintentionally.

After modifying routine for handling RD_KAFKA_RESP_ERR__IN_PROGRESS err, crash no longer occurred.

How to reproduce
================
Consuming messages and restart broker repeatedly.
(when partition count of topic is huge(>300), crashes more oftenly.)

**IMPORTANT**: Always try to reproduce the issue on the latest released version (see https://github.com/edenhill/librdkafka/releases), if it can't be reproduced on the latest version the issue has been fixed.

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `<1.0.1>`
 - [x] Apache Kafka version: `<2.12-2.2.1>`
 - [ ] librdkafka client configuration: 
 - [x] Operating system: `<windows server 2012 R2>`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [x] Critical issue

PR #2379, opened 2019-06-24
Prevents destroy toppar when retry OffsetRequest

binary85:
Fix for #2373.

Call `rd_kafka_handle_Offset` after checking op version and only when err is not `RD_KAFKA_RESP_ERR__OUTDATED`.

### DIFF
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -1291,11 +1291,11 @@ static void rd_kafka_toppar_handle_Offset (rd_kafka_t *rk,
 		    err = RD_KAFKA_RESP_ERR__OUTDATED;
 	}
 
-    if (err != RD_KAFKA_RESP_ERR__OUTDATED) {
-        /* Parse and return Offset */
-        err = rd_kafka_handle_Offset(rkb->rkb_rk, rkb, err,
-                                     rkbuf, request, offsets);
-    }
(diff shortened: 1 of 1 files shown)


src/rdkafka_partition.c
```diff
@@ -1295,6 +1291,12 @@ static void rd_kafka_toppar_handle_Offset (rd_kafka_t *rk,
 		    err = RD_KAFKA_RESP_ERR__OUTDATED;
 	}
 
+    if (err != RD_KAFKA_RESP_ERR__OUTDATED) {
```
edenhill: use 8 whitespaces for indent

binary85: Changed! :)

src/rdkafka_partition.c
```diff
@@ -1295,6 +1291,12 @@ static void rd_kafka_toppar_handle_Offset (rd_kafka_t *rk,
 		    err = RD_KAFKA_RESP_ERR__OUTDATED;
 	}
 
+        if (err != RD_KAFKA_RESP_ERR__OUTDATED) {
+            /* Parse and return Offset */
```
edenhill: each level of indent shall be 8 whitespaces

binary85: ok! changed it.

edenhill:
Thank you!

PR #2386, opened 2019-07-01
Fix message timeout check

jspdt0:
This check was added in the 1.0 release, but it ignores the special value of `message_timeout_ms = 0`, which indicates an infinite timeout. This patch makes the timeout check treat the special value appropriately.


edenhill:
Thank you!

PR #2393, opened 2019-07-04
Fill metadata in rd_kafka_committed & rd_kafka_offsets_for_times.

damour:
Currently rd_kafka_committed returns list with empty  metadata:

Выделение_021

### DIFF
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -3410,6 +3410,11 @@ rd_kafka_topic_partition_list_update (rd_kafka_topic_partition_list_t *dst,
 
                 d->offset = s->offset;
                 d->err    = s->err;
+                if (d->metadata) {
+                        rd_free(d->metadata);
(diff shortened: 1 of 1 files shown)


src/rdkafka_partition.c
```diff
@@ -3408,6 +3410,13 @@ rd_kafka_topic_partition_list_update (rd_kafka_topic_partition_list_t *dst,
 
                 d->offset = s->offset;
                 d->err    = s->err;
+                if (s->metadata_size > 0) {
```
edenhill: This will leak memory if the `d->metadata` is set. Above this line, add:
```
if (d->metadata) {
        rd_free(d->metadata);
        d->metadata = NULL;
        d->metadata_size = 0;
}
```

damour: Done, thanks.

edenhill:
Thank you!

PR #2396, opened 2019-07-08
 Add on_thread_start and on_thread_exit interceptors

edenhill:
CLIENTS-998


tests/0099-commit_metadata.c
```diff
@@ -0,0 +1,193 @@
...
+        rd_kafka_topic_partition_list_add(origin_toppar, topic, 0);
+
+        expected_toppar = rd_kafka_topic_partition_list_copy(origin_toppar);
+
+        expected_toppar->elems[0].offset = 42;
+        expected_toppar->elems[0].metadata = rd_strdup("Hello world!");
+        expected_toppar->elems[0].metadata_size =
+                strlen(expected_toppar->elems[0].metadata);
+
+        get_committed_metadata(group_id, origin_toppar, origin_toppar);
```
mhowlett: the logic here is convoluted (having the comparison assert inside the `get_committed_metadata` method, but it's just a test so whatever floats your boat)

mhowlett: also nice test, but i don't understand why it's in this PR

PR #2398, opened 2019-07-08
 Add support for sub-millisecond linger.ms

edenhill:
CLIENTS-998


CONFIGURATION.md
```diff
@@ -115,8 +115,8 @@ enable.idempotence                       |  P  | true, false     |         false
...
+queue.buffering.max.ms                   |  P  | 0 .. 900000     |           0.5 | high       | Delay in milliseconds to wait for messages in the producer queue to accumulate before constructing message batches (MessageSets) to transmit to brokers. A higher value allows larger and more effective (less overhead, improved compression) batches of messages to accumulate at the expense of increased message delivery latency. <br>*Type: float*
+linger.ms                                |  P  | 0 .. 900000     |           0.5 | high       | Alias for `queue.buffering.max.ms`: Delay in milliseconds to wait for messages in the producer queue to accumulate before constructing message batches (MessageSets) to transmit to brokers. A higher value allows larger and more effective (less overhead, improved compression) batches of messages to accumulate at the expense of increased message delivery latency. <br>*Type: float*
```
mhowlett: this is actually a breaking change in the strongly typed .net API.

mhowlett: though we could make it a special case (case int->float) and avoid that.

edenhill: int to double is an implicit conversion in .NET, so it should be fine in this case, right?

src/rdkafka_conf.c
```diff
@@ -59,6 +59,7 @@ struct rd_kafka_property {
 	enum {
 		_RK_C_STR,
 		_RK_C_INT,
+                _RK_C_DBL,  /* Double */
```
mhowlett: I guess this is better than specifying an integer number of microseconds.

src/rdkafka_conf.h
```diff
@@ -360,7 +360,8 @@ struct rd_kafka_conf_s {
         } eos;
 	int    queue_buffering_max_msgs;
 	int    queue_buffering_max_kbytes;
-	int    buffering_max_ms;
+        double buffering_max_ms_dbl; /**< This is the configured value */
```
mhowlett: having both of these here is unfortunate since they're redundant. i don't see a good way around it, but if you do, worth changing.

src/rdkafka_conf.c
edenhill: That would be a breaking change, changing linger.ms->linger.us, or worse; to add a new property linger.us.

src/rdtime.h
```diff
@@ -154,6 +180,31 @@ static RD_INLINE rd_ts_t rd_timeout_init (int timeout_ms) {
...
+/**
+ * @brief Initialize an absolute timespec timeout based on the provided
+ *        relative \p timeout_us.
+ *
+ * To be used with cnd_timedwait_abs().
+ *
+ * Honours RD_POLL_INFITE and RD_POLL_NOWAIT (reflected in tspec.tv_sec).
+ */
+static RD_INLINE void rd_timeout_init_timespec_us (struct timespec *tspec,
+                                                   int timeout_us) {
```
mhowlett: should be `rd_ts_t`

edenhill: Good catch!

CONFIGURATION.md
mhowlett: yeah you're right, silly me. it won't break provided you recompile (which is the standard we're holding ourselves to in .net).

PR #2401, opened 2019-07-09
Add documentation about ERR__PARTITION_EOF

ngrande:
clarify default behavior of ERR__PARTITION_EOF even/error since v1.0.0

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -188,8 +188,8 @@ enum ErrorCode {
 	ERR__MSG_TIMED_OUT = -192,
 	/** Reached the end of the topic+partition queue on
 	 * the broker. Not really an error. 
-	 * This code is not sent by default since v1.0.0
-	 * via "enable.partition.eof" in the configuration. */
+	 * This event is disabled by default,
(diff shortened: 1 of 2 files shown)


ngrande:
https://github.com/edenhill/librdkafka/issues/2397

src-cpp/rdkafkacpp.h
```diff
@@ -187,7 +187,9 @@ enum ErrorCode {
 	/** Produced message timed out*/
 	ERR__MSG_TIMED_OUT = -192,
 	/** Reached the end of the topic+partition queue on
-	 * the broker. Not really an error. */
+	 * the broker. Not really an error. 
+	 * This code is not sent by default since v1.0.0
```
edenhill: This is mentioned in the release notes for v1.0.0, no need to reference the version here.
Suggest rephrasing to: 
```
This event is disabled by default, see the `enable.partition.eof` configuration property.
```
Also update the same error code in src/rdkafka.h accordingly.

ngrande: have adjusted it accordingly. Should be fine now.

edenhill:
thank you!

PR #2405, opened 2019-07-10
Be consistent with rd_malloc/rd_free combos.

sarkanyi:
Sometimes rd_free is called on objects allocated with plain malloc. This is not consistent.

### DIFF
--- a/src-cpp/RdKafka.cpp
+++ b/src-cpp/RdKafka.cpp
@@ -50,3 +50,10 @@ int RdKafka::wait_destroyed (int timeout_ms) {
   return rd_kafka_wait_destroyed(timeout_ms);
 }
 
+void *RdKafka::mem_malloc (size_t size) {
+  return rd_kafka_mem_malloc(NULL, size);
+}
+
+void RdKafka::mem_free (void *ptr) {
+  rd_kafka_mem_free(NULL, ptr);
+}
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -146,6 +146,34 @@ std::string get_debug_contexts();
 RD_EXPORT
 int          wait_destroyed(int timeout_ms);
 
+/**
+ * @brief Allocate memory using the same allocator librdkafka uses.
+ *
+ * This is typically an abstraction for the malloc(3) call and makes sure
+ * the application can use the same memory allocator as librdkafka for
+ * allocating pointers that are used by librdkafka.
+ *
+ * @remark Memory allocated by mem_malloc() must be freed using
+ *         mem_free().
+ */
+RD_EXPORT
+void *mem_malloc (size_t size);
+
+/**
+ * @brief Free pointer returned by librdkafka
+ *
+ * This is typically an abstraction for the free(3) call and makes sure
+ * the application can use the same memory allocator as librdkafka for
+ * freeing pointers returned by librdkafka.
+ *
+ * In standard setups it is usually not necessary to use this interface
+ * rather than the free(3) function.
+ *
+ * @remark mem_free() must only be used for pointers returned by APIs
+ *         that explicitly mention using this function for freeing.
+ */
+RD_EXPORT
+void mem_free (void *ptr);
 
 /**@}*/
 
@@ -1736,6 +1764,33 @@ class RD_EXPORT Handle {
      * @sa RdKafka::Conf::set() \c "oauthbearer_token_refresh_cb"
      */
     virtual ErrorCode oauthbearer_set_token_failure (const std::string &errstr) = 0;
+
+   /**
+     * @brief Allocate memory using the same allocator librdkafka uses.
+     *
+     * This is typically an abstraction for the malloc(3) call and makes sure
+     * the application can use the same memory allocator as librdkafka for
+     * allocating pointers that are used by librdkafka.
+     *
+     * @remark Memory allocated by mem_malloc() must be freed using
+     *         mem_free().
+     */
+    virtual void *mem_malloc (size_t size) = 0;
+
+   /**
+     * @brief Free pointer returned by librdkafka
+     *
+     * This is typically an abstraction for the free(3) call and makes sure
+     * the application can use the same memory allocator as librdkafka for
+     * freeing pointers returned by librdkafka.
+     *
+     * In standard setups it is usually not necessary to use this interface
+     * rather than the free(3) function.
+     *
+     * @remark mem_free() must only be used for pointers returned by APIs
+     *         that explicitly mention using this function for freeing.
+     */
+    virtual void mem_free (void *ptr) = 0;
 };
 
 
@@ -2003,7 +2058,7 @@ public:
       value_size_ = other.value_size_;
 
       if (value_ != NULL)
-        free(value_);
+        mem_free(value_);
 
       value_ = copy_value(other.value_, value_size_);
 
@@ -2012,7 +2067,7 @@ public:
 
     ~Header() {
       if (value_ != NULL)
-        free(value_);
+        mem_free(value_);
     }
 
     /** @returns the key/name associated with this Header */
@@ -2046,7 +2101,7 @@ public:
       if (!value)
         return NULL;
 
-      char *dest = (char *)malloc(value_size + 1);
+      char *dest = (char *)mem_malloc(value_size + 1);
       memcpy(dest, (const char *)value, value_size);
       dest[value_size] = '\0';
 
@@ -3488,5 +3543,6 @@ class Metadata {
 /**@}*/
 
 }
+
 
 #endif /* _RDKAFKACPP_H_ */
--- a/src-cpp/rdkafkacpp_int.h
+++ b/src-cpp/rdkafkacpp_int.h
@@ -992,6 +992,13 @@ class HandleImpl : virtual public Handle {
                                                 rk_, errstr.c_str()));
   };
 
+  void *mem_malloc (size_t size) {
+    return rd_kafka_mem_malloc(rk_, size);
+  };
+
+  void mem_free (void *ptr) {
+    rd_kafka_mem_free(rk_, ptr);
+  };
 
   rd_kafka_t *rk_;
   /* All Producer and Consumer callbacks must reside in HandleImpl and
--- a/src/lz4.c
+++ b/src/lz4.c
@@ -198,10 +198,13 @@ void  LZ4_free(void* p);
 # define ALLOC_AND_ZERO(s) LZ4_calloc(1,s)
 # define FREEMEM(p)        LZ4_free(p)
 #else
-# include <stdlib.h>   /* malloc, calloc, free */
-# define ALLOC(s)          malloc(s)
-# define ALLOC_AND_ZERO(s) calloc(1,s)
-# define FREEMEM(p)        free(p)
+struct rdkafka_s;
+extern void *rd_kafka_mem_malloc(struct rdkafka_s *rk, size_t s);
+extern void *rd_kafka_mem_calloc(struct rdkafka_s *rk, size_t n, size_t s);
+extern void rd_kafka_mem_free(struct rdkafka_s *rk, void *p);
+# define ALLOC(s)          rd_kafka_mem_malloc(NULL, s)
+# define ALLOC_AND_ZERO(s) rd_kafka_mem_calloc(NULL, 1, s)
+# define FREEMEM(p)        rd_kafka_mem_free(NULL, p)
 #endif
 
 #include <string.h>   /* memset, memcpy */
--- a/src/rd.h
+++ b/src/rd.h
@@ -154,7 +154,7 @@ static RD_INLINE RD_UNUSED char *rd_strndup(const char *s, size_t len) {
 	char *n = strndup(s, len);
 	rd_assert(n);
 #else
-	char *n = (char *)malloc(len + 1);
+	char *n = (char *)rd_malloc(len + 1);
 	rd_assert(n);
 	memcpy(n, s, len);
 	n[len] = '\0';
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3093,6 +3093,43 @@ rd_kafka_offsets_for_times (rd_kafka_t *rk,
                             int timeout_ms);
 
 
+
+/**
+ * @brief Allocate and zero memory using the same allocator librdkafka uses.
+ *
+ * This is typically an abstraction for the calloc(3) call and makes sure
+ * the application can use the same memory allocator as librdkafka for
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p rk can be set to return memory allocated by a specific \c rk instance
+ * otherwise pass NULL for \p rk.
+ *
+ * @remark Memory allocated by rd_kafka_mem_calloc() must be freed using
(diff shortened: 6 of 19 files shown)


sarkanyi:
And I've found some more, updating soon.

sarkanyi:
Not sure why the unkpart test files, the commit has tehnically nothing to do with it.

sarkanyi:
I think the failure is related to the float-linger change, but have no time to investigate.

src-cpp/rdkafkacpp.h
```diff
@@ -81,13 +81,20 @@ typedef SSIZE_T ssize_t;
 
 /**@endcond*/
 
+#ifdef _MSC_VER    /* Visual Studio */
+#  pragma warning(disable : 4273)
```
edenhill, line 85: What's this for?

src-cpp/rdkafkacpp.h
```diff
@@ -81,13 +81,20 @@ typedef SSIZE_T ssize_t;
...
+
 extern "C" {
         /* Forward declarations */
         struct rd_kafka_s;
         struct rd_kafka_topic_s;
         struct rd_kafka_message_s;
         struct rd_kafka_conf_s;
         struct rd_kafka_topic_conf_s;
+
+        RD_EXPORT void *rd_kafka_mem_malloc(struct rd_kafka_s *rk, size_t size);
```
edenhill, line 96: I'd rather see a proper C++ wrapper API, we'll want to avoid conflating the two APIs unless necessary.

src/lz4hc.c
```diff
@@ -33,6 +33,7 @@
 */
 /* note : lz4hc is not an independent module, it requires lz4.h/lz4.c for proper compilation */
 
+#include "rd.h"   /* rd_malloc, rd_calloc, rd_free */
```
edenhill, line 36: Why is this needed?

src/rdkafka.h
```diff
@@ -3088,6 +3088,24 @@ rd_kafka_offsets_for_times (rd_kafka_t *rk,
                             int timeout_ms);
 
 
+/**
+ * @brief Allocate memory using the librdkafka internal allocator
```
edenhill, line 3092: "using the same allocator librdkafka uses." is more correct, it currently implies that there's an internal librdkafka allocator which there is not.

src/rdkafka.h
```diff
@@ -3088,6 +3088,24 @@ rd_kafka_offsets_for_times (rd_kafka_t *rk,
...
+ *
+ * This is typically an abstraction for the malloc(3) call and makes sure
+ * the application can use the same memory allocator as librdkafka for
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p rk must be set for memory returned by APIs that take an \c rk argument,
```
edenhill, line 3098: This does not apply to malloc, only free.

src/rdkafka.h
```diff
@@ -3088,6 +3088,24 @@ rd_kafka_offsets_for_times (rd_kafka_t *rk,
...
+ * for other APIs pass NULL for \p rk.
+ *
+ * @remark Memory allocated by rd_kafka_mem_alloc() must be freed using
```
edenhill, line 3101: mem_malloc

edenhill (changes requested):

src-cpp/rdkafkacpp.h
sarkanyi: To avoid this warning:
rdkafka_example.obj : error LNK2019: unresolved external symbol __imp_rd_kafka_mem_malloc referenced in function "public: __cdecl RdKafka::Headers::Header::Header(class RdKafka::Headers::Header const &)" (??0Header@Headers@RdKafka@@QEAA@AEBV012@@Z) [C:\projects\librdkafka\win32\rdkafka_example\rdkafka_example.vcxproj]
    13>rdkafka_example.obj : error LNK2019: unresolved external symbol __imp_rd_kafka_mem_free referenced in function "void __cdecl std::_Destroy_range<struct std::_Wrap_alloc<class std::allocator<class RdKafka::Headers::Header> > >(class RdKafka::Headers::Header *,class RdKafka::Headers::Header *,struct std::_Wrap_alloc<class std::allocator<class RdKafka::Headers::Header> > &)" (??$_Destroy_range@U?$_Wrap_alloc@V?$allocator@VHeader@Headers@RdKafka@@@std@@@std@@@std@@YAXPEAVHeader@Headers@RdKafka@@0AEAU?$_Wrap_alloc@V?$allocator@VHeader@Headers@RdKafka@@@std@@@0@@Z) [C:\projects\librdkafka\win32\rdkafka_example\rdkafka_example.vcxproj]
    13>

src-cpp/rdkafkacpp.h
```diff
@@ -146,6 +146,40 @@ std::string get_debug_contexts();
...
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p rk can be set to return memory allocated by a specific \c rk instance
+ * otherwise pass NULL for \p rk.
+ *
+ * @remark Memory allocated by mem_malloc() must be freed using
+ *         mem_free()
+ */
+RD_EXPORT
+void *mem_malloc(struct rd_kafka_s *rk, size_t size);
```
edenhill, line 163: I think it should take an optional Handle pointer, not an rk, no need to expose the C type here.

src-cpp/rdkafkacpp_int.h
```diff
@@ -964,7 +964,7 @@ class HandleImpl : virtual public Handle {
                                    std::string &errstr) {
           char errbuf[512];
           ErrorCode err;
-          const char **extensions_copy = new const char *[extensions.size()];
```
edenhill, line 967: This was actually a bug, new:ed memory should be freed with delete, not free().
Since this memory is not related to librdkafka and only lives for this call I think we should use the original new here, and change the free to delete;

src/rd.h
```diff
@@ -154,7 +154,7 @@ static RD_INLINE RD_UNUSED char *rd_strndup(const char *s, size_t len) {
 	char *n = strndup(s, len);
 	rd_assert(n);
 #else
-	char *n = (char *)malloc(len + 1);
+	char *n = (char *)rd_malloc(len + 1);
```
edenhill, line 157: This isn't great because if the system has strndup() we might be rd_free() that memory with another allocator (if plugged).

src-cpp/rdkafkacpp_int.h
sarkanyi: Well, since it's passed to the C function it makes sense to be malloc-ed, I guess, and free was called anyway, so in that case delete has to be called.

src/rd.h
sarkanyi: I was thinking to leave that to the other PR where you can set a different allocator. But calling system malloc is not really an option then, should call into however rd_malloc is called. But normally yes, all function should use the same allocator/free

src/rdkafka.h
edenhill: If we have the Handle* argument there we're free to start making use of it in the future, without asking users to fix their code.

edenhill: .. which gives us more options to work with in the future.

src-cpp/rdkafkacpp_int.h
edenhill: The allocator doesn't matter here since the memory will never be transfered to the application or librdkafka, it is just some temporary buffer, so let's use new/delete.

src/rd.h
edenhill: Right, we can fix it in the future :+1:

src-cpp/rdkafkacpp.h
```diff
@@ -146,6 +146,40 @@ std::string get_debug_contexts();
...
 int          wait_destroyed(int timeout_ms);
 
+/**
+ * @brief Allocate memory using the same allocator librdkafka uses.
+ *
+ * This is typically an abstraction for the malloc(3) call and makes sure
+ * the application can use the same memory allocator as librdkafka for
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p handle of type rd_kafka_t can be set to return memory allocated by a specific \c handle instance
```
edenhill, line 156: Why pass the C handle and not a C++ handle?

src/lz4.c
```diff
@@ -198,10 +198,13 @@ void  LZ4_free(void* p);
 # define ALLOC_AND_ZERO(s) LZ4_calloc(1,s)
 # define FREEMEM(p)        LZ4_free(p)
 #else
-# include <stdlib.h>   /* malloc, calloc, free */
-# define ALLOC(s)          malloc(s)
-# define ALLOC_AND_ZERO(s) calloc(1,s)
-# define FREEMEM(p)        free(p)
+struct rdkafka_s;
```
edenhill, line 201: Why is this different from lz4frame.c ?

src/rdkafka.h
```diff
@@ -3091,6 +3091,43 @@ rd_kafka_offsets_for_times (rd_kafka_t *rk,
                             int timeout_ms);
 
 
+
+/**
+ * @brief Allocate and zero memory using the same allocator librdkafka uses.
+ *
+ * This is typically an abstraction for the malloc(3) call and makes sure
```
edenhill, line 3098: calloc

src/rdkafka.h
```diff
@@ -3091,6 +3091,43 @@ rd_kafka_offsets_for_times (rd_kafka_t *rk,
...
+ * the application can use the same memory allocator as librdkafka for
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p rk can be set to return memory allocated by a specific \c rk instance
+ * otherwise pass NULL for \p rk.
+ *
+ * @remark Memory allocated by rd_kafka_mem_malloc() must be freed using
```
edenhill, line 3105: calloc

edenhill (changes requested):

src-cpp/rdkafkacpp.h
sarkanyi: There is no C++ handle for rd_kafka_t, or maybe I missed it.

src/lz4.c
sarkanyi: I don't exactly remember, but I think in this case we couldn't include "rd.h"

edenhill: It'd be good if we could use rd.h here

src-cpp/rdkafkacpp.h
edenhill: The RdKafka::Handle has a rk_ pointer.

src/lz4.c
sarkanyi: lz4.h is included only internally in the lz4 code, and I think some flags differ. Including either rd.h or rdkafka_int.h will open a huge can of worms (compiler errors)

src-cpp/rdkafkacpp.h
sarkanyi: Ok, replacing with RdKafka::Handle instance

src-cpp/RdKafka.cpp
```diff
@@ -50,3 +50,12 @@ int RdKafka::wait_destroyed (int timeout_ms) {
   return rd_kafka_wait_destroyed(timeout_ms);
 }
 
+void *RdKafka::mem_malloc (void *handle, size_t size) {
```
edenhill, line 53: Use RdKafka::Handle *handle as type, not void.

src-cpp/rdkafkacpp.h
```diff
@@ -146,6 +146,40 @@ std::string get_debug_contexts();
...
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p handle of type RdKafka::Handle can be set to return memory allocated by a
+ * specific \c handle instance otherwise pass NULL for \p handle.
+ *
+ * @remark Memory allocated by mem_malloc() must be freed using
+ *         mem_free()
+ */
+RD_EXPORT
+void *mem_malloc(void *handle, size_t size);
```
edenhill, line 163: style-nit: space before paren

src-cpp/rdkafkacpp.h
```diff
@@ -146,6 +146,40 @@ std::string get_debug_contexts();
...
+ *
+ * This is typically an abstraction for the malloc(3) call and makes sure
+ * the application can use the same memory allocator as librdkafka for
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p handle of type RdKafka::Handle can be set to return memory allocated by a
+ * specific \c handle instance otherwise pass NULL for \p handle.
+ *
+ * @remark Memory allocated by mem_malloc() must be freed using
+ *         mem_free()
```
edenhill, line 160: Add trailing .

edenhill (changes requested):

src-cpp/RdKafka.cpp
sarkanyi: Ok, but then you don't mind if I forward declare "Handle" in the header.

edenhill: Maybe it is better with a mem_malloc() method on the Handle class, and one without a handle in the RdKafka namespace.

sarkanyi: Without a handle completely? So we'd have:
void *RdKafka::mem_malloc(size_t size) and
void *mem_malloc(RdKafka:Handle *handle, size_t size)?
The C api takes a handle (unused at the moment, but still)

edenhill: Yep, that was the idea.

sarkanyi: Actually I can rearrange a bit I think so there would be no need for two allocators

sarkanyi: One change though I think I should make, let it accept a NULL handle...

edenhill: My point is that there should be malloc on the Handle, and one without a handle in the global namespace, covering both scenarios. We should avoid NULL-as-optional in C++ if we can.

sarkanyi: Ok, updated with two methods (with and w/o handle)

src-cpp/rdkafkacpp.h
```diff
@@ -3489,4 +3518,47 @@ class Metadata {
...
+ * allocating pointers that are used by librdkafka.
+ *
+ * \p handle of type RdKafka::Handle can be set to return memory allocated by a
+ * specific \c handle instance otherwise pass NULL for \p handle.
+ *
+ * @remark Memory allocated by mem_malloc() must be freed using
+ *         mem_free().
+ */
+RD_EXPORT
+void *mem_malloc (RdKafka::Handle *handle, size_t size);
```
edenhill, line 3541: I was thinking that these two are put as instance methods on the Handle class so that the Handle argument is not needed.

sarkanyi: That still wouldn't solve the NULL handle case. Since for example Headers has no handle.

sarkanyi: Or should I keep the void *RdKafka::mem_malloc (size_t size); version around for just these cases?

sarkanyi: Also my guess is that if I'll include it in Handle, then you want them pure virtual

edenhill (approved):
LGTM!

edenhill:
Thanks for your perseverence!

Issue #2403, opened 2019-07-10
Multiple -fsanitize=undefined failures


PlacidBox:
librdkafka version: `master @ 30c1f8f71fb79501d6cf697c4af323daa95b5834`, built with `-fsanitize=undefined`, running with `./tests/autotest.sh 2.3.0` on `debian_stretch` against x86_64. The issues also occur on librdkafka version 1.0.0.

Get the following failures from UBSan:

`rdvarint.h:69:54: runtime error: left shift of negative value -1`

```
67 | static RD_INLINE RD_UNUSED
68 | size_t rd_uvarint_enc_i64 (char *dst, size_t dstsize, int64_t num) {
69 |         return rd_uvarint_enc_u64(dst, dstsize, (num << 1) ^ (num >> 63));
70 | }
```

---

`rdkafka_buf.h:129:3: runtime error: null pointer passed as argument 2, which is declared to never be null`. 

This is a macro expansion from `rd_tmpabuf_write` in `rd_kafka_metadata_cache_insert`, where `mtopic->partitions` is `NULL`, and ends up passing `NULL` as a source to `memcpy`

```
        rkmce->rkmce_mtopic.partitions =
                rd_tmpabuf_write(&tbuf, mtopic->partitions, // <-- NULL
                                 mtopic->partition_cnt *
                                 sizeof(*mtopic->partitions));
```

```
123 | static RD_UNUSED void *
124 | rd_tmpabuf_write0 (const char *func, int line,
125 | 		   rd_tmpabuf_t *tab, const void *buf, size_t size) {
126 | 	void *ptr = rd_tmpabuf_alloc0(func, line, tab, size);
127 | 
128 | 	if (ptr)
129 | 		memcpy(ptr, buf, size);
130 | 
131 | 	return ptr;
132 | }
133 | #define rd_tmpabuf_write(tab,buf,size) \
134 | 	rd_tmpabuf_write0(__FUNCTION__, __LINE__, tab, buf, size)
```

---

`rdlist.c:250:9: runtime error: null pointer passed as argument 1, which is declared to never be null`

```
248 | void rd_list_sort (rd_list_t *rl, int (*cmp) (const void *, const void *)) {
249 | 	rd_list_cmp_curr = cmp;
250 |         qsort(rl->rl_elems, rl->rl_cnt, sizeof(*rl->rl_elems),
251 | 	      rd_list_cmp_trampoline);
252 | 	rl->rl_flags |= RD_LIST_F_SORTED;
[...]

PR #2406, opened 2019-07-10
Fix UB left shift in zigzag encoding

PlacidBox:
Fixes an issue in #2403 

Neither latest gcc/clang seem to have codegen differences based
on this change, and are unlikely ever to since left shift of
negative isn't UB in c++20. It's still worth fixing to get
less print outs when running with fsanitize=undefined


PlacidBox:
Looks like the `0002_unkpart` test is failing in other PRs?

PR #2407, opened 2019-07-11
Fix UBSan errors

PlacidBox:
rd_tmpabuf_write is sometimes called with a NULL buffer to clone,
leading to UB when calling memcpy.

It doesn't look like the func could just return NULL in this case
either. Almost all the callers ignore the possibility of NULL, or
are interpreting NULL as an allocation failure.

### DIFF
--- a/src/rdkafka_buf.h
+++ b/src/rdkafka_buf.h
@@ -125,8 +125,8 @@ rd_tmpabuf_write0 (const char *func, int line,
 		   rd_tmpabuf_t *tab, const void *buf, size_t size) {
 	void *ptr = rd_tmpabuf_alloc0(func, line, tab, size);
 
-	if (ptr && buf)
-		memcpy(ptr, buf, size);
+        if (likely(ptr && size))
+                memcpy(ptr, buf, size);
 
 	return ptr;
 }
--- a/src/rdkafka_buf.h
+++ b/src/rdkafka_buf.h
@@ -125,7 +125,7 @@ rd_tmpabuf_write0 (const char *func, int line,
 		   rd_tmpabuf_t *tab, const void *buf, size_t size) {
 	void *ptr = rd_tmpabuf_alloc0(func, line, tab, size);
 
-	if (ptr)
+	if (ptr && buf)
 		memcpy(ptr, buf, size);
 
 	return ptr;
--- a/src/rdlist.c
+++ b/src/rdlist.c
@@ -246,6 +246,10 @@ int rd_list_cmp_trampoline (const void *_a, const void *_b) {
 }
 
 void rd_list_sort (rd_list_t *rl, int (*cmp) (const void *, const void *)) {
+	if (rl->rl_elems == NULL) {
+		return;
+	}
+
 	rd_list_cmp_curr = cmp;
         qsort(rl->rl_elems, rl->rl_cnt, sizeof(*rl->rl_elems),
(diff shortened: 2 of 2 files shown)


PlacidBox:
bundled in the last fix for #2403, since it's an easy one

edenhill:
These are false positives, because why would memcpy or qsort read at any pointer address (null or not) if the size of the buffer or array is zero?

PlacidBox:
They're definitely not false positives, the compiler is very much free to optimise around the values being non-null, for example: https://godbolt.org/z/g1Lg0q. Note that there is no code path in `foo_bad` that calls `bar`, even if you pass in s=NULL count=0.

[Sometimes you end up with time travel too](https://devblogs.microsoft.com/oldnewthing/20140627-00/?p=633).

I have no idea why `memcpy` and `qsort` don't allow NULL args when the length is 0, but the unfortunate reality is that they don't, and you end up with UBSan triggers on them that need to be fixed.

An alternative fix would also be to guarantee that a lot of the internal lists and buffers are never nullable, but i'm not familiar enough with the code base to make that far reaching a change.

edenhill:
Okay, fair enough.

Let's fix the places where it passes NULL, preferably with something like:
```
if (likely(src))
 memcpy(dst, src, srclen);
```

src/rdkafka_buf.h
```diff
@@ -125,7 +125,7 @@ rd_tmpabuf_write0 (const char *func, int line,
 		   rd_tmpabuf_t *tab, const void *buf, size_t size) {
 	void *ptr = rd_tmpabuf_alloc0(func, line, tab, size);
 
-	if (ptr)
```
edenhill: I'd argue that `if (ptr && size)` is more correct, to catch bad callers.
If that messes up the compiler optimization, we could do:
```
 if (likely(ptr && buf)) {
       rd_dassert(size > 0);  <-- that's a ENABLE_DEVEL-only assert
       memcpy..
}
```

src/rdlist.c
```diff
@@ -246,6 +246,10 @@ int rd_list_cmp_trampoline (const void *_a, const void *_b) {
 }
 
 void rd_list_sort (rd_list_t *rl, int (*cmp) (const void *, const void *)) {
+	if (rl->rl_elems == NULL) {
```
edenhill: I know it is a mix of tabs and spaces in the code, and trhat will be fixed, but new code should use 8 whitespaces for indent.
Also drop the curly braces, we don't use them for single line blocks.

edenhill: and add unlikely()

PlacidBox: you should look in to clang-format, it's pretty good tool for this

PlacidBox:
Sorry about the delay, was on other projects for a bit

Do you want me to rebase https://github.com/edenhill/librdkafka/pull/2406 on to this PR? It's the same category of issue

src/rdkafka_buf.h
PlacidBox: I did a quick audit of other memcpy and friends usage, but couldn't see anything wrong. I'm not familiar enough with the code base though.

src/rdlist.c
edenhill: I did last year, but I couldn't get it to completely match the emacs c-style linux.

edenhill:
Thank you for this!

PR #2409, opened 2019-07-11
Add API to replace rd_kafka's default allocator

sarkanyi:
Add API to replace rd_kafka's default allocator with a custom implementation


src/rdallocator.c
```diff
@@ -0,0 +1,71 @@
...
+
+struct rd_allocator_state RD_CURRENT_ALLOCATOR = {
+	rd_default_malloc,
+	rd_default_calloc,
+	rd_default_realloc,
+	rd_default_free,
+	NULL
+};
+
+void rd_set_allocator(void *malloc, void *calloc, void *realloc, void *free, void *user_data) {
```
edenhill: keep below 80 columns.
rename arguments to avoid shadowing: `f_malloc`, `f_calloc`, etc.

src/rd.h
```diff
@@ -90,52 +90,56 @@
...
 
-static RD_INLINE RD_UNUSED char *rd_strdup(const char *s) {
-#ifndef _MSC_VER
-	char *n = strdup(s);
-#else
-	char *n = _strdup(s);
-#endif
-	rd_assert(n);
-	return n;
+#define rd_realloc(ptr, nsz)          \
```
edenhill: make all of these inline functions

src/rd.h
```diff
@@ -90,52 +90,56 @@
...
-#if HAVE_STRNDUP
-	char *n = strndup(s, len);
-	rd_assert(n);
-#else
-	char *n = malloc(len + 1);
+	if (s == NULL) {
+		return NULL;
+	}
+
+	char *n = rd_malloc(len + 1);
```
edenhill: variable declarations must be at the head of scope/block.

src/rdkafka.h
```diff
@@ -204,7 +204,18 @@ typedef enum rd_kafka_timestamp_type_t {
...
+ * This has to be set before calling any other rdkafka functions.
+ *
+ * @param malloc - malloc function pointer
+ * @param calloc - calloc function pointer
+ * @param realloc - realloc function pointer
+ * @param free - free function pointer
+ * @param user_data - extra user data that can be passed to the allocator
+ */
+RD_EXPORT
+void rd_set_allocator(void *malloc, void *calloc, void *realloc, void *free, void *user_data);
```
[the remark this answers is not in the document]
edenhill: `rd_kafka_set_allocator`

src/rdkafka.h
```diff
@@ -204,7 +204,18 @@ typedef enum rd_kafka_timestamp_type_t {
 	RD_KAFKA_TIMESTAMP_LOG_APPEND_TIME  /**< Log append time */
 } rd_kafka_timestamp_type_t;
 
-
+/**
+ * @brief Set external allocator for rd_malloc/rd_calloc/rd_realloc/rd_free.
```
edenhill: rd_malloc, et.al, are unknown to the users of librdkafka, rephrase as:
`Set external allocator for allocations made by librdkafka`

Also make a list of methods that are affected by this allocator, things like RD_KAFKA_MSG_F_FREE to produce(), rd_kafka_free(), etc.

src/rdkafka.h
```diff
@@ -204,7 +204,18 @@ typedef enum rd_kafka_timestamp_type_t {
...
+ * This has to be set before calling any other rdkafka functions.
+ *
+ * @param malloc - malloc function pointer
```
edenhill: Remove -

src/rdkafka.h
```diff
@@ -204,7 +204,18 @@ typedef enum rd_kafka_timestamp_type_t {
...
+ * @param calloc - calloc function pointer
+ * @param realloc - realloc function pointer
+ * @param free - free function pointer
+ * @param user_data - extra user data that can be passed to the allocator
```
edenhill: "opaque value passed to allocator functions"

src/rdkafka.h
```diff
@@ -204,7 +204,18 @@ typedef enum rd_kafka_timestamp_type_t {
 	RD_KAFKA_TIMESTAMP_LOG_APPEND_TIME  /**< Log append time */
 } rd_kafka_timestamp_type_t;
 
-
+/**
+ * @brief Set external allocator for rd_malloc/rd_calloc/rd_realloc/rd_free.
+ * This has to be set before calling any other rdkafka functions.
```
edenhill: `@warning This function may only be called once and must be called before calling any other rdkafka functions.`

src/rdallocator.c
edenhill: add an assert, like `rd_assert(!rd_allocator.is_set)` to make sure it is only set once, let this value be 0 for the default allocator.

src/rdallocator.c
```diff
@@ -0,0 +1,71 @@
...
+
+static void *
+rd_default_realloc(RD_UNUSED void *ud, void *ptr, size_t nsize) {
+	void *p = realloc(ptr, nsize);
+	rd_assert(p);
+	return p;
+}
+
+static void
+rd_default_free(RD_UNUSED void *ud, void *ptr) {
```
edenhill: code style for function defs/decl is `function_name<SPACE>(...`, which makes it easy to find the declaration with a search.

edenhill: We don't use RD_UNUSED for function arguments.

sarkanyi:
I think I covered all requests. Could you please re-check and merge if ok? Also please check and merge: https://github.com/edenhill/librdkafka/pull/2405 because it's related. I can rebase both if needed

edenhill:
Will re-review soon, but this will not be merged until after the next release.

edenhill:
With your PR merged to add mem_malloc() et.al. I'm guessing this PR is no longer needed, an application can use LD_PRELOAD.. to replace the rd_malloc()* calls, right?

sarkanyi:
> With your PR merged to add mem_malloc() et.al. I'm guessing this PR is no longer needed, an application can use LD_PRELOAD.. to replace the rd_malloc()* calls, right?

To be honest it would be preferable to have an API, either this - I'll change to have rd_str_dup & co also added, or function hooks something like:
"typeof (&rd_malloc) malloc_fn = rd_malloc_hook ? rd_malloc_hook : malloc;
    void *p = malloc_fn(sz);" -set via a setter function?

sarkanyi:
Most big libraries use some sort of API to set the custom allocator, ie: jansson has void json_set_alloc_funcs(json_malloc_t malloc_fn, json_free_t free_fn), it makes it more explicit. LD_PRELOAD is quite invasive, even if we'd overload rd_funcs

edenhill:
Okay, a single function to set the allocators once on program startup before calling any other librdkafka function (in particular before global_init is called) should be fine and avoid the need for locking.

sarkanyi:
Hi, is this still on track for 1.8? Do you need me to do anything else on it?

sarkanyi:
@edenhill won't this change make it into 1.9.0? Thanks

PR #2413, opened 2019-07-15
Static library bundles

edenhill:
Build static library bundles containing all external dependencies of librdkafka (or as many as possible).
This is to be used with confluent-kafka-go client ([demo here](https://github.com/confluentinc/confluent-kafka-go-example)).

Related Go PR:
https://github.com/confluentinc/confluent-kafka-go/pull/355

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,17 @@ language: c
 dist: trusty
 sudo: false
 cache: ccache
+
 env:
  global:
   - ARCH=x64
+
+addons:
+  apt:
+    packages:
+    # required by openssl installer
+    - perl
+
 matrix:
  include:
  - name: "Linux GCC: +centos +debian"
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,23 +10,31 @@ matrix:
  - name: "Linux GCC: +centos +debian"
    os: linux
    compiler: gcc
-   env: ADDITIONAL_BUILDS="centos debian"
+   env: ADDITIONAL_BUILDS="centos debian" LINKAGE=std
    before_script:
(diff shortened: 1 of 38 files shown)


.travis.yml
```diff
@@ -2,32 +2,52 @@ language: c
...
    os: osx
    compiler: gcc
+   env: LINKAGE=std
    before_script:
     - ./configure --install-deps --disable-lz4-ext --prefix="$PWD/dest"
  - name: "OSX clang: +static"
    os: osx
+   # Use an older image to and disable syslog for broader compatibility
+   # with old and new osx versions.
+   osx_image: xcode9.2
```
rnpridgeon: Is this still necessary with the `--disable-syslog` flag?

edenhill: Yes, older toolchains complained of .a files being generated by newer toolchains.

.travis.yml
```diff
@@ -2,32 +2,52 @@ language: c
 dist: trusty
 sudo: false
 cache: ccache
+
 env:
-- ARCH=x64
+ global:
```
mhowlett: inconsistent spacing/tab size of 1.

mikekamornikov:
@edenhill What is the status of this one? Should we expect it (and https://github.com/confluentinc/confluent-kafka-go/pull/355) to me merged anytime soon? The thing should simplify a lot `bazel` builds. Currently I have to build `librdkafka` separately via `rules_foreign_cc` and do some magic to tie it to `confluent-kafka-go`. I can't cross compile the result this way (`bazel` complains about missing toolchain). And for me it means a switch to some pure go client.

PR #2416, opened 2019-07-17
escape || inside markdown table

mhowlett:
alternatively could surround the default value in back ticks.


src/rdkafka_conf.c
```diff
@@ -746,7 +746,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           .sdef =
           /* First attempt to refresh, else acquire. */
           "kinit -R -t \"%{sasl.kerberos.keytab}\" "
-          "-k %{sasl.kerberos.principal} || "
+          "-k %{sasl.kerberos.principal} \\|\\| "
```
edenhill: I'm not sure this will work since this string is eventually used to refresh the ticket.

mhowlett: ahh, right.

one alternative might be to render all default values inside back ticks (or just this property as a special case), however i just checked how that markdown renders on github, and it doesn't work.

another alternative would be to escape | characters during table render. I don't see a problem with that. Will do that. Note: the intellisense in C# generated from this will include the escape characters, but I think that's pretty inconsequential.

mhowlett: let me know if that sounds like a good idea to you. i mean a string replace of | with \| about here: https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_conf.c#L2871

also, advice on how you would do a string replace would be appreciated as it seems like something you might have other ideas on than me.

AndyPook: If I'm understanding this correctly (?)
The `||` is confusing the "parsing" of this page in confluent-kafka (used to code-gen some Config classes).
In my PR https://github.com/confluentinc/confluent-kafka-dotnet/pull/979 I've made the parsing a little "smarter" to not treat the `||` as a column delimiter.
If that works then this change is not necessary :)

edenhill: Github markdown parser fails on this too, the pipes are considered column separators.

Adding backticks does not help, in fact it crashes the GH markdown previewer 🤷‍♂ 

Backticks and escaping works, which makes a markdown rendered file (such as on github) show the correct command, while anyone viewing the file as-is will see the backslashing quoting.
We might be fine with that, or we could generate a CONFIGURATION.txt without the markdown quoting.

As for now I suggest we go ahead with the escaping when generating the .md file (do this outside of rdkafka_conf.c with sed in Makefile).

mhowlett: ok, sounds good.

Makefile
```diff
@@ -27,6 +27,7 @@ CONFIGURATION.md: src/rdkafka.h examples
 	@printf "$(MKL_YELLOW)Updating$(MKL_CLR_RESET)\n"
 	@echo "# Configuration properties" > CONFIGURATION.md.tmp
 	@(examples/rdkafka_performance -X list >> CONFIGURATION.md.tmp; \
+		sed -i 's/||/\\|\\|/g' CONFIGURATION.md.tmp; \
```
edenhill: `-i` is non-standard, use a pipe from the previous command instead:
```
@(examples/rdkafka_performance -X list | \
		sed 's/||/\\|\\|/g' >> \
                 CONFIGURATION.md.tmp; \
```

PR #2420, opened 2019-07-18
use hostname when validating hostname

hunterjackson:
See issue https://github.com/edenhill/librdkafka/issues/2419 for more background


edenhill:
Thank you!

hunterjackson:
Do you have an ETA on when this will make it into a release?

edenhill:
mid/end august

PR #2421, opened 2019-07-18
don't fail kinit for ECHILD

hannip:
When librdkafka is used as a plugin and there is a signal handler that ignores ECHILD for plugins the kinit system call returns a -1 with errno ECHILD.  This can be safely ignored and just means the command finished before the wait for it could be issued.  This fixes an issue with pmacct using librdkafka as a plugin.

### DIFF
--- a/src/rdkafka_sasl_cyrus.c
+++ b/src/rdkafka_sasl_cyrus.c
@@ -209,8 +209,7 @@ static int rd_kafka_sasl_cyrus_kinit_refresh (rd_kafka_t *rk) {
                                      "returned ECHILD: %s: exit status "
                                      "unknown, assuming success",
                                      cmd);
-                }
-                else {
+                } else {
                         rd_kafka_log(rk, LOG_ERR, "SASLREFRESH",
                                      "Kerberos ticket refresh failed, "
                                      "errno=%d: Failed to execute %s",
--- a/src/rdkafka_sasl_cyrus.c
+++ b/src/rdkafka_sasl_cyrus.c
@@ -203,7 +203,13 @@ static int rd_kafka_sasl_cyrus_kinit_refresh (rd_kafka_t *rk) {
         mtx_unlock(&rd_kafka_sasl_cyrus_kinit_lock);
 
         if (r == -1) {
-                if (errno != ECHILD) {
+                if (errno == ECHILD) {
+                        rd_kafka_log(rk, LOG_WARNING, "SASLREFRESH",
+                                     "Kerberos ticket refresh command returned ECHILD errno which means ",
+                                     "success is unknown, but probably okay so continuing as such: %s", 
+                                     cmd);
+                }
+                else {
                         rd_kafka_log(rk, LOG_ERR, "SASLREFRESH",
                                      "Kerberos ticket refresh failed, errno=%d: "
                                      "Failed to execute %s",
--- a/src/rdkafka_sasl_cyrus.c
+++ b/src/rdkafka_sasl_cyrus.c
@@ -205,14 +205,15 @@ static int rd_kafka_sasl_cyrus_kinit_refresh (rd_kafka_t *rk) {
         if (r == -1) {
                 if (errno == ECHILD) {
                         rd_kafka_log(rk, LOG_WARNING, "SASLREFRESH",
-                                     "Kerberos ticket refresh command returned ECHILD errno which means ",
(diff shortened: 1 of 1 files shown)


src/rdkafka_sasl_cyrus.c
```diff
@@ -203,12 +203,14 @@ static int rd_kafka_sasl_cyrus_kinit_refresh (rd_kafka_t *rk) {
         mtx_unlock(&rd_kafka_sasl_cyrus_kinit_lock);
 
         if (r == -1) {
-                rd_kafka_log(rk, LOG_ERR, "SASLREFRESH",
-                             "Kerberos ticket refresh failed: "
-                             "Failed to execute %s",
-                             cmd);
-                rd_free(cmd);
-                return -1;
+                if (errno != ECHILD) {
```
edenhill: This assumes that kinit exited successfully in the case of ECHILD, which might not always be the case.

Honestly, this seems to be more of a problem in pmacct than librdkafka, since pmacct is breaking system() by waitpid:ing all exited pids, not just its own.

hannip: According to the documentation for the system call, they clearly state that ECHILD can safely be ignored.  I know I'm not the only one to waste a lot of time debugging this issue and I'm sure there will be more. It seems that this change is very low risk that will save people time.  At the very least the errno should be logged.
Reference:
https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.1.0/com.ibm.zos.v2r1.bpxbd00/systm.htm
"Note: When using a signal handler and setting the default handler for SIGCHLD to be SIG_IGN, an errno value of ECHILD will be returned. This is due to the speed at which the process executes and finishes before system() can call waitpid() to wait for the child process to end. In this case, the ECHILD can be ignored because the process has already completed successfully and returned."

edenhill: That note is not in the [Linux](http://man7.org/linux/man-pages/man3/system.3.html) or [OSX](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/system.3.html) man pages, and seems to be a Z/OS specific addition which we can't rely on elsewhere.

Logging these at the warning level and returning success seems to be a good middle ground though.

src/rdkafka_sasl_cyrus.c
```diff
@@ -203,12 +203,20 @@ static int rd_kafka_sasl_cyrus_kinit_refresh (rd_kafka_t *rk) {
...
         if (r == -1) {
-                rd_kafka_log(rk, LOG_ERR, "SASLREFRESH",
-                             "Kerberos ticket refresh failed: "
-                             "Failed to execute %s",
-                             cmd);
-                rd_free(cmd);
-                return -1;
+                if (errno == ECHILD) {
+                        rd_kafka_log(rk, LOG_WARNING, "SASLREFRESH",
+                                     "Kerberos ticket refresh command returned ECHILD errno which means ",
```
edenhill: Change to
```
"Kerberos ticket refresh failed with ECHILD: %s: exit status unknown",
cmd);
```
keep under 80 columns width.

src/rdkafka_sasl_cyrus.c
```diff
@@ -203,12 +203,20 @@ static int rd_kafka_sasl_cyrus_kinit_refresh (rd_kafka_t *rk) {
...
+                                     "success is unknown, but probably okay so continuing as such: %s", 
+                                     cmd);
+                }
+                else {
```
edenhill: `} else {`

src/rdkafka_sasl_cyrus.c
```diff
@@ -203,12 +203,20 @@ static int rd_kafka_sasl_cyrus_kinit_refresh (rd_kafka_t *rk) {
...
+                        rd_kafka_log(rk, LOG_ERR, "SASLREFRESH",
+                                     "Kerberos ticket refresh failed, errno=%d: "
```
edenhill:
```
...
"Kerberos ticket refresh failed: %s: %s",
cmd, rd_strerror(errno));
```

src/rdkafka_sasl_cyrus.c
hannip: Sorry about the z/OS reference, but I see similar wording for linux's man page on waitpid() which I assume is what the system call uses to wait for the child process to exit.
At any rate I've made the suggested changes.  Please review.

edenhill:
Thank you!

PR #2428, opened 2019-07-24
fix#2425 rdkafka_example_cpp -L to list metadata can't assign topic

njzcx:
./rdkafka_example_cpp -L -b 172.17.0.2:9092 -t test_topic
./rdkafka_example_cpp -L -b 172.17.0.2:9092
execute above two commands, the result is the same. The reason is topic!=NULL should be !topic in 661 line


edenhill:
Thank you!

PR #2430, opened 2019-07-25
Fix rd_kafka_handle_OffsetCommit

kenneth-jia:
To avoid consumer be stuck while closing if no response for outgoing OffsetCommit.

To reproduce the problem,
1. Start a consumer. (enable.auto.offset.store=false, enable.auto.commit=false, auto.commit.interval.ms=0)
2. Let consumer poll some messages.
3. Block all kafka brokers. (kill -STOP xxx. note: here we keep blocking it)
4. Let consumer commit these offsets (for previously polled messages)
5. Call `rdkafka_consumer_close`, and wait...wait...  It just could NEVER be finished!

Root cause:
Since the OffsetCommitRequest is in the `output queue`, it would keep retrying (`incr_retry` is 0) and refuse admitting failure. Thus, the `rkcg_wait_commit_cnt` could never become 0!

--------------------------------------------------------------------------
THE FOLLOWING DISCLAIMER APPLIES TO ALL SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE:

THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS” AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.

ONLY THE SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE, IF ANY, THAT ARE ATTACHED TO (OR OTHERWISE ACCOMPANY) THIS SUBMISSION (AND ORDINARY COURSE CONTRIBUTIONS OF FUTURES PATCHES THERETO) ARE TO BE CONSIDERED A CONTRIBUTION. NO OTHER SOFTWARE CODE OR MATERIALS ARE A CONTRIBUTION.

### DIFF
--- a/src/rdkafka_request.c
+++ b/src/rdkafka_request.c
@@ -734,8 +734,12 @@ rd_kafka_handle_OffsetCommit (rd_kafka_t *rk,
         int i;
 	int actions;
 
-        if (err)
-		goto err;
+        if (err) {
+		if ((rk->rk_cgrp->rkcg_flags & RD_KAFKA_CGRP_F_TERMINATE) && err == RD_KAFKA_RESP_ERR__TRANSPORT)
+			goto done;
+		else
+			goto err;
+	}
 
         rd_kafka_buf_read_i32(rkbuf, &TopicArrayCnt);
         for (i = 0 ; i < TopicArrayCnt ; i++) {


src/rdkafka_request.c
```diff
@@ -734,8 +734,12 @@ rd_kafka_handle_OffsetCommit (rd_kafka_t *rk,
         int i;
 	int actions;
 
-        if (err)
-		goto err;
+        if (err) {
+		if ((rk->rk_cgrp->rkcg_flags & RD_KAFKA_CGRP_F_TERMINATE) && err == RD_KAFKA_RESP_ERR__TRANSPORT)
```
edenhill: This will prevent retries for the final offset commits, which is something we definitely want, so this will not work.

edenhill: consumer_close() will block up to session.timeout.ms for the final offset commit to finish.

kenneth-jia: > This will prevent retries for the final offset commits, which is something we definitely want, so this will not work.

It would keep retrying (to send offset commits) infinitely (because `incr_retry` is 0), say if the brokers are down, then the consumer could never quit  `rdkafka_consumer_close`. It makes no sense.

kenneth-jia: > consumer_close() will block up to session.timeout.ms for the final offset commit to finish.

No, it would block there forever (while the brokers are down), even the `session.timeout.ms` expires. Only while we let the OffsetCommitRequest response with failure (instead of infinite retrying), the `rkcg_wait_commit_cnt` could get a chance to be `0`, thus trigger the `TERMINATE` event, and then `rdkafka_consumer_close` could quit.

kenneth-jia: > This will prevent retries for the final offset commits, which is something we definitely want, so this will not work.

While the network is down (`RD_KAFKA_RESP_ERR__TRANSPORT`) and the consumer is trying to close, why insist waiting for the network/cluster recovery to commit the offsets? Anyway, we get the OffsetCommit callback and we could get the offsets-commit-error, right?

edenhill: Great analysis of the code!

We can't bail out on ERR__TRANSPORT, the contract is for the consumer to commit its final offsets within the remaining session timeout, and it should retry even if the broker connection goes down. But you're right in that retrying the rkbuf will not work, unless the coordinator comes back up on the same broker.
So I'm suggesting a slightly modified fix where the request fails and the offset commit is put on the cgrp's deferred commit list: this list makes sure the commit is retried when the coordinator is back up, or that the commit is properly timed out after session.timeout.ms.

Please see https://github.com/edenhill/librdkafka/pull/2440
It would be great if you could try it out.

kenneth-jia: I just had a try, -- all our testcases could pass. Cheers! :-)

edenhill:
Thanks for your contribution, this has been fixed in #2440

PR #2433, opened 2019-07-26
fix for processing of enable.ssl.certificate.verification

salisbury-espinosa:
ignoring result of SSL_get_verify_result if **enable_verify == false**
lil chore: using SSL_VERIFY_NONE instead of 0 in SSL_CTX_set_verify
## the problem this patch closes is:
if I have a self-signed and expired certificate, then I can’t get the software to work with the option **enable.ssl.certificate.verification = false**, because the **rd_kafka_transport_ssl_verify** function will always return -1 and the process will always try to reconnect

### DIFF
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -532,6 +532,9 @@ rd_kafka_transport_ssl_io_event (rd_kafka_transport_t *rktrans, int events) {
  * @brief Verify SSL handshake was valid.
  */
 static int rd_kafka_transport_ssl_verify (rd_kafka_transport_t *rktrans) {
+        if (!rktrans->rktrans_rkb->rkb_rk->rk_conf.ssl.enable_verify) {
+                return 0;
+        }
         long int rl;
         X509 *cert;
 
@@ -544,7 +547,7 @@ static int rd_kafka_transport_ssl_verify (rd_kafka_transport_t *rktrans) {
                 return -1;
         }
 
-        if (rktrans->rktrans_rkb->rkb_rk->rk_conf.ssl.enable_verify && (rl = SSL_get_verify_result(rktrans->rktrans_ssl)) != X509_V_OK) {
+        if ((rl = SSL_get_verify_result(rktrans->rktrans_ssl)) != X509_V_OK) {
                 rd_kafka_broker_fail(rktrans->rktrans_rkb, LOG_ERR,
                                      RD_KAFKA_RESP_ERR__SSL,
                                      "Failed to verify broker certificate: %s",
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -544,7 +544,7 @@ static int rd_kafka_transport_ssl_verify (rd_kafka_transport_t *rktrans) {
                 return -1;
         }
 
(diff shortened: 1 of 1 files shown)


src/rdkafka_ssl.c
```diff
@@ -544,7 +544,7 @@ static int rd_kafka_transport_ssl_verify (rd_kafka_transport_t *rktrans) {
                 return -1;
         }
 
-        if ((rl = SSL_get_verify_result(rktrans->rktrans_ssl)) != X509_V_OK) {
+        if (rktrans->rktrans_rkb->rkb_rk->rk_conf.ssl.enable_verify && (rl = SSL_get_verify_result(rktrans->rktrans_ssl)) != X509_V_OK) {
```
edenhill: I suggest simplifying the logic by adding the enable_verify check to the beginning of this function, and if set simply return 0.

src/rdkafka_ssl.c
```diff
@@ -1070,7 +1070,7 @@ int rd_kafka_ssl_ctx_init (rd_kafka_t *rk, char *errstr, size_t errstr_size) {
 
         /* Set up broker certificate verification. */
         SSL_CTX_set_verify(ctx,
-                           rk->rk_conf.ssl.enable_verify ? SSL_VERIFY_PEER : 0,
+                           rk->rk_conf.ssl.enable_verify ? SSL_VERIFY_PEER : SSL_VERIFY_NONE,
```
edenhill: line break after `?` to keep < 80 cols

src/rdkafka_ssl.c
salisbury-espinosa: > and if set simply return 0  

opposite  **is not set** (set to false)

src/rdkafka_ssl.c
```diff
@@ -532,6 +532,9 @@ rd_kafka_transport_ssl_io_event (rd_kafka_transport_t *rktrans, int events) {
  * @brief Verify SSL handshake was valid.
  */
 static int rd_kafka_transport_ssl_verify (rd_kafka_transport_t *rktrans) {
+        if (!rktrans->rktrans_rkb->rkb_rk->rk_conf.ssl.enable_verify) {
```
edenhill: put this at line 541, variable definitions must be at the top of the scope.
remove braces {}
add empty line after

salisbury-espinosa: @edenhill 
> remove braces {}   

librdkafka C++ style guide follows rules from https://google.github.io/styleguide/cppguide.html#Conditionals
so there it is indicated:
`In general, curly braces are not required for single-line statements, but they are allowed if you like them; `
+
it may be worthwhile to connect a tool (linter, formatter) to auto-format a code as specified styleguide, for example https://clang.llvm.org/docs/ClangFormat.html

salisbury-espinosa: equested changes is resolved by https://github.com/edenhill/librdkafka/pull/2433/commits/347ae4a5e9f1a0d434b276af17a17010994b0c5d

edenhill: seems like brew hangs on travis osx sometimes, ignore it

edenhill:
Thank you for your contribution!

PR #2440, opened 2019-07-30
Defer commit on transport error to avoid consumer_close hang

edenhill:


kenneth-jia:
It works, thank a lot!

PR #2442, opened 2019-07-30
Fix memory leak in cpp:Headers

vsakharuk:
Change-Id: Ib0aab2e663a6ec63426605f4bfb527a6ee95fd94

https://github.com/edenhill/librdkafka/issues/2431
This proposed fix for the Headers memory leaks in some case.
Have a nice day.

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -1800,13 +1800,15 @@ public:
      *
      * @remark The error code is used for when the Header is constructed
      *         internally by using RdKafka::Headers::get_last which constructs
-     *         a Header encapsulating the ErrorCode in the process
+     *         a Header encapsulating the ErrorCode in the process.
+     *         If err is set, the value and value_size fields will be undefined.
      */
     Header(const std::string &key,
            const void *value,
            size_t value_size,
            const RdKafka::ErrorCode err):
-    key_(key), err_(err), value_size_(value_size) {
+    key_(key), err_(err), value_(NULL), value_size_(value_size) {
+      if (err == ERR_NO_ERROR)
         value_ = copy_value(value, value_size);
     }
 
@@ -1829,6 +1831,9 @@ public:
       key_ = other.key_;
       err_ = other.err_;
       value_size_ = other.value_size_;
+
+      if (value_ != NULL)
+        free(value_);
 
       value_ = copy_value(other.value_, value_size_);
 
--- a/tests/0085-headers.cpp
+++ b/tests/0085-headers.cpp
@@ -329,6 +329,25 @@ static void test_failed_produce () {
   delete headers;
 }
 
+static void test_assignment_op () {
+  Test::Say("Test Header assignment operator\n");
+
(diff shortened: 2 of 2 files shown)


src-cpp/rdkafkacpp.h
```diff
@@ -1800,14 +1800,17 @@ class RD_EXPORT Headers {
      *
      * @remark The error code is used for when the Header is constructed
      *         internally by using RdKafka::Headers::get_last which constructs
-     *         a Header encapsulating the ErrorCode in the process
+     *         a Header encapsulating the ErrorCode in the process.
+     *         In case of error, value should not be accessed, it ignores
```
edenhill: `value and value_size will be undefined in the case of error.`

src-cpp/rdkafkacpp.h
```diff
@@ -1800,14 +1800,17 @@ class RD_EXPORT Headers {
...
+     *         value argument is err set to error.
      */
     Header(const std::string &key,
            const void *value,
            size_t value_size,
            const RdKafka::ErrorCode err):
-    key_(key), err_(err), value_size_(value_size) {
-        value_ = copy_value(value, value_size);
+    key_(key), err_(err), value_(nullptr), value_size_(value_size) {
```
edenhill: nullptr is C++11, which may not be available, use NULL instead.

src-cpp/rdkafkacpp.h
```diff
@@ -1830,6 +1833,7 @@ class RD_EXPORT Headers {
       err_ = other.err_;
       value_size_ = other.value_size_;
 
+      free(value_);
```
edenhill:
```
if (value_)
  free(value_);
```

src-cpp/rdkafkacpp.h
```diff
@@ -1867,6 +1871,7 @@ class RD_EXPORT Headers {
     }
 
  private:
+    /** @remark: expects initialized value argument. */
```
edenhill: I think that is implicit (and true for most function arguments), so no need to comment it here.

src-cpp/rdkafkacpp.h
vsakharuk: free(nullptr) is defined, but i can change this.

src-cpp/rdkafkacpp.h
vsakharuk: Here is a catch, while your C api states above, Your C++ api does not.
For example, your destructor deallocates array undefined or not.
```
     if (value_ != NULL)
        free(value_);
```
While undefined value_size has no consequence access problems, undefined pointer has.
it is possible to do 2 different things.
1) set value_= NULL when error in single place.
2) if (err==ERR_NO_ERROR && value_ != NULL) <do some> else ...

edenhill: Yes, we should always set value_, either to the allocated space or NULL (on error).
But from the user's perspective they should consider value and size undefined in case of error.

edenhill:
Thank you!

PR #2457, opened 2019-08-06
Set thread names on macOS

benesch:
Like Linux, macOS supports the nonstandard pthread_setname_np, though
the signature is slightly different. Learn to detect and use macOS
variants to, to make multithreaded debugging on macOS a bit easier.


configure.self
```diff
@@ -217,6 +217,15 @@ const char *foo (void) {
 void foo (void) {
   pthread_setname_np(pthread_self(), "abc");
 }
+'
```
edenhill: If SETNAME_GNU works we shouldn't check for DARWIN, so change this to
`' || \`

edenhill:
Seems to be a missing include or feature define for the declaration:
https://travis-ci.org/edenhill/librdkafka/jobs/568176835#L230

configure.self
benesch: Done.

benesch:
> Seems to be a missing include or feature define for the declaration:
https://travis-ci.org/edenhill/librdkafka/jobs/568176835#L230

Indeed, looks like `_DARWIN_C_SOURCE` needs to be set. Note that simply setting it before including pthread.h is insufficient, as it needs to be set for all system headers or none of them (or else various things go wrong), so I added the define for it in rdkafka.h.

Also, heads up: the latest revision of this PR adds the same modifications to the CMake build system.

configure.self
```diff
@@ -217,6 +217,14 @@ const char *foo (void) {
 void foo (void) {
   pthread_setname_np(pthread_self(), "abc");
 }
+' || \
+    mkl_compile_check "pthread_setname_darwin" "HAVE_PTHREAD_SETNAME_DARWIN" disable CC "-lpthread" \
+'
+#include <pthread.h>
```
edenhill: this should have the DARWIN.. define too

packaging/cmake/try_compile/pthread_setname_darwin_test.c
```diff
@@ -0,0 +1,6 @@
+#include <pthread.h>
```
edenhill: and here

src/rd.h
```diff
@@ -38,6 +38,9 @@
 #ifndef _POSIX_C_SOURCE
 #define _POSIX_C_SOURCE 200809L  /* for timespec on solaris */
 #endif
+#ifndef _DARWIN_C_SOURCE
```
edenhill: only do this if on darwin

configure.self
benesch: Sure, done. I put the define in the compiler flags for congruency with the -D_GNU_SOURCE in the GNU check above.

packaging/cmake/try_compile/pthread_setname_darwin_test.c
benesch: Done, but again put the define in the CMake file for congruency with the GNU check.

src/rd.h
benesch: Done, but do note that this is a bit incongruent. _GNU_SOURCE is defined on all platforms but only makes sense on glibc systems, and _POSIX_C_SOURCE is defined on all platforms despite only being needed (according to comment) for Solaris. This inconsistency is fine with me, but wanted to point it out!

configure.self
edenhill: Good thinking, but we actually want to keep the command line argument defines to a minimum since they affect all files, exposing a wider, possibly incompatible API, where it is not desired.
So I suggest adding the ifdef-define-endif before the includes in tinycthread_extra.c

benesch: I'm not sure I follow? I meant I put `-D_DARWIN_C_SOURCE` in the compiler flags for this `mkl_compile_check` invocation, rather than putting a `#define _DARWIN_C_SOURCE` in the actual source file. I don't think that has any bearing on where we define the `_DARWIN_C_SOURCE` macro for compilation of librdkafka itself.

> So I suggest adding the ifdef-define-endif before the includes in tinycthread_extra.c

I think you mean moving the `#define _DARWIN_C_SOURCE` from rd.h to tinycthread_extra.c? That makes sense to me, and so that's what I've done. Let me know if you meant something else.

edenhill: >  I don't think that has any bearing on where we define the _DARWIN_C_SOURCE macro for compilation of librdkafka itself.

Any compiler flags you pass to mkl_compile_check for a passing check will be added to the final compiler flags for the librdkafka build.

benesch: > Any compiler flags you pass to mkl_compile_check for a passing check will be added to the final compiler flags for the librdkafka build.

I know you wrote mklove, so it's certainly possible that I'm wrong here, but I just don't see how that's true? The flags are `$5` on this line https://github.com/edenhill/librdkafka/blob/d979ad2eedfa9b276670dd4f3a0fa3d2505e36d6/mklove/modules/configure.base#L1332
and then `$5` isn't used elsewhere in the function, except to log the flags if the check fails. Are you perhaps thinking of `mkl_lib_check`, which does append its flags to LIBS if the check succeeds? 

Apologies if it seems like I'm being difficult! I just want to make sure this works the same way as the `-D_GNU_SOURCE` flag in the GNU check... and if `mkl_compile_check` works the way you suggest, then the GNU check is misbehaving and setting `-D_GNU_SOURCE` globally when it shouldn't.

edenhill: You are right! I had it confused with mkl_lib_check.
Current solution is fine :+1:

edenhill:
Thank you! :+1:

benesch:
Awesome, thanks @edenhill!

PR #2465, opened 2019-08-09
Fix deadlock between rd_kafka_curr_msgs_{add,sub}

lizthegrey:
If `rk_curr_msgs.cnt` is 0 and `cnt` is more than `rk_curr_msgs.max_cnt`, this method will otherwise never return, because `rd_kafka_curr_msgs_sub` ran too fast.

https://github.com/edenhill/librdkafka/issues/2464
https://github.com/confluentinc/confluent-kafka-go-dev/issues/2

### DIFF
--- a/src/rdkafka_int.h
+++ b/src/rdkafka_int.h
@@ -391,10 +391,12 @@ rd_kafka_curr_msgs_add (rd_kafka_t *rk, unsigned int cnt, size_t size,
 		return RD_KAFKA_RESP_ERR_NO_ERROR;
 
 	mtx_lock(&rk->rk_curr_msgs.lock);
-	while (unlikely(rk->rk_curr_msgs.cnt + cnt >
-			rk->rk_curr_msgs.max_cnt ||
-			(unsigned long long)(rk->rk_curr_msgs.size + size) >
-			(unsigned long long)rk->rk_curr_msgs.max_size)) {
+	/* If rk->rk_curr_msgs.cnt is already 0, nothing will wake us up. */
+	while (unlikely(rk->rk_curr_msgs.cnt > 0 && (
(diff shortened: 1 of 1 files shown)


src/rdkafka_int.h
```diff
@@ -391,10 +391,12 @@ rd_kafka_curr_msgs_add (rd_kafka_t *rk, unsigned int cnt, size_t size,
 		return RD_KAFKA_RESP_ERR_NO_ERROR;
 
 	mtx_lock(&rk->rk_curr_msgs.lock);
-	while (unlikely(rk->rk_curr_msgs.cnt + cnt >
-			rk->rk_curr_msgs.max_cnt ||
-			(unsigned long long)(rk->rk_curr_msgs.size + size) >
-			(unsigned long long)rk->rk_curr_msgs.max_size)) {
+	/* If rk->rk_curr_msgs.cnt is already 0, nothing will wake us up. */
+	while (unlikely(rk->rk_curr_msgs.cnt > 0 && (
```
edenhill: This will allow more than queue.buffering.max.messages on the queue.

But disregarding that, I don't think the case explained will ever happen because curr_msgs_add() is always called with a cnt of 1, which means the original code will never enter into the loop when curr_msgs.cnt is 0, since 0 + max(>0) <= 1.

Unless a message is produced that is larger than the maximum message size, but that should be enforced in produce(), et.al.

lizthegrey: Oh interesting. Sorry for the incorrect diagnosis, you're right that it always adds 1. This is surprising then why a low value of `queue.buffering.max.messages` results in deadlock after `queue.buffering.max.messages` messages are written, there must be some other thing going on.

PR #2468, opened 2019-08-12
Avoid max.poll.interval tracking when in Producer mode

edenhill:


src/rdkafka_int.h
```diff
@@ -658,9 +658,14 @@ rd_kafka_resp_err_t rd_kafka_subscribe_rkt (rd_kafka_itopic_t *rkt);
  */
 static RD_INLINE RD_UNUSED int
 rd_kafka_max_poll_exceeded (rd_kafka_t *rk) {
-        rd_ts_t last_poll = rd_atomic64_get(&rk->rk_ts_last_poll);
+        rd_ts_t last_poll;
         int exceeded;
 
+        if (rk->rk_type != RD_KAFKA_CONSUMER)
```
mhowlett: i don't think this branch will ever be hit, but anyway, it's harmless.

src/rdkafka_int.h
```diff
@@ -690,7 +695,8 @@ rd_kafka_max_poll_exceeded (rd_kafka_t *rk) {
  */
 static RD_INLINE RD_UNUSED void
 rd_kafka_app_poll_blocking (rd_kafka_t *rk) {
-        rd_atomic64_set(&rk->rk_ts_last_poll, INT64_MAX);
+        if (rk->rk_type == RD_KAFKA_CONSUMER)
```
mhowlett: seems fine, though i wonder if it's moderately unpleasant smell related to having all client types share `rd_kafka_t`.

mhowlett:
lgtm

Issue #2202, opened 2019-01-24
Fail faster if broker is down


Tezd:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
I try to produce to single broker, which is unavailable. With setting message.timeout.ms to 100 ms  expectations were that execution time will be be around 100 ms plus time it takes to destroy kafka resources. Due to https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.c#L1708 (master) or https://github.com/edenhill/librdkafka/blob/v0.11.6/src/rdkafka.c#L1399 (v0.11.6) shutdown time will be in steps of 1 second(ex. message.timeout.ms 1500ms will result in shutdown of 2 sec). Important part I do want to trigger dr_cb on failed messages. Question is there any way of speeding up failure in this kind of situations or maybe make window for **tmr_topic_scan** configurable?

How to reproduce
================

1. Bring up environment using [Dockerfile](https://github.com/edenhill/librdkafka/files/2792425/Dockerfile.txt)
2. Build librdkafka inside of the container: `./configure && make && make install`
3. Add following [producer.c](https://github.com/edenhill/librdkafka/files/2792400/producer.txt) to /app:` docker cp producer.c <container_name>:/app`
4. Build producer inside of the container: `gcc -g -O2 -fPIC -Wall -Wsign-compare -Wfloat-equal -Wpointer-arith -Wcast-align -I/usr/local/include -I./src producer.c -o producer_poc src/librdkafka.a -L/lib -lsasl2 -lssl -L/usr/local/lib -llz4 -lm -lcrypto -L/lib -lz -ldl -lpthread -lrt`
5. Run: `docker exec -it <container_name> time /app/producer_poc`
6. Verify that it took ~1s to complete

Checklist
=========

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `v0.11.6`
 - [x] Apache Kafka version: `1.0.0` (not used)
 - [x] librdkafka client configuration: `config: debug=all, topic_config: message.timeout.ms=100`
 - [x] Operating system: `alpine:2.8`
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka [producer.log](https://github.com/edenhill/librdkafka/files/2792417/producer.log)
 - [ ] Provide broker log excerpts (not used)
 - [ ] Critical issue

edenhill:
There currently isn't, the scanning will be on 1s intervals.

[...]

PR #2471, opened 2019-08-13
Producer: Messages were not timed out for leader-less partitions

edenhill:
https://github.com/confluentinc/confluent-kafka-dotnet/issues/1027

This also improves the granularity of message timeout scans to
match the actual message timeout. (#2202)
The minimum effective message timeout will still be 1000ms, but the
granularity after the initial 1000ms is now on millisecond precision.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3050,14 +3050,14 @@ static void rd_kafka_broker_ops_io_serve (rd_kafka_broker_t *rkb,
 
 
 /**
- * @brief Serve the toppar's assigned to this broker.
+ * @brief Consumer: Serve the toppars assigned to this broker.
  *
  * @returns the minimum Fetch backoff time (abs timestamp) for the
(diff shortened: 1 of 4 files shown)


src/rdkafka_broker.c
```diff
@@ -3074,43 +3074,39 @@ static rd_ts_t rd_kafka_broker_toppars_serve (rd_kafka_broker_t *rkb) {
...
         int xcnt, qcnt, cnt;
         uint64_t first, last;
+        rd_ts_t next;
+
+        *abs_next_timeout = 0;
 
         xcnt = rd_kafka_msgq_age_scan(rktp, &rktp->rktp_xmit_msgq,
-                                      &xtimedout, now);
+                                      &xtimedout, now, &next);
+        if (next && (!*abs_next_timeout || next < *abs_next_timeout))
```
mhowlett: `!*abs_next_timeout` is always true

Issue #2455, opened 2019-08-05
[Question] Is there any way to sync with results of async operations?


abyss7:
The operations of interest are:
- pause/resume
- subscribe/unsubscribe

The related calls to `rd_kafka_resume_partitions()`, `rd_kafka_pause_partitions()`, `rd_kafka_subscribe()`, etc. just put a task into some queue which is processed by another thread async. I can't find any high-level mechanism (like future-promise) to sync with a result of these operations, and can't figure out how to check if operation is completed within a couple lines of code.

The basic scenario:
1. Check if consumer is subscribed to anything.
2. If not, then pause all consumption - to not accidentally read any messages while polling for subscription messages. Wait for pause to apply.
3. Subscribe and poll with timeout until anything gets assigned.
4. Resume and wait for it - especially if subscription gone wrong and failed.

abyss7:
The better description of high-level scenario.

If consumer is not subscribed (or assigned) to anything, then subscribe it in sync way, make sure it's resumed, and make sure that next call to `poll()` will return new messages if there are any on broker-side.

edenhill:
pause/resume..() are currently asynchronous, as you say, but it is probably a good idea to make them synchronous.

edenhill:
Fixed

PR #2473, opened 2019-08-14
Make rd_kafka_pause|resume_partitions() synchronous (#2455)

edenhill:
This makes sure that a consumer_poll() call after pause() will not
return any messages.

### DIFF
--- a/src/rdkafka_subscription.c
+++ b/src/rdkafka_subscription.c
@@ -171,16 +171,22 @@ rd_kafka_subscription (rd_kafka_t *rk,
 
 rd_kafka_resp_err_t
 rd_kafka_pause_partitions (rd_kafka_t *rk,
-			   rd_kafka_topic_partition_list_t *partitions) {
-	return rd_kafka_toppars_pause_resume(rk, 1, RD_KAFKA_TOPPAR_F_APP_PAUSE,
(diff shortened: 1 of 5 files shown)


src/rdkafka_subscription.c
```diff
@@ -171,16 +171,22 @@ rd_kafka_subscription (rd_kafka_t *rk,
 
 rd_kafka_resp_err_t
 rd_kafka_pause_partitions (rd_kafka_t *rk,
-			   rd_kafka_topic_partition_list_t *partitions) {
-	return rd_kafka_toppars_pause_resume(rk, 1, RD_KAFKA_TOPPAR_F_APP_PAUSE,
-					     partitions);
+                           rd_kafka_topic_partition_list_t *partitions) {
+        return rd_kafka_toppars_pause_resume(rk,
+                                             rd_true/*pause*/,
+                                             rd_false/*sync*/,
```
mhowlett: this confused me on first glance. I read the sync comment to mean the argument name, not the behavior.

edenhill: I added a new enum type, rd_async_t, with RD_SYNC and RD_ASYNC to make this, and other places in the future, more readable.

Issue #2466, opened 2019-08-10
ERR__ALL_BROKERS_DOWN if no topics are referenced


sebwills:
Description
===========
Using the high-level producer API, if no messages are produced after the producer is created for longer than N times the server's connections.max.idle.ms (where N is the number of brokers), the client logs an ERROR like '4/4 brokers are down' before cycling round and reconnecting to the first broker. The client recovers, but the error being logged is undesirable (see #915, #1101, etc). Subsequently, the same error is logged each time a broker disconnects, until the first message is sent.

This may be happening because the high level producer does not start doing topic metadata refresh until the first message is produced. The issue does not occur with kafkacat because (I think) it is not using the high-level producer, and begins metadata refreshes immediately. The 'N x connections.max.idle.ms' behaviour comes from the fact that librdkafka connects to each broker in turn as the previous one disconnects, only logging the error when the last one disconnects.

How to reproduce
================

1. Create a producer using the high-level API (I used the Confluent.Kafka library in C#, but it's just a thin wrapper for librdkafka). Keep it open but don't send any messages:
```
var config = new ProducerConfig { BootstrapServers = "localhost:9092"};

using (var p = new ProducerBuilder<Null, string>(config).Build())
{
    Console.WriteLine("Sleeping");
    Thread.Sleep(TimeSpan.FromHours(10));
}
```
2. After (the number of brokers you have) x (connections.max.idle.ms), observe an ERROR 'N/N brokers are down'. For example, with default settings and 3 brokers, this happens after 30 minutes.
3. After this, the same error is logged every connections.max.idle.ms.

Log from repro with a single broker, so the first error happens after 10 minutes:

```
%7|1565381028.446|BRKMAIN|rdkafka#producer-1| [thrd::0/internal]: :0/internal: Enter main broker thread
%7|1565381028.446|WAKEUPFD|rdkafka#producer-1| [thrd:app]: localhost:9092/bootstrap: Enabled low-latency ops queue wake-ups
%7|1565381028.446|BROKER|rdkafka#producer-1| [thrd:app]: localhost:9092/bootstrap: Added new broker with NodeId -1
%7|1565381028.446|BRKMAIN|rdkafka#producer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/bootstrap: Enter main broker thread
[...]

PR #2476, opened 2019-08-15
Refresh broker list even if no topics to refresh (#2466)

edenhill:


CONFIGURATION.md
```diff
@@ -13,7 +13,7 @@ receive.message.max.bytes                |  *  | 1000 .. 2147483647 |     100000
...
 metadata.request.timeout.ms              |  *  | 10 .. 900000    |         60000 | low        | Non-topic request timeout in milliseconds. This is for metadata requests, etc. <br>*Type: integer*
-topic.metadata.refresh.interval.ms       |  *  | -1 .. 3600000   |        300000 | low        | Topic metadata refresh interval in milliseconds. The metadata is automatically refreshed on error and connect. Use -1 to disable the intervalled refresh. <br>*Type: integer*
+topic.metadata.refresh.interval.ms       |  *  | -1 .. 3600000   |        300000 | low        | Topic metadata refresh interval in milliseconds. The metadata is automatically refreshed on error and connect. Use -1 to disable the intervalled refresh. If there are no locally referenced topics (no topic objects created, no messages produced, no subscription or no assignment) then only the broker list will be refreshed every interval but no more often than every 10s. <br>*Type: integer*
```
mhowlett: This description caused me to pause. Initially, because I felt it was too verbose, but on reflection, I think it's because no reason for the refresh was stated. Here's an alternative to consider (the first part somewhat inspired by the java docs). I left it in, but use of the word 'error' is debatable I think since this refers to re-tryable error most of the time I think - that the user doesn't see - and so is not an error from the users perspective (or at best the word is ambiguous from the users perspective).

> "The period of time in milliseconds after which a metadata refresh is forced in order to proactively discover any new brokers or partitions. Metadata is automatically refreshed on error and connect. Use -1 to disable the intervalled refresh. If there are no locally referenced topics (no topic objects created, no messages produced, no subscription or no assignment) then only the broker list will be refreshed every interval but no more often than every 10s."

mhowlett:
lgtm. probably worth reworking the doc string a bit i think.

Issue #2478, opened 2019-08-16
if "rd_kafka_new" return NULL, unable to determine that app_conf was released.


David-xin:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
I read the code and found that if the return value of "rd_kafka_new" is NULL, the return value cannot be used to judge that app_conf is released.

How to reproduce
================
rd_kafka_t *rd_kafka_new (rd_kafka_type_t type, rd_kafka_conf_t *app_conf,
			  char *errstr, size_t errstr_size) {
	rd_kafka_t *rk;
	static rd_atomic32_t rkid;
        rd_kafka_conf_t *conf;
        rd_kafka_resp_err_t ret_err = RD_KAFKA_RESP_ERR_NO_ERROR;
        int ret_errno = 0;
        const char *conf_err;
#ifndef _MSC_VER
        sigset_t newset, oldset;
#endif
        char builtin_features[128];
        size_t bflen;

        rd_kafka_global_init();

        if (!app_conf)
                conf = rd_kafka_conf_new();
        else
                conf = app_conf;

        /* Verify and finalize configuration */
        if ((conf_err = rd_kafka_conf_finalize(type, conf))) {---------------if true
                /* Incompatible configuration settings */
                rd_snprintf(errstr, errstr_size, "%s", conf_err);
                if (!app_conf)--------------------------------------------------if app_conf is not NULL
                        rd_kafka_conf_destroy(conf);
                rd_kafka_set_last_error(RD_KAFKA_RESP_ERR__INVALID_ARG, EINVAL);
                return NULL;----------------------------------return NULL, not release app_conf.
        }
...
But, Other return NULL and release app_conf.

**IMPORTANT**: Always try to reproduce the issue on the latest released version (see https://github.com/edenhill/librdkafka/releases), if it can't be reproduced on the latest version the issue has been fixed.

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `1.0`

edenhill:
I don't think this is ever a problem in practice, the double free will only occur if the background threads fail to start after 60s, in which case all bets are off.
But fixed the code anyway.

David-xin:
Thank you for adopting my opinion.
In addition, I found that in "rd_kafka_new", if the function return NULL via "goto fail", there is no free app_conf, please check it.

edenhill:
[...]

PR #2479, opened 2019-08-16
Tidy up rd_kafka_new() conf freeing to make code clearer (#2478)

edenhill:
This is not a problem in practice, but the code was wrong, so fixed it.


mhowlett:
for reference: https://github.com/edenhill/librdkafka/issues/2478

PR #2482, opened 2019-08-19
Fix memory leak in aborted transactions

edenhill:


tests/0099-commit_metadata.c
```diff
@@ -157,6 +157,8 @@ get_committed_metadata (const char *group_id,
 
         compare_toppar_lists(committed_toppar, expected_toppar);
 
+        rd_kafka_topic_partition_list_destroy(committed_toppar);
```
mhowlett: ... glad to see i'm not alone

PR #2485, opened 2019-08-21
Fixes for issues found by Coverity static code analysis

edenhill:
https://scan.coverity.com/projects/edenhill-librdkafka?tab=overview

Let's get this merged after 1.2.0

### DIFF
--- a/src/rdbuf.c
+++ b/src/rdbuf.c
@@ -1483,7 +1483,7 @@ static int do_unittest_write_read_payload_correctness (void) {
                 for (i = 0 ; i < max_cnt ; i++) {
                         uint32_t buf_crc;
 
-                        crc = rd_crc32_update(crc, (void *)&i, sizeof(&i));
+                        crc = rd_crc32_update(crc, (void *)&i, sizeof(i));
 
                         if (pass == 0)
                                 r = rd_slice_peek(&slice, i * sizeof(buf_crc),
--- a/src/rdbuf.c
+++ b/src/rdbuf.c
@@ -1167,7 +1167,8 @@ void rd_buf_dump (const rd_buf_t *rbuf, int do_hexdump) {
                         rbuf->rbuf_segment_cnt);
                 TAILQ_FOREACH(seg, &rbuf->rbuf_segments, seg_link) {
                         rd_segment_dump(seg, "  ", 0, do_hexdump);
-                        rd_assert(++segcnt <= rbuf->rbuf_segment_cnt);
+                        segcnt++;
+                        rd_assert(segcnt <= rbuf->rbuf_segment_cnt);
                 }
         }
 }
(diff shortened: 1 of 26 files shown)


src/rdbuf.c
```diff
@@ -1167,7 +1167,8 @@ void rd_buf_dump (const rd_buf_t *rbuf, int do_hexdump) {
                         rbuf->rbuf_segment_cnt);
                 TAILQ_FOREACH(seg, &rbuf->rbuf_segments, seg_link) {
                         rd_segment_dump(seg, "  ", 0, do_hexdump);
-                        rd_assert(++segcnt <= rbuf->rbuf_segment_cnt);
+                        segcnt++;
+                        rd_assert(segcnt <= rbuf->rbuf_segment_cnt);
```
mhowlett: the old way seems marginally better in that it's terser? but either/or.

edenhill: Thing is some asserts may be disabled at compile time, in which case the increment wouldn't happen at all.
So better safe than sorry here.

src/rdkafka_broker.c
```diff
@@ -1676,7 +1676,10 @@ int rd_kafka_socket_cb_generic (int domain, int type, int protocol,
         if (s == -1)
                 return -1;
 #ifdef FD_CLOEXEC
-        fcntl(s, F_SETFD, FD_CLOEXEC, &on);
+        if (fcntl(s, F_SETFD, FD_CLOEXEC, &on) == -1)
+                fprintf(stderr, "WARNING: librdkafka: %s: "
```
mhowlett: can you do something else with logging here? people using a custom librdkafka log handler will not see this.

src/rdkafka_cgrp.c
```diff
@@ -3402,6 +3404,9 @@ void rd_kafka_cgrp_handle_SyncGroup (rd_kafka_cgrp_t *rkcg,
         if (rkbuf)
                 rd_kafka_buf_destroy(rkbuf);
 
+        if (assignment)
+                rd_kafka_topic_partition_list_destroy(assignment);
```
mhowlett: nice catch

src/rdkafka_idempotence.c
```diff
@@ -163,7 +163,7 @@ int rd_kafka_idemp_request_pid (rd_kafka_t *rk, rd_kafka_broker_t *rkb,
                 return 1;
         }
 
-        rd_kafka_wrunlock(rkb->rkb_rk);
+        rd_kafka_wrunlock(rk);
```
mhowlett: do you also want to make this change in the if block above?

src/rdkafka_broker.c
edenhill: Yeah that'd be good but unfortunately we don't have access to the rk here.
This warning is just a best effort thing to let the user know CLOEXEC failed, which should be extremely rare, if possible at all.

src/rdkafka_idempotence.c
edenhill: good catch!

PR #2495, opened 2019-08-27
Changed isolation.level default to read_committed

mhowlett:


moihn:
It seems the confluent page has a different default value for it, without clearly state difference in language bindings. Should the confluent page be updated? 
https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html

PR #2511, opened 2019-09-05
Fix VS 2017 build

nicklauslittle:
This replaces a couple c boolean values (true, false) with the built-in rdkafka boolean values (rd_true, rd_false).


edenhill:
Thank you!

Issue #993, opened 2017-01-10
Cannot produce message with size nearing message.max.bytes


ozac:
Description
===========
Using default `message.max.bytes` (1,000,000).
Trying to produce a message with size ranges from 999,918 to 1,000,000 gets into a loop.

Here is a sample output of rdkafka_example.exe:
```
Waiting for 1
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-PRODUCE: mybroker:9092/0: No more space in current message (0 messages)
Waiting for 1
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
Waiting for 1
LOG-7-PRODUCE: mybroker:9092/0: No more space in current message (0 messages)
LOG-7-TOPPAR: mybroker:9092/0: mytopic [0] 0+1 msgs
[...]

PR #2513, opened 2019-09-07
Less strict message.max.bytes check for individual messages (#993)

edenhill:
Since the final request size can't be known at produce() time
we allow ProduceRequests larger than message.max.bytes (overshot by at
most one message) and instead rely on the broker enforcing the
MessageSet size.


src/rdkafka_conf.c
```diff
@@ -271,9 +271,11 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
 	{ _RK_GLOBAL|_RK_HIGH, "bootstrap.servers", _RK_C_ALIAS, 0,
 	  "See metadata.broker.list",
 	  .sdef = "metadata.broker.list" },
-	{ _RK_GLOBAL|_RK_MED, "message.max.bytes", _RK_C_INT, _RK(max_msg_size),
-	  "Maximum Kafka protocol request message size.",
-	  1000, 1000000000, 1000000 },
+        { _RK_GLOBAL|_RK_MED, "message.max.bytes", _RK_C_INT, _RK(max_msg_size),
+          "Maximum Kafka protocol request message size. "
+          "For ProduceRequest this limit may be exceeded by the size of "
+          "one message, the actual MessageSet size is enforced by the broker.",
```
rnpridgeon: It's probably a good idea(nice to have) to include the reasoning for why this is an imperfect limit in the documentation. 

https://github.com/edenhill/librdkafka/issues/993#issuecomment-448205377

src/rdkafka_msg.c
```diff
@@ -169,14 +169,13 @@ static rd_kafka_msg_t *rd_kafka_msg_new0 (rd_kafka_itopic_t *rkt,
...
 
-	if (unlikely(len + keylen + hdrs_size >
-		     (size_t)rkt->rkt_rk->rk_conf.max_msg_size ||
-		     keylen > INT32_MAX)) {
-		*errp = RD_KAFKA_RESP_ERR_MSG_SIZE_TOO_LARGE;
-		if (errnop)
-			*errnop = EMSGSIZE;
-		return NULL;
-	}
+        if (unlikely(rd_kafka_msg_max_wire_size(keylen, len, hdrs_size) >
```
rnpridgeon: Should keylen still be restricted to 32bits?

src/rdkafka_conf.c
edenhill: Good point!

src/rdkafka_msg.c
edenhill: Yes man! And `len`, to avoid overflow in max_wire..

edenhill:
Fixed comments, added test. Please re-review

CONFIGURATION.md
```diff
@@ -7,7 +7,7 @@ builtin.features                         |  *  |                 | gzip, snappy,
...
 bootstrap.servers                        |  *  |                 |               | high       | Alias for `metadata.broker.list`: Initial list of brokers as a CSV list of broker host or host:port. The application may also use `rd_kafka_brokers_add()` to add brokers during runtime. <br>*Type: string*
-message.max.bytes                        |  *  | 1000 .. 1000000000 |       1000000 | medium     | Maximum Kafka protocol request message size. <br>*Type: integer*
+message.max.bytes                        |  *  | 1000 .. 1000000000 |       1000000 | medium     | Maximum Kafka protocol request message size. Due to differing framing overhead between protocol versions the producer is unable to reliably enforce a strict max message limit at produce time and may exceed the maximum size by one message in protocol ProduceRequests, the broker will enforce the the topic's `max.message.bytes` limit (see Apache Kafka documentation). <br>*Type: integer*
```
rnpridgeon: nit: link apache doc?

PR #2518, opened 2019-09-09
Fetch from follower

mhowlett:


mhowlett:
it looks like there's a problem with rapidjson & I've also made a minor change locally to trivup, but these are small things and this is otherwise ready to review.

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,294 @@
...
+  static bool should_capture_stats;
+  static bool has_captured_stats;
+  static std::map<int32_t, int64_t> rxbytes;
+
+  void event_cb (RdKafka::Event &event) {
+
+    switch (event.type())
+    {
+      case RdKafka::Event::EVENT_LOG:
+        Test::Say(tostr() << event.str().c_str() << "\n");
```
edenhill: `Test::Say(event.str() + "\n");` is sufficient

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,294 @@
...
+    idx++;
+  }
+
+  delete metadata;
+  delete topic;
+  delete p;
+}
+
+
+static void do_fff_test (void) {
```
edenhill: Add docstring outlining what is being tested and how.

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,294 @@
...
+  const int msgsize = 100;
+  std::vector<int> replica_ids;
+  int leader_id;
+  int client_rack_id;
+  std::string client_rack, topic_str;
+  RdKafka::Conf *conf;
+  RdKafka::KafkaConsumer *c;
+  std::vector<std::string> topics;
+  std::string errstr;
+  RdKafka::ErrorCode err;
```
edenhill: C++ on the other hand does not require all variable defs at top of scope

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,294 @@
...
+  topic_str = Test::mk_topic_name("0101-fetch-from-follower", 1);
+  test_create_topic(NULL, topic_str.c_str(), 1, 3);
+  test_produce_msgs_easy_size(topic_str.c_str(), 0, 0, msgcnt, msgsize);
+
+  get_brokers_info(topic_str, &leader_id, replica_ids);
+  test_assert(replica_ids.size() == 3, tostr() << "expecting three replicas, but " << replica_ids.size() << " were reported.");
+  Test::Say(tostr() << topic_str << " leader id: " << leader_id << ", all replica ids: [" << replica_ids[0] << ", " << replica_ids[1] << ", " << replica_ids[2] << "]\n");
+
+  /* arrange for the consumer's client.rack to align with a broker that is not the leader. */
+  for (i=0; i<replica_ids.size(); ++i) {
```
edenhill: Add error handling for when this is not possible, typically because env was not set up by trivup.

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,294 @@
...
+  Test::Say(tostr() << "client.rack: " << client_rack << "\n");
+
+  Test::conf_init(&conf, NULL, 10);
+  Test::conf_set(conf, "group.id", topic_str);
+  Test::conf_set(conf, "auto.offset.reset", "earliest");
+  Test::conf_set(conf, "enable.auto.commit", "false");
+  Test::conf_set(conf, "statistics.interval.ms", "1000");
+  conf->set("event_cb", &ex_event_cb, errstr);
+  Test::conf_set(conf, "client.rack", client_rack);
+  Test::conf_set(conf, "debug", "all");
```
edenhill: Remove. Use `TEST_DEBUG=all TESTS=0101 make` instead.

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,294 @@
...
+  c = RdKafka::KafkaConsumer::create(conf, errstr);
+  test_assert(c, "Failed to create KafkaConsumer: " + errstr);
+  delete conf;
+
+  /* Subscribe */
+  topics.push_back(topic_str);
+  if ((err = c->subscribe(topics)))
+    Test::Fail("subscribe failed: " + RdKafka::err2str(err));
+
+  /* Start consuming */
```
edenhill: It needs to validate that the correct messages are indeed returned, and in the correct order.

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,294 @@
...
+      test_assert(TestEvent2Cb::rxbytes[replica_ids[i]] < TestEvent2Cb::rxbytes[client_rack_id],
+                  "rxbytes was not highest on broker corresponding to client.rack.");
+
+  test_assert(TestEvent2Cb::rxbytes[client_rack_id] > msgcnt * msgsize,
+              tostr() << "expecting rxbytes of client.rack broker to be at least " << msgcnt * msgsize
+                      << " but it was " << TestEvent2Cb::rxbytes[client_rack_id]);
+
+  Test::Say("Done\n");
+
+  // Manual test 1:
```
edenhill: What's the plan with this?

tests/interactive_broker_version.py
```diff
@@ -52,6 +58,9 @@ def test_version (version, cmd=None, deploy=True, conf={}, debug=False, exec_cnt
 
     brokers = []
     for n in range(0, broker_cnt):
+        # Configure rack & replica selector if broker supports fetch-from-follower
+        if version_as_number(version) >= 2.4:
+            defconf.update({'conf': ['broker.rack=RACK${appid}', 'replica.selector.class=org.apache.kafka.common.replica.RackAwareReplicaSelector']})
```
edenhill: This needs to go in cluster_testing.py as well.

But, I'm not sure we want to change the default test environment to use fff for all tests since that changes our test coverage.
Instead I suggest you make this an option to interactive_broker_version.py and broker_version_test.py and add a new Make target to set up a rack-aware environment and run all tests, see the `broker_idempotence` target for an example.

src/rdinterval.h
```diff
@@ -95,6 +96,19 @@ static RD_INLINE RD_UNUSED void rd_interval_reset (rd_interval_t *ri) {
         ri->ri_backoff = 0;
 }
 
+/**
+ * Reset the interval to 'now'. If now is 0, the time will be gathered
+ * automatically.
+ */
+static RD_INLINE RD_UNUSED void rd_interval_reset_to_now (rd_interval_t *ri,
```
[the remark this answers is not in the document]
mhowlett: as you suggest was my first thought too. I specifically did it the way I have because i think it makes the api clearer. if done as you suggest, the now parameter would need to be tri-state (one value to say reset is immediate, another to say use current time, another to specify a time). also, it makes the comments more long winded and difficult to interpret in a number of places (the reset method, and the rd_interval0 method which explains when the interval will trigger immediately). as is, they are straightforward.

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+/**
+ * @brief Handle preferred replica in fetch response.
+ */
+void
+rd_kafka_fetch_preferred_replica_handle (rd_kafka_toppar_t *rktp,
+                                         rd_kafka_buf_t *rkbuf,
+                                         rd_kafka_broker_t *rkb,
+                                         int32_t preferred_id) {
+        if (rd_interval_immediate(&rktp->rktp_new_lease_intvl,
+                                  60*1000000/*1 minute*/, 0) < 0) {
```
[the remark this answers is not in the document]
mhowlett: good idea

src/rdkafka_broker.c
```diff
@@ -2539,12 +2540,13 @@ static void rd_kafka_broker_map_partitions (rd_kafka_broker_t *rkb) {
...
                         rd_kafka_toppar_lock(rktp);
-                        if (rktp->rktp_leader_id == rkb->rkb_nodeid &&
-                            !(rktp->rktp_leader && rktp->rktp_next_leader)) {
-                                rd_kafka_toppar_leader_update(
-                                        rktp, rktp->rktp_leader_id, rkb);
+                        if (rktp->rktp_broker_id == rkb->rkb_nodeid &&
+                            !(rktp->rktp_broker && rktp->rktp_next_broker)) {
+                                rd_kafka_toppar_broker_update(
+                                        rktp, rktp->rktp_broker_id, rkb,
+                                        rd_true);
```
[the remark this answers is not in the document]
mhowlett: right. i believe the behavior is correct as is because `rktp_leader_id` will never be different from `rktp_broker_id` if `rktp_broker` is `NULL`, however it's bad form to write this. i believe switching out `rktp_broker_id` for `rktp_leader_id` here is what we want.

generally, I found this all quite tricky to think through. I have a feeling a better structure is possible (but it's not immediately apparent what). Please check the interplay between `rktp_broker`, `rktp_leader_id`, and `rktp_broker_id` very carefully.

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+rd_kafka_fetch_preferred_replica_handle (rd_kafka_toppar_t *rktp,
+                                         rd_kafka_buf_t *rkbuf,
+                                         rd_kafka_broker_t *rkb,
+                                         int32_t preferred_id) {
+        if (rd_interval_immediate(&rktp->rktp_new_lease_intvl,
+                                  60*1000000/*1 minute*/, 0) < 0) {
+                rd_rkb_log(rkb, LOG_WARNING, "FETCH",
+                        "%.*s [%"PRId32"]: Not migrating to preferred "
+                        "replica %"PRId32" because previous lease was "
+                        "granted < 1 minute ago",
```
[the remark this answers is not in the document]
mhowlett: good idea

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+                        RD_KAFKAP_STR_PR(rktp->rktp_rkt->rkt_topic),
+                        rktp->rktp_partition, preferred_id);
+
+                rd_kafka_toppar_fetch_backoff(rkb,
```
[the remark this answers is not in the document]
mhowlett: this is partly addressing your concern above (logs being flooded with messages), though it's doing a fairly poor job of that. what are you thinking - a suppression `rd_interval` just for this? 

if we get here, something is up. we've bounced back from a preferred replica early indicating a problem there (though note: see below) and I think it's likely the leader is just going to bounce us right back again. regardless, this is an infrequent error scenario and 500ms is not a big deal in that context, so I was thinking better to err on the side of conservative. feel free to argue that's not a good way of looking at it.

I realized I made a mistake related to the above - this scenario can happen on a leader change since that could happen at any time after the preferred replica lease is granted (possibly before 1 minute). when that happens, there should be an `rd_interval_expedite` to prevent this case getting hit. will make that change.

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+ * @brief Handle preferred replica in fetch response.
+ */
+void
+rd_kafka_fetch_preferred_replica_handle (rd_kafka_toppar_t *rktp,
+                                         rd_kafka_buf_t *rkbuf,
+                                         rd_kafka_broker_t *rkb,
+                                         int32_t preferred_id) {
+        if (rd_interval_immediate(&rktp->rktp_new_lease_intvl,
+                                  60*1000000/*1 minute*/, 0) < 0) {
+                rd_rkb_log(rkb, LOG_WARNING, "FETCH",
```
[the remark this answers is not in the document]
mhowlett: see & respond in below comment.

src/rdinterval.h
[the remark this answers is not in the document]
edenhill: Sounds good :+1:

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+                        "%.*s [%"PRId32"]: Not migrating to preferred "
+                        "replica %"PRId32" because previous lease was "
+                        "granted < 1 minute ago",
+                        RD_KAFKAP_STR_PR(rktp->rktp_rkt->rkt_topic),
+                        rktp->rktp_partition, preferred_id);
+
+                rd_kafka_toppar_fetch_backoff(rkb,
+                        rktp, RD_KAFKA_RESP_ERR_NO_ERROR);
+        } else {
```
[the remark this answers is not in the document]
mhowlett: yep, of course. this is a hangover from when this wasn't in it's own method.

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+                rd_kafka_t *rk = rktp->rktp_rkt->rkt_rk;
+
+                rd_kafka_rdlock(rk);
+                preferred_rkb = rd_kafka_broker_find_by_nodeid(rk,
+                                                               preferred_id);
+                rd_kafka_rdunlock(rk);
+
+                if (preferred_rkb) {
+                        rd_kafka_topic_wrlock(rktp->rktp_rkt);
+                        rd_interval_reset_to_now(&rktp->rktp_lease_intvl, 0);
```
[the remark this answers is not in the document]
mhowlett: the `rkt` lock is there because it's required by `rd_kafka_toppar_broker_update` -> `rd_kafka_toppar_broker_delegate`. It would be possible to refactor the function structure in order to minimize how long the lock is held c.f. the current code, but it would be more verbose and I think the difference would be inconsequential.

you're right, no lock is required for the lease interval, and i've moved this up a line.

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+                        rd_kafka_toppar_broker_update(rktp, preferred_id,
```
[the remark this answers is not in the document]
mhowlett: yes, and @locks added.

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+                                                      preferred_rkb, rd_false);
+                        rd_kafka_topic_wrunlock(rktp->rktp_rkt);
+                        rd_kafka_broker_destroy(preferred_rkb);
+                } else {
+                        rd_kafka_metadata_refresh_brokers (
+                                rktp->rktp_rkt->rkt_rk, NULL,
+                                "preferred replica unavailable");
+
+                        rd_kafka_toppar_fetch_backoff(rkb,
+                                rktp, RD_KAFKA_RESP_ERR_NO_ERROR);
```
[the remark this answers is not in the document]
mhowlett: yep, re: `ERR__UNKNOWN_BROKER`, not sure why I was thinking that didn't deserve an error.

if we get here, we just received a response from the leader indicating that it's not going to be giving us any messages and we should get them from a replica - but this not known to us. The next time we fetch from the leader, it's most likely going to send back the same response. so I think it makes sense to back off. I say that being not 100% sure how long poll works in this scenario. If the broker responds to the fetch immediately with preferred replica (my assumption), i think we definitely want to slow it down. If not, probably not.

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,62 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+
+                if (preferred_rkb) {
+                        rd_kafka_topic_wrlock(rktp->rktp_rkt);
+                        rd_interval_reset_to_now(&rktp->rktp_lease_intvl, 0);
+                        rd_kafka_toppar_broker_update(rktp, preferred_id,
+                                                      preferred_rkb, rd_false);
+                        rd_kafka_topic_wrunlock(rktp->rktp_rkt);
+                        rd_kafka_broker_destroy(preferred_rkb);
+                } else {
+                        rd_kafka_metadata_refresh_brokers (
```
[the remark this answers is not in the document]
mhowlett: ah yes, whoops. we don't want this code doing the metadata request - it'll be called frequently. I believe `rd_kafka_toppar_broker_unavailable` is exactly what is required - flag that we need metadata, and let the getting of that be managed elsewhere.

src/rdkafka_broker.c
```diff
@@ -3658,6 +3709,13 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
 					  Throttle_Time);
 	}
 
+        if (rd_kafka_buf_ApiVersion(request) >= 7) {
+                int16_t ErrorCode;
+                int32_t SessionId;
+		rd_kafka_buf_read_i16(rkbuf, &ErrorCode);
```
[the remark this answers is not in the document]
mhowlett: hopefully fixed. VS Code tries to guess this based on existing file contents.

src/rdkafka_broker.c
```diff
@@ -3658,6 +3709,13 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
 					  Throttle_Time);
 	}
 
+        if (rd_kafka_buf_ApiVersion(request) >= 7) {
+                int16_t ErrorCode;
```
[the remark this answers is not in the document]
mhowlett: my implicit assumption was we'd never get an error here since it's only related to KIP-227, but that's a really bad assumption (though likely true at this point based on every thing I know). well pointed out... In the kip it says:

> When the top-level error code is set, the caller should assume that all the partitions in the fetch received the given error.

so i'm going to do that - blindly replace the per partition error variable with this if set. then let it fall through to the default handler which creates an error op (non-retryable error).

src/rdkafka_broker.c
```diff
@@ -3747,10 +3810,17 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
                                                 rd_kafka_aborted_txns_sort(aborted_txns);
                                         }
                                 }
-                        } else
+                        } else {
                                 hdr.LastStableOffset = -1;
+                                hdr.LogStartOffset = -1;
```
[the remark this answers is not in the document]
mhowlett: well spotted ...

src/rdkafka_broker.c
```diff
@@ -3789,13 +3859,43 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
 
                         rd_kafka_toppar_lock(rktp);
+
+                        if (hdr.PreferredReadReplica != -1) {
+                                rd_kafka_fetch_preferred_replica_handle(
+                                        rktp, rkbuf, rkb,
+                                        hdr.PreferredReadReplica);
+
+                                if (unlikely(hdr.MessageSetSize != 0)) {
+                                        rd_rkb_log(rkb, LOG_WARNING, "FETCH",
```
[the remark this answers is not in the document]
mhowlett: 1. because the logic is easier to write. 2. they should never be there. if they are, it's arguably most correct to ignore them.

src/rdkafka_broker.c
```diff
@@ -3882,13 +3982,66 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                                 * when fetching from follower.
+                                                 * Revert back to the leader.
+                                                 */
+                                                rd_kafka_toppar_revert_to_leader(
+                                                        rktp);
+                                                break;
+                                        }
+
+                                        /* Application error */
                                         int64_t err_offset =
```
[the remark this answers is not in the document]
mhowlett: the 80 col limit makes that awkward, avoiding the else.

src/rdkafka_broker.c
```diff
@@ -4088,12 +4243,21 @@ static int rd_kafka_broker_fetch_toppars (rd_kafka_broker_t *rkb, rd_ts_t now) {
...
+                /* ForgottenTopicsCnt */
+                4+
+                /* N x ForgottenTopicsData */
+                0);
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(
+                rkb, RD_KAFKAP_Fetch, 0, 11, NULL);
+
+        if (ApiVersion == 11)
+                rd_kafka_buf_ApiVersion_set(rkbuf, 11, 0);
```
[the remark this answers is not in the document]
mhowlett: i believe you mean in place of the `0`

src/rdkafka_broker.c
```diff
@@ -4106,19 +4270,26 @@ static int rd_kafka_broker_fetch_toppars (rd_kafka_broker_t *rkb, rd_ts_t now) {
...
+        if (rd_kafka_buf_ApiVersion(rkbuf) >= 4) {
                 /* MaxBytes */
                 rd_kafka_buf_write_i32(rkbuf,
                                        rkb->rkb_rk->rk_conf.fetch_max_bytes);
                 /* IsolationLevel */
                 rd_kafka_buf_write_i8(rkbuf, rkb->rkb_rk->rk_conf.isolation_level);
         }
+        if (rd_kafka_buf_ApiVersion(rkbuf) >= 7) {
+                /* SessionId */
+                rd_kafka_buf_write_i32(rkbuf, 0);
```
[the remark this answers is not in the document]
mhowlett: these values are consistent with: https://github.com/apache/kafka/blob/trunk/clients/src/main/resources/common/message/FetchRequest.json#L59

as well as the "FetchRequest Metadata meaning" table in KIP-227: https://cwiki.apache.org/confluence/display/KAFKA/KIP-227:+Introduce+Incremental+FetchRequests+to+Increase+Partition+Scalability#KIP-227:IntroduceIncrementalFetchRequeststoIncreasePartitionScalability-NewErrorCodes

src/rdkafka_broker.c
```diff
@@ -4181,6 +4360,17 @@ static int rd_kafka_broker_fetch_toppars (rd_kafka_broker_t *rkb, rd_ts_t now) {
                                            rktp, rktp_activelink)) !=
                  rkb->rkb_active_toppar_next);
 
+        if (rd_kafka_buf_ApiVersion(rkbuf) >= 7) {
+                /* Forgotten. broker only - not used by consumer. */
```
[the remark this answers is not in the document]
mhowlett: see kip-227. I added a better comment.

src/rdkafka_broker.c
```diff
@@ -4181,6 +4360,17 @@ static int rd_kafka_broker_fetch_toppars (rd_kafka_broker_t *rkb, rd_ts_t now) {
...
+                rd_kafka_buf_write_i32(rkbuf, 0);
+        }
+
+        if (rd_kafka_buf_ApiVersion(rkbuf) >= 11) {
+                /* RackId */
+                rd_kafka_buf_write_kstr(rkbuf,
```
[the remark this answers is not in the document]
mhowlett: empty string. i believe i have that interpretation and the implementation right.

https://github.com/apache/kafka/blob/trunk/clients/src/main/resources/common/message/FetchRequest.json#L88

https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java#L339

src/rdkafka_partition.c
```diff
@@ -998,37 +1001,34 @@ void rd_kafka_toppar_broker_leave_for_remove (rd_kafka_toppar_t *rktp) {
...
- * 'rkb' may be NULL to undelegate leader.
+ * Delegates toppar 'rktp' to broker 'rkb'. 'rkb' may be NULL to
+ * undelegate broker.
  *
  * Locks: Caller must have rd_kafka_topic_wrlock(rktp->rktp_rkt) 
  *        AND rd_kafka_toppar_lock(rktp) held.
  */
 void rd_kafka_toppar_broker_delegate (rd_kafka_toppar_t *rktp,
-				      rd_kafka_broker_t *rkb,
-				      int for_removal) {
```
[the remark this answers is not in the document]
mhowlett: unsure what you want me to do with this

src/rdkafka_partition.c
```diff
@@ -1842,6 +1843,21 @@ rd_ts_t rd_kafka_toppar_fetch_decide (rd_kafka_toppar_t *rktp,
 		goto done;
 	}
 
+        /* Check for preferred replica lease expiry */
+        if (rktp->rktp_leader_id != rktp->rktp_broker_id &&
+            rd_interval(&rktp->rktp_lease_intvl,
+                        5*60*1000000/*5 minutes*/, 0) > 0) {
+                rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "FETCHDEC",
```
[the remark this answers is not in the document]
mhowlett: good point

src/rdkafka_partition.c
```diff
@@ -2425,25 +2442,25 @@ rd_kafka_broker_t *rd_kafka_toppar_leader (rd_kafka_toppar_t *rktp,
...
-        rd_kafka_dbg(rkt->rkt_rk, TOPIC, "LEADERUA",
-                     "%s [%"PRId32"]: leader unavailable: %s: %s",
+        rd_kafka_dbg(rkt->rkt_rk, TOPIC, "BROKERUA",
+                     "%s [%"PRId32"]: broker unavailable: %s: %s",
                      rkt->rkt_topic->str, rktp->rktp_partition, reason,
                      rd_kafka_err2str(err));
 
         rd_kafka_topic_wrlock(rkt);
-        rkt->rkt_flags |= RD_KAFKA_TOPIC_F_LEADER_UNAVAIL;
+        rkt->rkt_flags |= RD_KAFKA_TOPIC_F_BROKER_UNAVAIL;
```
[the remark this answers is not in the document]
mhowlett: ahh, i didn't follow the code through far enough and assumed it's behavior incorrectly. I've reverted this and the function name back to leader and will need to think of how best to cause the broker metadata refresh.

src/rdkafka_topic.c
```diff
@@ -476,63 +476,91 @@ const char *rd_kafka_topic_name (const rd_kafka_topic_t *app_rkt) {
...
+        if (is_leader)
+                rktp->rktp_leader_id = broker_id;
 
 	if (!rkb) {
-		int had_leader = rktp->rktp_leader ? 1 : 0;
+		int had_broker = rktp->rktp_broker ? 1 : 0;
 
-		rd_kafka_toppar_broker_delegate(rktp, NULL, 0);
+                if (had_broker)
+                        rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "TOPICUPD",
```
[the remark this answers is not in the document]
mhowlett: yes, good point, i didn't see that branch.

src/rdkafka_topic.c
```diff
@@ -552,15 +580,71 @@ static int rd_kafka_toppar_leader_update2 (rd_kafka_itopic_t *rkt,
...
 	return r;
 }
 
 
+/**
+ * @brief Revert the topic+partition delegation to the leader.
+ * @returns 1 if the broker delegation was changed, -1 if the broker
+ *          delegation was changed and is now undelegated, else 0.
+ *
+ * @locks caller must have rd_kafka_toppar_lock(rktp)
```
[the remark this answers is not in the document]
mhowlett: didn't appreciate the lock ordering.

the topic lock was intended to synchronize access to `rkt_rk` and `rkt_topic` only, but they won't change unless i'm mistaken (as you suggest), so I've removed it

src/rdkafka_topic.c
```diff
@@ -552,15 +580,71 @@ static int rd_kafka_toppar_leader_update2 (rd_kafka_itopic_t *rkt,
...
+        rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "BROKER",
+                "Topic %s [%"PRId32"]: Reverting from preferred "
+                "replica %"PRId32" to leader %"PRId32,
+                rktp->rktp_rkt->rkt_topic->str, rktp->rktp_partition,
+                rktp->rktp_broker_id, rktp->rktp_leader_id);
+
+        leader = rd_kafka_broker_find_by_nodeid(rktp->rktp_rkt->rkt_rk,
+                                                rktp->rktp_leader_id);
+
+        int r = rd_kafka_toppar_broker_update(
```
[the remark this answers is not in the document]
mhowlett: i'll just return the result of this instead.

tests/0101-fetch-from-follower.cpp
mhowlett: it's not really possible to write tests to cover most of the error scenarios. I did some manual tests that included simple modifications to the code in order to verify some behavior, and i thought it worth noting the procedure here.

src/rdkafka_topic.c
```diff
@@ -552,15 +580,71 @@ static int rd_kafka_toppar_leader_update2 (rd_kafka_itopic_t *rkt,
...
+ * @locality any
+ */
+int rd_kafka_toppar_revert_to_leader (rd_kafka_toppar_t *rktp) {
+        rd_kafka_broker_t *leader;
+
+        rd_kafka_topic_wrlock(rktp->rktp_rkt);
+
+        if (rktp->rktp_leader_id == rktp->rktp_broker_id) {
+                rd_kafka_log(rktp->rktp_rkt->rkt_rk, LOG_WARNING, "BROKER",
+                        "Topic %s [%"PRId32"]: Cannot revert toppar to "
```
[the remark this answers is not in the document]
mhowlett: I think it's valid (also imo better) to have this utility function have leader_id == broker_id as a valid precondition - it may be useful in another context in the future where that is an expected input. as is, it happens to be the case that this is never true. I don't see much value in using assert here instead, but i'll change it to decrease the number of things we need to argue about by one.

if seen from my perspective though, this shouldn't be a warning message, it would better be at a debug level (so I did get it wrong whichever way you look at it).

tests/interactive_broker_version.py
mhowlett: notes:
- all the other tests will have `client.rack` set to empty and `org.apache.kafka.common.replica.RackAwareReplicaSelector` will never result in a preferred replica being sent in a FETCH response in that case, given the configured `broker.rack` values, so in practice the changes in this PR won't change code coverage of the other tests.
- the fff test checks that messages were fetched from a replica (via rxbytes in stats), so it will fail if run in a non rack aware environment. This test would need to be aware of it's environment (I guess that is not hard, but I don't know).

tests/0101-fetch-from-follower.cpp
mhowlett: i need access to a good reference on how to drive your test env.

tests/0101-fetch-from-follower.cpp
mhowlett: much easier said than done. i need to do a describe configs to see if the broker.rack is set to be completely sure of this.

mhowlett: can i use `test_getenv` for this?

src/rdkafka_broker.c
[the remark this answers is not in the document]
mhowlett: i've added code to rate-limit this to 5 seconds.

src/rdkafka_partition.c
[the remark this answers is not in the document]
mhowlett: (done)

src/rdkafka_broker.c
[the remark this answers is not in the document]
mhowlett: I've changed the behavior + added a comment.

I doubt you will like this but I don't know what's better. Note: there could be an unnecessary back-off in the event there was a problem on a preferred replica, the toppar was delegated back to the leader and the leader then delegates it to a different preferred replica, but I don't think the complexity in tracking that case is worth the benefit.

src/rdkafka_partition.h
```diff
@@ -325,6 +327,19 @@ struct rd_kafka_toppar_s { /* rd_kafka_toppar_t */
...
 						  * timer */
 
+        rd_interval_t      rktp_lease_intvl;     /**< Preferred replica lease
+                                                  *   period */
+        rd_interval_t      rktp_new_lease_intvl; /**< Controls max frequency
+                                                  *   at which a new preferred
+                                                  *   replica lease can be
+                                                  *   created for a toppar.
+                                                  */
+        rd_interval_t      rktp_metadata_intvl;  /**< Controls max frequency
```
edenhill: Good commentry

src/rdkafka_topic.c
```diff
@@ -476,63 +476,81 @@ const char *rd_kafka_topic_name (const rd_kafka_topic_t *app_rkt) {
...
+	if (rktp->rktp_broker) {
+		if (rktp->rktp_broker == rkb) {
 			/* No change in broker */
 			return 0;
 		}
 
 		rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "TOPICUPD",
-			     "Topic %s [%"PRId32"] migrated from "
-			     "broker %"PRId32" to %"PRId32,
+			     "Topic %s [%"PRId32"]: migrating from "
```
edenhill: When changing a log/dbg thing that has tabs for indent, reindent the whole thing with spaces so at least this one call is consistently indented.

src/rdkafka_topic.c
```diff
@@ -476,63 +476,81 @@ const char *rd_kafka_topic_name (const rd_kafka_topic_t *app_rkt) {
...
+ *          delegation was changed and is now undelegated, else 0.
+ * 
+ * @locks caller must have rd_kafka_toppar_lock(rktp)
  * @locality any
  */
-int rd_kafka_toppar_leader_update (rd_kafka_toppar_t *rktp,
-                                   int32_t leader_id, rd_kafka_broker_t *rkb) {
-
-        rktp->rktp_leader_id = leader_id;
-        if (rktp->rktp_leader_id != leader_id) {
```
edenhill: This log is still useful for the leader change, down in `if (is_leader)`

src/rdkafka_topic.c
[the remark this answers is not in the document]
edenhill: rkt_rk, rkt_topic, et.al, will not change.

src/rdkafka_partition.c
```diff
@@ -998,37 +1002,33 @@ void rd_kafka_toppar_broker_leave_for_remove (rd_kafka_toppar_t *rktp) {
...
-
 /**
- * Delegates broker 'rkb' as leader for toppar 'rktp'.
- * 'rkb' may be NULL to undelegate leader.
+ * @brief Delegates toppar 'rktp' to broker 'rkb'. 'rkb' may be NULL to
+ *        undelegate broker.
  *
- * Locks: Caller must have rd_kafka_topic_wrlock(rktp->rktp_rkt) 
- *        AND rd_kafka_toppar_lock(rktp) held.
+ * @locks: Caller must have rd_kafka_toppar_lock(rktp) held.
```
edenhill: `@locks`, no :

src/rdkafka_topic.c
```diff
@@ -552,15 +570,57 @@ static int rd_kafka_toppar_leader_update2 (rd_kafka_itopic_t *rkt,
...
+
+        rd_assert(rktp->rktp_leader_id != rktp->rktp_broker_id);
+
+        rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "BROKER",
+                "Topic %s [%"PRId32"]: Reverting from preferred "
+                "replica %"PRId32" to leader %"PRId32,
+                rktp->rktp_rkt->rkt_topic->str, rktp->rktp_partition,
+                rktp->rktp_broker_id, rktp->rktp_leader_id);
+
+        leader = rd_kafka_broker_find_by_nodeid(rktp->rktp_rkt->rkt_rk,
```
edenhill: this reference, if not NULL, is never released.
Call rd_kafka_broker_destroy after broker_update

src/rdkafka_topic.c
```diff
@@ -1126,24 +1186,24 @@ void rd_kafka_topic_partitions_remove (rd_kafka_itopic_t *rkt) {
...
  * @locks rd_kafka_toppar_lock MUST be held
  */
 static const char *rd_kafka_toppar_needs_query (rd_kafka_t *rk,
                                                 rd_kafka_toppar_t *rktp) {
-        int leader_state;
+        int broker_state;
 
-        if (!rktp->rktp_leader)
+        if (!rktp->rktp_broker)
                 return "not assigned";
```
edenhill: maybe change to "not delegated"

src/rdkafka_topic.h
```diff
@@ -69,7 +69,7 @@ struct rd_kafka_itopic_s {
 	} rkt_state;
 
         int               rkt_flags;
-#define RD_KAFKA_TOPIC_F_LEADER_UNAVAIL   0x1 /* Leader lost/unavailable
+#define RD_KAFKA_TOPIC_F_LEADER_UNAVAIL   0x1 /* Broker lost/unavailable
```
edenhill: This still depicts the actual leader, not a replica, right? (so the comment should not be changed)

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,305 @@
...
+ * @brief A basic test of fetch from follower funtionality
+ *  - produces a bunch of messages to a replicated topic.
+ *  - configure the consumer such that `client.rack` is different from the
+ *    broker's `broker.rack` (and use
+ *    org.apache.kafka.common.replica.RackAwareReplicaSelector).
+ *  - consume the messages, and check they are as expected.
+ *  - use rxbytes from the statistics event to confirm that 
+ *    the messages were retrieved from the replica broker (not the
+ *    leader).
+ */
```
edenhill: Good summary!

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+
+static void test_assert(bool cond, std::string msg) {
```
edenhill: I'm making the C asserts available in C++ in my branch, will just need to remember to update these when that gets merged.

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,305 @@
...
+  RdKafka::Producer *p = RdKafka::Producer::create(pConf, errstr);
+  delete pConf;
+  test_assert(p, tostr() << "Failed to create producer: " << errstr);
+
+  RdKafka::Conf *tconf = RdKafka::Conf::create(RdKafka::Conf::CONF_TOPIC);
+  RdKafka::Topic *topic = RdKafka::Topic::create(p, topic_str, tconf, errstr);
+  delete tconf;
+  test_assert(topic, tostr() << "Failed to create topic: " << errstr);
+
+  err = p->metadata(0, topic, &metadata, 5000);
```
edenhill: use `tmout_multip(5000)` which gives code like this more time on CI, under valgrind, etc.

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,305 @@
...
+  }
+};
+
+bool TestEvent2Cb::should_capture_stats;
+bool TestEvent2Cb::has_captured_stats;
+std::map<int32_t, int64_t> TestEvent2Cb::rxbytes;
+static TestEvent2Cb ex_event_cb;
+
+
+static void get_brokers_info(std::string &topic_str, int *leader, std::vector<int> &brokers) {
```
edenhill: leader should be int32_t

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,305 @@
...
+  int client_rack_id = -1;
+  size_t i;
+  for (i=0; i<replica_ids.size(); ++i) {
+    if (replica_ids[i] != leader_id) {
+      client_rack_id = replica_ids[i];
+      break;
+    }
+  }
+  std::string client_rack = tostr() << "RACK" << client_rack_id;
+  Test::Say(tostr() << "client.rack: " << client_rack << "\n");
```
edenhill: nit: dont need tostr() stream when it is all strings, can use + string concat instead.

tests/0101-fetch-from-follower.cpp
edenhill: Not addressed

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,305 @@
...
+        break;
+      }
+
+    delete msg;
+  }
+
+  /* rely on the test timeout to prevent an infinite loop in
+   * the (unlikely) event that the statistics callback isn't
+   * called. */
+  Test::Say("Capturing rxbytes statistics");
```
edenhill: Missing newline in string

tests/0101-fetch-from-follower.cpp
edenhill: Manual tests, albeit documented, does not provide any value since they're never run again (in practice) and thus do not catch regression.

Try to see if any of these can be materialized as automatic tests.

tests/test.c
```diff
@@ -326,6 +327,7 @@ struct test tests[] = {
         _TEST(0098_consumer_txn, 0),
         _TEST(0099_commit_metadata, 0),
         _TEST(0100_thread_interceptors, TEST_F_LOCAL),
+        _TEST(0101_fetch_from_follower, 0),
```
edenhill: specify the minimum broker version here

tests/0101-fetch-from-follower.cpp
mhowlett: what happened here is that i addressed it, then unaddressed it because that was easier than remembering the env var.

src/rdkafka_broker.c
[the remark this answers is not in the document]
mhowlett: I've added an additional dbg message instead - not sure what error code to use. the root cause is `RD_KAFKA_RESP_ERR_OFFSET_OUT_OF_RANGE`, but that seems likely to cause confusion if used here.

tests/0101-fetch-from-follower.cpp
mhowlett: ok, done.

tests/0101-fetch-from-follower.cpp
mhowlett: using these notes I was just able to quickly re-verify behavior following recent changes, so they do provide some value :-). but of course very limited compared to automated regression tests since in practice no one will make use of them unless they're explicitly updating the related code.

it would be very difficult to automate this testing. the java implementation of fff is tested exclusively with unit tests.

INTRODUCTION.md
```diff
@@ -911,6 +911,14 @@ the logging level will be LOG_WARNING (4), else LOG_INFO (6).
 but it is recommended to instead rely on the above heuristics.
 
 
+#### Fetch From Follower
+
+librdkafka supports consuming messages from follower replicas. This is enabled
+by setting the `client.rack` configuration property which corresponds to
+`broker.rack` on the broker. The actual assignment of consumers to replicas is
+determined by the configured `replica.selector.class` on the broker.
```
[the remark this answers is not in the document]
mhowlett: linked to the kip, which is also the best docs on the replica.selector class afaict.

src/rdkafka_broker.c
```diff
@@ -3687,39 +3757,55 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
                                 int32_t MessageSetSize;
+                                int32_t PreferredReadReplica;   /* v11 */
                         } hdr;
                         rd_kafka_resp_err_t err;
                         int64_t end_offset;
 
 			rd_kafka_buf_read_i32(rkbuf, &hdr.Partition);
 			rd_kafka_buf_read_i16(rkbuf, &hdr.ErrorCode);
+                        if (ErrorCode)
+                                hdr.ErrorCode = ErrorCode;
```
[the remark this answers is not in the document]
mhowlett: whoops

src/rdkafka_broker.c
```diff
@@ -3885,11 +4007,69 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                         * Handle by retrying FETCH (with
+                                         * backoff).
+                                         */
+                                        rd_rkb_dbg(rkb, MSG, "FETCH",
+				                "Topic %.*s [%"PRId32"]: "
+                                                "Offset %"PRId64" not "
+                                                "available on broker %"PRId32
+                                                " (leader %"PRId32"): "
+                                                "retrying",
+				                RD_KAFKAP_STR_PR(&topic),
```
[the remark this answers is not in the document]
mhowlett: not in my editor.. hopefully fixed.

src/rdkafka_broker.c
```diff
@@ -3885,11 +4007,69 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
+                                                hdr.Partition,
+                                                rktp->rktp_offsets
```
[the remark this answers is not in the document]
mhowlett: ok. i don't agree with either convention, but doing my best to fit in ...

src/rdkafka_topic.c
mhowlett: ok, i had decided it was overly verbose, but when SHTF, additional logging here may indeed prove useful.

src/rdkafka_topic.c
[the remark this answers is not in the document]
mhowlett: removing `and NOT rd_kafka_topic_wrlock(rkt)` since that is now out of date.

src/rdkafka_partition.c
mhowlett: so many rules

src/rdkafka_topic.c
mhowlett: ouch, that's a bad mistake.

src/rdkafka_topic.h
mhowlett: yep, that's a mistake.

tests/0101-fetch-from-follower.cpp
mhowlett: yeah, i know. i find it interesting how many elementary mistakes i'm making. i think because little bits of mental effort add up to a lot over many places.

src/rdkafka_broker.c
```diff
@@ -3687,39 +3757,55 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
 			rd_kafka_buf_read_i32(rkbuf, &hdr.Partition);
 			rd_kafka_buf_read_i16(rkbuf, &hdr.ErrorCode);
+                        if (ErrorCode)
+                                hdr.ErrorCode = ErrorCode;
 			rd_kafka_buf_read_i64(rkbuf, &hdr.HighwaterMarkOffset);
 
                         end_offset = hdr.HighwaterMarkOffset;
 
-                        if (rd_kafka_buf_ApiVersion(request) == 4) {
+                        hdr.LastStableOffset = -1;
```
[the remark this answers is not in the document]
mhowlett: yeah totally agree. this will change what's reported in stats from the previous release which is technically breaking, but not a big deal. Also appropriate for LogStartOffset, though that has no consequence.

tests/0101-fetch-from-follower.cpp
mhowlett: i'm choosing to reading an implicit 'I' into that sentence just after the comma.

tests/test.c
mhowlett: oh cool, missed that feature

mhowlett: ... but a bit inconvenient currently as this doesn't work in the `./interactive_broker_version.py trunk` env.

src/rdkafka_partition.c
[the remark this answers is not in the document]
mhowlett: sometimes i need very funny jokes explained to me

tests/0101-fetch-from-follower.cpp
mhowlett: i'm checking for `TRIVUP_ROOT` and skipping the test if not found. seems ok.

tests/test.c
```diff
@@ -326,6 +327,7 @@ struct test tests[] = {
         _TEST(0098_consumer_txn, 0),
         _TEST(0099_commit_metadata, 0),
         _TEST(0100_thread_interceptors, TEST_F_LOCAL),
+        _TEST(0101_fetch_from_follower, TEST_BRKVER(2,4,0,0)),
```
mhowlett: note: need to comment out `TEST_BRKVER(2,4,0,0)` to test against trunk (2.4 not out yet).

src/rdkafka_broker.c
```diff
@@ -3789,13 +3859,43 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
                         rktp = rd_kafka_toppar_s2i(s_rktp);
 
                         rd_kafka_toppar_lock(rktp);
+
+                        if (hdr.PreferredReadReplica != -1) {
+                                rd_kafka_fetch_preferred_replica_handle(
+                                        rktp, rkbuf, rkb,
+                                        hdr.PreferredReadReplica);
+
```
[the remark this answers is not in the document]
mhowlett: in fact, it should be moved into `rd_kafka_fetch_preferred_replica_handle`

src/rdkafka_broker.c
```diff
@@ -3631,13 +3633,74 @@ static void rd_kafka_toppar_fetch_backoff (rd_kafka_broker_t *rkb,
...
+                 * relatively quickly instructed the client back to 
+                 * a preferred replica, quite possibly the same one
+                 * as before (possibly resulting from stale metadata),
+                 * so we back off the toppar to slow down potential
+                 * back-and-forth.
+                 */
+                rd_kafka_toppar_fetch_backoff(rkb,
+                        rktp, RD_KAFKA_RESP_ERR_NO_ERROR);
+        
+        rd_kafka_rdlock(rk);
```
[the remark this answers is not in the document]
mhowlett: wonderful plan, this is was indeed an issue as i've no doubt you were aware.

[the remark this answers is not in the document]
mhowlett: i've changed this to only lock the rktp over the `rd_kafka_toppar_broker_update` call.

please confirm locks aren't necessary over the `rd_kafka_toppar_fetch_backoff` calls (since they're all on the broker thread)

src/rdkafka.c
```diff
@@ -3565,12 +3569,12 @@ int rd_kafka_queue_poll_callback (rd_kafka_queue_t *rkqu, int timeout_ms) {
 static void rd_kafka_toppar_dump (FILE *fp, const char *indent,
 				  rd_kafka_toppar_t *rktp) {
 
-	fprintf(fp, "%s%.*s [%"PRId32"] leader %s\n",
+	fprintf(fp, "%s%.*s [%"PRId32"] broker %s\n",
```
edenhill: it would be useful to also print the leader id here

src/rdkafka.c
```diff
@@ -1278,6 +1280,7 @@ static RD_INLINE void rd_kafka_stats_emit_toppar (struct _stats_emit *st,
 
 	_st_printf("%s\"%"PRId32"\": { "
 		   "\"partition\":%"PRId32", "
+		   "\"broker\":%"PRId32", "
```
edenhill: update STATISTICS.md

src/rdkafka.h
```diff
@@ -363,6 +363,8 @@ typedef enum {
         RD_KAFKA_RESP_ERR__GAPLESS_GUARANTEE = -148,
         /** Maximum poll interval exceeded */
         RD_KAFKA_RESP_ERR__MAX_POLL_EXCEEDED = -147,
+	/** Unknown broker */
```
edenhill: use space for indent

src/rdkafka_partition.c
```diff
@@ -128,7 +128,7 @@ static void rd_kafka_toppar_consumer_lag_req (rd_kafka_toppar_t *rktp) {
         if (rktp->rktp_wait_consumer_lag_resp)
                 return; /* Previous request not finished yet */
 
-        rkb = rd_kafka_toppar_leader(rktp, 1/*proper brokers only*/);
+        rkb = rd_kafka_toppar_broker(rktp, 1/*proper brokers only*/);
```
edenhill: Do we want to base the consumer lag on the replica's view of the High-Watermark or the leader's?

src/rdkafka_topic.c
```diff
@@ -476,63 +476,88 @@ const char *rd_kafka_topic_name (const rd_kafka_topic_t *app_rkt) {
...
+                                   rd_bool_t is_leader) {
 
-        rktp->rktp_leader_id = leader_id;
-        if (rktp->rktp_leader_id != leader_id) {
+        rktp->rktp_broker_id = broker_id;
+        if (is_leader) {
                 rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "TOPICUPD",
-                             "Topic %s [%"PRId32"] migrated from "
-                             "leader %"PRId32" to %"PRId32,
+                             "Topic %s [%"PRId32"] migrating leader: "
```
edenhill: "migrating from "

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,320 @@
...
+                       rapidjson::GetParseError_En(d.GetParseError()) <<
+                       " at " << d.GetErrorOffset());
+
+          /* iterate over brokers. */
+          rapidjson::Pointer jpath((const char *)"/brokers");
+          rapidjson::Value *pp = rapidjson::GetValueByPointer(d, jpath);
+          if (pp == NULL)
+            return;
+
+          for (rapidjson::Value::ConstMemberIterator itr = pp->MemberBegin(); itr != pp->MemberEnd(); ++itr) {
```
edenhill: C++ is so noisy and verbose

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,320 @@
...
+  /* Determine the ids of the brokers that the partition has replicas
+   * on and which one of those is the leader.
+   */
+  RdKafka::Conf *pConf;
+  Test::conf_init(&pConf, NULL, 10);
+  RdKafka::Producer *p = RdKafka::Producer::create(pConf, errstr);
+  delete pConf;
+  test_assert(p, tostr() << "Failed to create producer: " << errstr);
+
+  RdKafka::Conf *tconf = RdKafka::Conf::create(RdKafka::Conf::CONF_TOPIC);
```
edenhill: tconf is not really used

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    switch (msg->err())
+      {
+      case RdKafka::ERR__TIMED_OUT:
+        break;
+
+      case RdKafka::ERR_NO_ERROR:
+        {
+          test_assert(msg->len() == 100, "expecting message value size to be 100");
+          char* cnt_str_start_ptr = strstr((char *)msg->payload(), "msg=") + 4;
```
edenhill: `char *`

tests/0101-fetch-from-follower.cpp
```diff
@@ -0,0 +1,320 @@
...
+  //     - that metadata being periodically received and not interfering with anything.
+
+  c->close();
+  delete c;
+}
+#endif
+
+extern "C" {
+  int main_0101_fetch_from_follower (int argc, char **argv) {
+    if (!test_getenv("TRIVUP_ROOT", NULL)) {
```
edenhill: This is a bit too broad-edged (and TRIVUP_ROOT can be set outside a trivup environment to point out the path to where it should store its local state (which is bad reuse of an env, I know)).

Since you're already collecting the broker racks in the test itself I suggest you Skip the test from there instead if there are less than 2 racks configured.

tests/cluster_testing.py
```diff
@@ -53,6 +58,9 @@ def __init__(self, version, conf={}, num_brokers=3, debug=False):
         self.conf = defconf
 
         for n in range(0, num_brokers):
+            # Configure rack & replica selector if broker supports fetch-from-follower
+            if version_as_number(version) >= 2.4:
```
edenhill: nit: I'd recommend using a list, e.g: `if version_as_list(version) >= [2, 4, 0]`

src/rdkafka.c
mhowlett: i added leader_id. Also added broker_id because timing of that is consistent with leader_id and broker_id may not align with (broker *). please confirm this makes sense to you.

mhowlett: consistent with the statistics, on second thoughts I've removed broker_id.

tests/0101-fetch-from-follower.cpp
mhowlett: IIRC C++26 is slated to sort this out with additional languages features and a rewrite of the standard library

tests/0101-fetch-from-follower.cpp
mhowlett: i'm not actually doing that - need to do a describe configs to get that information.

tests/cluster_testing.py
mhowlett: i just learnt something about python

src/rdkafka_partition.c
mhowlett: this is a good question. an even better question is will it even work, given the Offsets API - at least before FFF was implemented - requires the request to be directed at the leader. it turns out this is still the case, and this doesn't even work.

can you think of anything else that uses rktp_broker that may also break? I will check as well.

src/rdkafka_offset.c
```diff
@@ -785,23 +791,23 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
 			rktp, RD_KAFKA_TOPPAR_FETCH_NONE);
 
 	} else {
-                /* Use cached log start offset if available */
+                /* Use cached log start / end offsets from last FETCH
+                   if available */
                 if (offset == RD_KAFKA_OFFSET_BEGINNING &&
                     rktp->rktp_lo_offset >= 0) {
-                        rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "OFFSET",
```
edenhill: Please don't remove debug logs, they're useful for post-mortem troubleshooting.

src/rdkafka_offset.c
```diff
@@ -785,23 +791,23 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
 			rktp, RD_KAFKA_TOPPAR_FETCH_NONE);
 
 	} else {
-                /* Use cached log start offset if available */
+                /* Use cached log start / end offsets from last FETCH
+                   if available */
```
edenhill: subsequent multiline comment lines start with `* if available`

src/rdkafka_offset.c
```diff
@@ -743,10 +743,16 @@ rd_kafka_offset_reset_op_cb (rd_kafka_t *rk, rd_kafka_q_t *rkq,
...
+ *        error, or offset is logical).
  *
- * Locality: toppar handler thread
- * Locks: toppar_lock() MUST be held
+ * @param rktp the toppar
+ * @param err_offset a logical offset, or offset corresponding to the error.
+ * @param err the error, or RD_KAFKA_RESP_ERR_NO_ERROR if offset is logical.
+ * @param reason a reason string for logging.
+ *
+ * @locality: any. if not main thread, work will be enqued on main thread.
```
edenhill: `@locality`, drop the :, same for locks.

src/rdkafka_offset.c
```diff
@@ -785,23 +791,23 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
...
-                                     "%s [%"PRId32"]: offset reset (at offset %s) "
-                                     "to BEGINNING (%"PRId64") from cache",
-                                     rktp->rktp_rkt->rkt_topic->str, rktp->rktp_partition,
-                                     rd_kafka_offset2str(err_offset),
-                                     rktp->rktp_lo_offset);
-                        rd_kafka_toppar_next_offset_handle(rktp, rktp->rktp_lo_offset);
-                        return;
+                        offset = rktp->rktp_lo_offset;
+                        rd_kafka_toppar_next_offset_handle(rktp, offset);
+                }
```
edenhill: should be on same line:
`} else if (`

src/rdkafka_offset.c
```diff
@@ -785,23 +791,23 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
...
+                else if (offset == RD_KAFKA_OFFSET_END &&
+                         rktp->rktp_ls_offset > 0) {
```
edenhill: Isn't it possible that the LSO is in fact 0 for an empty partition?

src/rdkafka_offset.c
```diff
@@ -785,23 +791,23 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
...
+                        offset = rktp->rktp_ls_offset;
+                        rd_kafka_toppar_next_offset_handle(rktp, offset);
+                /* Else query cluster for offset */
```
edenhill: The comment should be in the else block

src/rdkafka_offset.c
```diff
@@ -812,6 +818,10 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
                      rd_kafka_offset2str(offset),
                      reason, rd_kafka_err2str(err));
 
+        /* Note: If rktp is not delegated to the leader, then low and high
+           offsets will necessarily be cached from the last FETCH request,
```
edenhill: dito *

src/rdkafka_offset.c
```diff
@@ -812,6 +818,10 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
...
+           and so this offset query will never occur in that case for
+           BEGINNING / END logical offsets. */
```
edenhill: This is only true of the partition has been fetched at least once from the follower, right? So is there a  race condition here between the delegate-to-follower and first-valid-fetch?

src/rdkafka_topic.c
```diff
@@ -478,21 +478,22 @@ const char *rd_kafka_topic_name (const rd_kafka_topic_t *app_rkt) {
...
- * 
+ *
  * @locks caller must have rd_kafka_toppar_lock(rktp)
  * @locality any
  */
 int rd_kafka_toppar_broker_update (rd_kafka_toppar_t *rktp,
                                    int32_t broker_id,
-                                   rd_kafka_broker_t *rkb) {
+                                   rd_kafka_broker_t *rkb,
+                                   const char *reason) {
```
mhowlett: good idea. needs @param comment.

src/rdkafka_partition.c
```diff
@@ -1913,24 +1913,24 @@ rd_ts_t rd_kafka_toppar_fetch_decide (rd_kafka_toppar_t *rktp,
...
         lease_expired =
                 rktp->rktp_leader_id != rktp->rktp_broker_id &&
                 rd_interval(&rktp->rktp_lease_intvl,
-                        5*60*1000*1000/*5 minutes*/, 0) > 0;
-        rd_kafka_toppar_unlock(rktp);
-
+                            5*60*1000*1000/*5 minutes*/, 0) > 0;
         if (lease_expired) {
+                /* delete_to_leader() requires no locks to be held */
+                rd_kafka_toppar_unlock(rktp);
```
mhowlett: ahh whoops, i think i screwed this up late in the game trying to simplify it.

src/rdkafka_offset.c
```diff
@@ -795,15 +795,32 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
                    if available */
                 if (offset == RD_KAFKA_OFFSET_BEGINNING &&
                     rktp->rktp_lo_offset >= 0) {
+                        rd_kafka_dbg(rktp->rktp_rkt->rkt_rk, TOPIC, "OFFSET",
```
mhowlett: haha, i preemptively didn't put debug statements in these if blocks because i sensed you'd have DRY complaints.

there is now no reason not to return early from these blocks so we should do that.

src/rdkafka_offset.c
```diff
@@ -795,15 +795,32 @@ void rd_kafka_offset_reset (rd_kafka_toppar_t *rktp, int64_t err_offset,
...
+                                     rd_kafka_offset2str(err_offset),
+                                     rktp->rktp_lo_offset);
                         offset = rktp->rktp_lo_offset;
                         rd_kafka_toppar_next_offset_handle(rktp, offset);
-                }
-                else if (offset == RD_KAFKA_OFFSET_END &&
-                         rktp->rktp_ls_offset > 0) {
+
+                } else if (offset == RD_KAFKA_OFFSET_END &&
+                           rktp->rktp_ls_offset >= 0) {
```
mhowlett: whoops

src/rdkafka_broker.c
```diff
@@ -3661,28 +3662,32 @@ rd_kafka_fetch_preferred_replica_handle (rd_kafka_toppar_t *rktp,
         const rd_ts_t five_seconds = 5*1000*1000;
         rd_kafka_broker_t *preferred_rkb;
         rd_kafka_t *rk = rktp->rktp_rkt->rkt_rk;
-        rd_ts_t new_intvl = rd_interval_immediate(&rktp->rktp_new_lease_intvl,
-                                                  one_minute, 0);
+        rd_ts_t new_intvl = rd_interval(&rktp->rktp_new_lease_intvl,
```
mhowlett: just wanted to check you're 100% about this. AFAICT looking at it quickly again now, this will operate as required (first call will still return > 0, which is what is needed). regardless, i'd argue `rd_interval_immediate` is semantically more correct.

tests/0104-fetch_from_follower_mock.c
```diff
@@ -0,0 +1,171 @@
...
+        test_conf_set(conf, "auto.offset.reset", auto_offset_reset);
+        /* Make sure we don't consume the entire partition in one Fetch */
+        test_conf_set(conf, "fetch.message.max.bytes", "100");
+
+        c = test_create_consumer("mygroup", NULL, conf, NULL);
+
+        /* The first fetch will go to the leader which will redirect
+         * the consumer to the follower, the second and sub-sequent fetches
+         * will go to the follower. We want the third fetch, second one on
+         * the follower, to fail and trigger an offset reset. */
```
mhowlett: I think you might mean to use `RD_KAFKA_RESP_ERR_OFFSET_NOT_AVAILABLE` here?

`RD_KAFKA_RESP_ERR_OFFSET_OUT_OF_RANGE` shouldn't trigger an offset reset unless `fetch_offset` is less than `HighwaterMarkOffset` (which would seem the least important case of the two to test, but we could have tests for both). It should bounce back to the leader, which then decides what to do next - there could be different test cases for different scenarios, which don't seem to be defined in this test either.

tests/0104-fetch_from_follower_mock.c
```diff
@@ -0,0 +1,171 @@
...
+}
+
+
+/**
+ * @brief Test delegating consumer to a follower that does not exist,
+ *        the consumer should not be able to consume any messages (which
+ *        is questionable but for a later PR). Then change to a valid
+ *        replica and verify messages can be consumed.
+ */
+static void do_test_unknown_follower (void) {
```
mhowlett: good test

src/rdkafka_offset.c
edenhill: Due to separation of concerns we can't assume that next_offset_handle() does not trigger an offset query, so I'll keep the fallthru now to be safe, but adjusting the duplicatory logs.

tests/0104-fetch_from_follower_mock.c
edenhill: I added an additional test for lagging replicas (trigger ERR_OFFSET_NOT_AVAILABLE)

src/rdkafka_broker.c
edenhill: Sorry, don't know why I changed that. Reverting.

mhowlett: very good. i spent a long time making this work correctly / fprintf debugging it. although I couldn't see any problem with the code just reading the code, i wasn't sure i wasn't missing something...

Issue #2509, opened 2019-09-05
Signifcant performance degradation after upgrading librdkafka from 0.11.0 to 1.1.0


Eliyahu-Machluf:
Description
===========
We've upgraded our librdkafka version from 0.11.0 to 1.1.0, and doing performance tests, we've noticed a major performance degradation. librdkafka 1.1.0 is about 50% slower compared to 0.11.0 for our scenario

How to reproduce
================
We are running a produce session, producing 500,000 messages, and telling librdkafka to transfer them. When using librdkafka 1.1.0 it takes ~25 seconds, using 0.11.0 it takes ~12 seconds. 
This is after we configured the `request.required.acks` to be '1' (as we've seen the default was changed, and want to compare the same configuration)

Checklist
=========
 - [x] librdkafka version (release number or git tag): `1.1.0`
 - [x ] Apache Kafka version: `2.0 we've also tried it with 2.3`
 - [ ] librdkafka client configuration:
`api.version.request=true
request.required.acks=1
broker.version.fallback=0.9.0
message.max.bytes=1000000
queue.buffering.max.ms=1000
api.version.request=true
request.required.acks=1
broker.version.fallback=0.9.0
message.max.bytes=1000000
queue.buffering.max.ms=1000`
 - [x] Operating system: `Windows10`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [x ] Critical issue

edenhill:
How many partitions?
What kind of machine are you running the producer on, how many cores does it have?

Eliyahu-Machluf:
There is 1 partition. 
We are running the producer on windows10 machine, which is a virtual machine (running on Hyper-V), with 8 virtual processors.
The hosting machine has 4 cores, 8 logical processors.

edenhill:
That thruput sounds low even for the old version.
What is the cpu usage on the client and broker? Network load?

Eliyahu-Machluf:
We are doing some processing when preparing the messages, that's why the throughput is low, but it is OK for us. The point is that when we do the **same** processing and use the new librdkafka we see poor performance as described.
The CPU usage is not relevant, as our process work on preparing the messages, and this also loads the CPU.

edenhill:
Suggest cutting out your processing to troubleshoot this issue

Eliyahu-Machluf:
I'm currently busy with other tasks and cannot isolate only rdkafka at our product.

[...]

PR #2520, opened 2019-09-10
Rate limit IO-based queue wakeups to linger.ms (#2509)

edenhill:


shanson7:
Is the `0086_purge_remote` test failure related to this change? Should I hold off on applying this patch for further testing on my side?

edenhill:
@shanson7 It is a real problem. The patch will need to be modified, in its current form it rate limits all queue wakeups, not just due to msg enqueuing.

shanson7:
If it's not too problematic to get the config element there, I think it should probably happen in `rd_kafka_toppar_enq_msg` since that's where the decision is made for `queue_len == 1` as well.

PR #2525, opened 2019-09-13
Kip 345: Static group membership 

rnpridgeon:


rnpridgeon:
oddy enough I'm unable to reproduce the `0050_subscribe_adds`

[0050_subscribe_adds         / 20.764s] Closing consumer
[0050_subscribe_adds         / 20.769s] CONSUMER.CLOSE: duration 5.200ms
[0050_subscribe_adds         / 20.769s] 0050_subscribe_adds: duration 20769.172ms
[0050_subscribe_adds         / 20.769s] ================= Test 0050_subscribe_adds PASSED =================
[<MAIN>                      / 21.044s] ALL-TESTS: duration 21044.305ms
TEST 20190913130820 (bare) SUMMARY
| <MAIN>                                   |     PASSED |  21.044s |
| 0050_subscribe_adds                      |     PASSED |  20.769s |
[<MAIN>                      / 21.045s] # Test report written to test_report_20190913130820.json
[<MAIN>                      / 21.045s] 0 thread(s) in use by librdkafka
[<MAIN>                      / 21.045s] 
###
###  ./merged in bare mode PASSED! ###
###

edenhill:
re 0050_subscribe_adds: if you find an intermittent test failure you can use `TESTS=0050 ./until-fail.sh bare` to try to reproduce it.

src/rdkafka_request.c
```diff
@@ -1179,6 +1221,9 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
         if (ApiVersion >= 1)
                 rd_kafka_buf_write_i32(rkbuf, rk->rk_conf.max_poll_interval_ms);
         rd_kafka_buf_write_kstr(rkbuf, member_id);
+        if (ApiVersion >= 5)
```
edenhill: ^ add group_instance_id to buf_new_request() above

src/rdkafka_request.c
```diff
@@ -1211,6 +1256,20 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
                            rk->rk_conf.max_poll_interval_ms,
                            rk->rk_conf.group_session_timeout_ms);
 
+
+        if (ApiVersion < 5 &&
+            rk->rk_conf.group_instance_id &&
```
edenhill: Use your IS_STATIC .. macro

src/rdkafka_request.c
```diff
@@ -1211,6 +1256,20 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
...
+            rd_interval(&rkb->rkb_suppress.unsupported_kip345,
+                        /* at most once per day */
+                        (rd_ts_t)86400 * 1000 * 1000, 0) > 0)
+                rd_rkb_log(rkb, LOG_NOTICE, "MAXPOLL",
+                           "Broker does not support KIP-345 "
+                           "(requires Apache Kafka >= v0.2.3.0): "
```
edenhill: That's a really old version

src/rdkafka_request.c
```diff
@@ -1211,6 +1256,20 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
...
+                           "consumer configuration "
+                           "`group.instance.id` (%s) "
+                           "will not take effect.",
```
edenhill: no trailing periods in log messages

src/rdkafka_request.c
```diff
@@ -1240,17 +1299,34 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
...
                                  const rd_kafkap_str_t *member_id,
+                                 const rd_kafkap_str_t *group_instance_id,
                                  rd_kafka_replyq_t replyq,
                                  rd_kafka_resp_cb_t *resp_cb,
                                  void *opaque) {
         rd_kafka_buf_t *rkbuf;
+        int16_t ApiVersion = 0;
+        int features;
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(rkb,
```
edenhill: need to update handle_LeaveGroup ..

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test.h"
+
+/* Typical include path would be <librdkafka/rdkafka.h>, but this program
```
edenhill: no need for this include anymore, included from test.h

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+ * is built from within the librdkafka source tree and thus differs. */
+#include "rdkafka.h" /* for Kafka driver */
+
+/**
+ * KafkaConsumer static group membership tests
```
edenhill: `@name`

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * Runs two consumers subscribing to  a topic simulating various
```
edenhill: multiple whitespaces

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+ * reblance scenerios.
```
edenhill: typo

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+ * reblance scenerios. 
+ */
+
+typedef struct _consumer_s {
+        rd_kafka_t *rk;
+        int assign_cnt;
+        int max_rebalance_cnt;
+} _consumer_t;
+
+rd_list_t worker_list;
```
edenhill: make it static

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+                TEST_FAIL("rebalance failed: %s", rd_kafka_err2str(err));
+                break;
+        }
+
+}
+
+#define _CONSUMER_CNT 2
+int main_0102_static_group_rebalance (int argc, char **argv) {
+        rd_kafka_conf_t *conf;
+
```
edenhill: remove empty line

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        const int msgcnt = 100;
+        const char *topic = test_mk_topic_name("0101_static_group_rebalance", 1);
```
edenhill: 0102

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        _consumer_t c[_CONSUMER_CNT] = RD_ZERO_INIT;
+
+        test_conf_init(&conf, NULL, 60);
+
+        test_conf_set(conf, "session.timeout.ms", "1000");
```
edenhill: This is too aggressive on lousy CI systems, use 5s

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        test_conf_set(conf, "auto.offset.reset", "earliest");
+        test_conf_set(conf, "topic.metadata.refresh.interval.ms", "500");
+
+        rd_kafka_conf_set_opaque(conf, &c[0]);
+        test_conf_set(conf, "group.instance.id", "consumer1");
+        c[0].rk = test_create_consumer("static", &rebalance_cb,
```
edenhill: the group.id must be unique within the test suite since tests may run in parallel. Use the topic name.

edenhill: drop `&` from rebalance_cb

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        _consumer_t c[_CONSUMER_CNT] = RD_ZERO_INIT;
+
+        test_conf_init(&conf, NULL, 60);
+
+        test_conf_set(conf, "session.timeout.ms", "1000");
+        test_conf_set(conf, "auto.offset.reset", "earliest");
+        test_conf_set(conf, "topic.metadata.refresh.interval.ms", "500");
+
+        rd_kafka_conf_set_opaque(conf, &c[0]);
+        test_conf_set(conf, "group.instance.id", "consumer1");
```
edenhill: Is instance.id required to be unique cluster-wide or group-wide?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+                rd_kafka_topic_partition_list_destroy(parts);
+                break;
+        }
+}
+
+static void rebalance_cb (rd_kafka_t *rk,
+                          rd_kafka_resp_err_t err,
+                          rd_kafka_topic_partition_list_t *parts,
+                          void *opaque) {
+
```
edenhill: remove empty line

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        _consumer_t *rk_c = opaque;
```
edenhill: This is not a good variable name, `rk_*` is typically an rd_kafka_t instance:
`git grep 'rd_kafka_t \*rk_.*;'`

Try to use the same variable name for the same thing across the code, so in this case go with `c`

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+                 parts->cnt);
+
+        switch (err)
+        {
+        case RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS:
+                rk_c->assign_cnt++;
+
+                rd_kafka_assign(rk, parts);
+
+                TEST_ASSERT(rk_c->assign_cnt <= rk_c->max_rebalance_cnt,
```
edenhill: ASAP: Assert As Soon As Possible, namely before the assign.

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        TEST_SAY("Waiting for partition assignment\n");
+        await_assignment(&c[0]);
+        await_assignment(&c[1]);
+
+        TEST_SAY("Bouncing c[1] instance.\n");
+        test_consumer_close(c[1].rk);
+        rd_kafka_destroy(c[1].rk);
+
+        /** 
+         * Removeing and adding a new member with the same group instance id
```
edenhill: typo

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        await_assignment(&c[0]);
+        await_assignment(&c[1]);
+
+        TEST_SAY("Bouncing c[1] instance.\n");
+        test_consumer_close(c[1].rk);
+        rd_kafka_destroy(c[1].rk);
+
+        /** 
+         * Removeing and adding a new member with the same group instance id 
+         * should not prompt a rebalance for the original remaining member.
```
edenhill: -original

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+
+        rd_kafka_conf_set_opaque(conf, &c[1]);
+        test_conf_set(conf, "group.instance.id", "consumer2");
+        c[1].rk = test_create_consumer("static", &rebalance_cb,
+                                       rd_kafka_conf_dup(conf), NULL);
+
+        c[0].max_rebalance_cnt = 1;
+        c[1].max_rebalance_cnt = 1;
+
+        test_consumer_subscribe(c[0].rk, rd_strdup(tsprintf("^%s.*", topic)));
```
edenhill: This leaks memory. Sure you ran this with asan?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+         * should not prompt a rebalance for the original remaining member.
+         */
+        c[1].max_rebalance_cnt++;
+        c[1].rk = test_create_consumer("static", &rebalance_cb,
+                                       conf, NULL);
+        test_consumer_subscribe(c[1].rk, rd_strdup(tsprintf("^%s.*", topic)));
+
+        /* Wait for c[1] to rejoin, serve c[0] ops */
+        await_assignment(&c[1]);
+        do_consume(&c[0], 5/*5s*/);
```
edenhill: we typically use millisecond timeouts.

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        TEST_ASSERT(c[0].assign_cnt == c[0].max_rebalance_cnt,
+                    "c[0] rebalanced %d times, expected %d",
+                    c[0].assign_cnt, c[0].max_rebalance_cnt);
+
+        TEST_SAY("Closing remaining consumers\n");
+        test_consumer_close(c[0].rk);
+        rd_kafka_destroy(c[0].rk);
+
+        return 0;
+}
```
edenhill: missing newline

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,217 @@
...
+        do_consume(&c[0], 2/*1s*/);
+
+        TEST_ASSERT(c[0].assign_cnt == c[0].max_rebalance_cnt,
+                    "c[0] rebalanced %d times, expected %d",
+                    c[0].assign_cnt, c[0].max_rebalance_cnt);
+
+        TEST_SAY("Closing remaining consumers\n");
+        test_consumer_close(c[0].rk);
+        rd_kafka_destroy(c[0].rk);
+
```
edenhill: conf is leaked

src/rdkafka_cgrp.c
```diff
@@ -1016,7 +1034,7 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
         actions = rd_kafka_err_action(rkb, ErrorCode, request,
                                       RD_KAFKA_ERR_ACTION_IGNORE,
                                       RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID,
-
+                                      RD_KAFKA_RESP_ERR_MEMBER_ID_REQUIRED,
```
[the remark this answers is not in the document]
rnpridgeon: It's safe for the application to ignore from an application perspective but librdkafka must set the member id upon encountering this error. Perhaps I am not flagging that appropriately however. 

This is true for both static and dynamic group members per KIP-394. 

>Effectively speaking, previously we accept anonymous member joining as new member, and now we require one more bounce to justify new member identity.

tests/0102-static_group_rebalance.c
rnpridgeon: group

src/rdkafka_cgrp.c
```diff
@@ -912,11 +925,12 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
                      member_cnt,
                      ErrorCode ? rd_kafka_err2str(ErrorCode) : "(no error)");
 
+
+
         if (!ErrorCode) {
-                char *my_member_id;
```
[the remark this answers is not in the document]
rnpridgeon: It's also used in the event we receive a `RD_KAFKA_RESP_ERR_MEMBER_ID_REQUIRED`. Although I see you mention I should use a local variable there as well. Is there a particular reason for doing it this way rather than placing it in the larger scope.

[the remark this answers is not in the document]
edenhill: Yes, logical scoping, it is a temp variable that only has relevance where it is used.
Having it locally scoped helps the reader of the code know what is relevant where.

src/rdkafka.h
```diff
@@ -2538,6 +2541,19 @@ RD_EXPORT
...
+ *
+ * @remark This requires the high-level KafkaConsumer
+ *
+ * @returns An allocated string containing the current user-assigned group
+ *          instance id, or NULL if not available.
+ *          The application must free the string with \p free() or
+ *          rd_kafka_mem_free()
+ */
+RD_EXPORT
+char *rd_kafka_group_instance_id (const rd_kafka_t *rk);
```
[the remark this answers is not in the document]
rnpridgeon: Mostly I just wanted it to be consistent with how a user would access the member id. It's not explicitly necessary but it is sort of nice to treat it the same way.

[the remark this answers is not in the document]
edenhill: But the member_id is generated by the broker, while the group.instance.id is configured by the application, so it already knows it. This is equivalent to conf_get("group.instance.id") and does not provide any value afaict.

tests/0102-static_group_rebalance.c
rnpridgeon: do_consume uses seconds. I've reduced the timeout

tests/0102-static_group_rebalance.c
rnpridgeon: conf is handed over (without a call to dup) to the last instance of c1

src/rdkafka_conf.c
```diff
@@ -843,6 +843,12 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           _RK(group_id_str),
           "Client group id string. All clients sharing the same group.id "
           "belong to the same group." },
+        { _RK_GLOBAL|_RK_CGRP|_RK_MED, "group.instance.id", _RK_C_STR,
+          _RK(group_instance_id),
+          "Setting this property enables static group membership. "
+          "Static group members are able to leave and rejoin a group "
+          "within the configured session.timeout without prompting a "
+          "group rebalance. Each member of the group must have a unique id." },
```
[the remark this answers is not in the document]
edenhill: Requires broker version >= ..

src/rdkafka_cgrp.c
```diff
@@ -1356,6 +1383,9 @@ void rd_kafka_cgrp_handle_Heartbeat (rd_kafka_t *rk,
                 goto err;
         }
 
+
```
edenhill: move this empty line below throttle time for consistency.

src/rdkafka_request.c
```diff
@@ -867,6 +869,14 @@ int rd_kafka_OffsetCommitRequest (rd_kafka_broker_t *rkb,
 	int tot_PartCnt = 0;
         int i;
```
edenhill: remove empty line

src/rdkafka_request.c
```diff
@@ -867,6 +869,14 @@ int rd_kafka_OffsetCommitRequest (rd_kafka_broker_t *rkb,
...
 
+        int16_t ApiVersion;
+        int features;
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(rkb,
+                                                          RD_KAFKAP_OffsetCommit,
+                                                          0, 7,
```
edenhill: is the code below supporting all versions between 0 and 7 (inclusive)?

src/rdkafka_request.c
```diff
@@ -1041,6 +1061,13 @@ void rd_kafka_SyncGroupRequest (rd_kafka_broker_t *rkb,
...
+        int16_t ApiVersion;
+        int features;
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(rkb,
+                                                          RD_KAFKAP_SyncGroup,
+                                                          0, 3,
+                                                          &features);
 
         rkbuf = rd_kafka_buf_new_request(rkb, RD_KAFKAP_SyncGroup,
                                          1,
```
edenhill: add STR_SIZE(group_instance_id)

src/rdkafka_request.c
```diff
@@ -1041,6 +1061,13 @@ void rd_kafka_SyncGroupRequest (rd_kafka_broker_t *rkb,
                                 void *opaque) {
         rd_kafka_buf_t *rkbuf;
         int i;
+        int16_t ApiVersion;
+        int features;
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(rkb,
+                                                          RD_KAFKAP_SyncGroup,
+                                                          0, 3,
```
edenhill: are 0..3 fully supported below? (and in the response handlers)

src/rdkafka_request.c
```diff
@@ -1102,6 +1133,14 @@ void rd_kafka_handle_SyncGroup (rd_kafka_t *rk,
                 goto err;
         }
 
+        if (request->rkbuf_reqhdr.ApiVersion >= 1) {
+                int32_t Throttle_Time;
+                rd_kafka_buf_read_i32(rkbuf, &Throttle_Time);
```
edenhill: use throttle reader

src/rdkafka_request.c
edenhill: Most of that code predates the dynamic reallocations, so it is no longer needed. But if we know the size upfront we should provide it to avoid a reallocation / overallocation.

src/rdkafka_request.c
```diff
@@ -1211,6 +1256,20 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
...
                            rk->rk_conf.group_session_timeout_ms);
 
+
+        if (ApiVersion < 5 &&
+            rk->rk_conf.group_instance_id &&
+            rd_interval(&rkb->rkb_suppress.unsupported_kip345,
+                        /* at most once per day */
+                        (rd_ts_t)86400 * 1000 * 1000, 0) > 0)
+                rd_rkb_log(rkb, LOG_NOTICE, "MAXPOLL",
+                           "Broker does not support KIP-345 "
```
edenhill: Isn't this an error? The user configured the consumer to be a static member but that can't be realized due to outdated broker version.
I'd say you want to raise a consumer error and back off the next join 30s or so.

src/rdkafka_request.c
```diff
@@ -1240,17 +1299,26 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
...
                                  rd_kafka_replyq_t replyq,
                                  rd_kafka_resp_cb_t *resp_cb,
                                  void *opaque) {
         rd_kafka_buf_t *rkbuf;
+        int16_t ApiVersion = 0;
+        int features;
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(rkb,
+                                                          RD_KAFKAP_LeaveGroup,
+                                                          0, 1,
```
edenhill: are all versions supported here and in the handler?

src/rdkafka_request.c
```diff
@@ -1325,10 +1393,18 @@ void rd_kafka_HeartbeatRequest (rd_kafka_broker_t *rkb,
...
                                 rd_kafka_replyq_t replyq,
                                 rd_kafka_resp_cb_t *resp_cb,
                                 void *opaque) {
         rd_kafka_buf_t *rkbuf;
+        int16_t ApiVersion = 0;
+        int features;
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(rkb,
+                                                          RD_KAFKAP_Heartbeat,
+                                                          0, 3,
```
edenhill: dito, all versions.

tests/0102-static_group_rebalance.c
edenhill: not resolved

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,219 @@
...
+        rd_kafka_message_destroy(rkm);
+
+        if (timeout_s > 0) {
+                TEST_SAY("%s: simulate processing by sleeping for %ds\n",
+                         rd_kafka_name(cons->rk), timeout_s);
+                rd_sleep(timeout_s);
+        }
+}
+
+/* Serve op queue until we have an assignment */
```
edenhill: use doxygen syntax

edenhill: also see test_consumer_wait_assignment(), might be usable.

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,219 @@
...
+static void await_assignment (_consumer_t *c) {
+        TEST_SAY("Waiting for partition assignment\n");
+        while (1) {
+                rd_kafka_topic_partition_list_t *parts = NULL;
+
+                do_consume(c, 1/*1s*/);
```
edenhill: This doesn't really rhyme well within an await_assignment() function

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,219 @@
...
+                         rd_kafka_name(cons->rk), timeout_s);
+                rd_sleep(timeout_s);
+        }
+}
+
+/* Serve op queue until we have an assignment */
+static void await_assignment (_consumer_t *c) {
+        TEST_SAY("Waiting for partition assignment\n");
+        while (1) {
+                rd_kafka_topic_partition_list_t *parts = NULL;
```
edenhill: no need to initialize

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,219 @@
...
+        c[1].rk = test_create_consumer("static", &rebalance_cb,
+                                       rd_kafka_conf_dup(conf), NULL);
+
+        c[0].max_rebalance_cnt = 1;
+        c[1].max_rebalance_cnt = 1;
+
+        test_consumer_subscribe(c[0].rk, subscription);
+        test_consumer_subscribe(c[1].rk, subscription);
+
+        test_create_topic(NULL, topic, 3, 1);
```
edenhill: Suggest you create and seed the topic before subscribing to make the test faster to run (otherwise subscribe will need to slowly poll metadata for the topic to become available)

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,219 @@
...
+        await_assignment(&c[0]);
+
+        /* Wait until session.timeout.ms is exceeded to force a rebalance. */
+        TEST_SAY("Closing c[1], waiting for static instance to be evicted.\n");
+        c[0].max_rebalance_cnt++;
+
+        test_consumer_close(c[1].rk);
+        rd_kafka_destroy(c[1].rk);
+
+        /* 2x heartbeat interval to give time for c[0] to recognize rebalance */
```
edenhill: heartbeat is 3s

INTRODUCTION.md
```diff
@@ -1156,6 +1156,18 @@ The latest stored offset will be automatically committed every
 Broker based consumer groups (requires Apache Kafka broker >=0.9) are supported,
 see KafkaConsumer in rdkafka.h or rdkafkacpp.h
 
+##### Static Consumer groups
+
+By default Kafka consumer groups are rebalanced each time a new consumer joins 
+the group or an existing member leaves. This is what is known as a "dynamic"
```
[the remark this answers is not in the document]
rnpridgeon: KIP-345 refers to it as static or dynamic membership which is more accurate so I'll fix this.

src/rdkafka_request.c
edenhill: Great! 💯

src/rdkafka_request.c
edenhill: I'd say max.poll is not so much of a feature as it is a control mechanism for the client to monitor the application, while static membership is a feature an application might want to rely on.

But we can leave it as-is for now and improve in the future if necessary.

src/rdkafka_request.c
edenhill: Nice rice!

src/rdkafka_request.c
```diff
@@ -1240,17 +1296,26 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
...
+                                                          0, 1,
+                                                          &features);
 
         rkbuf = rd_kafka_buf_new_request(rkb, RD_KAFKAP_LeaveGroup,
                                          1,
                                          RD_KAFKAP_STR_SIZE(group_id) +
                                          RD_KAFKAP_STR_SIZE(member_id));
         rd_kafka_buf_write_kstr(rkbuf, group_id);
         rd_kafka_buf_write_kstr(rkbuf, member_id);
+        rd_kafka_buf_ApiVersion_set(rkbuf, ApiVersion, 0);
```
edenhill: Add an empty line above this call to separate the buf writers from the apiversion, for consistency.

src/rdkafka_request.c
edenhill: :fire: !

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,191 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test.h"
+
+/**
+ * @name KafkaConsumer static membership tests
+ *
+ * Runs two consumers subscribing to a topic simulating various
+ * rebalance scenerios.
```
edenhill: scenarios

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,191 @@
...
+ */
+
+typedef struct _consumer_s {
+        rd_kafka_t *rk;
+        int assign_cnt;
+        int max_rebalance_cnt;
+} _consumer_t;
+
+/**
+ * Calls poll to serve any pending callbacks. Blocks up to timeout_s + 100 ms
```
edenhill: doxygen syntax, `@brief `

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,191 @@
...
+ */
+static void do_consume (_consumer_t *cons, int timeout_s) {
+        rd_kafka_message_t *rkm;
+
+        rkm = rd_kafka_consumer_poll(cons->rk, 100+(timeout_s*1000));
+        if (!rkm)
+                return;
+
+        if (rkm->err)
+                TEST_SAY("%s consumer error: %s",
```
edenhill: Are you expecting errors? If so, handle them specifically, otherwise fail the test on errors.

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,191 @@
...
+        do_consume(&c[0], 1/*1s*/);
+
+        TEST_SAY("Creating a new topic.\n");
+        /* Expanding the subscription forces a rebalance */
+        c[0].max_rebalance_cnt++;
+        c[1].max_rebalance_cnt++;
+
+        /* The topic prefix uses the test id which is "random" */
+        test_create_topic(c[0].rk, tsprintf("%snew", topic), 1, 1);
+        do_consume(&c[0], 1);
```
edenhill: add 1s comment here for consistency

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,191 @@
...
+} _consumer_t;
+
+/**
+ * Calls poll to serve any pending callbacks. Blocks up to timeout_s + 100 ms
+ */
+static void do_consume (_consumer_t *cons, int timeout_s) {
+        rd_kafka_message_t *rkm;
+
+        rkm = rd_kafka_consumer_poll(cons->rk, 100+(timeout_s*1000));
+        if (!rkm)
```
edenhill: I think you want to be more stringent here, if you're expecting a message and don't get one, then fail the test.
Tests should not be forgiving.

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,191 @@
...
+        TEST_SAY("Closing c[1], waiting for static instance to be evicted.\n");
+        c[0].max_rebalance_cnt++;
+
+        test_consumer_close(c[1].rk);
+        rd_kafka_destroy(c[1].rk);
+
+        /* 3x heartbeat interval to give time for c[0] to recognize rebalance */
+        rd_sleep(9);
+
+        rd_kafka_flush(c[0].rk, 5000);
```
edenhill: Flush is really a producer API.

tests/test.c
```diff
@@ -326,7 +326,7 @@ struct test tests[] = {
         _TEST(0098_consumer_txn, 0),
         _TEST(0099_commit_metadata, 0),
         _TEST(0100_thread_interceptors, TEST_F_LOCAL),
-
+        _TEST(0102_static_group_rebalance, 0, TEST_BRKVER(2,3,0,0)),
```
edenhill: reinsert the empty new line that separated automatic tests from manual tests

src/rdkafka_conf.c
```diff
@@ -3457,6 +3457,12 @@ int rd_kafka_conf_warn (rd_kafka_t *rk) {
                                      rk->rk_conf.socket_timeout_ms);
         }
 
+        if (rd_kafka_conf_is_modified(rk->rk_conf, "max.poll.interval.ms") &&
+            rd_kafka_conf_is_modified(rk->rk_conf, "group.instance.id"))
```
edenhill: This is not symmetrical to how group.instance.id is used: what happens if group.instance.id is set to the empty string (e.g., disabled)?

You should check if group.instance.id is non-empty here,.

tests/0102-static_group_rebalance.c
edenhill: But the test should also verify that messages are received as expected, not only that rebalances happen.
Otherwise we don't have any test coverage for a static member consumer that consumes messages, which is the functionality that we're offering.

tests/0102-static_group_rebalance.c
```diff
@@ -177,7 +177,8 @@ int main_0102_static_group_rebalance (int argc, char **argv) {
         /* 3x heartbeat interval to give time for c[0] to recognize rebalance */
         rd_sleep(9);
 
-        rd_kafka_flush(c[0].rk, 5000);
+        do_consume(c[0].rk, 5/*5s*/);
```
edenhill: Since the function name is not suitable for what we are doing here I suggest you add a comment saying: Wait for rebalance, or similar.

src/rdkafka_cgrp.c
```diff
@@ -2227,7 +2227,10 @@ static void rd_kafka_cgrp_unassign_done (rd_kafka_cgrp_t *rkcg,
         if (rd_kafka_destroy_flags_no_consumer_close(rkcg->rkcg_rk))
                 rkcg->rkcg_flags &= ~RD_KAFKA_CGRP_F_LEAVE_ON_UNASSIGN;
 
-        /* KIP-345: Static group members do not send LeaveGroupRequests */
+        /*
+         * KIP-345: Static group members will call unsubscribe to release their partions
```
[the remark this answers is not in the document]
rnpridgeon: I see, that makes much more sense. I was curious why we would want to comment about this after the fact but wasn't sure it was worth questioning. 

I agree we should only disable this flag when `RD_KAFKA_CGRP_F_TERMINATE` is set. I'll add an additional check and retest.

src/rdkafka_conf.c
```diff
@@ -843,6 +843,14 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           _RK(group_id_str),
           "Client group id string. All clients sharing the same group.id "
           "belong to the same group." },
+        { _RK_GLOBAL|_RK_CGRP|_RK_MED, "group.instance.id", _RK_C_STR,
+          _RK(group_instance_id),
+          "Enables static group membership. "
```
edenhill: "Enable", to be consistent with existing properties

INTRODUCTION.md
```diff
@@ -1278,6 +1278,17 @@ The latest stored offset will be automatically committed every
 Broker based consumer groups (requires Apache Kafka broker >=0.9) are supported,
 see KafkaConsumer in rdkafka.h or rdkafkacpp.h
 
+##### Static Consumer groups
+
+By default Kafka consumers are rebalanced each time a new consumer joins
+the group or an existing member leaves. This is what is known as a dynamic
+membership. Apache Kafka >= 2.3.0 introduces static membership.
+Unlike dynamic membership, members can leave and rejoin a group
```
edenhill: maybe "..membership, _static_ members can leave .." to make it clearer

INTRODUCTION.md
```diff
@@ -1278,6 +1278,17 @@ The latest stored offset will be automatically committed every
...
+within the `session.timeout.ms` without triggering a rebalance.
```
edenhill: " and without loosing their partition assignment", but in proper english, which is the net effect

src/rdkafka.c
```diff
@@ -3744,7 +3744,6 @@ char *rd_kafka_memberid (const rd_kafka_t *rk) {
 	return memberid;
 }
 
-
```
edenhill: remove whitespace diff

src/rdkafka.h
```diff
@@ -1414,9 +1414,12 @@ typedef enum {
  * The properties are identical to the Apache Kafka configuration properties
  * whenever possible.
  *
+ * @remark A successfull call to rd_kafka_new() will assume ownership of
```
edenhill: successful

src/rdkafka_cgrp.c
```diff
@@ -1717,12 +1748,14 @@ rd_kafka_cgrp_partitions_fetch_start0 (rd_kafka_cgrp_t *rkcg,
...
-                 * The timer interval is 2 hz */
+                 * The timer interval is 2 hz.
+                 */
+
                 rd_kafka_timer_start(&rkcg->rkcg_rk->rk_timers,
-                             &rkcg->rkcg_max_poll_interval_tmr,
-                             500 * 1000ll /* 500ms */,
-                             rd_kafka_cgrp_max_poll_interval_check_tmr_cb,
-                             rkcg);
+                                &rkcg->rkcg_max_poll_interval_tmr,
```
edenhill: avoid whitespace diffs  (including the comment above)

src/rdkafka_cgrp.c
```diff
@@ -2612,8 +2653,13 @@ rd_kafka_cgrp_max_poll_interval_check_tmr_cb (rd_kafka_timers_t *rkts,
         /* Leave the group before calling rebalance since the standard leave
          * will be triggered first after the rebalance callback has been served.
          * But since the application is blocked still doing processing
-         * that leave will be further delayed. */
-        rd_kafka_cgrp_leave(rkcg);
+         * that leave will be further delayed.
+         *
+         * KIP-345: static group members should continue to respect
+         * `max.poll.interval.ms`but should not send a LeaveGroupRequest.
```
edenhill: missing space

src/rdkafka_conf.c
```diff
@@ -843,6 +843,14 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
+          "Static group members are able to leave and rejoin a group "
+          "within the configured `session.timeout.ms` without prompting a "
+          "group rebalance. Each member of the group must have a "
+          "unique group instance id."
```
edenhill: Missing space

src/rdkafka_request.c
```diff
@@ -876,16 +885,21 @@ int rd_kafka_OffsetCommitRequest (rd_kafka_broker_t *rkb,
...
-                        rd_kafka_buf_write_i64(rkbuf, -1);
         }
 
+        /* v7: GroupInstanceId */
+        if (ApiVersion >= 7)
+                rd_kafka_buf_write_kstr(rkbuf, rkcg->rkcg_group_instance_id);
+
+        /* v2-4: RetentionTime */
+        if (ApiVersion >= 2 && ApiVersion <= 4)
+                        rd_kafka_buf_write_i64(rkbuf, -1);
```
edenhill: indent

src/rdkafka_request.c
```diff
@@ -1102,6 +1134,10 @@ void rd_kafka_handle_SyncGroup (rd_kafka_t *rk,
                 goto err;
         }
 
+        if (request->rkbuf_reqhdr.ApiVersion >= 1)
+                rd_kafka_buf_read_throttle_time(rkbuf);
+
+
```
edenhill: inconsistent extra line, remove.

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2012-2019, Magnus Edenhill
```
edenhill: nit: 2019 (no backdating them files)

tests/test.c
edenhill: Not addressed

src/rdkafka_request.c
```diff
@@ -1211,6 +1253,20 @@ void rd_kafka_JoinGroupRequest (rd_kafka_broker_t *rkb,
...
 
+
+        if (ApiVersion < 5 &&
+            rk->rk_conf.group_instance_id &&
+            rd_interval(&rkb->rkb_suppress.unsupported_kip345,
+                        /* at most once per day */
+                        (rd_ts_t)86400 * 1000 * 1000, 0) > 0)
+                rd_rkb_log(rkb, LOG_NOTICE, "STATICMEMBER",
+                           "Broker does not support KIP-345 "
+                           "(requires Apache Kafka >= v2.3.0): "
```
edenhill: the KIP list says KIP-345 is only partially supported in 2.3.0, fully supported in 2.4.0.
What's the story?

tests/test.c
```diff
@@ -2258,7 +2259,6 @@ test_consume_msgs_easy (const char *group_id, const char *topic,
         test_msgver_clear(&mv);
 }
 
-
```
edenhill: whitespace. Please scan through your diff on github, I find it really helps seeing the same code in a different setting.

tests/test.c
```diff
@@ -3277,7 +3277,43 @@ void test_consumer_poll_no_msgs (const char *what, rd_kafka_t *rk,
 	TEST_ASSERT(cnt == 0, "Expected 0 messages, got %d", cnt);
 }
 
+/**
+ * Consumer poll with expectation that a \p err will be reached
```
edenhill: `@brief ..`

tests/test.c
```diff
@@ -3277,7 +3277,43 @@ void test_consumer_poll_no_msgs (const char *what, rd_kafka_t *rk,
...
+ * within \p timeout_ms.
+ */
+void test_consumer_poll_expect_err (rd_kafka_t *rk, uint64_t testid,
+                                    int timeout_ms, rd_kafka_resp_err_t err) {
+        int64_t tmout = test_clock() + timeout_ms * 1000;
+
+        TEST_SAY("%s: expecting err %s within %dms\n",
```
edenhill: error

tests/test.c
```diff
@@ -3277,7 +3277,43 @@ void test_consumer_poll_no_msgs (const char *what, rd_kafka_t *rk,
...
+                 rd_kafka_name(rk), rd_kafka_err2name(err), timeout_ms);
+
+        do {
+                rd_kafka_message_t *rkmessage;
+                rkmessage = rd_kafka_consumer_poll(rk, timeout_ms);
+                if (!rkmessage)
+                        continue;
+
+                if (rkmessage->err == err) {
+                        TEST_SAY("ERROR: %s\n", rd_kafka_err2str(rkmessage->err));
```
edenhill: There's more information available here, do something like:
`TEST_SAY("Got expected error: %s: %s\n", rd_kafka_err2name(rkmessage->err), rd_kafka_message_errstr());`

tests/test.c
```diff
@@ -3277,7 +3277,43 @@ void test_consumer_poll_no_msgs (const char *what, rd_kafka_t *rk,
...
+                        rd_kafka_message_destroy(rkmessage);
+
+                        return;
+                } else if (rkmessage->err) {
+                        TEST_FAIL("%s [%"PRId32"] error (offset %"PRId64
```
edenhill: `unexpected error`, and print err2name too so we know what error code hit.

tests/test.c
```diff
@@ -3277,7 +3277,43 @@ void test_consumer_poll_no_msgs (const char *what, rd_kafka_t *rk,
...
+                                 rd_kafka_topic_name(rkmessage->rkt) :
+                                 "(no-topic)",
+                                 rkmessage->partition,
+                                 rkmessage->offset,
+                                 rd_kafka_message_errstr(rkmessage));
+                }
 
+                rd_kafka_message_destroy(rkmessage);
+        } while (test_clock() <= tmout);
+        TEST_FAIL("Error %s expected but never returned.", err);
```
edenhill: no ending periods in log messages

edenhill: "never" is incorrect.
"Expected error %s not seen in %dms" or something

tests/0102-static_group_rebalance.c
edenhill: not addressed

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+
+
+        test_msgver_init(&mv, testid);
+        c[0].mv = &mv;
+        c[1].mv = &mv;
+
+        test_create_topic(NULL, topic, 3, 1);
+        test_produce_msgs_easy(topic, testid, RD_KAFKA_PARTITION_UA, msgcnt);
+
+        test_conf_init(&conf, NULL, 60);
```
edenhill: this typically comes first in the test (since it sets the timeout)

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+        test_conf_set(conf, "session.timeout.ms", "5000");
+        test_conf_set(conf, "max.poll.interval.ms", "10001");
```
edenhill: why ..01?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+        test_conf_set(conf, "auto.offset.reset", "earliest");
+        test_conf_set(conf, "topic.metadata.refresh.interval.ms", "500");
+        test_conf_set(conf, "enable.partition.eof", "true");
+        test_conf_set(conf, "isolation.level", "read_uncommitted");
```
edenhill: What is this for? You're not producing transactional messages to the topic, right?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+        rd_kafka_conf_set_opaque(conf, &c[0]);
+        test_conf_set(conf, "group.instance.id", "consumer1");
+        c[0].rk = test_create_consumer(topic, rebalance_cb,
+                                       rd_kafka_conf_dup(conf), NULL);
+
+        rd_kafka_conf_set_opaque(conf, &c[1]);
+        test_conf_set(conf, "group.instance.id", "consumer2");
+        c[1].rk = test_create_consumer(topic, rebalance_cb,
+                                       rd_kafka_conf_dup(conf), NULL);
+        rd_kafka_conf_destroy(conf);
```
edenhill: Why are you dupping conf just to destroy it?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+
+        test_consumer_subscribe(c[0].rk, topics);
+        test_consumer_subscribe(c[1].rk, topics);
+
+        /*
+         * Static members continue to enforce `max.poll.interval.ms`.
+         * These members remain in the member list however so we  must
```
edenhill: spacing

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+         * interleave calls to poll while awaiting our assignment to avoid
+         * unexpected rebalances being triggered.
+         */
+        rebalance_start = test_clock();
+        while(!static_member_wait_rebalance(&c[0], rebalance_start,
```
edenhill: `while (`

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+              &c[0].assigned_at, 1000))
```
edenhill: this indent does not look correct

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+                /* keep consumer 2 alive while we await our assignment */
```
edenhill: `we` -> `consumer 1` ?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+                test_consumer_poll_once(c[1].rk, &mv, 10);
```
edenhill: Do you need a timeout?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+        test_consumer_close(c[1].rk);
+        rd_kafka_destroy(c[1].rk);
+
+        c[1].rk = test_create_handle(RD_KAFKA_CONSUMER, conf);
+        rd_kafka_poll_set_consumer(c[1].rk);
+        test_consumer_subscribe(c[1].rk, topics);
+
+        /* Assign */
+        rebalance_start = test_clock();
+        while(!static_member_wait_rebalance(&c[1], rebalance_start,
```
edenhill: same

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+
+        conf = rd_kafka_conf_dup(rd_kafka_conf(c[1].rk));
+        test_consumer_close(c[1].rk);
+        rd_kafka_destroy(c[1].rk);
+
+        c[1].rk = test_create_handle(RD_KAFKA_CONSUMER, conf);
+        rd_kafka_poll_set_consumer(c[1].rk);
+        test_consumer_subscribe(c[1].rk, topics);
+
+        /* Assign */
```
edenhill: This comment isnt very helpful.
Is it "Wait for assignment"?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+         * group rebalance
+         */
+        c[0].max_rebalance_cnt++;
+        c[1].max_rebalance_cnt++;
+
+        rebalance_start = test_clock();
+        test_create_topic(c->rk, tsprintf("%snew", topic), 1, 1);
+
+        /* Revoke */
+        while(!static_member_wait_rebalance(&c[0], rebalance_start,
```
edenhill: again

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+         * New topics matching the subscription pattern should cause
+         * group rebalance
+         */
+        c[0].max_rebalance_cnt++;
+        c[1].max_rebalance_cnt++;
+
+        rebalance_start = test_clock();
+        test_create_topic(c->rk, tsprintf("%snew", topic), 1, 1);
+
+        /* Revoke */
```
edenhill: Wait for revoke

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+        while(!static_member_wait_rebalance(&c[0], rebalance_start,
+              &c[0].revoked_at, 1000))
+                test_consumer_poll_once(c[1].rk, &mv, 10);
+        static_member_wait_rebalance(&c[1], rebalance_start,
+                                     &c[1].revoked_at, -1);
+
+        /* Assign */
```
edenhill: ditto dito

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+         * These members remain in the member list however so we  must
+         * interleave calls to poll while awaiting our assignment to avoid
+         * unexpected rebalances being triggered.
+         */
+        rebalance_start = test_clock();
+        while(!static_member_wait_rebalance(&c[0], rebalance_start,
+              &c[0].assigned_at, 1000))
+                /* keep consumer 2 alive while we await our assignment */
+                test_consumer_poll_once(c[1].rk, &mv, 10);
+        static_member_wait_rebalance(&c[1], rebalance_start,
```
edenhill: could do with an empty line

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+        /* Assign */
+        while(!static_member_wait_rebalance(&c[1], rebalance_start,
+                                            &c[1].assigned_at, 1000))
+                test_consumer_poll_once(c[0].rk, &mv, 10);
+
+        static_member_wait_rebalance(&c[0], rebalance_start,
+                                     &c[0].assigned_at, -1);
+
+        TEST_SAY("== Testing max poll violation ==\n");
+        /* max.poll.interval.ms should still be enforced by the consumer */
```
edenhill: It is usually better to move this into its own test (in the same file), which allows the tests to run in parallel (faster!) and makes this test more focused.

But don't change it, just wanted to point it out for future tests.

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+        c[0].max_rebalance_cnt++;
+        c[1].max_rebalance_cnt++;
+
+        rebalance_start = test_clock();
+        test_consumer_poll_no_msgs("wait.max.poll", c[0].rk, testid, 15000);
```
edenhill: A comment wouldnt hurt

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,289 @@
...
+                        TEST_SAY0("%s%s[%"PRId32"]",
+                                  i == 0 ? "" : ", ",
+                                  parts->elems[i].topic,
+                                  parts->elems[i].partition);
+                TEST_SAY0("\n");
+
+                break;
+
+        case RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS:
+                TEST_ASSERT(++c->rebalance_cnt == c->max_rebalance_cnt,
```
edenhill: Never pass arguments with side-effects to (evident) macros, since they may be evaluated multiple times.
Instead do:
```
c->rebalance_cnt++;
TEST_ASSERT(c->rebalance_cnt == ...
```

edenhill:
Also update the protocol versions and supported KIPs in INTRODUCTION.md

edenhill:
Bump to latest Kafka version in .travis.yml since your test depends on &gt;=2.3.0

CONFIGURATION.md
```diff
@@ -88,6 +88,7 @@ oauthbearer_token_refresh_cb             |  *  |                 |
 plugin.library.paths                     |  *  |                 |               | low        | List of plugin libraries to load (; separated). The library search path is platform dependent (see dlopen(3) for Unix and LoadLibrary() for Windows). If no filename extension is specified the platform-specific extension (such as .dll or .so) will be appended automatically. <br>*Type: string*
 interceptors                             |  *  |                 |               | low        | Interceptors added through rd_kafka_conf_interceptor_add_..() and any configuration handled by interceptors. <br>*Type: *
 group.id                                 |  C  |                 |               | high       | Client group id string. All clients sharing the same group.id belong to the same group. <br>*Type: string*
+group.instance.id                        |  C  |                 |               | medium     | Enables static group membership. Static group members are able to leave and rejoin a group within the configured `session.timeout.ms` without prompting a group rebalance. Each member of the group must have a unique group instance id.Requires broker version >= 2.3.0 <br>*Type: string*
```
edenhill: Not re-generated after conf.c change

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,297 @@
...
+                                  i == 0 ? "" : ", ",
+                                  parts->elems[i].topic,
+                                  parts->elems[i].partition);
+                TEST_SAY0("\n");
+
+                break;
+
+        case RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS:
+                c->rebalance_cnt++;
+                TEST_ASSERT(c->rebalance_cnt == c->max_rebalance_cnt,
```
edenhill: Shouldn't this be `rebalance_cnt <= max_rebalance_cnt` ?
Otherwise max is not a max

tests/0102-static_group_rebalance.c
edenhill: still not addressed

tests/0102-static_group_rebalance.c
edenhill: If that is a bug in the finalization, fix it now or it won't happen.

tests/0102-static_group_rebalance.c
edenhill: still looks off, what is that, 6 whitespaces and not aligning with paren?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,297 @@
...
+        while (!static_member_wait_rebalance(&c[0], rebalance_start,
+               &c[0].revoked_at, 1000))
+                test_consumer_poll_once(c[1].rk, &mv, 0);
+
+        static_member_wait_rebalance(&c[1], rebalance_start,
+                                     &c[1].revoked_at, -1);
+
+        /* Await assignment */
+        while (!static_member_wait_rebalance(&c[0], rebalance_start,
+                                             &c[0].assigned_at, 1000))
```
edenhill: this indent is better, but also inconsistent with the previous identical while-loops.

tests/0102-static_group_rebalance.c
rnpridgeon: what about renaming it `expected_rebalance_cnt`. We want to ensure we do not rebalance too often or not often enough.

edenhill: practically it can only be set to 1 (or rebalance_cnt+1), right?
So maybe change the logic to check for which rebalance err you are getting, .e.g above the switch do:
`TEST_ASSERT(c->expected_rebalance_event == err)`

which is either ASSIGN_PART.. or REVOKE_PA..

This also makes the test more rigid since it will verify the rebalance event ordering (revoke, assign, revoke, assign)

tests/0102-static_group_rebalance.c
rnpridgeon: yeah I see I was aligning with the wrong opening paren

tests/0102-static_group_rebalance.c
rnpridgeon: If I miss a rebalance that I expect to happen but it didn't I could end up with  a rebalance_cnt of 1 which is less than the expected value of 3. This would falsely advertise that static group membership is working as expected. 

We could track assign and revoke although it seems like overkill so I won't have an assignment without a revocation and I won't actually be a member of that group if I don't have an assignment.

edenhill: That's sort of my point: we don't want the test to assume and skim over important details such as the order of rebalance events, the more rigid the test the better the test coverage.

rnpridgeon: consider it done

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,323 @@
...
+        test_conf_set(conf, "group.instance.id", "consumer2");
+        c[1].rk = test_create_consumer(topic, rebalance_cb,
+                                       rd_kafka_conf_dup(conf), NULL);
+        rd_kafka_conf_destroy(conf);
+
+        test_consumer_subscribe(c[0].rk, topics);
+        test_consumer_subscribe(c[1].rk, topics);
+
+        /*
+         * Static members continue to enforce `max.poll.interval.ms`.
```
edenhill: continue in relation to what? Should this just be "Static members enforce .." ?

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,323 @@
...
+        test_msgver_t mv;
+        int64_t rebalance_start;
+        _consumer_t c[_CONSUMER_CNT] = RD_ZERO_INIT;
+        const int msgcnt = 100;
+        uint64_t testid  = test_id_generate();
+        const char *topic = test_mk_topic_name("0102_static_group_rebalance",
+                                               1);
+        char *topics = rd_strdup(tsprintf("^%s.*", topic));
+
+        test_conf_init(&conf, NULL, 60);
```
edenhill: What's the runtime of the test today? Is it close to 60?
It is usually a good idea to allow an extra 20-30s for long tests like this to make it more robust.

rnpridgeon: Typically 34-40 seconds locally. Which would be 20-30s less than 60. I can check the travis output though to ensure it aligns with that

tests/0102-static_group_rebalance.c
```diff
@@ -0,0 +1,323 @@
...
+                    rd_kafka_name(rk),
+                    rd_kafka_err2name(c->expected_rb_event),
+                    rd_kafka_err2name(err));
+
+        switch (err)
+        {
+        case RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS:
+                TEST_SAY("%s Assignment (%d partition(s)): ",
+                         rd_kafka_name(rk), parts->cnt);
+                for (i = 0 ; i < parts->cnt ; i++)
```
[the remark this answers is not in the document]
rnpridgeon: good to know, sad I missed it

src/rdkafka_conf.c
```diff
@@ -848,6 +848,14 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           _RK(group_id_str),
           "Client group id string. All clients sharing the same group.id "
           "belong to the same group." },
+        { _RK_GLOBAL|_RK_CGRP|_RK_MED, "group.instance.id", _RK_C_STR,
+          _RK(group_instance_id),
+          "Enable static group membership. "
+          "Static group members are able to leave and rejoin a group "
+          "within the configured `session.timeout.ms` without prompting a "
+          "group rebalance. Each member of the group must have a "
+          "unique group instance id. "
```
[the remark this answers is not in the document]
rnpridgeon: good point.

tests/0102-static_group_rebalance.c
rnpridgeon: Bumped to 70 since the additional session timeout test more consistently takes 40 seconds

src/rdkafka_conf.c
edenhill: This is not resolved

src/rdkafka_conf.c
```diff
@@ -850,12 +850,16 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
           _RK(group_instance_id),
-          "Enable static group membership. "
+          "Enables static group membership. "
           "Static group members are able to leave and rejoin a group "
           "within the configured `session.timeout.ms` without prompting a "
-          "group rebalance. Each member of the group must have a "
+          "group rebalance. In order to take advantage of this new behavior "
+          "`session.timeout.ms` should be raised to a larger value. How large "
+          "you set this value will largely be contingent on your orchestration "
+          "environment's ability to detect and react to process failure."
```
edenhill: missing space after .

src/rdkafka_conf.c
```diff
@@ -850,12 +850,16 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
         { _RK_GLOBAL|_RK_CGRP|_RK_MED, "group.instance.id", _RK_C_STR,
           _RK(group_instance_id),
-          "Enable static group membership. "
+          "Enables static group membership. "
           "Static group members are able to leave and rejoin a group "
           "within the configured `session.timeout.ms` without prompting a "
-          "group rebalance. Each member of the group must have a "
+          "group rebalance. In order to take advantage of this new behavior "
+          "`session.timeout.ms` should be raised to a larger value. How large "
+          "you set this value will largely be contingent on your orchestration "
```
edenhill: Don't beat about the bush, this is too vague. Tell them what the timeout should be set to.
E.g., the maximum allowed time for a group member application to restart without triggering a rebalance, or whatever it is.

We're not using "you" form in any other configuration properties, formalize.

tests/0102-static_group_rebalance.c
```diff
@@ -182,9 +183,10 @@ int main_0102_static_group_rebalance (int argc, char **argv) {
 
         TEST_SAY("== Testing consumer restart ==\n");
         conf = rd_kafka_conf_dup(rd_kafka_conf(c[1].rk));
-
+
```
edenhill: remove whitespaces from line

tests/0102-static_group_rebalance.c
```diff
@@ -199,6 +201,11 @@ int main_0102_static_group_rebalance (int argc, char **argv) {
         while (!static_member_wait_rebalance(&c[1], rebalance_start,
                                              &c[1].assigned_at, 1000))
                 test_consumer_poll_once(c[0].rk, &mv, 0);
+        TIMING_STOP(&t_close);
+
+        /* Catch slow close issues earlier (don't wait to call poll on c[0]) */
+        TIMING_ASSERT(&t_close, 0, 6000);
```
edenhill: What's 6000? Is it an arbitrary value or based off some sleep or config?

tests/0102-static_group_rebalance.c
```diff
@@ -302,20 +309,32 @@ int main_0102_static_group_rebalance (int argc, char **argv) {
...
 
+        c[1].expected_rb_event = RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS;
+        static_member_wait_rebalance(&c[1], rebalance_start,
+                                     &c[1].revoked_at, 7000);
+        
+        c[1].expected_rb_event = RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS;
+        static_member_wait_rebalance(&c[1], rebalance_start,
+                                     &c[1].assigned_at, 2000);
+
+        /* Should take at least as long as `session.timeout.ms but less than
```
edenhill: unbalanced backtick

src/rdkafka_conf.c
rnpridgeon: Sort of , I took the liberty of adding a `s` when changing this to use `Enable`as it sounded better to me. Fixed to your exact specification this time.

src/rdkafka_conf.c
```diff
@@ -850,15 +850,12 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
+          "Enable static group membership. "
           "Static group members are able to leave and rejoin a group "
           "within the configured `session.timeout.ms` without prompting a "
-          "group rebalance. In order to take advantage of this new behavior "
-          "`session.timeout.ms` should be raised to a larger value. How large "
-          "you set this value will largely be contingent on your orchestration "
-          "environment's ability to detect and react to process failure."
-          "Each member of the group must have a "
-          "unique group instance id. "
+          "group rebalance. This be used in combination with a larger "
```
edenhill: "This be used", is there a missing can, should, may, must?

CONFIGURATION.md
```diff
@@ -88,6 +88,7 @@ oauthbearer_token_refresh_cb             |  *  |                 |
...
 interceptors                             |  *  |                 |               | low        | Interceptors added through rd_kafka_conf_interceptor_add_..() and any configuration handled by interceptors. <br>*Type: *
 group.id                                 |  C  |                 |               | high       | Client group id string. All clients sharing the same group.id belong to the same group. <br>*Type: string*
+group.instance.id                        |  C  |                 |               | medium     | Enables static group membership. Static group members are able to leave and rejoin a group within the configured `session.timeout.ms` without prompting a group rebalance. In order to take advantage of this new behavior `session.timeout.ms` should be raised to a larger value. How large you set this value will largely be contingent on your orchestration environment's ability to detect and react to process failure.Each member of the group must have a unique group instance id. Requires broker version >= 2.3.0. <br>*Type: string*
```
edenhill: not regenerated

Issue #2508, opened 2019-09-04
Recovery takes too long after stopping and restarting a broker in a cluster


sarkanyi:
Description
===========
Stopping a broker for a while (while producing with a high rate) and then restarting it leads to message losses when the broker rejoins the cluster (basically because librdkafka sees the brokers as DOWN). I don't think this should be the case.

How to reproduce
================
Stop a broker in a cluster for a while then restart it.

Checklist
=========
Please provide the following information:

 - [x] librdkafka version (release number or git tag): `1.0.1`
 - [x] Apache Kafka version: `2.3.0`
 - [x ] librdkafka client configuration: `"linger.ms=1;queue.buffering.max.messages=10000000;batch.num.messages=100000;message.max.bytes=20971520;queued.max.messages.kbytes=2097151;enable.idempotence=true"`
 - [x] Operating system: `Centos 7 (x64)`
 
Attaching stack trace and logs.
[stacktrace.txt](https://github.com/edenhill/librdkafka/files/3574206/stacktrace.txt)

I will try to provide more data as I figure out more details.

sarkanyi:
examples/rdkafka_performance -P -t test -s 343 -c 500000000  -b <> -X linger.ms=1 -X queue.buffering.max.messages=10000000 -X batch.num.messages=1000000 -X message.max.bytes=20971520 -X queued.max.messages.kbytes=2097151 -X enable.idempotence=true -D
Stop broker, and after a few seconds restart it. Cluster needs to have at least 2 brokers.

sarkanyi:
Set min.insync.replicas=1. After the second broker restarts, sending hangs for a while and the cause looks like it's somewhere related to rebalancing:

%7|1568019280.284|REQERR|rdkafka#producer-1| [thrd:192.168.4.2:9095/bootstrap]: 192.168.4.2:9095/1: ProduceRequest failed: Broker: Not leader for partition: actions Refresh,MsgNotPersisted
%7|1568019280.284|MSGSET|rdkafka#producer-1| [thrd:192.168.4.2:9095/bootstrap]: 192.168.4.2:9095/1: nioTopicUP [1]: MessageSet with 59131 message(s) (MsgId 29904096, BaseSeq 29904095) encountered error: Broker: Not leader for partition (actions Refresh,MsgNotPersisted)
[...]

PR #2528, opened 2019-09-16
Fix msgq (re)insertion code to avoid O(N^2) insert sort operations on retry (#2508)

edenhill:
The msgq insert code now properly handles interleaved and overlapping
message range inserts, which may occur during Producer retries for
high-throughput applications.

### DIFF
--- a/src/rdkafka_msg.c
+++ b/src/rdkafka_msg.c
@@ -654,25 +654,91 @@ int rd_kafka_msgq_enq_sorted (const rd_kafka_itopic_t *rkt,
 }
 
 /**
- * @brief Find the insert position (i.e., the previous element)
- *        for message \p rkm.
+ * @brief Find the insert before position (i.e., the msg which comes
+ *        after \p rkm sequencially) for message \p rkm.
+ *
+ * @param rkmq insert queue.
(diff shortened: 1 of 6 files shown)


edenhill:
This will not make v1.2.0.

sarkanyi:
Looks great, no issues with it so far!

kenneth-jia:
> This will not make v1.2.0.
The on-coming release `v1.2.0` would not include the fix? It's such a pity.
During our throughput test (with EOS enabled), we found that once the kafka cluster stuck  for a few seconds, the cpu rate for some internal broker thread would reach nearly 100%!

sarkanyi:
@kenneth-jia For our use I've applied this on top of 1.1.0 in my fork and will use that until we have a stable release. Maybe you can do the same or something similar. Also @edenhill promised that there is a chance of an 1.2.1 in a short while after 1.2.0 with this fix in it.

kenneth-jia:
> @kenneth-jia For our use I've applied this on top of 1.1.0 in my fork and will use that until we have a stable release. Maybe you can do the same or something similar. Also @edenhill said that there is a chance of an 1.2.1 in a short while after 1.2.0 with this fix in it.

Sounds good. Thanks :-)

PR #2531, opened 2019-09-19
Include rack field in broker metadata

mezzi:
Added the "rack" field to broker metadata struct.


src/rdkafka.h
```diff
@@ -3838,6 +3838,7 @@ typedef struct rd_kafka_metadata_broker {
         int32_t     id;             /**< Broker Id */
         char       *host;           /**< Broker hostname */
         int         port;           /**< Broker listening port */
+        char       *rack;           /**< Broker rack */
```
edenhill: Due to ABI guarantees we can't modify public structs (defined in rdkafka.h).
This is a known issue with the metadata API and we need to conceive something with private types and accessor methods, but that is a bigger change.

mezzi: Hi @edenhill, thanks for the quick feedback and for bringing this to my attention, as it led me to learn more about the subject. Do you mind expanding a little on the use of private types and accessor methods?

If I'm interpreting that correctly, that means:
1. using (and allocating memory for) a new struct type internally for broker metadata containing the `char *rack` field, while casting pointers used and returned by public APIs to the public `rd_kafka_metadata_broker *` type, and 
2. introducing an accessor method to the public API to get `rack` from the underlying private struct.

Is that something like what you had in mind?

edenhill: A private type, like rd_kafka_t, does not have a public struct so that application's can see or access individual fields, instead accessor methods are used on the rd_kafka_t instance to get and set stuff.

The metadata API has public structs, and due to ABI compatibility they must not be changed in any way or applications will break when librdkafka is upgraded, which is unacceptable under our guarantees.

So there is currently no way of extending the metadata API at this point, short of rewriting it which is a bigger effort.

As a workaround I suggest to add the rack information to the [statistics broker object](https://github.com/edenhill/librdkafka/blob/master/STATISTICS.md#brokers).
Since stats is just JSON there is no problem in extending these objects.

mezzi: Understood, thank you. I'll go ahead and close this PR and open a new one with the stats workaround you suggested (or a proposal if I come up with one).

PR #2537, opened 2019-09-25
Improve test build and run instructions

edenhill:


tests/README.md
```diff
@@ -0,0 +1,386 @@
...
+leaks, these shall be ignored. The memory leak report is only relevant
+when the test suite passes.
+
+**NOTE**: The OSX version of ASAN does not provide memory leak protection,
+          you will need to run the test suite on Linux (native or in Docker).
+
+**NOTE**: ASAN, TSAN and valgrind are mutually exclusive.
+
+
+#### Valgrind - memory checker
```
rnpridgeon: > Valgrind only runs on Linux.

https://www.gungorbudak.com/blog/2018/04/28/how-to-install-valgrind-on-macos-high-sierra/

tests/README.md
```diff
@@ -0,0 +1,386 @@
...
+    $ ./run-test.sh ./merged valgrind   # memory checking
+
+or with helgrind (the valgrind thread checker):
+
+    $ ./run-test.sh ./merged helgrind   # thread checking
+
+
+To run the tests in gdb:
+
+**NOTE**: gdb support is flaky on OSX due to signing issues.
```
rnpridgeon: Given the many caveats associated with OSX it may be nice to include a recommended environment in the intro in addition to the notes scattered through the README.md.

tests/README.md
rnpridgeon: Although admittedly this work around isn't super helpful since you should run *SAN on linux.

tests/README.md
```diff
@@ -0,0 +1,386 @@
...
+This script will rebuild librdkafka and the test suite with ASAN enabled.
+
+Then run tests as usual. Memory access issues will be reported on stderr
+in real time as they happen (and the test will fail eventually), while
+memory leaks will be reported on stderr when the test run exits successfully,
+i.e., no tests failed.
+
+Test failures will typically cause the current test to exit hard without
+cleaning up, in which case there will be a large number of reported memory
+leaks, these shall be ignored. The memory leak report is only relevant
```
rnpridgeon: in my experience ASAN simply doesn't print anything on test failure. Valgrind will report a bunch of extra junk but ASAN remained silent.

tests/README.md
```diff
@@ -0,0 +1,386 @@
...
+## Test numbers
+
+Automated tests: 0000-0999
+Manual tests:    8000-8999
+
+
+
+## A guide to testing, verifying, and troubleshooting, librdkafka
+
+
```
rnpridgeon: It would be nice to cover the additional files you must touch to get your tests to run locally and on ci. 

win32/tests/tests.vcxproj
tests/test.c

tests/README.md
edenhill: Until the brew version works out of the box we'll treat it as non-working.

As for "you should run *SAN on linux.": that is not correct, you'll use valgrind when ASAN's level of instrumentation is not enough.

tests/README.md
edenhill: It depends.

tests/Makefile
```diff
@@ -113,3 +115,31 @@ realclean: clean clean-output
 
 java: .PHONY
 	make -C java
+
+# Run test-suite with ASAN
+asan:
```
mhowlett: will anyone ever use these options here? might be an improvement to leave them out, either way minor point.

tests/README.md
```diff
@@ -0,0 +1,427 @@
...
+## Automated broker cluster setup using trivup
+
+A local broker cluster can be set up using
+[trivup](https://github.com/edenhill/trivup), which is a Python package
+available on PyPi.
+These self-contained clusters are used to run the librdkafka test suite
+on a number of different broker versions or with specific broker configs.
+
+trivup will download the specified Kafka version into its root directory,
```
mhowlett: side note: this is broken on my machine for older versions not kept on mirrors. At a glance, the code in trivup looks like it was written to try a mirror first then the main archive, but it's not working - it's giving a not a tgz archive, so it's probably getting back a 404 page or something.

tests/README.md
```diff
@@ -0,0 +1,427 @@
...
+the root directory is also used for cluster instances, where Kafka will
+write messages, lgos, etc.
```
mhowlett: lgos -> logs or explain what lgos is.

tests/README.md
```diff
@@ -0,0 +1,427 @@
...
+    $ make quick
+
+    # Run tests in sequence
+    $ make run_seq
+
+    # Run specific test
+    $ TESTS=0004 make
+
+    # Run test(s) with helgrind, valgrind, gdb
+    $ TESTS=0009 ./run-test.sh ./merged valgrind|helgrind|gdb
```
mhowlett: but how do you integrate that with VS Code for a beautiful interactive debugging experience?

tests/README.md
```diff
@@ -0,0 +1,427 @@
...
+
+
+All tests in the 0000-0999 series are run automatically with `make`.
+
+Tests 1000-1999 are subject to specific non-standard setups or broker
+configuration, these tests are run with `TESTS=1nnn make`.
+See comments in the test's source file for specific requirements.
+
+To insert test results into SQLite database make sure the `sqlite3` utility
```
mhowlett: all data should flow through kafka on it's way to a destination, not go direct.

mhowlett: that was an attempt at humor, which failed badly. should i remove it?

tests/README.md
```diff
@@ -0,0 +1,427 @@
...
+## A guide to testing, verifying, and troubleshooting, librdkafka
+
+
+### Creating a development build
+
+The [dev-conf.sh](../dev-conf.sh) script configures and builds librdkafka and
+the test suite for development use, enabling extra runtime
+checks (`ENABLE_DEVEL`, `rd_dassert()`, etc), disabling optimization
+(to get accurate stack traces and line numbers), etc.
+It also makes it easy to enable ASAN or TSAN (see below) by
```
mhowlett: this line is out of place / grammer not great. maybe another note below?

tests/README.md
```diff
@@ -0,0 +1,427 @@
...
+    # bare means to run the test without valgrind
+    $ TESTS=0061 TEST_DEBUG=cgrp,fetch ./until-fail.sh bare
+
+
+### How to run tests
+
+The standard way to run the test suite is firing up a trivup cluster
+in an interactive shell:
+
+    $ ./interactive_broker_version.py 2.3.0   # Broker version
```
mhowlett: maybe document `trunk` explicitly here.

tests/README.md
```diff
@@ -0,0 +1,427 @@
...
+
+
+Run a single test:
+
+    $ TESTS=0034 make
+
+
+Run test suite with valgrind (see instructions below):
+
+    $ ./run-test.sh ./merged valgrind   # memory checking
```
mhowlett: without digging, i'm wondering what `./merged` is. might want to note that in passing.

edenhill: It is a stupid name, that's what it is. I'm changing it to test-runner.

tests/README.md
edenhill: I tried to ignore trunk runs since they are more complex and error prone.
Do you think we need them documented here?

tests/README.md
edenhill: That should be fixed on the master version of trivup.

edenhill:
Please give this a final glance before merging

tests/README.md
mhowlett: i mentioned it because it would have been helpful for me with #2518 and we'll more than likely need to test against trunk in the future as we move to working in parallel with java.

edenhill: Makes sense, I'd like to add that in another PR later though since I think there are some necessary improvements to source paths, etc.

mhowlett: sounds good

PR #2540, opened 2019-09-26
Add supported KIPs and ApiVersions to manual

edenhill:
...removed some outdated benchmarks and transferred some wiki pages to the manual.

### DIFF
--- a/INTRODUCTION.md
+++ b/INTRODUCTION.md
@@ -5,93 +5,104 @@ librdkafka is a high performance C implementation of the Apache
 Kafka client, providing a reliable and performant client for production use.
 librdkafka also provides a native C++ interface.
 
-## Contents
+<!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
+**Table of Contents**
 
-The following chapters are available in this document
+- [Introduction to librdkafka - the Apache Kafka C/C++ client library](#introduction-to-librdkafka---the-apache-kafka-cc-client-library)
+    - [Performance](#performance)
+        - [High throughput](#high-throughput)
+        - [Low latency](#low-latency)
+            - [Latency measurement](#latency-measurement)
+        - [Compression](#compression)
+    - [Message reliability](#message-reliability)
+        - [Producer message delivery success](#producer-message-delivery-success)
+        - [Producer message delivery failure](#producer-message-delivery-failure)
+            - [Error: Timed out in transmission queue](#error-timed-out-in-transmission-queue)
+            - [Error: Timed out in flight to/from broker](#error-timed-out-in-flight-tofrom-broker)
+            - [Error: Temporary broker-side error](#error-temporary-broker-side-error)
+            - [Error: Temporary errors due to stale metadata](#error-temporary-errors-due-to-stale-metadata)
+            - [Error: Local time out](#error-local-time-out)
+            - [Error: Permanent errors](#error-permanent-errors)
+        - [Producer retries](#producer-retries)
+        - [Reordering](#reordering)
+        - [Idempotent Producer](#idempotent-producer)
+            - [Guarantees](#guarantees)
+            - [Ordering and message sequence numbers](#ordering-and-message-sequence-numbers)
+            - [Partitioner considerations](#partitioner-considerations)
+            - [Message timeout considerations](#message-timeout-considerations)
+            - [Leader change](#leader-change)
+            - [Error handling](#error-handling)
+                - [RD_KAFKA_RESP_ERR_OUT_OF_ORDER_SEQUENCE_NUMBER](#rdkafkaresperroutofordersequencenumber)
+                - [RD_KAFKA_RESP_ERR_DUPLICATE_SEQUENCE_NUMBER](#rdkafkaresperrduplicatesequencenumber)
+                - [RD_KAFKA_RESP_ERR_UNKNOWN_PRODUCER_ID](#rdkafkaresperrunknownproducerid)
+                - [Standard errors](#standard-errors)
+                - [Message persistence status](#message-persistence-status)
+    - [Usage](#usage)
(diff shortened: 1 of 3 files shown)


INTRODUCTION.md
```diff
@@ -5,93 +5,102 @@ librdkafka is a high performance C implementation of the Apache
...
-	  **230000 messages/second**, **23 MB/second**
+or if a low latency service is required, or a balance between the two, all
+through the configuration property interface.
 
+The single most important configuration properties for performance tuning is
+`linger.ms` - how long to wait for `batch.num.messages` to fill up in the
+local per-partition queue before sending the batch to the broker.
 
+A lower value improves latency at the cost of lower throughput and higher
+per-message overhead.
```
mhowlett: this commentary isn't quite right. maybe replace these two paragraphs with:

> In low throughput scenarios, a lower value improves latency. As throughput increases, the cost of each broker request becomes significant impacting both maximum throughput and latency. For higher throughput applications, latency will actually be lower using a higher `linger.ms` due to decreased load on the broker. A good general purpose setting is 5ms. For applications seeking maximum throughput, the recommended value is >= 50ms.

edenhill: thank you!

INTRODUCTION.md
```diff
@@ -743,14 +752,123 @@ Configuration is applied prior to object creation using the
...
+ * `TopicPartition`
+ * `Event`
+ * `Queue`
+
+
+#### High-level KafkaConsumer
+
+Proper termination sequence for the high-level KafkaConsumer is:
+```c
+     /* 1) Close the consumer, committing final offsets, etc. */
```
mhowlett: maybe replace "Close the consumer" which is a tautology here with "Leave the group, ..."

INTRODUCTION.md
```diff
@@ -743,14 +752,123 @@ Configuration is applied prior to object creation using the
...
+```
+
+**NOTE**: There is no need to unsubscribe prior to calling `rd_kafka_consumer_close()`.
+
+**NOTE**: Any topic objects created must be destroyed prior to rd_kafka_destroy()
+
+Effects of not doing the above, for:
+ 1. Final offsets are not committed. It is okay to omit the
+    `rd_kafka_consumer_close()` call if an immediate termination of
+    the consumer, without committing final offsets or leaving the group, is
```
mhowlett: this may be misinterpreted to mean the consumer stays in the group. maybe just leave out "without committing final offsets or leaving the group" and make the edit suggested above to mention leaving the group.

mhowlett: actually no, maybe change first sentence to "Final offsets are not committed and the consumer will leave the group uncleanly" or maybe better something that mentions session timeout

INTRODUCTION.md
edenhill: Tautology is so my thing, I'm really great at repeating myself, saying the same thing at least twice. Finally there is a word for it, thank you Matt for putting words to my feelings.

INTRODUCTION.md
```diff
@@ -1258,70 +1376,234 @@ The purged messages in queue will fail with error code set to
...
-**30 megabytes per second**.
+For librdkafka >= v1.0.0 there is no need to set any api.version-related
+configuration parameters, the defaults are tailored for broker version 0.10.0.0
+or later.
 
+For librdkafka < v1.0.0, please specify:
+```
+api.version.request=true
+api.version.fallback.ms=0
+```
```
mhowlett: in the client examples, we still specify:

```
    'broker.version.fallback': '0.10.0.0',
    'api.version.fallback.ms': 0,
```

by my reading of these docs, that is not required. is that correct?

if so, we should update our examples. I also think it'd be worth updating the cut-n-paste docs in confluent cloud to have these commented out with an explanation of when they're needed (pre librdkafka 1.0 only) as they come across as a bit weird.

INTRODUCTION.md
```diff
@@ -1258,70 +1376,234 @@ The purged messages in queue will fail with error code set to
...
+| 12      | Heartbeat               | 3           | 3                       |
+| 13      | LeaveGroup              | 3           | 1                       |
+| 14      | SyncGroup               | 3           | 3                       |
+| 15      | DescribeGroups          | 4           | 0                       |
+| 16      | ListGroups              | 2           | 0                       |
+| 17      | SaslHandshake           | 1           | 1                       |
+| 18      | ApiVersions             | 2           | 0                       |
+| 19      | CreateTopics            | 4           | 2                       |
+| 20      | DeleteTopics            | 3           | 1                       |
+| 21      | DeleteRecords           | 1           | -                       |
```
mhowlett: can't wait for `DeleteRecords`

INTRODUCTION.md
edenhill: Our examples should definitely be updated, as for CCloud, that's trickier: not sure multiple variants is less confusing, and it is likely that the user does not know the version of its Kafka client.

INTRODUCTION.md
```diff
@@ -1258,70 +1376,234 @@ The purged messages in queue will fail with error code set to
...
+| 35      | DescribeLogDirs         | 1           | -                       |
+| 36      | SaslAuthenticate        | 1           | 0                       |
+| 37      | CreatePartitions        | 1           | 0                       |
+| 38      | CreateDelegationToken   | 1           | -                       |
+| 39      | RenewDelegationToken    | 1           | -                       |
+| 40      | ExpireDelegationToken   | 1           | -                       |
+| 41      | DescribeDelegationToken | 1           | -                       |
+| 42      | DeleteGroups            | 1           | -                       |
+| 43      | ElectPreferredLeaders   | 0           | -                       |
+| 44      | IncrementalAlterConfigs | 0           | -                       |
```
mhowlett: nice!

PR #2546, opened 2019-10-01
Support for logging TLS session keys using SSLKEYLOGFILE environment variable

piotrsmolinski:
SSLKEYLOGFILE environment variable is standard way of declaring the file to write the TLS session keys. This mechanism is used by major web browsers as well as by cURL and other projects. The session key logging is a powerful mechanism to debug network services in such tools as Wireshark.

In Java the same solution is achieved using jSslKeyLog agent (https://svn.code.sf.net/p/jsslkeylog/code). The patch provides the feature to be available in librdkafka based Kafka applications.

The code in the PR was written similar way as the same feature in cURL.


edenhill:
Please follow the coding style guide: https://github.com/edenhill/librdkafka/blob/master/CONTRIBUTING.md#librdkafka-c-style-guide

PR #2559, opened 2019-10-09
Reorganized examples and added cleaner consumer and producer examples

edenhill:
Also removed the zookeeper example since it is of no use anymore.


examples/.gitignore
```diff
@@ -1,9 +1,11 @@
 rdkafka_example
 rdkafka_performance
 rdkafka_example_cpp
-rdkafka_consumer_example
-rdkafka_consumer_example_cpp
+rdkafka_complex_consumer_example
```
rnpridgeon: No ccloud example?

examples/producer.c
```diff
@@ -225,7 +210,7 @@ int main (int argc, char **argv) {
                 } else {
                         fprintf(stderr, "%% Enqueued message (%zd bytes) "
                                 "for topic %s\n",
-                                len, rd_kafka_topic_name(rkt));
+                                len, topic);
                 }
```
rnpridgeon: You could also use the deliver report event to serve up your reports from the background queue no? We should provide an example of that.

edenhill: These are aimed to the be bare minimal examples, there's a bunch of existing more complex examples.

As for background thread delivery reports; another day :)

examples/.gitignore
edenhill: The [ccloud examples](https://github.com/confluentinc/examples/blob/5.3.1-post/clients/cloud/README.md) are linked from the top-level [README](https://github.com/edenhill/librdkafka#usage-in-code) thanks to @ybyzek

PR #2560, opened 2019-10-09
Fix warnings from newer GCC

edenhill:


src/rdaddr.c
```diff
@@ -39,7 +39,7 @@ const char *rd_sockaddr2str (const void *addr, int flags) {
 	const rd_sockaddr_inx_t *a = (const rd_sockaddr_inx_t *)addr;
 	static RD_TLS char ret[32][INET6_ADDRSTRLEN + 16];
 	static RD_TLS int  reti = 0;
-	char portstr[64];
+	char portstr[32];
```
rnpridgeon: Mind commenting on the warning you are resolving here?

Issue #2450, opened 2019-08-02
CPU increase from 250% to 360% after upgrade from 0.11.6 to 1.1.0


eelyaj:
Description
===========
I upgraded librdkafka from 0.11.6 to 1.1.0. I found that the CPU increased from 250% to 360%.
The 'perf top' shows the cpu of 'rd_kafka_toppar_producer_serve' is around 10%.

perf top result in librdkafka 1.1.0:
  10.00%  librdkafka.so.1       [.] rd_kafka_toppar_producer_serve
   6.59%  libc-2.17.so          [.] __memcpy_ssse3
   5.02%  [kernel]              [k] _raw_spin_unlock_irqrestore
   4.90%  libpthread-2.17.so    [.] pthread_mutex_lock
   4.54%  UfdrUdpCollector      [.] PbMessage::serialize

but in librdkafka 0.11.6, the perf top result shows:
   9.42%  libc-2.17.so          [.] __memcpy_ssse3
   6.72%  UfdrUdpCollector      [.] PbMessage::serialize
   5.60%  liblz4.so.1.8.0       [.] LZ4_compress_fast_extState
   4.65%  UfdrUdpCollector      [.] std::_Hashtable<UfdrJoinKey, std::pair<UfdrJoinKey const...
   2.88%  [kernel]              [k] _raw_spin_unlock_irqrestore
   2.82%  libc-2.17.so          [.] __memset_sse2

Maybe there are something wrong with the function 'rd_kafka_toppar_producer_serve'.

How to reproduce
================
1. install librdkafka 1.1.0
2. clean and make my application
3. run application

Checklist
=========
Please provide the following information:

 - [x] librdkafka version (release number or git tag): `0.11.6`
 - [x] Apache Kafka version: `1.1.1`
 - [x] librdkafka client configuration: `"queue.buffering.max.ms": "1000", "queue.buffering.max.messages": "1000000","queue.buffering.max.kbytes": "2048000","batch.num.messages": "10000","compression.codec": "lz4","socket.send.buffer.bytes": "3200000","socket.receive.buffer.bytes": "3200000","message.max.bytes": "209715200"`
 - [ ] Operating system: `<Centos 7.2 (x64)>`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
Can you provide more detailed profiling?

eelyaj:
[...]

PR #2573, opened 2019-10-15
Fix producer insert msgq regression in v1.2.1 (#2450)

edenhill:
The common case where the source queue is simply appended
to the destination queue was not optimized, causing O(n) scans
to find the insert position.

Issue introduced in v1.2.1

### DIFF
--- a/src/rdkafka_msg.c
+++ b/src/rdkafka_msg.c
@@ -1831,9 +1831,13 @@ unittest_msgq_insert_all_sort (const char *what,
         ut_rd_kafka_msgq_purge(&srcq);
         ut_rd_kafka_msgq_purge(&destq);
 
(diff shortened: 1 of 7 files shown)


rnpridgeon:
How confident are we that the us_per_msg max value is sufficiently tolerant of to avoid frequent test flakiness?  

RDUT: FAIL: rdkafka_msg.c:1942: unittest_msgq_insert_each_sort: assert failed: !(us_per_msg > max_us_per_msg + 0.0001): maximum us/msg exceeded: 0.7500 > 0.3000 us/msg


Other than that LGTM

edenhill:
> How confident are we that the us_per_msg max value is sufficiently tolerant of to avoid frequent test flakiness?

Since that unittest runs an isolated part of the code base it is not subject to external factors (such as networks), other than the system load.

PR #2605, opened 2019-11-05
Transactional Producer (full EOS, KIP-98)

edenhill:
This is work in progress.

Current status:
 * :heavy_check_mark: Functionally complete - try it out in action!
 * :heavy_check_mark: Functional tests are complete.
 * :heavy_check_mark: Public API is settled.
 * :heavy_check_mark: Proper error handling across, including controller switches during transactions
 * :construction: API docs and manual are not done and will be finalized in a sub-sequent PR.

Review guide:
 * The transactional producer relies on the idempotent producer: to acquire the initial ProducerID, and for the message guarantees. Most fatal errors in the idempotent producer are only fatal to the current transaction when using the transactional producer, resulting in the ABORTABLE_ERROR state.
 * Most in-transaction errors, be they from the underlying idempotent producer, or failure to add offsets or partitions to the transaction, result in an Abortable Error, which is a transactional error state that requires the application to call abort_transaction() and then start over again with begin_transaction().
 * All the transactional logic is the transaction manager, rdkafka_txnmgr.c, with some leaky abstraction in rdkafka_idempotence.c for the interplay between the idempotent producer and the transaction manager.
 * The use of coord_req() (rdkafka_coord.c), which is an abstraction on top of coordinator brokers makes it easier to perform retries, etc. External community work on the Admin API has already started using it (based on the txns branch).
 * Matt's 0098 transactional consumer test has been extended to run identical tests with the Java txn producer and the librdkafka txn producer.
 * Ignore documentation FIXMEs for now, docs will be finalized after this PR has been merged.

Miscellaneous:
 * Adds a builtin mock cluster with error injection.
 * Manual unittest code coverage functionality as an alternative to object inspection.


INTRODUCTION.md
```diff
@@ -701,15 +701,106 @@ which returns one of the following values:
...
 
+### Transactional Producer
 
+#### FIXME: misc
+
+To make sure messages time out (in case of connectivity problems, etc) within
+the transaction, the `message.timeout.ms` configuration property must be
+set lower than the `transaction.timeout.ms`.
+If `message.timeout.ms` is not explicitly configured it will be adjusted
+automatically.
```
mhowlett: I would add the text "(this is enforced)" after "set lower than the `transaction.timeout.ms`" - i assume it is. it's nice to know from the documentation that this is not something I need to worry about (i.e. if i screw up the config i can be sure to get an error).

src/rdkafka.h
```diff
@@ -6225,6 +6231,350 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ * the transaction aware consumer (\c isolation.level=read_committed).
+ *
+ * A producer instance is configured for transactions by setting the
+ * \c transactional.id to an identifier unique for the application. This
+ * id will be used to fence stale transactions from previous instances of
+ * the application, typically following an outage or crash.
+ *
+ * After creating the transactional produce instance using rd_kafka_new()
+ * the transactional state must be initialized by calling
+ * rd_kafka_init_transactions(). This is a blocking call that will
```
mhowlett: we should perhaps start accommodating cancellable versions of sync calls (no time like the present?)

To do this, you'd need variants of the sync methods that return immediately and signal their completion via an event on a queue. you then need some way for the application to signal to librdkafka a cancellation is desired and a mechanism for librdkafka to pick up on that. there are various ways that could be set up. for example, you could pass in a cancellation token similar to .net (see `CancellationTokenSource` + `CancellationToken` classes). Or `rd_kafka_init_transactions` could be responsible for generating a token, returning it, and then if the application wants to cancel the operation, it uses the returned token to pass into a `rd_kafka_init_transactions_cancel` call. there's precedent for this approach too - e.g. timers in javascirpt work somewhat like this.

my instincts at the moment are erring on something like the second approach. you'd want to avoid allocs, so the token could perhaps be an incrementing id.

edenhill: I think the queue interface makes most sense since it is what we use in other places, and it is very flexible and caters to all kinds of asynchronous needs.
I.e., you can use it to wait on a single object (the queue) for multiple outstanding operations, you can cancel it at any time by destroying the queue, you can forward the results to another queue, etc.

But, all this is outside the scope of the initial API.

mhowlett: sounds good.

CONFIGURATION.md
```diff
@@ -113,6 +113,8 @@ offset_commit_cb                         |  C  |                 |
...
 client.rack                              |  *  |                 |               | low        | A rack identifier for this client. This can be any string value which indicates where this client is physically located. It corresponds with the broker config `broker.rack`. <br>*Type: string*
+transactional.id                         |  P  |                 |               | high       | The TransactionalId to use for transactional delivery. This enables reliability semantics which span multiple producer sessions since it allows the client to guarantee that transactions using the same TransactionalId have been completed prior to starting any new transactions. If no TransactionalId is provided, then the producer is limited to idempotent delivery. Note that enable.idempotence must be enabled if a TransactionalId is configured. The default is null, which means transactions cannot be used. Note that, by default, transactions require a cluster of at least three brokers which is the recommended setting for production; for development you can change this, by adjusting broker setting transaction.state.log.replication.factor. <br>*Type: string*
```
mhowlett: > Note that enable.idempotence must be enabled if a TransactionalId is configured.

this is either not how it works (i've been apparently using transactions without setting enable.idemptence) or you should return an error on incorrect configuration to alert the user there's a problem.

----

Also, suggested alternate text (i think clearer):

"The transactional.id is used to identify the same producer instance across process restarts. It allows the client to guarantee that transactions corresponding earlier instances of the client have been finalized prior to starting any new transactions, and that any zombie instances are fenced off. If no transactional.id is provided, then the producer is limited to idempotent delivery. Note that by default, transactions require a cluster of at least three brokers which is the recommended setting for production; for development you can change this by adjusting broker setting transaction.state.log.replication.factor."

CONFIGURATION.md
```diff
@@ -113,6 +113,8 @@ offset_commit_cb                         |  C  |                 |
...
+transaction.timeout.ms                   |  P  | 1000 .. 2147483647 |         60000 | medium     | The maximum amount of time in milliseconds that the transaction coordinator will wait for a transaction status update from the producer before proactively aborting the ongoing transaction. If this value is larger than the `transaction.max.timeout.ms` setting in the broker, the init_transactions() call will fail with ERR_INVALID_TRANSACTION_TIMEOUT. The transaction timeout must be at least 1000 ms larger than `message.timeout.ms` and `socket.timeout.ms`. <br>*Type: integer*
```
mhowlett: 'setting in the broker' -> 'setting on the broker' i think

CONFIGURATION.md
edenhill: This was copied verbatim from the Java client and subject for rewrite, thanks for beating me to it!

And yes, enable.idempotence is set automatically when transactional.id is set.

INTRODUCTION.md
```diff
@@ -705,15 +705,106 @@ which returns one of the following values:
...
+    rd_kafka_t *producer = rd_kafka_new(RD_KAFKA_PRODUCER, pconf);
+
+    rd_kafka_init_transactions(producer);
+
+
+    /* Consumer */
+    rd_kafka_conf_t *cconf = rd_kafka_conf_new();
+    rd_kafka_conf_set(cconf, "bootstrap.servers", "mybroker");
+    rd_kafka_conf_set(cconf, "group.id", "my-group-id");
+    rd_kafka_conf_set(cconf, "enable.auto.commit", "false");
```
mhowlett: ... this is an unfortunate api flaw. `enable.auto.commit=true` is such a bad default.... this is just one more reason why.

INTRODUCTION.md
```diff
@@ -705,15 +705,106 @@ which returns one of the following values:
 This method should be called by the application on delivery report error.
 
 
+### Transactional Producer
 
+#### FIXME: misc
```
mhowlett: I think you probably want to do a search for 'FIXME'.

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ * sub-sequent transactional API calls will return FIXME (what error??).
+ *
+ * @{
+ */
+
+
+/**
+ * @brief Initialize transactions for the producer instance.
+ *
+ * Needs to be called before any other methods when the \c transactional.id
```
mhowlett: maybe longer winded is better "this method needs to be ..."

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ *   * rd_kafka_produce() (et.al)
+ *   * rd_kafka_send_offsets_to_transaction()
+ *   * rd_kafka_commit_transaction()
+ *   * rd_kafka_abort_transaction()
+ *
+ * @param rk Producer instance.
+ * @param timeout_ms The maximum time to block. On timeout the operation
+ *                   may continue in the background, depending on state,
+ *                   and it is okay to call init_transactions() again.
+ * @param errstr A human readable error string (nul-terminated) is written to
```
mhowlett: null-terminated

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ *               this location that must be of at least \p errstr_size bytes.
+ *               The \p errstr is only written to if there is a fatal error.
+ * @param errstr_size Writable size in \p errstr.
+ *
+ * @remark This function may block up to \p timeout_ms milliseconds.
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success,
```
mhowlett: nice docs

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+
+/**
+ * @brief Begin a new transaction.
+ *
+ * rd_kafka_init_transactions() must have been called successfully (once)
+ * before this function is called.
+ *
+ * Any messages produced, offsets sent (rd_kafka_send_offsets_to_transaction()),
+ * etc, after the successful return of this function will be part of
+ * the transaction and committed or aborted atomatically.
```
mhowlett: optional extra: it would be nice to explicitly state whether or not produce requests must be part of a transaction following a call to init transactions.

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ *
+ * Finish the transaction by calling rd_kafka_commit_transaction() or
+ * abort the transaction by calling rd_kafka_abort_transaction().
+ *
+ * @param rk Producer instance.
+ * @param errstr A human readable error string (nul-terminated) is written to
```
mhowlett: nul -> null or i will start calling you 'disagreeable frenchman'

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+
+/**
+ * @brief Sends a list of topic partition offsets to the consumer group
+ *        coordinator for \p consumer_group_id, and marks the offsets as part
+ *        part of the current transaction.
+ *        These offsets will be considered committed only if the transaction is
+ *        committed successfully.
+ *
+ *        The offsets should be the next message your application will consume,
+ *        i.e., the last processed message's offset + 1 for each partition.
```
mhowlett: this is really dangerous - people are going to screw this up by default because everyone expects committed offset to be the last messages successfully read (and it's what you naturally have in your code without doing any arithmetic, which people won't do by default unless they're thinking hard about it, which they won't be). when offsets are committed using the consumer, this isn't such a big deal because the consumer does the calculation for you behind the scenes (for most variants). But in this case, the user will be in the loop.

given this, i think in the higher level bindings we should have an overload of `SendOffsetsToTransaction` which accepts a consumer instance. the java producer doesn't have this.

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ *        rd_kafka_position() (on the consumer) to get the current offsets for
+ *        the partitions assigned to the consumer.
+ *
+ *        Use this method at the end of a consume-transform-produce loop prior
+ *        to committing the transaction with rd_kafka_commit_transaction().
+ *
+ * @param rk Producer instance.
+ * @param offsets List of offsets to commit to the consumer group upon
+ *                successful commit of the transaction. Offsets should be
+ *                the next message to consume, e.g., last processed message + 1.
```
mhowlett: the java producer allows you to commit `OffsetAndMetadata`. what's all that about and why don't we have it.

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ * @param errstr_size Writable size in \p errstr.
+ *
+ * @remark This function must be called on the transactional producer instance,
+ *         not the consumer.
+ *
+ * @remark The consumer must disable auto commits
+ *         (set \c enable.auto.commit to false on the consumer).
+ *
+ * @remark Logical and invalid offsets (such as RD_KAFKA_OFFSET_INVALID) in
+ *         \p offsets will be ignored, if there are no valid offsets in
```
mhowlett: haven't looked at the code yet to check actual behavior, but I think this should fail immediately if there are **any** logic offsets (and this comment updated accordingly).

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ * @param rk Producer instance.
+ * @param timeout_ms The maximum time to block. On timeout the operation
+ *                   may continue in the background, depending on state,
+ *                   and it is okay to call this function again.
+ * @param errstr A human readable error string (nul-terminated) is written to
+ *               this location that must be of at least \p errstr_size bytes.
+ *               The \p errstr is only written to if there is a fatal error.
+ * @param errstr_size Writable size in \p errstr.
+ *
+ * @remark This function will block until all outstanding messages are
```
mhowlett: nice doc

src/rdkafka_timer.h
```diff
@@ -62,14 +62,14 @@ int rd_kafka_timer_stop (rd_kafka_timers_t *rkts,
...
+                            rd_bool_t oneshot, rd_bool_t restart,
                             void (*callback) (rd_kafka_timers_t *rkts,
                                               void *arg),
                             void *arg);
 #define rd_kafka_timer_start(rkts,rtmr,interval,callback,arg) \
-        rd_kafka_timer_start0(rkts,rtmr,interval,rd_false,callback,arg)
-#define rd_kafka_timer_start_oneshot(rkts,rtmr,interval,callback,arg)   \
-        rd_kafka_timer_start0(rkts,rtmr,interval,rd_true,callback,arg)
+        rd_kafka_timer_start0(rkts,rtmr,interval,rd_false,rd_true,callback,arg)
+#define rd_kafka_timer_start_oneshot(rkts,rtmr,restart,interval,callback,arg) \
```
mhowlett: when reading librdkafka code, I keep thinking that it would be much quicker if documentation was in the header files with the declarations + #defines were also documented. this is a very specific example of this where this would have saved a time.

i'm not sure i like how `rd_kafka_timer_start0` is taking on so much capability, though I get there is efficiency justification for it. my 50ms instinct was why do you have two booleans for what should be a mutually exclusive thing (restart vs oneshot)? of course with a bit more reading, I get what's going on, but i don't want to have to do the bit-more-reading. note: following on from the first comment, if the #defines had have been documented, I wouldn't have even got to the point of bothering to consider `rd_kafka_timer_start0` enough to make this second comment.

anyway, all good, no changes necessary, just telling you what I'm thinking in case it's useful.

src/rdkafka_broker.c
```diff
@@ -5929,10 +5999,120 @@ rd_kafka_broker_persistent_connection_del (rd_kafka_broker_t *rkb,
...
+ *          of op queues.
+ *
+ * @locks none
+ * @locality any
+ */
+void rd_kafka_broker_monitor_del (rd_kafka_broker_monitor_t *rkbmon) {
+        rd_kafka_broker_t *rkb = rkbmon->rkbmon_rkb;
+
+        if (!rkb)
+                return;
```
mhowlett: i can't easily work out if you want `assert(!rkb)` here instead, so will just ask the question since i'm wondering it.

src/rdkafka_request.c
```diff
@@ -3588,18 +3830,233 @@ rd_kafka_InitProducerIdRequest (rd_kafka_broker_t *rkb,
...
+        /* PID */
+        rd_kafka_buf_write_i64(rkbuf, pid.id);
+        rd_kafka_buf_write_i16(rkbuf, pid.epoch);
+
+        /* Group Id */
+        rd_kafka_buf_write_str(rkbuf, group_id, -1);
+
+        rd_kafka_buf_ApiVersion_set(rkbuf, ApiVersion, 0);
+
+        rkbuf->rkbuf_max_retries = 3;
```
mhowlett: maybe consider putting a `RD_KAFKA_BUF_TXN_RETRIES` constant or something in rdkafka_buf.h for this?

src/rdkafka_txnmgr.c
```diff
@@ -0,0 +1,2255 @@
...
+                rd_kafka_op_destroy(rko);
+                return RD_KAFKA_RESP_ERR__NO_OFFSET;
+        }
+
+        rd_kafka_topic_partition_list_log(rk, "TXNOFFSET", RD_KAFKA_DBG_EOS,
+                                          rko->rko_u.txn.offsets);
+
+        rd_kafka_buf_ApiVersion_set(rkbuf, ApiVersion, 0);
+
+        rkbuf->rkbuf_max_retries = 3;
```
mhowlett: 3 is the magic number

src/rdkafka_txnmgr.c
```diff
@@ -0,0 +1,2255 @@
...
+        }
+
+        va_start(ap, fmt);
+        vsnprintf(errstr, sizeof(errstr), fmt, ap);
+        va_end(ap);
+
+        rd_kafka_wrlock(rk);
+        if (rk->rk_eos.txn_err) {
+                rd_kafka_dbg(rk, EOS, "TXNERR",
+                             "Ignoring sub-sequent abortable transaction "
```
mhowlett: `sub-sequent` -> `subsequent`

src/rdkafka.h
edenhill: This is intentional, the ASCII null byte symbol name is nul, not null. And it helps disambiguating between NULL pointers and NUL bytes.

src/rdkafka.h
edenhill: I'll be a proud bearer of that title,

Bon voyage,
Disagreeable Frenchman

src/rdkafka.h
edenhill: I hear you, but we have existing offset interfaces that take the actual offset (commit, store_offsets), without doing +1, so we need to careful with being consistent in this regard.

Best thing would be if users did not have to care about individual partitions and offsets, but instead, as you propose, provide a helper method taking a consumer.

In fact, we really need three things:
 1. set start-of-transaction snapshot/restore point
 2. commit/send-offsets for transaction
 3. rewind to snapshot/restore point on transaction abort


Which would allow you to do something along the lines of:
```python

while the_truth:
    producer.begin_transaction()
    consumer.set_restorepoint()

  try:
    while time_in_txn < max_time_per_txn:
       input = consumer.poll(remaining_time)

       if consumer.is_rebalancing():
           producer.abort_transaction()
           consumer.clear_restorepoint()
            continue outer_loop

       output = process(input)
       producer.produce(output)

    producer.commit_offsets(consumer)
    producer.commit_transaction()

  except AbortableError:
          producer.abort_transaction()
          consumer.rewind_to_restorepoint()
          continue outer_loop

```

src/rdkafka.h
edenhill: We do have it, rd_kafka_topic_partition has a metadata field.

src/rdkafka_timer.h
edenhill: The reason the doc strings are in the implementat (.c) rather than the declaration (.h) is to keep it as close to the implementation as possible, to minimize the risk of docstrings not being updated when code is changed.

As for parameters; You are right, but I'll ignore it for this case.

src/rdkafka_broker.c
edenhill: This allows multiple monitor_del() calls which is needed since monitor_del() is called from various code paths where the broker is being decommissioned.

src/rdkafka_conf.c
```diff
@@ -3391,6 +3429,21 @@ const char *rd_kafka_topic_conf_finalize (rd_kafka_type_t cltype,
                 } else {
                         tconf->queuing_strategy = RD_KAFKA_QUEUE_FIFO;
                 }
+
+                if (conf->eos.transactional_id) {
+                        /* Make sure at least one message can be produced
+                         * before the transaction times out. */
+                        if (conf->eos.transaction_timeout_ms - 1000 <=
```
rnpridgeon: Nit: conf->eos.transaction_timeout_ms  >=  tconf->message_timeout_ms + 1000 is easier to read IMO. 

> The transaction timeout must be at least 1000 ms larger than `message.timeout.ms` and `socket.timeout.ms

src/rdkafka_conf.c
```diff
@@ -3284,6 +3302,26 @@ const char *rd_kafka_conf_finalize (rd_kafka_type_t cltype,
...
+                                                "`enable.idempotence=true`";
+
+                                conf->eos.idempotence = rd_true;
+                        }
+
+                        /* Make sure at least one request can be sent
+                         * before the transaction times out. */
+                        if (conf->eos.transaction_timeout_ms <
+                            conf->socket_timeout_ms)
+                                return "`transaction.timeout.ms` must be "
```
rnpridgeon: does `transaction.timeout.ms` need to be greater than equal to `socket.timeout.ms` or greater than equal to `socket.timeout.ms` + 1000

>The transaction timeout must be at least 1000 ms larger than `message.timeout.ms` and `socket.timeout.ms`

src/rdkafka_broker.h
```diff
@@ -35,6 +35,47 @@
...
+ */
+typedef struct rd_kafka_broker_monitor_s {
+        TAILQ_ENTRY(rd_kafka_broker_monitor_s) rkbmon_link; /**< rkb_monitors*/
+        struct rd_kafka_broker_s *rkbmon_rkb; /**< Broker being monitored. */
+        rd_kafka_q_t *rkbmon_q;               /**< Queue to enqueue op on. */
+
+        /**< Callback triggered on the monitoree's op handler thread.
+         *   Do note that the callback might be triggered even after
+         *   it has been deleted due to the queueing nature of op queues. */
+        void (*rkbmon_cb) (rd_kafka_broker_t *rkb);
```
rnpridgeon: If monitors are watching for state changes does it make sense to pass in in the new state as opposed to the broker handle. The hope being you can limit the number of times the broker lock is help since presumably a monitoree will call `rd_kafka_broker_get_state` after the callback is served off its ops queue.

rnpridgeon: In retrospect there is no guarantee the state is the same by the time the op is actually served by the listener.

INTRODUCTION.md
```diff
@@ -705,15 +705,106 @@ which returns one of the following values:
...
+    rd_kafka_t *consumer = rd_kafka_new(RD_KAFKA_CONSUMER, cconf);
+    rd_kafka_poll_set_consumer(consumer);
+
+    rd_kafka_subscribe(consumer, "inputTopic");
+
+    /* Consume-Process-Produce loop */
+    while (run) {
+
+       /* Begin transaction */
+       rd_kafka_begin_transaction(producer);
```
rnpridgeon: No timeout? 

Also the example should probably include at least a check for `RD_KAFKA_RESP_ERR_NO_ERROR` prior to beginning a transaction.

INTRODUCTION.md
```diff
@@ -705,15 +705,106 @@ which returns one of the following values:
...
+following pseudo-code:
+
+```c
+    /* Producer */
+    rd_kafka_conf_t *pconf = rd_kafka_conf_new();
+    rd_kafka_conf_set(pconf, "bootstrap.servers", "mybroker");
+    rd_kafka_conf_set(pconf, "transactional.id", "my-transactional-id");
+    rd_kafka_t *producer = rd_kafka_new(RD_KAFKA_PRODUCER, pconf);
+
+    rd_kafka_init_transactions(producer);
```
rnpridgeon: Not timeout?

Also should we not verify success by checking for `RD_KAFKA_RESP_ERR_NO_ERROR` prior to calling `rd_kafka_begin_transaction`

src/rdkafka.h
rnpridgeon: It would be nice to expose `rd_kafka_is_transactional` for such cases so we can fail fast on eos consumer instantiation.

rnpridgeon: It's not terribly clear to me from the mega design doc what happens in the event of a rebalance when a transaction is still open. Do we need to abort as implied by the snipped above, or is there a way for a consumer to pick up the last offset sent to a transaction for a given partition upon assignment.

rnpridgeon: Mystery solved 

https://cwiki.apache.org/confluence/display/KAFKA/KIP-447%3A+Producer+scalability+for+exactly+once+semantics

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ *
+ * @remark The consumer must disable auto commits
+ *         (set \c enable.auto.commit to false on the consumer).
+ *
+ * @remark Logical and invalid offsets (such as RD_KAFKA_OFFSET_INVALID) in
+ *         \p offsets will be ignored, if there are no valid offsets in
+ *         \p offsets the function will return RD_KAFKA_RESP_ERR_NO_ERROR
+ *         and no action will be taken.
+ *
+ * @remark This function will block until the offsets are registered on the
```
rnpridgeon: @remark It's unsafe to send an `OffsetFetchRequest` for the consumer group prior to calling `rd_kafka_init_transactions`. See KIP-447 for additional details.

INTRODUCTION.md
```diff
@@ -705,15 +705,106 @@ which returns one of the following values:
...
+
+
+    /* Consumer */
+    rd_kafka_conf_t *cconf = rd_kafka_conf_new();
+    rd_kafka_conf_set(cconf, "bootstrap.servers", "mybroker");
+    rd_kafka_conf_set(cconf, "group.id", "my-group-id");
+    rd_kafka_conf_set(cconf, "enable.auto.commit", "false");
+    rd_kafka_t *consumer = rd_kafka_new(RD_KAFKA_CONSUMER, cconf);
```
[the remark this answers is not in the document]
edenhill: That's the default

INTRODUCTION.md
edenhill: See the note below, this pseudo code is intentionally left minimal to show-case the program structure.

src/rdkafka.h
edenhill: > It would be nice to expose rd_kafka_is_transactional for such cases so we can fail fast on eos consumer instantiation.

Can you elaborate on this? Where and how would it be used?

src/rdkafka.h
edenhill: The idea here is that you should be able to pass the current position() directly to this function, and some partitions may not have been consumed at all (due to emptyness, no messages, paused, whatever) and thus may have a logical current position.

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ *          RD_KAFKA_RESP_ERR__NOT_CONFIGURED if transactions have not been
+ *          configured for the producer instance,
+ *          RD_KAFKA_RESP_ERR__INVALID_ARG if \p rk is not a producer instance,
+ *          or if the \p consumer_group_id or \p offsets are empty.
+ *
+ * FIXME: add more broker-induced error codes
+ */
+RD_EXPORT
+rd_kafka_resp_err_t
+rd_kafka_send_offsets_to_transaction (
```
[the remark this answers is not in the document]
edenhill: This follows the Java API.
Relevant discussion: https://github.com/edenhill/librdkafka/pull/2605/files#r359907215

src/rdkafka_conf.c
edenhill: Good catch :+1:

To keep things simple for the user we probably want to make transaction.timeout.ms the defining property, automatically adjusting socket.timeout.ms (unless explicitly configured < transaction.timeout.ms) and message.timeout.ms (set to transaction.timeout.ms).

src/rdkafka.h
edenhill: Thank you, we'll handle KIP-447 in the second iteration of transactions.

src/rdkafka.h
```diff
@@ -6240,6 +6246,351 @@ rd_kafka_oauthbearer_set_token_failure (rd_kafka_t *rk, const char *errstr);
...
+ *        committing the transaction.
+ *
+ *        If any of the outstanding messages fail permanently the current
+ *        transactional will enter the abortable error state and this
+ *        function will return FIXME, in this case the application
+ *        must call rd_kafka_abort_transaction() before attempting a new
+ *        transaction with rd_kafka_begin_transaction().
+ *
+ * @param rk Producer instance.
+ * @param timeout_ms The maximum time to block. On timeout the operation
```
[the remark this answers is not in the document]
edenhill: I think it makes sense to keep here with the param documentation.

src/rdkafka_request.c
edenhill: If it was consistent across txn protocol requests, yes, but it is not.

src/rdkafka_timer.h
edenhill: Restart and oneshot are not mutually exclusive, restart dictates if a timer should be restarted if already started, while oneshot dictates if the timer should be automatically restarted after it fires (i.e., a recurring timer)

tests/Makefile
```diff
@@ -68,6 +72,13 @@ idempotent_seq: $(BIN)
 
 idempotent: idempotent_par
 
+transactions: $(BIN)
+	for _test in 0098 0101; do TESTS=$$_test ./run-test.sh ./$(BIN) ; done
```
edenhill: The run-test syntax changed, and the list of tests is incomplete. This is a rather pointless target.

alex88:
🎉 glad to see this merged! Thanks for the awesome work!

PR #2613, opened 2019-11-08
Print compression type per message-set when debug=msg

edenhill:
This is handy

### DIFF
--- a/src/rdkafka_msgset_writer.c
+++ b/src/rdkafka_msgset_writer.c
@@ -1363,8 +1363,10 @@ rd_kafka_msgset_writer_finalize (rd_kafka_msgset_writer_t *msetw,
         msetw->msetw_rkbuf->rkbuf_u.Produce.batch.pid = msetw->msetw_pid;
 
         /* Compress the message set */
-        if (msetw->msetw_compression)
-                rd_kafka_msgset_writer_compress(msetw, &len);
+        if (msetw->msetw_compression) {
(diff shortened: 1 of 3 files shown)


src/rdkafka_msgset_writer.c
```diff
@@ -1363,8 +1363,10 @@ rd_kafka_msgset_writer_finalize (rd_kafka_msgset_writer_t *msetw,
         msetw->msetw_rkbuf->rkbuf_u.Produce.batch.pid = msetw->msetw_pid;
 
         /* Compress the message set */
-        if (msetw->msetw_compression)
-                rd_kafka_msgset_writer_compress(msetw, &len);
+        if (msetw->msetw_compression) {
+                if (rd_kafka_msgset_writer_compress(msetw, &len) == -1)
+                        msetw->msetw_compression = 0;
```
mhowlett: i think it'd be worth having "none because of error" as a different type to "none", but optional.

edenhill: I thought about that too but left it out for now to keep the change minimal.
The information is implicitly there though; if a compression codec is configured but the compression log says "uncompressed" it means compression did not work (errors are debug-logged) or produced larger results than the original.

mhowlett: sounds good. the other thought i had here btw is the meaning of `msetw_compression` changes (i.e. it's not immutable) and immutable things make things easier to follow (this comment said without considering/looking at broader context at all).

PR #2619, opened 2019-11-11
Mock Cluster

edenhill:


edenhill:
This was integrated into mhowlett/fff

Issue #2607, opened 2019-11-05
RdKafka::OAuthBearerTokenRefreshCb interface needs an RdKafka::Handle to call oauthbearer_set_token()


chrisbeard:
Description
===========
According to the [`RdKafka::OAuthBearerTokenRefreshCb`](https://github.com/edenhill/librdkafka/blob/v1.2.1/src-cpp/rdkafkacpp.h#L548-L583) docs, a class implementing this interface should call [`RdKafka::oauthbearer_set_token()`](https://github.com/edenhill/librdkafka/blob/v1.2.1/src-cpp/rdkafkacpp.h#L1519-L1562), in the callback.

Since `oauthbearer_set_token()` is a member function of `RdKafka::Handle`, the `RdKafka::OAuthBearerTokenRefreshCb` implementation needs access to the `RdKafka::Handle` which triggered the callback. However, this `RdKafka::Handle` object can't exist when the `OAuthBearerTokenRefreshCb` object is used to configure `RdKafka::Conf`, since the Producer/Consumer is created from this `RdKafka::Conf`, and no `RdKafka::Handle` is provided as an argument to `oauthbearer_token_refresh_cb`.

This leads to code like the following:

```c++
RdKafka::Conf *conf = RdKafka::Conf::create(RdKafka::Conf::CONF_GLOBAL);
std::string errstr;
...
conf->set("sasl.mechanisms", "OAUTHBEARER", errstr);

// Provide our OAuthBearerTokenRefreshCb implementation. We can't bind the RdKafka::Handle* to our callback
// because the RdKafka::Handle which will trigger the token refresh has not been created yet.
MyOAuthBearerTokenRefreshCb cb;
conf->set("oauthbearer_token_refresh_cb", &cb, errstr);

// Create the RdKafka::Handle which will be using OAUTHBEARER
RdKafka::Producer *producer = RdKafka::Producer::create(conf, errstr);

// Store the RdKafka::Handle* which will be used in our implementation of oauthbearer_token_refresh_cb() to call oauthbearer_set_token().
cb.setHandle(producer);

// Trigger the token refresh
producer->poll(1000);
```

This seems really clunky, especially if a task is creating multiple producers and consumers.

In comparison, the [C API](https://github.com/edenhill/librdkafka/blob/v1.2.1/src/rdkafka.h#L1816-L1856) provides a `rd_kafka_t` as a callback parameter.

[...]

PR #2620, opened 2019-11-11
Provide RdKafka::Handle* to OAuthBearerTokenRefreshCb callback (#2607)

chrisbeard:
This change provides  `RdKafka::OAuthBearerTokenRefreshCb::oauthbearer_token_refresh_cb()` with an `RdKafka::Handle*`. This gives the callback the proper `RdKafka::Handle` to call `oauthbearer_set_token()`/`oauthbearer_set_token_failure()`.

I've also updated the docs to more explicitly refer to the `RdKafka::Handle` methods.

Addresses issue: #2607

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -707,6 +707,10 @@ class RD_EXPORT OAuthBearerTokenRefreshCb {
  public:
   /**
    * @brief SASL/OAUTHBEARER token refresh callback class.
+   *
+   * @param handle The RdKafka::Handle which requires a refreshed token.
+   * @param oauthbearer_config The value of the
+   * \p sasl.oauthbearer.config configuration property for \p handle.
    */
   virtual void oauthbearer_token_refresh_cb (RdKafka::Handle* handle,
                                              const std::string &oauthbearer_config) = 0;
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -707,9 +707,23 @@ class RD_EXPORT OAuthBearerTokenRefreshCb {
  public:
   /**
    * @brief SASL/OAUTHBEARER token refresh callback class.
+   *
+   * @deprecated Use the overloaded version of this function instead. This
+   * version will be unused, even though an implementation is required.
+   */
+  virtual void oauthbearer_token_refresh_cb (const std::string &oauthbearer_config) = 0;
+
+  /**
+   * @brief SASL/OAUTHBEARER token refresh callback class.
+   *
+   * @param handle The RdKafka::Handle which requires a refreshed token.
+   * @param oauthbearer_config The value of the
+   * \p sasl.oauthbearer.config configuration property for \p handle.
    */
   virtual void oauthbearer_token_refresh_cb (RdKafka::Handle* handle,
-                                             const std::string &oauthbearer_config) = 0;
+                                             const std::string &oauthbearer_config) {
+    oauthbearer_token_refresh_cb(oauthbearer_config);
+  }
 
   virtual ~OAuthBearerTokenRefreshCb() { }
 };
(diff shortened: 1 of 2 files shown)


src-cpp/rdkafkacpp.h
```diff
@@ -577,7 +578,8 @@ class RD_EXPORT OAuthBearerTokenRefreshCb {
   /**
    * @brief SASL/OAUTHBEARER token refresh callback class.
    */
-  virtual void oauthbearer_token_refresh_cb (const std::string &oauthbearer_config) = 0;
+  virtual void oauthbearer_token_refresh_cb (RdKafka::Handle* handle,
```
edenhill: This is a breaking change that is not allowed by our API compatibility guarantees.
The old method needs to stay in place (but can be deprecated so we can remove it in v2).

chrisbeard: Ah, right...

How about something like this:
* `virtual void oauthbearer_token_refresh_cb(const std::string&) = 0;` remains intact
* Add `virtual void oauthbearer_token_refresh_cb(RdKafka::Handle*, const std::string&);` with a default implementation that simply forwards to `oauthbearer_token_refresh_cb(const std::string&)`

This would allow for an implementation of this interface the option of using the newer API, while requiring no changes from existing implementations. Because the original API is pure virtual, you'd still have to supply an implementation for `oauthbearer_token_refresh_cb(const std::string &)` until it is removed, though it can be empty if using the newer API.

Thoughts?

rondagostino: That makes sense to me; it would become `virtual void oauthbearer_token_refresh_cb(RdKafka::Handle*, const std::string&) = 0;` when the deprecated method is removed (i.e. in v2).

chrisbeard:
@edenhill It doesn't looks like the travis-ci failure is related as far as I can tell, is this okay?

src-cpp/rdkafkacpp.h
edenhill: Reiterating on this I'm leaning towards actually breaking the API just this once since the proposed fix here will break existing applications that make use of this interface anyway, right?

And I don't think many people are using the C++ OAUTHBEARER interface.

Thoughts?

chrisbeard: The changes I proposed on Nov 12th would allow applications currently using this interface to remain unchanged with the additional method. Once the deprecated method is removed in favor of the additional one, they'll break.

I don't have strong opinions on whether it's best to "fix" it now or later though, I'm okay with either.

chrisbeard: After a week or so, I'm more and more of the opinion that this is a reasonable breaking change to make if you're okay with it.

edenhill: @rondagostino What is your take on this?
We want to break the current C++ API, which is a semver sin, but there are two limiting factors:
 1. very few people, if any, are using the C++ OAUTHBEARER interface.
 2. We don't provide any ABI guarantees for the C++ library, so applications need to be recompiled anyway on upgrade, and this is a small fix for the developer to make.

rondagostino: @edenhill I think that is an accurate summary of the situation.  I'm disappointed that this problem exists -- I wish I had realized it back when I did this. :-(

There really isn't a great solution other than bumping the major version and fixing it/breaking existing stuff, but if that isn't possible right now then I say just go for it and break things now without bumping the major version as opposed to jumping through the hoop imposed by semantic versioning making a messy situation messier.

edenhill: Thanks @rondagostino .

@chrisbeard Do you want to update the PR with the outcome of this discussion, namely that we'll break API backwards compat and change the existing method signature?

chrisbeard: Hey @edenhill, finally got back to this. I've reverted back to the non backwards compatible interface that passes the `RdKafka::Handle*` to the callback.

Let me know if there's anything that needs cleaning up. 0105_transactions_mock seems to be failing, but seems (hopefully) unrelated?

edenhill:
Thank you, @chrisbeard !

PR #2635, opened 2019-11-25
Static consumer fencing and fatal consumer errors

edenhill:
This PR introduces fatal errors on the consumer (previously only used on the idempotent producer), for use when a static group member is fenced off by a newer instance.

There's also a stability fix for the KIP-345 test suite
The lower timeout for when a group rebalance will occur can't be known from the test since it relies on the time of the latest SyncGroup or Heartbeat request.
This PR allows a larger low end timespan to accomodate for this uncertainty.

Also did some minor refactoring to help catch the issue, which I think might be useful to keep.


src-cpp/rdkafkacpp.h
```diff
@@ -463,6 +463,9 @@ enum ErrorCode {
         ERR_PREFERRED_LEADER_NOT_AVAILABLE = 80,
         /** Consumer group has reached maximum size */
         ERR_GROUP_MAX_SIZE_REACHED = 81,
+        /** Static consumer fenced by other consumer with same
+         * group.instance.id. */
+        ERR_FENCED_INSTANCE_ID = 82,
```
mhowlett: this is in KIP-345, did our initial implementation take a short-cut?

src/rdkafka_cgrp.c
```diff
@@ -2568,6 +2568,10 @@ void rd_kafka_cgrp_handle_heartbeat_error (rd_kafka_cgrp_t *rkcg,
                 reason = "group is rebalancing";
                 break;
 
+        case RD_KAFKA_RESP_ERR_FENCED_INSTANCE_ID:
+                reason = "consumer fenced by newer instance: rebalancing";
```
mhowlett: `FENCED_INSTANCE_ID` is also evidently a potential error code for JoinGroup, SyncGroup, OffsetCommit and LeaveGroup - do you want to explicitly consider it in those as well? If nothing else, for documentation purposes (I haven't checked for correctness, but assume behavior is correct without).

mhowlett: is the word `: rebalancing` is redundant?

src-cpp/rdkafkacpp.h
edenhill: I'm not sure, will need to ask @rnpridgeon when he's back.

src/rdkafka_cgrp.c
```diff
@@ -1086,14 +1090,20 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
                                           "JoinGroup failed: %s",
                                           rd_kafka_err2str(ErrorCode));
 
-                if (ErrorCode == RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID)
+                if (ErrorCode == RD_KAFKA_RESP_ERR_FENCED_INSTANCE_ID)
+                        rd_kafka_set_fatal_error(rkcg->rkcg_rk, ErrorCode,
```
mhowlett: it might be good if every time you call poll on a client in a failed state, you got a fatal error. is that how it works? I looked quickly and it didn't seem like it, but i might have missed it.

edenhill:
@rnpridgeon @mhowlett Please give this PR a complete re-review since it now also contains fenced error handling.

INTRODUCTION.md
```diff
@@ -1288,6 +1292,26 @@ The latest stored offset will be automatically committed every
 Broker based consumer groups (requires Apache Kafka broker >=0.9) are supported,
 see KafkaConsumer in rdkafka.h or rdkafkacpp.h
 
+##### Static consumer groups
+
+By default Kafka consumers are rebalanced each time a new consumer joins
+the group or an existing member leaves. This is what is known as a dynamic
+membership. Apache Kafka >= 2.3.0 introduces static membership.
+Unlike dynamic membership, static members can leave and rejoin a group
+within the `session.timeout.ms` without triggering a rebalance retaining
```
mhowlett: comma after rebalance.

INTRODUCTION.md
```diff
@@ -1288,6 +1292,26 @@ The latest stored offset will be automatically committed every
...
+their existing partitions assignment.
```
mhowlett: I would add something like: "Consumers with `group.instance.id` set will not send a leave group request on close - session timeout is the only mechanism that will trigger the group rebalance in this case" (as a n00b, this wasn't at all clear to me without digging deep into the kip).

INTRODUCTION.md
```diff
@@ -1348,9 +1372,7 @@ is returned.
 
 ### Fatal errors
 
-The added guarantee of ordering and no duplicates also requires a way for
-the client to fail gracefully when these guarantees can't be satisfied.
-If an unresolvable error occurs a fatal error is triggered in one
+If an unrecoverable error occurs a fatal error is triggered in one
```
mhowlett: comma after occurs

INTRODUCTION.md
```diff
@@ -1387,6 +1420,27 @@ The purged messages in queue will fail with error code set to
 `RD_KAFKA_RESP_ERR__PURGE_QUEUE`.
 
 
+#### Fatal consumer errors
+
+A consumer configured for static group membership (`group.instance.id`) may
+raise a fatal error if a new consumer instance is started with the same
```
mhowlett: clearer if you change 'a new' to 'another'

src/rdkafka_cgrp.c
edenhill: I dont think that's desirable since the poll call is also used to trigger things like logs, stats, etc, which the application might want to handle before shutting down.
What it will do though is enqueue a single ERR__FATAL event on the poll queue, which should be sufficient for the application to add fatal error handling, if so desired.

INTRODUCTION.md
edenhill: I think new is more clear, since that is what it technically is, a newer instance of the app.
 keeping it.

rnpridgeon: Wouldn't the new consumer be the one that's fenced rather than the existing one?

> The edge case is that if we could have members with the same `group.instance.id` (for example mis-configured instances with a valid `member.id` but added a used `group.instance.id` on runtime). When `group.instance.id` has duplicates, we could refuse join request from members with an outdated `member.id`, since we update the mapping upon each join group request. In an edge case where the client hits this exception in the response, it is suggesting that some other consumer takes its spot. The client should immediately fail itself to inform end user that there is a configuration bug which is generating duplicate consumers with same identity.

edenhill: The way I read that is the old consumer is fenced off, which is what makes sense, you want to terminate old stale instances when new are created.

rnpridgeon: Yeah you are right.

PR #2636, opened 2019-11-26
Add getaddrinfo callback

bdaheb:
Here is a proposal for a callback to getaddrinfo(). This is extension to librdkafka existing socket() and connect() callbacks. 

This callback is used to control the whole connection process starting from getaddrinfo_cb then socket_cb and finally connect_cb. 
  
In details, it provides rd_kafka_conf_set_getaddrinfo_cb() to add the getaddrinfo_cb callback. 

Moreover, the typical usage is for a router that has 2 wan interfaces to reach the kafka broker. Each wan interface is in a different network namespace and can be up or down.
Since, the kafka client is launched in the root namespace, it needs to be called (via getaddrinfo_cb()) in order to resolve the broker name in a network namespace where the wan interface is up. 
Otherwise, if the kafka client starts in a network namespace where the wan interface is down, it gets stuck there retrying getaddrinfo() indefinitely without success.

### DIFF
--- a/src/rdaddr.c
+++ b/src/rdaddr.c
@@ -143,7 +143,10 @@ const char *rd_addrinfo_prepare (const char *nodesvc,
 
 
 
-rd_sockaddr_list_t *rd_getaddrinfo (const char *nodesvc, const char *defsvc,
+rd_sockaddr_list_t *rd_getaddrinfo (int(*getaddrinfo_cb)(const char *node, const char *service,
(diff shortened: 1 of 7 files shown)


edenhill:
Thanks for your contribution.

While I understand the use of this it still a very niche use-case and the changes exposes the netdb interface in rdkafka.h, which will have portability concerns, so I'll close this PR until there's bigger demand for this feature.

Thank you.

Issue #556, opened 2016-02-19
configure with "--enable-{sasl,ssl}" should fail when sasl/ssl isn't available


Whissi:
Commit 8d141ef30a18df3d953a052e3fd1b1a47298feb3 and 92ec079c58bff7928ec5e3c4684bb487478a4173 added support for disabling sasl/ssl, however if you have enabled sasl but sasl wasn't found configure doesn't fail.

I expect the same for ssl but haven't verified.

edenhill:
--enable-sasl is the default and it was actually added to avoid installing sasl/ssl by specifying --disable-xxx

But I hear you, it would need to be a tri-state:  defaulty, y, n, and fail if 'y' && notfound.

edenhill:
added trivup issue

PR #2640, opened 2019-12-02
Fail configuration when requested library isn't present 

benesch:
Two commits in here. Messages reproduced below. The tl;dr is this fixes edenhill/mklove#21 (which was spawned from #556). It seems like the change to configure.base should be incorporated into mklove, but it seems like more mklove development happens in this repository than there?

----

**Fail configuration when requested library isn't present**

If the user specifies --enable-XXX, fail configuration if XXX isn't
present. The prior behavior required the user to manually inspect the
configure output to determine whether the feature was actually enabled.
This was particularly problematic with downstream build systems, like
Cargo with rust-rdkafka, where the configure output is hidden from users
if configuration is successful.

The default behavior remains unchanged; if a user does not explicitly
request a feature with --enable-XXX nor dis-request it with
--disable-XXX, and the feature is enabled by default, then it is not be
an error if the feature cannot be found.

Note that this commit additionally excludes the `ENABLE_*` options from
the confugration summary, as they were misleading. For example, if a
user passed --enable-zstd, but zstd was not found on the system, the
configuration summary would report "ENABLE_ZSTD y"--merely reporting
that the user had requested zstd support, not that zstd support would
actually be available in the built library. (If the user does desire
that information, they can find it in the BUILT_WITH line, which
includes only the features that were actually possible to enable.)

---

**Allow disabling zlib when building with mklove**

For symmetry with the CMake build, which allows disabling zlib
explicitly.

### DIFF
--- a/configure.self
+++ b/configure.self
@@ -39,7 +39,7 @@ mkl_toggle_option "Feature" ENABLE_LZ4_EXT "--enable-lz4" "Deprecated: alias for
 # librdkafka with TSAN won't work with glibc C11 threads on Ubuntu 19.04.
 # This option allows disabling libc-based C11 threads and instead
 # use the builtin tinycthread alternative.
-mkl_toggle_option "Feature" ENABLE_C11THREADS "--enable-c11threads" "Enable detection of C11 threads support in libc" "y"
+mkl_toggle_option "Feature" ENABLE_C11THREADS "--enable-c11threads" "Enable detection of C11 threads support in libc" "try"
 
 mkl_toggle_option "Feature" ENABLE_SYSLOG "--enable-syslog" "Enable logging to syslog" "y"
 
@@ -53,10 +53,15 @@ function checks {
     mkl_lib_check "libpthread" "" fail CC "-lpthread" \
                   "#include <pthread.h>"
 
-    if [[ $ENABLE_C11THREADS == "y" ]]; then
+    if [[ $ENABLE_C11THREADS != n ]]; then
+        case "$ENABLE_C11THREADS" in
+            y) local action=fail ;;
+            try) local action=disable ;;
+            *) mkl_err "mklove internal error: invalid value for ENABLE_C11THREADS: $ENABLE_C11THREADS"; exit 1 ;;
+        esac
         # Use internal tinycthread if C11 threads not available.
         # Requires -lpthread on glibc c11 threads, thus the use of $LIBS.
-        mkl_lib_check "c11threads" WITH_C11THREADS disable CC "$LIBS" \
+        mkl_lib_check "c11threads" WITH_C11THREADS $action CC "$LIBS" \
                       "
 #include <threads.h>
 
@@ -94,10 +99,10 @@ void foo (void) {
     fi
 
     # optional libs
-    mkl_check "zlib" disable
-    mkl_check "libssl" disable
-    mkl_check "libsasl2" disable
-    mkl_check "libzstd" disable
+    mkl_check "zlib"
+    mkl_check "libssl"
+    mkl_check "libsasl2"
+    mkl_check "libzstd"
 
     if mkl_lib_check "libm" "" disable CC "-lm" \
                      "#include <math.h>"; then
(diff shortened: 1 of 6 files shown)


edenhill:
Sorry for the long delay, can you rebase this on latest master?

benesch:
No problem, done!

configure.self
```diff
@@ -53,10 +53,14 @@ function checks {
     mkl_lib_check "libpthread" "" fail CC "-lpthread" \
                   "#include <pthread.h>"
 
-    if [[ $ENABLE_C11THREADS == "y" ]]; then
+    if [[ $ENABLE_C11THREADS != n ]]; then
+        case "$ENABLE_C11THREADS" in
```
edenhill, line 57: Maybe make a helper function, `mkl_ynd "some-string"` that returns `y`, `n` or `d` depending on the string.
E.g.:
```bash
local action=$(mkl_ynd "$ENABLE_C11THREADS")
```

mklove/modules/configure.base
```diff
@@ -912,7 +912,7 @@ Configuration summary:"
     local n=
     for n in $MKL_MKVARS ; do
         # Skip the boring booleans
-        if [[ $n == WITH_* || $n == WITHOUT_* || $n == HAVE_* || $n == def_* ]]; then
+        if [[ $n == ENABLE_* || $n == WITH_* || $n == WITHOUT_* || $n == HAVE_* || $n == def_* ]]; then
```
edenhill, line 915: I've been meaning to do this for years!

configure.self
```diff
@@ -53,10 +53,14 @@ function checks {
...
+            y) local action=fail ;;
+            *) local action=disable ;;
```
edenhill, line 59: This has a risk of swallowing typos, e.g., Y or 1, what do you think about using 'd' as the default (which also coincides with disable, which is the default action)?
And setting the default to 'd' in mkl_check if not provided.

benesch:
Reworking this to avoid the potential for typos makes a lot of sense! I wasn't quite sure how to apply your suggestions exactly as written, so a took a stab at capturing their spirit in the way that made sense to me. Toggle options that should be a tristate now set their default value to `"d"`. Then the decision on what to do has four branches:

  * `n` -> don't perform the check, leave the library disabled
  * `y` -> fail the configure if the library is missing
  * `d` -> disable the library and continue if the library is missing
  * `*` -> internal mklove error, explode

I wasn't sure how to wrap that logic up in a helper function, because handling the `n` case requires returning from the function early.

By the way, if it would be easier for you, feel free to just take over this PR and adapt it to your liking!

mklove/modules/configure.libsasl2
```diff
@@ -7,23 +7,26 @@
 #
 #
 # And then call the following function from the correct place/order in checks:
-#   mkl_check libsasl2 [<action>]
+#   mkl_check libsasl2
 #
 
-mkl_toggle_option "Feature" ENABLE_GSSAPI "--enable-gssapi" "Enable SASL GSSAPI support with Cyrus libsasl2" "y"
+mkl_toggle_option "Feature" ENABLE_GSSAPI "--enable-gssapi" "Enable SASL GSSAPI support with Cyrus libsasl2" "d"
```
edenhill, line 13: I'm guessing this means the --help output will show "(d)" as the default, which doesn't mean much to the user.
We should probably have --help show "try", "if available", or somesuch as the default string for "d".

mklove/modules/configure.libzstd
```diff
@@ -6,15 +6,18 @@
 #   mkl_require libzstd
 #
 # And then call the following function from the correct place/order in checks:
-#   mkl_check libzstd [<action>]
+#   mkl_check libzstd
 #
 
-mkl_toggle_option "Feature" ENABLE_ZSTD "--enable-zstd" "Enable support for ZSTD compression" "y"
+mkl_toggle_option "Feature" ENABLE_ZSTD "--enable-zstd" "Enable support for ZSTD compression" "d"
```
edenhill, line 12: Also need to make sure this still works with automatic dependency installation, e.g, `./configure --install-deps` and `./configure --install-deps --source-deps-only`.

mklove/modules/configure.libsasl2
benesch: Ah, sure. I didn't see an obvious way to plumb that through `mkl_toggle_option` and `mkl_option`, so... instead I just changed the default from "d" to "try". That seems to have the desired effect and doesn't require any additional code changes. What do you think?

mklove/modules/configure.libzstd
benesch: Yep, `--install-deps` and `--install-deps` --source-deps-only` both are unaffected by this change!

edenhill:
This is great stuff, thank you!

PR #2452, opened 2019-08-02
allow compilation with a non-standard cyrus-sasl

yhager:
This should fix #2451. Notice that I updated the minimum cmake requirement, so I can use the `IMPORTED_TARGET` functionality. I found that to be the easiest solution, and cmake 3.6 should be widely available.

PR #2641, opened 2019-12-02
Support finding headers in nonstandard directories in CMake build

### LINKED ISSUE #2451: Does not compile with non-standard instal of cyrus-sasl

Description
===========
I have built cyrus-sasl and installed it in a non-standard directory. Building with `-DCMAKE_PREFIX_PATH=<cyrus-sasl install dir>` works for the most part.

But it is missing the `-I` flag for the non-standard directory, which leads to compilation failure:
```
/usr/bin/ccache /usr/lib/ccache/bin/gcc -DLIBRDKAFKA_STATICLIB -I/home/yuval/work/js/sonarw.1/build/rdkafka-prefix/src/rdkafka/src -Igenerated/dummy -isystem /home/yuval/work/js/sonarw.1/build/frontend/openssl-prefix/include -O2 -g -DNDEBUG -fPIC -MD -MT src/CMakeFiles/rdkafka.dir/rdkafka_sasl_cyrus.c.o -MF src/CMakeFiles/rdkafka.dir/rdkafka_sasl_cyrus.c.o.d -o src/CMakeFiles/rdkafka.dir/rdkafka_sasl_cyrus.c.o   -c /home/yuval/work/js/sonarw.1/build/rdkafka-prefix/src/rdkafka/src/rdkafka_sasl_cyrus.c
/home/yuval/work/js/sonarw.1/build/rdkafka-prefix/src/rdkafka/src/rdkafka_sasl_cyrus.c:46:10: fatal error: sasl/sasl.h: No such file or directory
   46 | #include <sasl/sasl.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
```

How to reproduce
================
- remove your sasl package, or simply rename away `/usr/include/sasl/sasl.h`
- build and install cyrus-sasl in a non-standard directory:
  - download and extract https://github.com/cyrusimap/cyrus-sasl/releases/download/cyrus-sasl-2.1.27/cyrus-sasl-2.1.27.tar.gz
  - build using `./configure --prefix /tmp/cyrus-sasl && make && make install`
- try to build rdkafka v1.1.0 with `cmake -DCMAKE_PREFIX_PATH=/tmp/cyrus-sasl`. You will get the above error

The following patch fixes the issue for me:
```
--- rdkafka.orig/src/CMakeLists.txt	2019-08-02 11:40:15.348245463 -0700
+++ rdkafka/src/CMakeLists.txt	2019-08-02 11:50:24.937713227 -0700
@@ -201,6 +201,7 @@
 target_link_libraries(rdkafka PUBLIC Threads::Threads)
 
 if(WITH_SASL_CYRUS)
+  target_include_directories(rdkafka PUBLIC ${SASL_INCLUDE_DIRS})
   target_link_libraries(rdkafka PUBLIC ${SASL_LIBRARIES})
 endif()
```

Checklist
=========

 - [x] librdkafka version (release number or git tag): `v1.1.0`
 - [x] Apache Kafka version: `N/A`
 - [x] librdkafka client configuration: `N/A`
 - [x] Operating system: `redhat enterprise linux 7, archlinux`
 - [ ] Critical issue

benesch:
CMake may discover libraries, like zlib/openssl/libsasl2 in nonstandard
directories, e.g., if the user has specified the path to a custom build
of one of these libraries via CMAKE_PREFIX_PATH. In these cases, both the
library search paths and the header search paths must be updated, but
the CMake build system was forgetting to update the header search paths
for all libraries besides zstd.

Fix #2451.
Supersedes #2452.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -172,6 +172,7 @@ endif()
 
 if(WITH_ZLIB)
   find_package(ZLIB REQUIRED)
+  target_include_directories(rdkafka PUBLIC ${ZLIB_INCLUDE_DIRS})
   target_link_libraries(rdkafka PUBLIC ZLIB::ZLIB)
 endif()
 
@@ -191,6 +192,7 @@ if(WITH_SSL)
     add_dependencies(rdkafka bundled-ssl)
   else()
     find_package(OpenSSL REQUIRED)
+    target_include_directories(rdkafka PUBLIC ${OPENSSL_INCLUDE_DIR})
     target_link_libraries(rdkafka PUBLIC OpenSSL::SSL OpenSSL::Crypto)
   endif()
 endif()
@@ -203,6 +205,7 @@ find_package(Threads REQUIRED)
 target_link_libraries(rdkafka PUBLIC Threads::Threads)
 
 if(WITH_SASL_CYRUS)
+  target_include_directories(rdkafka PUBLIC ${SASL_INCLUDE_DIRS})
   target_link_libraries(rdkafka PUBLIC ${SASL_LIBRARIES})
 endif()
 
@@ -211,6 +214,7 @@ if(WITH_LIBDL)
 endif()
 
 if(WITH_LZ4_EXT)
+  target_include_directories(rdkafka PUBLIC ${LZ4_INCLUDE_DIRS})
   target_link_libraries(rdkafka PUBLIC LZ4::LZ4)
 endif()
 


src/CMakeLists.txt
```diff
@@ -172,6 +172,7 @@ endif()
 
 if(WITH_ZLIB)
   find_package(ZLIB REQUIRED)
+  target_include_directories(rdkafka PUBLIC ${ZLIB_INCLUDE_DIRS})
```
edenhill: This is called .._DIRS while OPENSSL.. has .._DIR.
Are these standard names for these libraries? If not we should probably use the same form, _DIR or _DIRS.

benesch: The standard name is `...DIRS`, but the OpenSSL find module is apparently inconsistent, and just uses `_DIR`. Compare:

  * https://cmake.org/cmake/help/latest/module/FindOpenSSL.html
  * https://cmake.org/cmake/help/latest/module/FindZLIB.html

AFAICT there just is literally no `OPENSSL_INCLUDE_DIRS` variable.

edenhill: Okay, thanks fort he clarification 👍 

Can you add a docker-based build script to packaging/tools/build-mingw.sh ?

edenhill: Sorry, wrong PR, nevermind the buildscript!

edenhill:
Is this good to go?

benesch:
Yep, definitely!

edenhill:
Thank you for this!

PR #2642, opened 2019-12-02
consumer_lag fix, and some test compat fixes

edenhill:


src/rdkafka_broker.c
```diff
@@ -3911,7 +3911,11 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
                         rd_kafka_toppar_lock(rktp);
                         rktp->rktp_lo_offset = hdr.LogStartOffset;
                         rktp->rktp_hi_offset = hdr.HighwaterMarkOffset;
-                        rktp->rktp_ls_offset = hdr.LastStableOffset;
+                        /* Let the LastStable offset be the effective
+                         * end_offset based on protocol version, that is:
+                         * if connected to a broker that does not support
+                         * LastStableOffset we use the HighwaterMarkOffset. */
+                        rktp->rktp_ls_offset = end_offset;
```
mhowlett: I can't see a problem with this. My instinct is they are better separate, since combining them makes the definition of ls slightly more nebulous but it seems natural/justifiable to use hi_offset in place of ls_offset in protocol versions where it's not available.

note: I am unsure why there would be an issue related to this, since the behavior in offset_reset should revert to the old way in the case of ls_offset being invalid. but i'm probably blind.

mhowlett: btw: the comment is wrong here: https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_offset.c#L803

edenhill: I'd consider LastStableOffset == HwMarkOffset for broker versions that predates transactions,
 and ls_offset will behave correctly in both offset reset as well as stats/consumer_lag with this fix, I believe.

PR #2646, opened 2019-12-04
Update librdkafka.redist.targets

goldenbull:
add librdkafkacpp.lib into VisualStudio targets file


edenhill:
I'm not sure we want C applications to link to the CPP library, which I assume this is doing.

goldenbull:
maybe my knowledge of MSVC compiler is out dated, but I guess C application will not be linked to the cpp lib file if no function is used in the lib, compiler will ignore the cpp lib file even it's included in project setting.

PR #2657, opened 2019-12-15
Fix Coverity parse warnings

nicklauslittle:
This fixes the parse warnings generated by Coverity during a scan.


edenhill:
Thank you!

PR #2659, opened 2019-12-17
Improve pkg-config support

gridaphobe:
1. Add required libraries (zlib, lz4, zstd, openssl, sasl2) to the
   pkg-config `Requires` field instead of to `Libs` if they were
   discovered via pkg-config. This allows pkg-config (and other tools
   built on top of the pkg-config metadata) to see librdkafka's full
   dependency graph.

2. Prefix full paths to librdkafka.a with the pkg-config variable
   ${pc_sysrootdir}. pkg-config automatically applies this prefix to
   paths passed to -I and -L, but nowhere else. This means that the
   previous link line in the -static.pc files is inconsistent, and
   would expand to

     -L${pc_sysrootdir}${libdir} ${libdir}/librdkafka.a

3. #include <stddef.h> when checking for sasl so that detection via
   pkg-config works. This is needed as sasl.h references some types
   that are defined in stddef.h.

### DIFF
--- a/mklove/Makefile.base
+++ b/mklove/Makefile.base
@@ -176,7 +176,7 @@ includedir=$(includedir)
 Name: $(LIBNAME)-static
 Description: $(MKL_APP_DESC_ONELINE) (static)
 Version: $(MKL_APP_VERSION)
-Requires: $(MKL_PKGCONFIG_REQUIRES_PRIVATE:rdkafka=rdkafka-static)
+Requires: $(MKL_PKGCONFIG_REQUIRES:rdkafka=rdkafka-static)
 Cflags: -I$${includedir}
 Libs: -L$${libdir} $${pc_sysrootdir}$${libdir}/$(LIBNAME)-static.a $(MKL_PKGCONFIG_LIBS_PRIVATE)
 endef
@@ -269,7 +269,7 @@ lib-uninstall:
 bin-install:
 	@printf "$(MKL_YELLOW)Install $(BIN) to $$DESTDIR$(prefix)$(MKL_CLR_RESET)\n"
 	$(INSTALL) -d $$DESTDIR$(bindir) && \
-	$(INSTALL) $(BIN) $$DESTDIR$(bindir) 
+	$(INSTALL) $(BIN) $$DESTDIR$(bindir)
 
 bin-uninstall:
 	@printf "$(MKL_YELLOW)Uninstall $(BIN) from $$DESTDIR$(prefix)$(MKL_CLR_RESET)\n"
--- a/mklove/Makefile.base
+++ b/mklove/Makefile.base
@@ -160,9 +160,10 @@ includedir=$(includedir)
 Name: $(LIBNAME)
 Description: $(MKL_APP_DESC_ONELINE)
 Version: $(MKL_APP_VERSION)
+Requires.private: $(MKL_PKGCONFIG_REQUIRES_PRIVATE)
 Cflags: -I$${includedir}
 Libs: -L$${libdir} -l$(LIBNAME0)
-Libs.private: $(LIBS)
+Libs.private: $(MKL_PKGCONFIG_LIBS_PRIVATE)
 endef
 
 export _PKG_CONFIG_DEF
@@ -175,8 +176,9 @@ includedir=$(includedir)
 Name: $(LIBNAME)-static
 Description: $(MKL_APP_DESC_ONELINE) (static)
 Version: $(MKL_APP_VERSION)
+Requires: $(MKL_PKGCONFIG_REQUIRES_PRIVATE:rdkafka=rdkafka-static)
 Cflags: -I$${includedir}
-Libs: -L$${libdir} $${libdir}/$(LIBNAME)-static.a $(MKL_DYNAMIC_LIBS)
+Libs: -L$${libdir} $${pc_sysrootdir}$${libdir}/$(LIBNAME)-static.a $(MKL_PKGCONFIG_LIBS_PRIVATE)
 endef
 
 export _PKG_CONFIG_STATIC_DEF
--- a/mklove/modules/configure.base
+++ b/mklove/modules/configure.base
@@ -1724,6 +1724,11 @@ $cflags"
         local stlibs=$(mkl_lib_check_static $1 "$libs")
         if [[ -n $stlibs ]]; then
             libs=$stlibs
+        else
+            # if we don't find a static library to bundle into the
+            # -static.a, we need to export a pkgconfig dependency
+            # so it can be resolved when linking downstream packages
+            mkl_mkvar_append $1 "MKL_PKGCONFIG_REQUIRES" "$libname"
         fi
     fi
 
@@ -2338,8 +2343,3 @@ function mkl_toggle_option_lib {
     eval "function _tmp_func { mkl_lib_check \"$2\" \"$3\" \"$4\" CC \"$7\"; }"
     mkl_func_push MKL_CHECKS "$MKL_MODULE" _tmp_func
 }
-
-
-
-
-
--- a/mklove/modules/configure.base
+++ b/mklove/modules/configure.base
@@ -1578,6 +1578,7 @@ function mkl_lib_check0 {
         # E.g., check for crypto and then ssl should result in -lssl -lcrypto
         mkl_dbg "$1: from lib_check: LIBS: prepend $libs"
         mkl_mkvar_prepend "$1" LIBS "$libs"
+        mkl_mkvar_prepend "$1" MKL_PKGCONFIG_LIBS_PRIVATE "$libs"
     fi
 
     return 0
@@ -1714,6 +1715,8 @@ $cflags"
             return 1
         fi
     fi
+
+    mkl_mkvar_append $1 "MKL_PKGCONFIG_REQUIRES_PRIVATE" "$libname"
 
     mkl_mkvar_append $1 "CFLAGS" "$cflags"
 
--- a/mklove/modules/configure.libsasl2
+++ b/mklove/modules/configure.libsasl2
@@ -25,8 +25,8 @@ function manual_checks {
     mkl_meta_set "libsasl2" "rpm" "cyrus-sasl"
     mkl_meta_set "libsasl2" "apk" "cyrus-sasl-dev"
 
-    if ! mkl_lib_check "libsasl2" "WITH_SASL_CYRUS" $action CC "-lsasl2" "#include <sasl/sasl.h>" ; then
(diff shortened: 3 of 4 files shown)


edenhill:
@gridaphobe As I'm no expert in pkg-config (obviously), I'm relying on the assumption that you've tested this in the real world?

Can you rebase this on latest master and I'll review it, if you still think it should be added?

gridaphobe:
I wouldn't call myself a pkg-config expert either, but I've tested these changes in the context of our internal package system at Bloomberg, where we have ~900 reverse dependencies of librdkafka, using both the regular and -static pc files from a variety of languages.

If you have contacts in the debian or redhat communities that handle packaging of librdkafka, I'd be happy for them to review this patch too!

edenhill:
@paravoid @vincentbernat Could you take a quick look at this to see if it would pose problems for the Debian packages?

gridaphobe:
For reference, here's the diff of the pc files that I get on my macbook.

```diff
❯ diff -ur old new
diff -ur old/rdkafka++-static.pc new/rdkafka++-static.pc
--- old/rdkafka++-static.pc	2020-09-04 09:35:01.000000000 -0400
+++ new/rdkafka++-static.pc	2020-09-04 09:35:45.000000000 -0400
@@ -5,5 +5,6 @@
 Name: librdkafka++-static
 Description: The Apache Kafka C/C++ library (static)
 Version: 1.5.0
+Requires: rdkafka-static
 Cflags: -I${includedir}
-Libs: -L${libdir} ${libdir}/librdkafka++-static.a -llz4 -lm -lzstd -lsasl2 -lssl -lcrypto -lz -ldl -lpthread
+Libs: -L${libdir} ${pc_sysrootdir}${libdir}/librdkafka++-static.a -lm -ldl -lpthread

diff -ur old/rdkafka++.pc new/rdkafka++.pc
--- old/rdkafka++.pc	2020-09-04 09:35:01.000000000 -0400
+++ new/rdkafka++.pc	2020-09-04 09:35:45.000000000 -0400
@@ -5,6 +5,7 @@
 Name: librdkafka++
 Description: The Apache Kafka C/C++ library
 Version: 1.5.0
+Requires.private: rdkafka
 Cflags: -I${includedir}
 Libs: -L${libdir} -lrdkafka++
-Libs.private: -L../src -lrdkafka
+Libs.private: -lm -ldl -lpthread

diff -ur old/rdkafka-static.pc new/rdkafka-static.pc
--- old/rdkafka-static.pc	2020-09-04 09:35:04.000000000 -0400
+++ new/rdkafka-static.pc	2020-09-04 09:35:48.000000000 -0400
@@ -5,5 +5,6 @@
 Name: librdkafka-static
 Description: The Apache Kafka C/C++ library (static)
 Version: 1.5.0
+Requires: zlib libcrypto libssl libzstd liblz4
 Cflags: -I${includedir}
-Libs: -L${libdir} ${libdir}/librdkafka-static.a -llz4 -lm -lzstd -lsasl2 -lssl -lcrypto -lz -ldl -lpthread
+Libs: -L${libdir} ${pc_sysrootdir}${libdir}/librdkafka-static.a -lm -ldl -lpthread

diff -ur old/rdkafka.pc new/rdkafka.pc
--- old/rdkafka.pc	2020-09-04 09:35:04.000000000 -0400
+++ new/rdkafka.pc	2020-09-04 09:35:48.000000000 -0400
@@ -5,6 +5,7 @@
 Name: librdkafka
 Description: The Apache Kafka C/C++ library
 Version: 1.5.0
+Requires.private: zlib libcrypto libssl libzstd liblz4
 Cflags: -I${includedir}
 Libs: -L${libdir} -lrdkafka
-Libs.private: -L/usr/local/Cellar/lz4/1.9.2/lib -llz4 -lm -L/usr/local/Cellar/zstd/1.4.5//lib -lzstd -lsasl2 -L/usr/local/Cellar/openssl@1.1/1.1.1g/lib -lssl -L/usr/local/Cellar/openssl@1.1/1.1.1g/lib -lcrypto -L/usr/local/Cellar/zlib/1.2.11/lib -lz -ldl -lpthread
+Libs.private: -lm -ldl -lpthread
```

edenhill:
@remicollet Do these changes look ok from an RPM perspective?

vincentbernat:
For Debian, I think the changes are fine.

remicollet:
Seems **libsasl2** disappear from dependencies

BTW for RPM packaging, static/private stuff are not used (by Guidelines), so no real impact.

Notice: some old libsasl version doesn't provide the .pc file so cannot be used as "requires" (but very old version, such as 2.1.23 on RHEL / CentOS 6, which is close to its EOL), at least version >= 2.1.26 are ok)

gridaphobe:
Good catch @remicollet, it looks like the change I made to the include check for `sasl.h` was producing invalid C code somehow.

I also just noticed that when you `--enable-static` rdkafka now bundles the dependencies into `librdkafka-static.a`, which means we don't need to add them to the `Requires` field for `rdkafka-static.pc`.

I'll try to fix both issues tomorrow.

edenhill:
> I also just noticed that when you --enable-static rdkafka now bundles the dependencies into librdkafka-static.a

Only for the dependencies that are available as static libraries, the remaining deps (if any) still needs to be linked dynamically and thus listed in the .pc file.

gridaphobe:
Ok, I've fixed the `sasl` include check and removed libs bundled into the `-static.a` from the Requires clause. The diff (under --enable-static) is now:

```diff
diff -ur old/rdkafka++-static.pc new/rdkafka++-static.pc
--- old/rdkafka++-static.pc	2020-09-10 19:54:54.000000000 -0400
+++ new/rdkafka++-static.pc	2020-09-10 19:51:46.000000000 -0400
@@ -5,5 +5,6 @@
 Name: librdkafka++-static
 Description: The Apache Kafka C/C++ library (static)
 Version: 1.5.0
+Requires: rdkafka-static
 Cflags: -I${includedir}
-Libs: -L${libdir} ${libdir}/librdkafka++-static.a -lm -lsasl2 -ldl -lpthread
+Libs: -L${libdir} ${pc_sysrootdir}${libdir}/librdkafka++-static.a -lm -lsasl2 -ldl -lpthread

diff -ur old/rdkafka++.pc new/rdkafka++.pc
--- old/rdkafka++.pc	2020-09-10 19:54:54.000000000 -0400
+++ new/rdkafka++.pc	2020-09-10 19:51:46.000000000 -0400
@@ -5,6 +5,7 @@
 Name: librdkafka++
 Description: The Apache Kafka C/C++ library
 Version: 1.5.0
+Requires.private: rdkafka
 Cflags: -I${includedir}
 Libs: -L${libdir} -lrdkafka++
-Libs.private: -L../src -lrdkafka
+Libs.private: -lm -lsasl2 -ldl -lpthread

diff -ur old/rdkafka-static.pc new/rdkafka-static.pc
--- old/rdkafka-static.pc	2020-09-10 19:54:50.000000000 -0400
+++ new/rdkafka-static.pc	2020-09-10 19:51:50.000000000 -0400
@@ -5,5 +5,6 @@
 Name: librdkafka-static
 Description: The Apache Kafka C/C++ library (static)
 Version: 1.5.0
+Requires:
 Cflags: -I${includedir}
-Libs: -L${libdir} ${libdir}/librdkafka-static.a -lm -lsasl2 -ldl -lpthread
+Libs: -L${libdir} ${pc_sysrootdir}${libdir}/librdkafka-static.a -lm -lsasl2 -ldl -lpthread

diff -ur old/rdkafka.pc new/rdkafka.pc
--- old/rdkafka.pc	2020-09-10 19:54:50.000000000 -0400
+++ new/rdkafka.pc	2020-09-10 19:51:50.000000000 -0400
@@ -5,6 +5,7 @@
 Name: librdkafka
 Description: The Apache Kafka C/C++ library
 Version: 1.5.0
+Requires.private: zlib libcrypto libssl libzstd
 Cflags: -I${includedir}
 Libs: -L${libdir} -lrdkafka
-Libs.private: -lm /usr/local/Cellar/zstd/1.4.5//lib/libzstd.a -lsasl2 /usr/local/Cellar/openssl@1.1/1.1.1g/lib/libssl.a /usr/local/Cellar/openssl@1.1/1.1.1g/lib/libcrypto.a /usr/local/Cellar/zlib/1.2.11/lib/libz.a -ldl -lpthread
+Libs.private: -lm -lsasl2 -ldl -lpthread
```

And if I hide e.g. `libzstd.a` I get

```diff
diff -ur old/rdkafka-static.pc new/rdkafka-static.pc
--- old/rdkafka-static.pc	2020-09-10 19:54:50.000000000 -0400
+++ new/rdkafka-static.pc	2020-09-10 20:07:20.000000000 -0400
@@ -5,5 +5,6 @@
 Name: librdkafka-static
 Description: The Apache Kafka C/C++ library (static)
 Version: 1.5.0
+Requires: libzstd
 Cflags: -I${includedir}
-Libs: -L${libdir} ${libdir}/librdkafka-static.a -lm -lsasl2 -ldl -lpthread
+Libs: -L${libdir} ${pc_sysrootdir}${libdir}/librdkafka-static.a -lm -lsasl2 -ldl -lpthread

diff -ur old/rdkafka.pc new/rdkafka.pc
--- old/rdkafka.pc	2020-09-10 19:54:50.000000000 -0400
+++ new/rdkafka.pc	2020-09-10 20:07:20.000000000 -0400
@@ -5,6 +5,7 @@
 Name: librdkafka
 Description: The Apache Kafka C/C++ library
 Version: 1.5.0
+Requires.private: zlib libcrypto libssl libzstd
 Cflags: -I${includedir}
 Libs: -L${libdir} -lrdkafka
-Libs.private: -lm /usr/local/Cellar/zstd/1.4.5//lib/libzstd.a -lsasl2 /usr/local/Cellar/openssl@1.1/1.1.1g/lib/libssl.a /usr/local/Cellar/openssl@1.1/1.1.1g/lib/libcrypto.a /usr/local/Cellar/zlib/1.2.11/lib/libz.a -ldl -lpthread
+Libs.private: -lm -lsasl2 -ldl -lpthread
```

Note that `libzstd` is marked as a public dependency for `rdkafka-static.pc` since it's not bundled in `librdkafka-static.a`, but still as a private dependency for `rdkafka.pc`. This is still correct as `librdkafka.so` encodes its own dependency on zstd and we never expose zstd types, but `librdkafka.a` (not the static bundle) still needs to link against `libzstd.a`.

edenhill (approved):
LGTM.
Is this good to go?

gridaphobe:
I just did another full rebuild of our internal packages with the new pkgconfig files, and it looks good to me 👍

edenhill:
Thank you @gridaphobe ½

PR #2660, opened 2019-12-19
fix: cleanup conf object if fails creating producer

fboranek:
I notice there is memory leak in `RdKafka::Producer::create` if `rd_kafka_new` returns `NULL`.

### DIFF
--- a/src-cpp/ConsumerImpl.cpp
+++ b/src-cpp/ConsumerImpl.cpp
@@ -58,6 +58,9 @@ RdKafka::Consumer *RdKafka::Consumer::create (RdKafka::Conf *conf,
   if (!(rk = rd_kafka_new(RD_KAFKA_CONSUMER, rk_conf,
                           errbuf, sizeof(errbuf)))) {
     errstr = errbuf;
+    // rd_kafka_new() takes ownership only if succeeds
+    if (rk_conf)
+      rd_kafka_conf_destroy(rk_conf);
     delete rkc;
     return NULL;
   }
--- a/src-cpp/KafkaConsumerImpl.cpp
+++ b/src-cpp/KafkaConsumerImpl.cpp
@@ -41,7 +41,7 @@ RdKafka::KafkaConsumer *RdKafka::KafkaConsumer::create (RdKafka::Conf *conf,
   rd_kafka_conf_t *rk_conf = NULL;
   size_t grlen;
 
-  if (!confimpl->rk_conf_) {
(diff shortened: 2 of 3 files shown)


edenhill:
Great! Can you fix the same issues in ConsumerImpl.cpp, KafkaConsumerImpl.cpp

fboranek:
You are right. The same issue. Done.

src-cpp/ConsumerImpl.cpp
```diff
@@ -58,6 +58,8 @@ RdKafka::Consumer *RdKafka::Consumer::create (RdKafka::Conf *conf,
   if (!(rk = rd_kafka_new(RD_KAFKA_CONSUMER, rk_conf,
                           errbuf, sizeof(errbuf)))) {
     errstr = errbuf;
+    // rd_kafka_new() takes ownership only if succeeds
+    rd_kafka_conf_destroy(rk_conf);
```
edenhill: Sorry forgot to comment on this last time around, rk_conf may be NULL for the Producer and simple Consumer, so need and if-guard:

```c
if (rk_conf)
  rd_kafka_conf_destroy(rk_conf);
```

fboranek: Good catch. I didn't notice.

src-cpp/KafkaConsumerImpl.cpp
```diff
@@ -63,6 +63,9 @@ RdKafka::KafkaConsumer *RdKafka::KafkaConsumer::create (RdKafka::Conf *conf,
   if (!(rk = rd_kafka_new(RD_KAFKA_CONSUMER, rk_conf,
                           errbuf, sizeof(errbuf)))) {
     errstr = errbuf;
+    // rd_kafka_new() takes ownership only if succeeds
+    if (rk_conf)
```
edenhill: Not needed, rk_conf is always set in this constructor.
Revert to unambigiuate intent.
Thanks

fboranek: Then there can be null pointer dereferention on line 

:44 `if (!confimpl->rk_conf_) {`

Since this is probably called only once during live time of application, I'am goint test `confimpl` on NULL and rise exception if requiaremts are not met.

fboranek: Not rise exception but return error.

edenhill:
Thank you!

PR #2670, opened 2020-01-05
Use _WIN32 preprocessor macro instead of _MSC_VER

neptoess:
_MSC_VER is only defined by MSVC, but other compilers, e.g. gcc from MinGW, can also be used to build librdkafka on Windows. _WIN32 is not compiler-specific.

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -55,7 +55,7 @@ extern "C" {
 #endif
 #endif
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 #include <basetsd.h>
 #ifndef WIN32_MEAN_AND_LEAN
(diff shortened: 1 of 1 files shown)


neptoess:
It looks like https://github.com/edenhill/librdkafka/pull/2553 is already pending and covers a lot more cases of _MSC_VER in the code. I’m not using a cross compiler. I only went far enough to allow cgo to link to rdkafka when I reference confluent-kafka-go. However, it looks like the #ifdef checks in that PR are still compiler-based, instead of OS-target based. I still believe that PR may be the smarter one to merge, but I will propose that the author use the _WIN32 macro instead. If he agrees, I will close this PR.

PR #2673, opened 2020-01-06
Corrected the reference to the broker/topic setting

sknop:
It is `message.max.bytes`, not `max.message.bytes`


CONFIGURATION.md
```diff
@@ -7,7 +7,7 @@ builtin.features                         |  *  |                 | gzip, snappy,
...
-message.max.bytes                        |  *  | 1000 .. 1000000000 |       1000000 | medium     | Maximum Kafka protocol request message size. Due to differing framing overhead between protocol versions the producer is unable to reliably enforce a strict max message limit at produce time and may exceed the maximum size by one message in protocol ProduceRequests, the broker will enforce the the topic's `max.message.bytes` limit (see Apache Kafka documentation). <br>*Type: integer*
+message.max.bytes                        |  *  | 1000 .. 1000000000 |       1000000 | medium     | Maximum Kafka protocol request message size. Due to differing framing overhead between protocol versions the producer is unable to reliably enforce a strict max message limit at produce time and may exceed the maximum size by one message in protocol ProduceRequests, the broker will enforce the the topic's `message.max.bytes` limit (see Apache Kafka documentation). <br>*Type: integer*
```
edenhill: The topic-level config is called `max.message.bytes`, the broker-level config is called `message.max.bytes`, so I believe the existing description is correct since it specifically says `topic's ..`.

PR #2679, opened 2020-01-08
minor fix to rdkafka_example usage: add lz4 and zstd compression codec to usage.

Eliyahu-Machluf:
update the usage with the compression codec of lz4 and zstd


Eliyahu-Machluf:
The check which failed is not relevant to the fix (simply fix of usage text)

edenhill:
Thank you!

PR #2683, opened 2020-01-10
OpenBSD support

jaspervandenberg:
Both points that reference FreeBSD should also check for OpenBSD as OS.

I was able to compile the library on OpenBSD with these two minor changes.


edenhill:
Thank you!

PR #2661, opened 2019-12-19
Store offsets corresponding to ctrl messages

mhowlett:
resolves: https://github.com/edenhill/librdkafka/issues/2539

PR #2684, opened 2020-01-10
Broaded check for no-op when no messages read

mhowlett:
Additionally, i think it's misleading to use `RD_KAFKA_RESP_ERR_MSG_SIZE_TOO_LARGE` in the error op below, as it's a broker error implying it originated from the broker. This had me confused for some time. I'll let you work out what to do with that.


edenhill:
Can you rebase this on latest master?

mhowlett:
this one is important (the symptom was the branch with the `RD_KAFKA_RESP_ERR_MSG_SIZE_TOO_LARGE` error was getting hit errantly). unfortunately, i can't remember / don't have a link to somewhere with more info for context.

edenhill:
Is this really needed when #2661 is fixed?

edenhill:
Fixed by #2743

PR #2686, opened 2020-01-15
Fix format specifier for printing size_t

pponnuvel:
size_t being an unsigned type, '%zu' is the correct format specifier.
 Using '%zd' is technically undefined behaviour (due to sign mismatch).

### DIFF
--- a/src/lz4frame.c
+++ b/src/lz4frame.c
@@ -834,7 +834,7 @@ size_t LZ4F_compressUpdate(LZ4F_cctx* cctxPtr,
     LZ4F_lastBlockStatus lastBlockCompressed = notDone;
     compressFunc_t const compress = LZ4F_selectCompression(cctxPtr->prefs.frameInfo.blockMode, cctxPtr->prefs.compressionLevel);
 
-    DEBUGLOG(4, "LZ4F_compressUpdate (srcSize=%zu)", srcSize);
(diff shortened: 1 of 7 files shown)


edenhill:
Should use `"...%"PRIusz".."` for portability

pponnuvel:
PRIu macros are for fixed width integers (C99). Unless there's some non-standard platform/toolchain which doesn't support C99, '%zu' should work everywhere for `size_t`.

Checking MSVS, [Visual studio 2019](https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=vs-2019) documents it. Perhaps, older versions of Visual Studio and/or somewhere else it doesn't work?

edenhill:
Older versions of MSVC use `%Iu` (as in i, not L). See rdwin32.h

pponnuvel:
Ah, rdwin32.h and rdposix.h makes sense.

Modified to use the macro - left out the examples as that requires more headers from src/ (I can't test for all platforms). 

On another note, it's probably OK now (at least for lack of `z` modifier support) given that it's been using `zd` currently (which wasn't supported on older Visual Studio).

 I'll leave the above two your consideration. Thanks.

edenhill:
Thank you!

PR #2688, opened 2020-01-15
Use sig_atomic_t for vars modified in signal handlers

pponnuvel:
Another minor tweak to a type:
For proper atomicity (with respect to signal handler), the type of variables modified in signal handlers should be 'sig_atomic_t' with 'volatile' qualifier (modifying other types such as a plain 'int' is undefined behaviour even if it works as expected on most platforms). This is atomicity between main code & signal handler i.e. required even in single-threaded program.

In C++ files, used 0 for false and 1 for true respectively just to avoid any confusion and implicit conversions (which is otherwise fine).

sig_atomic_t type is available since C89, so should be reasonably portable.

### DIFF
--- a/examples/consumer.c
+++ b/examples/consumer.c
@@ -44,7 +44,7 @@
 #include "rdkafka.h"
 
 
-static int run = 1;
+static volatile sig_atomic_t run = 1;
 
 /**
  * @brief Signal termination of program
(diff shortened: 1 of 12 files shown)


edenhill:
Thanks for this!

PR #2690, opened 2020-01-17
Add vcpkg installation instructions

NancyLi1013:
librdkafka is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for librdkafka and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build librdkafka, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, arm) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/librdkafka/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,15 +72,15 @@ On Windows, reference [librdkafka.redist](https://www.nuget.org/packages/librdka
 For other platforms, follow the source building instructions below.
 
 
-## Building librdkafka - Using vcpkg
+## Installing librdkafka - Using vcpkg
 
 You can download and install librdkafka using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
 
-    git clone https://github.com/Microsoft/vcpkg.git
-    cd vcpkg
-    ./bootstrap-vcpkg.sh
-    ./vcpkg integrate install
-    vcpkg install librdkafka
+    $ git clone https://github.com/Microsoft/vcpkg.git
+    $ cd vcpkg
+    $ ./bootstrap-vcpkg.sh
+    $ ./vcpkg integrate install
+    $ vcpkg install librdkafka
 
 The librdkafka port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,6 +72,19 @@ On Windows, reference [librdkafka.redist](https://www.nuget.org/packages/librdka
 For other platforms, follow the source building instructions below.
 
 
+## Building librdkafka - Using vcpkg
```
edenhill: Change to `## Installing librdkafka using vcpkg`

README.md
```diff
@@ -72,6 +72,19 @@ On Windows, reference [librdkafka.redist](https://www.nuget.org/packages/librdka
...
+
+You can download and install librdkafka using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
```
edenhill: Use a tripple-tick bash block with `$` prompts to be consistent with the other examples.

edenhill:
It'd be good if the vcpkg was upgraded to the latest version (v1.3.0) before this PR gets merged.

NancyLi1013:
Hi @edenhill thanks for your review and so quick feedback.
I have updated the changes you requested above. Please help check them again.

As for the upgrade to the latest version in vcpkg, we will try to update it later.
Thanks again for your help and support.

README.md
```diff
@@ -72,15 +72,15 @@ On Windows, reference [librdkafka.redist](https://www.nuget.org/packages/librdka
...
+## Installing librdkafka - Using vcpkg
 
 You can download and install librdkafka using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
 
-    git clone https://github.com/Microsoft/vcpkg.git
-    cd vcpkg
-    ./bootstrap-vcpkg.sh
-    ./vcpkg integrate install
-    vcpkg install librdkafka
+    $ git clone https://github.com/Microsoft/vcpkg.git
```
edenhill: Please use bash formatting, as so:
```
\`\`\`bash
$ cd vcpkg
$ ..
..
\`\`\`

(remove backslashes)

edenhill:
Thank you!

PR #2697, opened 2020-01-27
Fix test 0052 and rename test executable generated with CMake

dadufour:
Test 0052 is broken since 01JAN2020 - fix test by generating dynamically expected timestamps

Test executable generated by CMake was rdkafka_test while when generated through configure is test-runner. Rename rdkafka_test to test-runner to be consistent

### DIFF
--- a/tests/0052-msg_timestamps.c
+++ b/tests/0052-msg_timestamps.c
@@ -32,11 +32,6 @@
 /**
  * Verify message timestamp behaviour on supporting brokers (>=0.10.0.0).
  * Issue #858
- *
- * FIXME: Intermittent failures:
- *  "consume.easy: consumer_poll() timeout (1/-1 eof, 0/20 msgs)"
- * are due to the really old timestamps being used (my_timestamp, 1234)
- * causing the offset retention cleaner on the broker to kick in.
  */
 struct timestamp_range {
         int64_t min;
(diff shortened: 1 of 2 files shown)


tests/0052-msg_timestamps.c
```diff
@@ -32,24 +32,26 @@
...
 
+static void prepare_timestamps() {
+        struct timeval ts;
+        rd_gettimeofday(&ts, NULL);
 
+        /* broker timestamps expected to be within 120 seconds */
+        broker_timestamp.min = ((int64_t)ts.tv_sec * 1000LLU);
+        broker_timestamp.max = broker_timestamp.min + 120 * 1000LLU;
+
+        my_timestamp.min = my_timestamp.max = ((int64_t)ts.tv_sec * 1000LLU);
```
edenhill: We actually want these timestamps to be different from the broker-assigned timestamps so we can check that they were indeed assigned by the producer and not the broker. To avoid the retention cleaner we could probably use future timestamps, e.g., now + 1 day.

tests/0052-msg_timestamps.c
```diff
@@ -32,24 +32,26 @@
...
-};
-const struct timestamp_range my_timestamp = { 1234, 1234 };
+struct timestamp_range broker_timestamp;
+struct timestamp_range my_timestamp;
 
+static void prepare_timestamps() {
+        struct timeval ts;
+        rd_gettimeofday(&ts, NULL);
 
+        /* broker timestamps expected to be within 120 seconds */
```
edenhill: 120 is a reasonable test duration time on slow CIs, so make this 600 seconds or so.

tests/0052-msg_timestamps.c
```diff
@@ -32,24 +32,26 @@
...
         int64_t min;
         int64_t max;
 };
 
 const struct timestamp_range invalid_timestamp = { -1, -1 };
-const struct timestamp_range broker_timestamp = {
-        946684800000/* 2000-01-01 */, 1577836800000 /* 2020-01-01 */
-};
-const struct timestamp_range my_timestamp = { 1234, 1234 };
+struct timestamp_range broker_timestamp;
```
edenhill: Make these static (and the invalid timestamp too)

edenhill:
Thank you for your contribution!

PR #2701, opened 2020-02-04
implement admin DeleteRecords & DeleteGroups APIs with fanout

### LINKED ISSUE #2056: KIP-107: Add DeleteRecords in Admin API
Add a request type to the Admin API for the DeleteRecords protocol request (https://kafka.apache.org/protocol#The_Messages_DeleteRecords) - https://cwiki.apache.org/confluence/display/KAFKA/KIP-107%3A+Add+deleteRecordsBefore%28%29+API+in+AdminClient

I'm currently working on a patch for this.

gridaphobe:
Fixes #2056

This adds support for the DeleteRecords admin API. The request is mostly asynchronous -- there's a brief blocking period where we look up partition leaders -- and consolidates the per-broker responses into a single response that is sent back to the caller.

This PR supercedes https://github.com/edenhill/librdkafka/pull/2065.


src/rdkafka_admin.c
```diff
@@ -838,6 +841,29 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
 }
 
 
+typedef struct rd_kafka_AdminFanoutState_s {
+        rd_list_t requests;
```
gridaphobe: The `requests` list is not required by DeleteRecords, but we are using it in our implementation of DeleteGroups (which I'll be sending as soon as the `txns` branch lands).

gridaphobe:
Looks like the CI failure is due to a [timing issue](https://travis-ci.org/edenhill/librdkafka/jobs/646070494#L5478).

gridaphobe:
@edenhill I've implemented DeleteGroups as well and attached it to this PR since it uses the same underlying fanout mechanism for admin requests.

src/rdkafka_aux.c
```diff
@@ -97,3 +97,76 @@ void rd_kafka_topic_result_destroy (rd_kafka_topic_result_t *terr) {
 void rd_kafka_topic_result_free (void *ptr) {
         rd_kafka_topic_result_destroy((rd_kafka_topic_result_t *)ptr);
 }
+
+rd_kafka_resp_err_t
+rd_kafka_group_result_error (const rd_kafka_group_result_t *groupres) {
+        return groupres->err;
```
edenhill, line 103: Have this return an `rd_kafka_error_t *` object (or NULL) and then get rid of error_string below.

src/rdkafka_aux.c
```diff
@@ -97,3 +97,76 @@ void rd_kafka_topic_result_destroy (rd_kafka_topic_result_t *terr) {
...
+
+rd_kafka_group_result_t *
+rd_kafka_group_result_new (const char *group, ssize_t group_size,
+                           rd_kafka_resp_err_t err,
+                           const char *errstr) {
+        size_t tlen = group_size != -1 ? (size_t)group_size : strlen(group);
+        size_t elen = errstr ? strlen(errstr) + 1 : 0;
+        rd_kafka_group_result_t *terr;
+
+        terr = rd_malloc(sizeof(*terr) + tlen + 1 + elen);
```
edenhill, line 141: For non-fast-path objects like this I prefer to use rd_calloc() to avoid missing an initialization if the object is ever extended.

src/rdkafka_aux.c
```diff
@@ -97,3 +97,76 @@ void rd_kafka_topic_result_destroy (rd_kafka_topic_result_t *terr) {
...
+ * @param errstr Optional error string.
+ *
+ * All input arguments are copied.
+ */
+
+rd_kafka_group_result_t *
+rd_kafka_group_result_new (const char *group, ssize_t group_size,
+                           rd_kafka_resp_err_t err,
+                           const char *errstr) {
+        size_t tlen = group_size != -1 ? (size_t)group_size : strlen(group);
```
edenhill, line 137: tlen ->  glen

src/rdkafka_broker.c
```diff
@@ -5463,11 +5463,10 @@ rd_kafka_broker_t *rd_kafka_broker_add (rd_kafka_t *rk,
 
 	rd_kafka_broker_unlock(rkb);
 
-        /* Add broker state monitor for the (txn) coordinator request to use */
-        if (rd_kafka_is_transactional(rk))
-                rd_kafka_broker_monitor_add(&rkb->rkb_coord_monitor, rkb,
-                                            rk->rk_ops,
-                                            rd_kafka_coord_rkb_monitor_cb);
+        /* Add broker state monitor for the coordinator request to use */
```
edenhill, line 5466: Can you add something like:
`This is used by the transaction code and DeleteRecords(??) Admin API`

src/rdkafka_op.h
```diff
@@ -394,8 +396,13 @@ struct rd_kafka_op_s {
...
                                             *   the controller.
-                                            *   Defaults to -1:
-                                            *   look up and use controller. */
+                                            *   Special values:
+                                            *   -1:
+                                            *   look up and use controller.
+                                            *   -2:
+                                            *   look up and use coordinator.
+                                            *   */
+                        char *coordname; /**< Which coordinator to look up.  */
```
edenhill, line 405: Looking at rdkafka_coord.c we should probably have both coordtype and coordkey here.

src/rdkafka_request.c
```diff
@@ -3527,6 +3527,116 @@ rd_kafka_DeleteTopicsRequest (rd_kafka_broker_t *rkb,
 }
 
 
+/**
+ * @brief Construct and send DeleteRecordsRequest to \p rkb
+ *        with the DeleteRecords objects (DeleteRecords_t *) in
+ *        \p delete_records, using \p options.
```
edenhill, line 3533: No such param

src/rdkafka_request.c
```diff
@@ -3527,6 +3527,116 @@ rd_kafka_DeleteTopicsRequest (rd_kafka_broker_t *rkb,
...
+                              rd_kafka_resp_cb_t *resp_cb,
+                              void *opaque) {
+        rd_kafka_buf_t *rkbuf;
+        int16_t ApiVersion = 0;
+        size_t of_TopicArrayCnt = 0, of_PartArrayCnt = 0;
+        int32_t topic_cnt = 0, part_cnt = 0;
+        const char *last_topic = "";
+        int features;
+        int i = 0;
+        rd_kafka_topic_partition_list_t *partitions;
```
edenhill, line 3557: This probably needs to be const since offsets_list is const

src/rdkafka_request.c
```diff
@@ -3527,6 +3527,116 @@ rd_kafka_DeleteTopicsRequest (rd_kafka_broker_t *rkb,
...
+        int op_timeout;
+
+        partitions = rd_list_elem(offsets_list, 0);
+
+        if (partitions == NULL) {
+                rd_snprintf(errstr, errstr_size, "Partition vector is NULL");
```
edenhill, line 3563: This should be checked in the Admin API, by the time we get here input params should be proper.

src/rdkafka_request.c
```diff
@@ -3527,6 +3527,116 @@ rd_kafka_DeleteTopicsRequest (rd_kafka_broker_t *rkb,
...
+                return RD_KAFKA_RESP_ERR__INVALID_ARG;
+        }
+
+        if (partitions->cnt == 0) {
```
edenhill, line 3567: This too

src/rdkafka_request.c
edenhill: If it needs sorting we should make a copy of it (but we should already have a copy, so maybe make the offsets_list argument non-const instead).

src/rdkafka_request.c
```diff
@@ -3527,6 +3527,116 @@ rd_kafka_DeleteTopicsRequest (rd_kafka_broker_t *rkb,
...
+
+        rkbuf = rd_kafka_buf_new_request(rkb, RD_KAFKAP_DeleteRecords, 1,
+                                         4 +
+                                         (partitions->cnt * 100) +
+                                         4);
+
+        rd_kafka_topic_partition_list_sort_by_topic(partitions);
+
+        /* #topics */
+        of_TopicArrayCnt = rd_kafka_buf_write_i32(rkbuf, 0); /* updated later */
```
edenhill, line 3589: You can use `rd_kafka_write_topic_partitions()` instead.

src/rdkafka_request.c
```diff
@@ -3840,6 +3950,66 @@ rd_kafka_DescribeConfigsRequest (rd_kafka_broker_t *rkb,
...
+                              rd_kafka_replyq_t replyq,
+                              rd_kafka_resp_cb_t *resp_cb,
+                              void *opaque) {
+        rd_kafka_buf_t *rkbuf;
+        int16_t ApiVersion = 0;
+        int features;
+        int i = 0;
+        rd_kafka_DeleteGroup_t *delt;
+
+        if (rd_list_cnt(del_groups) == 0) {
```
edenhill, line 3979: Should be checked in the Admin API

src/rdkafka_request.c
```diff
@@ -3840,6 +3950,66 @@ rd_kafka_DescribeConfigsRequest (rd_kafka_broker_t *rkb,
...
+                rd_snprintf(errstr, errstr_size, "No groups to delete");
+                rd_kafka_replyq_destroy(&replyq);
+                return RD_KAFKA_RESP_ERR__INVALID_ARG;
+        }
+
+        ApiVersion = rd_kafka_broker_ApiVersion_supported(
+                rkb, RD_KAFKAP_DeleteGroups, 0, 1, &features);
+        if (ApiVersion == -1) {
+                rd_snprintf(errstr, errstr_size,
+                            "DeleteGroups Admin API (KIP-229) not supported "
```
edenhill, line 3989: Update the list of supported KIPs in INTRODUCTION.md

src/rdkafka_request.h
```diff
@@ -330,4 +339,24 @@ rd_kafka_EndTxnRequest (rd_kafka_broker_t *rkb,
...
+                               const rd_list_t *offsets_list /*(rd_topic_partition_list_t*)*/,
+                               rd_kafka_AdminOptions_t *options,
+                               char *errstr, size_t errstr_size,
+                               rd_kafka_replyq_t replyq,
+                               rd_kafka_resp_cb_t *resp_cb,
+                               void *opaque);
+
+
+rd_kafka_resp_err_t
+rd_kafka_DeleteRecordsRequest (rd_kafka_broker_t *rkb,
```
edenhill, line 354: So many!

tests/0081-admin.c
```diff
@@ -1131,6 +1131,419 @@ static void do_test_unclean_destroy (rd_kafka_type_t cltype, int with_mainq) {
...
+
+        TIMING_START(&timing, "DeleteRecords");
+        TEST_SAY("Call DeleteRecords\n");
+        err = rd_kafka_DeleteRecords(rk, del_records, options, q, tmout_multip(20*1000));
+        TEST_ASSERT(!err, "%s", rd_kafka_err2str(err));
+        TIMING_ASSERT_LATER(&timing, 0, 1000);
+
+        TIMING_START(&timing, "DeleteRecords.queue_poll");
+
+        const rd_kafka_DeleteRecords_result_t *res;
```
edenhill, line 1226: All variable defs need to go at head of scope to be compliant with old compilers

tests/0081-admin.c
```diff
@@ -1131,6 +1131,419 @@ static void do_test_unclean_destroy (rd_kafka_type_t cltype, int with_mainq) {
...
+
+        /* Poll result queue for DeleteRecords result.
+         * Print but otherwise ignore other event types
+         * (typically generic Error events). */
+        while(1) {
```
edenhill, line 1231: style nit: `while (1) {`

tests/0081-admin.c
```diff
@@ -1131,6 +1131,419 @@ static void do_test_unclean_destroy (rd_kafka_type_t cltype, int with_mainq) {
...
+                rd_kafka_resp_err_t expected_err = 0;
+
+                if (expected_offset == RD_KAFKA_OFFSET_END)
+                        expected_offset = msgs_cnt;
+
+                /* Expect Offset out of range error */
+                if(input->offset < RD_KAFKA_OFFSET_END || input->offset > msgs_cnt)
+                        expected_err = 1;
+
+                TEST_SAY("DeleteRecords Returned [%s] %s-%d low-watermark = %d\n",
```
edenhill, line 1291: We typically format topic partition as `%s [%d]`, good to be consistent.

tests/0081-admin.c
```diff
@@ -1131,6 +1131,419 @@ static void do_test_unclean_destroy (rd_kafka_type_t cltype, int with_mainq) {
...
+                if (output->err == 0 && output->offset != expected_offset)
+                        TEST_FAIL_LATER("%s-%d: "
+                                        "expected offset %"PRId64", got %"PRId64,
+                                        output->topic,
+                                        output->partition,
+                                        expected_offset,
+                                        output->offset);
+        }
+
+        /* Check watermarks for partitions */
```
edenhill, line 1331: :+1:

tests/0081-admin.c
```diff
@@ -1131,6 +1131,419 @@ static void do_test_unclean_destroy (rd_kafka_type_t cltype, int with_mainq) {
...
+        for (i = 0 ; i < MY_DEL_RECORDS_CNT; i++ ) {
+                int32_t partition;
+                for (partition = 0 ; partition < partitions_cnt; partition++ ) {
+                        const rd_kafka_topic_partition_t *del =
+                                rd_kafka_topic_partition_list_find(results, topics[i], partition);
+                        int64_t expected_low = 0;
+                        int64_t expected_high = msgs_cnt;
+                        int64_t low, high;
+
+                        if(del && del->err == 0) {
```
edenhill, line 1341: style-nit: `if (..`

tests/0081-admin.c
```diff
@@ -1131,6 +1131,419 @@ static void do_test_unclean_destroy (rd_kafka_type_t cltype, int with_mainq) {
...
+        }
+
+        TIMING_START(&timing, "DeleteGroups");
+        TEST_SAY("Call DeleteGroups\n");
+        rd_kafka_DeleteGroups(rk, del_groups, MY_DEL_GROUPS_CNT, options, q);
+        TIMING_ASSERT_LATER(&timing, 0, 1000);
+
+        TIMING_START(&timing, "DeleteGroups.queue_poll");
+
+        const rd_kafka_DeleteGroups_result_t *res;
```
edenhill, line 1470: move to head of scope

tests/0081-admin.c
```diff
@@ -1183,6 +1596,14 @@ static void do_test_apis (rd_kafka_type_t cltype) {
...
         do_test_DescribeConfigs(rk, mainq);
 
+        /* Delete records */
+        do_test_DeleteRecords("temp queue, op timeout 0", rk, NULL, 0);
+        do_test_DeleteRecords("main queue, op timeout 1500", rk, mainq, 1500);
+
+        /* Delete groups */
+        do_test_DeleteGroups("temp queue, op timeout 0", rk, NULL, 0);
+        do_test_DeleteGroups("main queue, op timeout 1500", rk, mainq, 1500);
+
```
edenhill, line 1606: Please also add some dry API testing to 0080, like passing empty or NULL lists, test timeouts, etc, look at the other APIs for inspiration.

src/rdkafka_admin.c
```diff
@@ -727,14 +785,32 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
-                if (rko->rko_u.admin_request.broker_id != -1) {
-                        /* Specific broker */
-                        rko->rko_u.admin_request.state =
-                                RD_KAFKA_ADMIN_STATE_WAIT_BROKER;
-                } else {
+                if (rko->rko_u.admin_request.broker_id == -1) {
                         /* Controller */
                         rko->rko_u.admin_request.state =
                                 RD_KAFKA_ADMIN_STATE_WAIT_CONTROLLER;
+		} else if (rko->rko_u.admin_request.broker_id == -2 ) {
```
edenhill, line 792: Use whitespace not tabs

src/rdkafka_admin.c
```diff
@@ -838,6 +917,29 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+typedef struct rd_kafka_AdminFanoutState_s {
+        rd_list_t requests;
+        rd_list_t responses;
+        size_t outstanding;
+} rd_kafka_AdminFanoutState_t;
+
+
+rd_kafka_AdminFanoutState_t*
+rd_kafka_AdminFanoutState_new (size_t cnt, void (*req_free_cb)(void*), void (*resp_free_cb)(void*)) {
+        rd_kafka_AdminFanoutState_t *state = rd_malloc(sizeof(rd_kafka_AdminFanoutState_t));
```
edenhill, line 929: Use rd_calloc()

src/rdkafka_admin.c
```diff
@@ -838,6 +917,29 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
 
 
+typedef struct rd_kafka_AdminFanoutState_s {
+        rd_list_t requests;
+        rd_list_t responses;
+        size_t outstanding;
+} rd_kafka_AdminFanoutState_t;
+
+
+rd_kafka_AdminFanoutState_t*
```
edenhill, line 927: style nit: `type *`

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+        rd_kafka_AdminFanoutState_t *state =
+                rd_kafka_confval_get_ptr(&rko_req->rko_u.admin_request.options.opaque);
+        rd_kafka_topic_partition_list_t *offsets = rd_list_elem(&state->responses, 0);
+
+        /* throttle_time_ms */
+        rd_kafka_buf_read_i32(reply, &Throttle_Time);
+        rd_kafka_op_throttle_time(rkb, rk->rk_rep, Throttle_Time);
+
+        /* #topics */
+        rd_kafka_buf_read_i32(reply, &topic_cnt);
```
edenhill, line 2881: Use rd_kafka_buf_read_topic_partitions()

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                                                                 partition);
+                        /* low_watermark */
+                        rd_kafka_buf_read_i64(reply, &skel->offset);
+
+                        /* error_code */
+                        rd_kafka_buf_read_i16(reply, &skel->err);
+                }
+        }
+
+        if (--state->outstanding == 0) {
```
edenhill, line 2912: add:
`rd_assert(state->outstanding > 0)`

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                rd_kafka_op_t *rko_result = rd_kafka_admin_result_new(rko_req);
```
edenhill, line 2913: This shadows the parent scope variable.

Also, suggest reversing this if-clause to cut down on sub-scoping, as so:

```c
if (--state->outstanding > 0) {
    /* still.. */
   return ..PARTIAL;
}

rko_result = ...;
...

return ..NO_ERROR;
```

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                *rko_resultp = rko_result;
+                rd_kafka_AdminFanoutState_free(state);
+                return RD_KAFKA_RESP_ERR_NO_ERROR;
+        } else {
+                /* Still waiting for outstanding requests to return */
+                return RD_KAFKA_RESP_ERR__PARTIAL;
+        }
+
+err_parse:
+        if (rko_result)
```
edenhill, line 2927: Dont think there's a case where this can happen.

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                     (void *)rd_kafka_partition_leader_destroy);
+        err = rd_kafka_topic_partition_list_query_leaders(rk, offsets, &leaders,
+                                                          timeout_ms);
+
+        if (err) {
+                rd_list_destroy(&leaders);
+                return err;
+        }
+
+        rd_kafka_AdminFanoutState_t *state =
```
edenhill, line 2962: Vardefs at head of scope

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                rd_kafka_AdminFanoutState_new(rd_list_cnt(&leaders),
+                                              rd_kafka_topic_partition_list_free,
+                                              rd_kafka_topic_partition_list_free);
+
+        rd_kafka_topic_partition_list_t *results =
+                rd_kafka_topic_partition_list_new(offsets->cnt);
+        rd_list_add(&state->responses, results);
+
+        /* For each leader send a request for its partitions */
```
edenhill, line 2971: We should (eventually) add a leader_req(), much like coord_req().

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+        RD_LIST_FOREACH(leader, &leaders, i) {
+                rd_kafka_op_t *rko;
+
+                rko = rd_kafka_admin_request_op_new(rk,
+                                                    RD_KAFKA_OP_DELETERECORDS,
+                                                    RD_KAFKA_EVENT_DELETERECORDS_RESULT,
+                                                    &cbs, options, rkqu);
+                rko->rko_u.admin_request.broker_id = leader->rkb->rkb_nodeid;
+                rd_kafka_AdminOptions_set_opaque(&rko->rko_u.admin_request.options, state);
```
edenhill, line 2980: This'll overwrite the application's opaque, suggest adding a new field to the internal AdminOptions for this purpose.
Don't need accessor functions for it since it is just used internally.

edenhill: Add a test-case to verify that the user's opaque makes it through to the result.

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+        };
+        rd_kafka_resp_err_t err;
+        int i;
+        rd_list_t leaders;
+        struct rd_kafka_partition_leader *leader;
+
+
+        rd_list_init(&leaders, offsets->cnt,
+                     (void *)rd_kafka_partition_leader_destroy);
+        err = rd_kafka_topic_partition_list_query_leaders(rk, offsets, &leaders,
```
edenhill, line 2954: Admin APIs need to be completely asynchronous, so we can't do this blocking thing here.

edenhill: Let me add the previously mentioned leader_req() to master branch and you can use that instead.

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+        const rd_kafka_op_t *rko = (const rd_kafka_op_t *) result;
+        size_t cntp;
+
+        rd_kafka_op_type_t reqtype =
+                rko->rko_u.admin_result.reqtype & ~RD_KAFKA_OP_FLAGMASK;
+        rd_assert(reqtype == RD_KAFKA_OP_DELETERECORDS);
+
+        cntp = rd_list_cnt(&rko->rko_u.admin_result.results);
+
+        rd_assert(cntp == 1);
```
edenhill, line 3016: Why 1 and not > 0?

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+ *        of all live replicas).
+ *
+ * The returned object's life-time is the same as the \p result object.
+ */
+const rd_kafka_topic_partition_list_t *
+rd_kafka_DeleteRecords_result_offsets (
+        const rd_kafka_DeleteRecords_result_t *result) {
+        const rd_kafka_topic_partition_list_t *offsets;
+        const rd_kafka_op_t *rko = (const rd_kafka_op_t *) result;
+        size_t cntp;
```
edenhill, line 3008: Use `cnt`, the `p` suffix is used for value-returning-pointers.

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                                     char *errstr, size_t errstr_size) {
+        const int log_decode_errors = LOG_ERR;
+        rd_kafka_resp_err_t err = RD_KAFKA_RESP_ERR_NO_ERROR;
+        rd_kafka_broker_t *rkb = reply->rkbuf_rkb;
+        rd_kafka_t *rk = rkb->rkb_rk;
+        int32_t group_cnt;
+        int i;
+        int32_t Throttle_Time;
+
+        rd_kafka_buf_read_i32(reply, &Throttle_Time);
```
edenhill, line 3096: Use `rd_kafka_buf_read_throttle_time()` (here and in DeleteRecs)

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+
+        if (group_cnt > rd_list_cnt(&rko_req->rko_u.admin_request.args))
+                rd_kafka_buf_parse_fail(
+                        reply,
+                        "Received %"PRId32" groups in response "
+                        "when only %d were requested", group_cnt,
+                        rd_list_cnt(&rko_req->rko_u.admin_request.args));
+
+
+        rd_kafka_AdminFanoutState_t *state =
```
edenhill, line 3110: vardefs at head of scope

src/rdkafka_admin.c
```diff
@@ -2735,3 +2837,400 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                        rd_kafka_buf_parse_fail(
+                                reply,
+                                "Broker returned group %.*s multiple times",
+                                RD_KAFKAP_STR_PR(&kgroup));
+                }
+
+                rd_list_set(&state->responses, orig_pos, terr);
+        }
+
+        if (state->outstanding == 0) {
```
edenhill, line 3159: Same comment as for DelRecords, reverse the if and unscope this part.

edenhill (changes requested):
Solid work here, really impressed by the code quality and sticking with the librdkafka code style!

Left some comments that need addressing, but the main thing is we need to make DeleteRecords asynchronous, I'll add a leader_req() to move this burden from the Admin code.

src/rdkafka.h
```diff
@@ -5592,6 +5627,36 @@ rd_kafka_topic_result_error_string (const rd_kafka_topic_result_t *topicres);
...
+/**
+ * @returns the name of the group for the given group result.
+ * @remark lifetime of the returned string is the same as the \p groupres.
+ *
+ */
+RD_EXPORT const char *
+rd_kafka_group_result_name (const rd_kafka_group_result_t *groupres);
+
+RD_EXPORT rd_kafka_group_result_t *
+rd_kafka_group_result_copy (const rd_kafka_group_result_t *groupres);
```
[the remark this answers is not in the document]
gridaphobe: It's used in `rd_kafka_DeleteGroupsResponse_parse` to copy the accumulated group results into the result rko, I'll add a docstring.

src/rdkafka.h
```diff
@@ -6505,8 +6572,138 @@ rd_kafka_DescribeConfigs_result_resources (
...
+ *
+ * @remark The result event type emitted on the supplied queue is of type
+ *         \c RD_KAFKA_EVENT_DELETERECORDS_RESULT
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_DeleteRecords (rd_kafka_t *rk,
+                        rd_kafka_topic_partition_list_t *offsets,
+                        const rd_kafka_AdminOptions_t *options,
+                        rd_kafka_queue_t *rkqu,
+                        int timeout_ms);
```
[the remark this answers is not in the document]
gridaphobe: This is for the initial lookup of the partition leaders, using `rd_kafka_topic_partition_list_query_leaders` which is synchronous. If we add an async function to query partition leaders, we can make DeleteRecords fully async too.

src/rdkafka_admin.c
gridaphobe: We use an `rd_list` to match the existing `admin_result` interface, but are really just accumulating results into a single `topic_partition_list` at the head of the `rd_list`.

src/rdkafka.h
```diff
@@ -6505,8 +6572,138 @@ rd_kafka_DescribeConfigs_result_resources (
...
+ * @param del_group_cnt Number of elements in \p groups array.
+ * @param options Optional admin options, or NULL for defaults.
+ * @param rkqu Queue to emit result on.
+ *
+ * @remark The result event type emitted on the supplied queue is of type
+ *         \c RD_KAFKA_EVENT_DELETEGROUPS_RESULT
+ */
+RD_EXPORT
+void rd_kafka_DeleteGroups (rd_kafka_t *rk,
+                            rd_kafka_DeleteGroup_t **del_groups,
```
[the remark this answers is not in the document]
gridaphobe: I tried making this const, but it leads to gross warnings about discarding nested qualifiers when we call `rd_kafka_DeleteGroups` in the test suite. It looks like C has an overly simplistic notion of where you can safely insert `const` qualifiers :/

src/rdkafka.h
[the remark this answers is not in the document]
edenhill: But do we need it in the public API?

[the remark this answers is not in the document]
gridaphobe: Oh good point, no I don't think it needs to be public. 👍

src/rdkafka.h
```diff
@@ -5592,6 +5627,36 @@ rd_kafka_topic_result_error_string (const rd_kafka_topic_result_t *topicres);
...
 
+/**
+ * @brief Group result provides per-group operation result information.
+ *
+ */
+
+/**
+ * @returns the error code for the given group result.
+ */
+RD_EXPORT rd_kafka_resp_err_t
```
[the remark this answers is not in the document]
gridaphobe: How would you model `rd_kafka_group_result_t` with the new error type? My inclination would be to have `group_result_t` own an `error_t` and return `const error_t *` from `group_result_error()`, but then it seems we lose the ability to make `group_result_new` be a single-allocation.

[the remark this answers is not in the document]
edenhill: The single-allocation thing is not important (just a convenience), so I'm fine with the group_result_t owning an error_t object.

edenhill:
Let's leave the current blocking functionality as-is and get it merged to master and then work on making it non-blocking from there.
Can you rebase this PR on latest master?

edenhill:
Please address the other comments, rebase, and then we'll get this merged :100:

src/rdkafka.h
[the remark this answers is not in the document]
gridaphobe: I also see that `rd_kafka_CreateTopics`, `rd_kafka_DeleteTopics`, etc. take an array of non-const objects, probably for the same reason.

tests/0081-admin.c
gridaphobe: Aha, testing the timeout path revealed that `rd_kafka_admin_worker` circumvents the fanout logic that the `*Response_parse` functions implement :)

Each fanned-out request that times out enqueues its own timeout error on the replyq, will have to do some extra work to consolidate them into a single response.

gridaphobe: I think the best way to ensure that a single response is always returned on the replyq is to do what I had originally intended and introduce a new type of ADMIN_FANOUT op with its own state machine similar to `rd_kafka_admin_worker`. The idea would be that if we want to fanout a request into N chunks, we create N+1 ops, the N chunks plus one ADMIN_FANOUT op that waits for and accumulates the responses from the chunks. 

The chunks operate like any other admin op and can succeed or fail due to broker errors, timeouts, etc. But instead of sending their responses to the client, they send them to a new queue that the ADMIN_FANOUT op owns.

The ADMIN_FANOUT op listens for response chunks on its queue, assembles them into a single response, and once it has received them all it sends the response back to the client. It can also be responsible for reordering the responses to match the order of the requests and propagating errors from chunks that failed outright (eg due to a timeout) to the specific elements of the request (ie if I try to delete 10 groups, and a chunk of 5 times out, the client response should always contain all 10 groups, with 5 marked as timed out). Note that since each chunk has its own timeout, the ADMIN_FANOUT op doesn't need to worry about timing out itself.

This is looking like a larger change though, so I wanted to run the design by you before getting too far into it. I also have some questions about the finer details.

1. I'm a little fuzzy on how ops are triggered from external sources like broker responses. Ideally, the ADMIN_FANOUT op would be triggered only when one of the chunks enqueues its response, because otherwise it can't have any work to do. But I'm not sure how to arrange that.

2. Even though chunks can fail individually for different reasons, the client response still needs an encompassing error code. How would you go about setting that? Just document that it will always be `NO_ERROR` and that the individual elements need to be checked?

edenhill:
Now with v1.5.2 released we should try to get this into v1.6.0.

What's the current state of this PR?
If ready for review, please rebase on latest master.

src/rdkafka_admin.c
```diff
@@ -480,6 +571,15 @@ static void rd_kafka_admin_common_worker_destroy (rd_kafka_t *rk,
                 rd_kafka_enq_once_destroy(rko->rko_u.admin_request.eonce);
                 rko->rko_u.admin_request.eonce = NULL;
         }
+
+        if (rko->rko_u.admin_request.admin_fanout_parent) {
```
edenhill, line 575: Have you run the Admin API tests with helgrind or tsan enabled? Just to make sure there is no concurrency issues introduced.

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
 
 }
 
+static rd_kafka_op_res_t
+rd_kafka_admin_fanout_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko);
+
+static rd_kafka_op_t *
+rd_kafka_admin_fanout_op_new (rd_kafka_t *rk,
```
edenhill, line 964: Doc string

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+                              rd_kafka_op_type_t req_type,
+                              rd_kafka_event_type_t reply_event_type,
+                              const struct rd_kafka_admin_fanout_worker_cbs* cbs,
```
edenhill, line 967: * ->

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+                              const rd_kafka_AdminOptions_t *options,
+                              rd_kafka_queue_t *rkqu) {
+        rd_kafka_op_t *rko;
+
+        rd_assert(rk);
```
edenhill, line 972: This assert is not necessary, there is very little risk that someone calls this with a NULL rk.  The rkqu is okay though.

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+                rd_kafka_enq_once_new(rko, RD_KAFKA_REPLYQ(rk->rk_ops, 0));
+
+        /* Set up replyq */
+        rd_kafka_set_replyq(&rko->rko_u.admin_fanout.replyq,
+                            rkqu->rkqu_q, 0);
+
+        return rko;
+}
+
+static rd_kafka_op_res_t
```
edenhill, line 1000: doc stirng

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+rd_kafka_admin_fanout_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
+        const char *name = rd_kafka_op2str(rko->rko_type);
+        rd_kafka_event_t* evt;
```
edenhill, line 1003: `* -> `

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+        rd_kafka_op_t* rko_result;
```
edenhill, line 1004: ` * >> 1`

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+        if (rko->rko_err == RD_KAFKA_RESP_ERR__DESTROY)
+                goto destroy; /* rko being destroyed (silent) */
+
+        rd_kafka_dbg(rk, ADMIN, name,
+                     "%s worker called with %"PRIusz" requests outstanding: %s",
+                     name,
+                     rko->rko_u.admin_fanout.outstanding,
+                     rd_kafka_err2str(rko->rko_err));
+
+        rd_assert(thrd_is_current(rko->rko_rk->rk_thread));
```
edenhill, line 1025: You can use `rk` here

src/rdkafka_admin.c
```diff
@@ -839,6 +957,114 @@ rd_kafka_admin_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko) {
...
+        for (i = 0; i < rd_list_cnt(&rko->rko_u.admin_fanout.responses); i++) {
+                const void *elem = rd_list_elem(&rko->rko_u.admin_fanout.responses, i);
+                rd_list_add(&rko_result->rko_u.admin_result.results,
+                            rko->rko_u.admin_fanout.cbs->copy_result(elem));
+        }
+
+        /* Enqueue result on application queue, we're done. */
+        rd_kafka_replyq_enq(&rko->rko_u.admin_fanout.replyq, rko_result,
+                            rko->rko_u.admin_fanout.replyq.version);
+
```
edenhill, line 1062: add `/*FALLTHRU*/` comment to make it clear that we do indeed want to fall thru to the destroy and we didnt just miss a return.

src/rdkafka_admin.c
```diff
@@ -2737,3 +2964,474 @@ rd_kafka_DescribeConfigs_result_resources (
...
+/**
+ * @name Delete Records
+ * @{
+ *
+ *
+ *
+ *
+ */
+
+static void rd_kafka_topic_partition_list_free (void *ptr) {
```
edenhill, line 2977: There's a rd_kafka_topic_partition_list_destroy_free() already that you can use.

src/rdkafka_admin.c
```diff
@@ -2737,3 +2964,474 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                            rd_kafka_topic_partition_list_copy(leader->partitions));
+                rd_kafka_q_enq(rk->rk_ops, rko);
+        }
+
+        rd_list_destroy(&leaders);
+}
+
+
+const rd_kafka_topic_partition_list_t *
+rd_kafka_DeleteRecords_result_offsets (
```
edenhill, line 3148: docstring

src/rdkafka_admin.c
```diff
@@ -2737,3 +2964,474 @@ rd_kafka_DescribeConfigs_result_resources (
...
+                int orig_pos;
+                rd_kafka_error_t *error;
+
+                rd_kafka_buf_read_str(reply, &kgroup);
+                rd_kafka_buf_read_i16(reply, &error_code);
+
+                if (error_code) {
+                        error = rd_kafka_error_new(error_code, "%s", rd_kafka_err2str(error_code));
+                } else {
+                        error = rd_kafka_error_new(error_code, NULL);
```
edenhill, line 3269: Don't think we should create an error object unless there was an error.

src/rdkafka_admin.c
```diff
@@ -2737,3 +2964,474 @@ rd_kafka_DescribeConfigs_result_resources (
...
+        *rko_resultp = rko_result;
+        return RD_KAFKA_RESP_ERR_NO_ERROR;
+
+err_parse:
+        if (rko_result)
+                rd_kafka_op_destroy(rko_result);
+
+        rd_snprintf(errstr, errstr_size,
+                    "DeleteGroups response protocol parse failure: %s",
+                    rd_kafka_err2str(err));
```
edenhill, line 3312: reply->rkbuf_err

src/rdkafka_admin.c
```diff
@@ -2737,3 +2964,474 @@ rd_kafka_DescribeConfigs_result_resources (
...
+
+        return err;
+}
+
+/** @brief Merge the DeleteGroups response from a single broker
+ *         into the user response list.
+ */
+void rd_kafka_DeleteGroups_merge (rd_kafka_op_t* rko_req,
```
edenhill, line 3320: .._response_merge (..)  would read even better

edenhill (changes requested):
Magnificient!
Left a couple or 43 comments that need some sort of addressing, but they're all minor. This is looking good :+1:.

src/rdkafka.h
```diff
@@ -6082,8 +6130,8 @@ rd_kafka_CreateTopics (rd_kafka_t *rk,
                        rd_kafka_queue_t *rkqu);
 
 
-/**
- * @brief CreateTopics result type and methods
+/*
+ * CreateTopics result type and methods
```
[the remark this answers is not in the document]
gridaphobe: It's explained in the last commit https://github.com/edenhill/librdkafka/pull/2701/commits/77f0b0c7a6e9495972704ed5e51472e29187c69e. I noticed that the Admin API docs were rendering strangely. It turns out that doxygen will collapse consecutive doc comments into a single item, and then apply the formatting. This results in comments like

```
/**
 * @section CreateTopics Header
 */

/**
 * @brief creates topics
 *
 * longer description
 */
rd_kafka_CreateTopics()
```

rendering like

```
rd_kafka_CreateTopics()
creates topics

CreateTopics Header

longer description
```

Happy to revert that commit if you want to sort out the doc rendering separately.

src/rdkafka.h
```diff
@@ -6648,8 +6696,139 @@ rd_kafka_DescribeConfigs_result_resources (
...
+ * @param rk Client instance.
+ * @param offsets Offsets to delete up to (exclusive).
+ * @param options Optional admin options, or NULL for defaults.
+ * @param rkqu Queue to emit result on.
+ *
+ * Supported admin options:
+ *  - rd_kafka_AdminOptions_set_operation_timeout() - default 0
+ *  - rd_kafka_AdminOptions_set_request_timeout() - default socket.timeout.ms
+ *
+ * @remark The call may block for up to
```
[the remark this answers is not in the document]
gridaphobe: Yes, it will still block for the duration of `rd_kafka_topic_partition_list_query_leaders`. You mentioned earlier that you were going to implement an async version of that function we could call, but that we could merge this PR in the meantime. Is that API available already?

src/rdkafka_request.c
```diff
@@ -3614,6 +3621,71 @@ rd_kafka_DeleteTopicsRequest (rd_kafka_broker_t *rkb,
...
+ * @remark The rd_kafka_topic_partition_list_t in \p offsets_list must already
+ *          be sorted.
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR if the request was enqueued for
+ *          transmission, otherwise an error code and errstr will be
+ *          updated with a human readable error string.
+ */
+rd_kafka_resp_err_t
+rd_kafka_DeleteRecordsRequest (rd_kafka_broker_t *rkb,
+                               const rd_list_t *offsets_list /*(rd_kafka_topic_partition_list_t*)*/,
```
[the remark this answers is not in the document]
gridaphobe: This function is used as an admin worker cb, so its type has to match `rd_kafka_admin_Request_cb_t`.

tests/0080-admin_ut.c
```diff
@@ -367,15 +367,227 @@ static void do_test_DeleteTopics (const char *what,
...
+                    err ? errstr2 : "n/a");
+
+        /* Extract groups, should return MY_DEL_GROUPS_CNT groups. */
+        resgroups = rd_kafka_DeleteGroups_result_groups(res, &resgroup_cnt);
+        TEST_ASSERT(resgroups && resgroup_cnt == MY_DEL_GROUPS_CNT,
+                    "expected %d result_groups, got %p cnt %"PRIusz,
+                    MY_DEL_GROUPS_CNT, resgroups, resgroup_cnt);
+
+        for (i = 0 ; i < MY_DEL_GROUPS_CNT ; i++) {
+                TEST_SAY(_C_MAG "[ group %d = %s ]\n", i, group_names[i]);
```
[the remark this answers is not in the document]
gridaphobe: Oops, those were for my own personal debugging. It took me a while to realize that you have to `rd_strdup` the result of `test_mk_topic_name()` if you want multiple distinct names :)

src/rdkafka_admin.c
gridaphobe: This is the same pattern as the other `*Response_parse` functions, should I update them all?

src/rdkafka_op.h
```diff
@@ -413,14 +417,63 @@ struct rd_kafka_op_s {
...
+
+                        /** For collecting incremental responses from
+                         *  admin requests that fan out to multiple brokers. */
+                        rd_list_t responses;
+
+                        /** Number of outstanding requests remaining to wait for. */
+                        size_t outstanding;
+
+                        /** Internal queue for incremental responses. */
+                        rd_kafka_queue_t* responseq;
```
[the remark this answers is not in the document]
gridaphobe: In this case, the queue needs to be passed to the fanned out requests, which all expect a `queue_t` as part of the usual admin_request_op machinery. So it's probably best to just keep a `queue_t` here. Will fix the pointer formatting though :)

src/rdkafka_request.c
```diff
@@ -3614,6 +3621,71 @@ rd_kafka_DeleteTopicsRequest (rd_kafka_broker_t *rkb,
...
+        }
+
+        rkbuf = rd_kafka_buf_new_request(rkb, RD_KAFKAP_DeleteRecords, 1,
+                                         4 +
+                                         (partitions->cnt * 100) +
+                                         4);
+
+        rd_kafka_buf_write_topic_partitions(rkbuf, partitions,
+                                            rd_false /*don't skip invalid offsets*/,
+                                            rd_false /*don't write offsets*/,
```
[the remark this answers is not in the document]
gridaphobe: Yes! Good catch! This is actually pretty concerning, the DeleteRecords tests were triggering a `TEST_FAIL_LATER` assertion as a result, but somehow the test harness still reported success for the whole `0081_admin` suite.. Not sure what's going on here..

src/rdkafka.h
```diff
@@ -6648,8 +6696,139 @@ rd_kafka_DescribeConfigs_result_resources (
...
+ */
+
+/**
+ * @brief Delete records in topics and partitions as specified by
+ *        the \p offsets
+ *
+ * \p offsets should contain \c topic, \c partition, and \c offset
+ * \c offset is the offset before which the messages will be deleted.
+ * Set \c offset to RD_KAFKA_OFFSET_END (high-watermark) in order to
+ * delete all data in the partition.
```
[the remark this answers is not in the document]
gridaphobe: It appears to be the actual high-watermark according to https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/cluster/Partition.scala#L1201-L1202

src/rdkafka_request.c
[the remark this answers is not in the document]
edenhill: You should use TEST_LATER_CHECK() once you are done to act on any previously accumualated later test failures.

src/rdkafka_admin.c
gridaphobe: Yep, I've been running with tsan on my mac and it seems happy.

src/rdkafka.h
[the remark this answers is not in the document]
edenhill: Thanks! Keep it!

src/rdkafka.h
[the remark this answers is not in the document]
edenhill: Ah, I see, forgot about that. Let me add a PR for that.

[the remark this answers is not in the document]
gridaphobe: @edenhill any progress on the async leader_req API?

edenhill:
I'm adding the async leader lookups to your branch, hope that's ok.

gridaphobe:
> I'm adding the async leader lookups to your branch, hope that's ok.

Sounds good, thanks!!

edenhill:
If you shake the tree....

Found a bunch of AdminAPI and rko pitfalls and issues that was unearthed by the new broker target types, fixed them, made some minor API adjustments for DeleteRecords, added DeleteConsumerGroupOffsets while I was at it.

Eric, could you try this out in your environment and see that the new Admin APIs are to your liking.

I had to make the DeleteRecords a bit more complicated by adding a DeleteRecords_t* type, this is so that we can extend the calling convention without breaking the API in case the protocol request requires extra parameters in the future.

gridaphobe:
Awesome! The addition of the DeleteRecords_t type seems very sensible. I can't imagine that will cause us any problems, but I'll test the new branch later today or early next week to be sure.

gridaphobe:
@edenhill the API changes were easy to adapt to and everything looks good on our end!

edenhill:
Big thanks for this contribution, @gridaphobe ! :100:

PR #2707, opened 2020-02-10
KIP-511: client.software.name and client.software.version

edenhill:
The JMX metrics read-out in test 0108 was verified with AK trunk as of now.

### DIFF
--- a/INTRODUCTION.md
+++ b/INTRODUCTION.md
@@ -86,6 +86,12 @@ librdkafka also provides a native C++ interface.
             - [Detailed description](#detailed-description)
         - [Supported KIPs](#supported-kips)
         - [Supported protocol versions](#supported-protocol-versions)
+- [Recommendations for language binding developers](#recommendations-for-language-binding-developers)
+    - [Expose the configuration interface pass-thru](#expose-the-configuration-interface-pass-thru)
+    - [Error constants](#error-constants)
+    - [Reporting client software name and version to broker](#reporting-client-software-name-and-version-to-broker)
+    - [Documentation reuse](#documentation-reuse)
+    - [Community support](#community-support)
 
 <!-- markdown-toc end -->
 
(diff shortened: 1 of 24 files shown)


INTRODUCTION.md
```diff
@@ -1858,3 +1867,72 @@ release of librdkafka.
...
+
+If your language binding/applications allows configuration properties to be set
+in a pass-through fashion without any pre-checking done by your binding code it
+means that a simple upgrade of the underlying librdkafka library (but not your
+bindings) will provide new features to the user.
+
+## Error constants
+
+The error constants, both the official (value >= 0) errors as well as the
+internal (value < 0) evolve constantly.
```
mhowlett: just in case i don't pick up on anything else in this PR:
"The error constants, both official (value >= 0) and internal (value < 0), evolve constantly."

edenhill: This, is the first time, I've under-punctualized.

INTRODUCTION.md
```diff
@@ -1858,3 +1867,72 @@ release of librdkafka.
...
+## Reporting client software name and version to broker
+
+[KIP-511](https://cwiki.apache.org/confluence/display/KAFKA/KIP-511%3A+Collect+and+Expose+Client%27s+Name+and+Version+in+the+Brokers) introduces a means for a
+Kafka client to report its implementation name and version to the broker, the
+broker then exposes this as metrics (e.g., through JMX) to help Kafka operators
+troubleshoot problematic clients, understand the impact of broker and client
+upgrades, etc.
+This requires broker version 2.4.0 or later (metrics added in 2.5.0).
+
+librdkafka will send its name (`librdkafka`) and version (e.g., `v1.3.0`)
```
mhowlett: "By default, ..."

src/rdkafka_conf.c
```diff
@@ -253,16 +253,41 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
 	  "Client identifier.",
 	  .sdef =  "rdkafka" },
+        { _RK_GLOBAL|_RK_HIDDEN, "client.software.name", _RK_C_STR,
+          _RK(sw_name),
+          "Client software name as reported to broker version >= v2.4.0. "
+          "Broker-side character restrictions apply, as of broker version "
+          "v2.4.0 the allowed characters are `a-zA-Z0-9.-`. The local client "
+          "will replace any other character with `-` and strip leading and "
+          "trailing non-alphanumeric characters before tranmission to "
+          "the broker. "
```
mhowlett: yeah, i guess that is a good idea. on the one hand it seems like overkill and trying too hard. on the other, on the off-chance a deviant client is let out in the wild, you probably want to do as good as you can with what you have.

mhowlett: it may be better to catastrophically fail on startup, but minor point either way.

src/rdkafka_request.c
```diff
@@ -1892,25 +1894,26 @@ rd_kafka_handle_ApiVersion (rd_kafka_t *rk,
 			    rd_kafka_buf_t *request,
 			    struct rd_kafka_ApiVersion **apis,
 			    size_t *api_cnt) {
-        const int log_decode_errors = LOG_ERR;
+        const int log_decode_errors = LOG_DEBUG;
```
mhowlett: `LOG_DEBUG` here seems like something you probably didn't mean to leave in.

INTRODUCTION.md
edenhill: Initially I had that in, but to me it implied that it could be disabled, which it cant.

src/rdkafka_conf.c
edenhill: The reason it is not a hard error is that it would make it error prone to pass auto-generated version strings, such as from git tags

PR #2714, opened 2020-02-13
Added SSL chapter to manual

edenhill:


INTRODUCTION.md
```diff
@@ -1132,6 +1133,37 @@ A DNS record containing all broker address can thus be used to provide a
...
+   `/usr/local/etc/openssl` on Mac OSX (Homebrew).
+ * On Windows the Root certificate store is used.
+
+If the system-provided default CA root certificates are not sufficient to
+verify the broker's certificate, such as when a self-signed certificate
+or a local CA authority is used, the CA certificate must be specified
+explicitly so that the client can find it.
+This can be done either by providing a PEM file (e.g., `cacert.pem`)
+as the `ssl.ca.location` configuration property, or by passing an in-memory
+PEM, X.509/DER or PKCS#12 certificate to `rd_kafka_conf_set_ssl_cert()`.
```
rnpridgeon: nit: they are DER/PEM encoded X.509 certificates and PKCS#12 is an archive file format primarily used to bind the private key with the X/509 certificate. 

It may be nice to mention common extensions as well; .crt, .cer, .p12, .pfx

I think it's worth mentioning mtls as well  as this is where the private key comes into play and where pkcs#12 shines.

edenhill: Thanks for the input, but I'm not comfortable to write that out, feel free to improve this in a future PR :+1:

PR #2727, opened 2020-02-20
Fully support transactions in rd_kafka_produce_batch()

mekleo:
When messages were manually assigned to partitions (i.e. no partitioner invocation), rd_kafka_produce_batch() did not include them in the transaction. Consequently, rd_kafka_commit_transaction() failed to flush all outstanding messages.


edenhill:
Thanks for identifying this mistake!

I'm thinking we might want to move this logic to toppar_enq_msg() instead, need to give it a think-thru.

edenhill:
Merged to master manually. Thank you for your contribution!

PR #2728, opened 2020-02-21
Fixes for the transactional producer, et.al.

edenhill:
Issues that surfaced during development of the Go client transactions.


src-cpp/rdkafkacpp.h
```diff
@@ -2228,6 +2228,17 @@ class RD_EXPORT Queue {
 
 /**@}*/
 
+/**
+ * @name ConsumerGroupMetadata holds a consumer instance's group metadata state.
+ * @{
+ *
+ */
+class RD_EXPORT ConsumerGroupMetadata {
```
mhowlett: ok. i believe this is for binary api compatibility, very good.

src-cpp/rdkafkacpp.h
```diff
@@ -523,6 +523,78 @@ class KafkaConsumer;
...
+  * @returns true if the operation may be retried, else false.
+  */
+ virtual bool is_retriable () const = 0;
+
+ /**
+  * @returns true if the error is an abortable transaction error in which case
+  *          the application may call RdKafka::Producer::abort_transaction()
+  *          and start a new transaction with
+  *          RdKafka::Producer::begin_transaction().
+  *          Else returns false.
```
mhowlett: but what is the application developer expected to do if `false`?

src/rdkafka_error.c
```diff
@@ -0,0 +1,190 @@
...
+const char *rd_kafka_error_name (const rd_kafka_error_t *error) {
+        return rd_kafka_err2name(error->code);
+}
+
+const char *rd_kafka_error_string (const rd_kafka_error_t *error) {
+        return error->errstr ? error->errstr : rd_kafka_err2str(error->code);
+}
+
+int rd_kafka_error_is_fatal (const rd_kafka_error_t *error) {
+        return error->fatal ? 1 : 0;
```
mhowlett: why the ternary operator?  x3

edenhill: Avoids implicit cast warnings on various compilers.

src-cpp/rdkafkacpp.h
edenhill: That is a very good question.
For some error types it indicates a bug in the app (e.g invalid arg), for others it might be a broker error code that we don't explicitly handle. 

What should the recommendation be? Always treat as fatal?

src-cpp/rdkafkacpp.h
```diff
@@ -523,6 +523,78 @@ class KafkaConsumer;
...
+  *
+  * @remark The return value of this method is only valid for errors returned
+  *         by the transactional API.
```
mhowlett: unless the result can be spuriously true, i think pedantically this remark is technically incorrect. i'd omit it, or maybe the change word 'valid' to 'makes sense'.

mhowlett: also, agree that having a series bools is a more ergonomic api than an enum, even if they're mutually exclusive.

also, agree that it's a good idea here on balance to have unnecessary methods exposed in places where they have no relevance.

src-cpp/rdkafkacpp.h
mhowlett: i don't know that's why i'm looking for guidance :-). maybe document the possible causes, and suggest that treating as fatal is a safe thing to do?

src-cpp/rdkafkacpp.h
```diff
@@ -2927,50 +3031,150 @@ class RD_EXPORT Producer : public virtual Handle {
...
+
 
   /**
-   * FIXME blocking?
+   * @brief init_transactions() must have been called successfully
+   *        (once) before this function is called.
+   *
+   * @returns an RdKafka::Error object on error, or NULL on success.
+   *          Check whether a fatal error has been raised by calling
+   *          RdKafka::Error::is_fatal_error().
```
mhowlett: Some indication of why a fatal error could occur would be very helpful here i think (pointer to, or brief summary).

mhowlett: i felt like i wanted this here, but not in the other methods for some reason. i think because here we're not already in a transaction.

src-cpp/rdkafkacpp.h
```diff
@@ -2927,50 +3031,150 @@ class RD_EXPORT Producer : public virtual Handle {
...
+   * @param timeout_ms The maximum time to block. On timeout the operation
+   *                   may continue in the background, depending on state,
+   *                   and it is okay to call this function again.
    *
-   * FIXME blocking?
+   * @returns an RdKafka::Error object on error, or NULL on success.
+   *          Check whether the returned error object permits retrying
+   *          by calling RdKafka::Error::is_retriable(), or whether an abortable
+   *          or fatal error has been raised by calling
+   *          RdKafka::Error::is_txn_abortable() or RdKafka::Error::is_fatal()
```
mhowlett: what does it mean to have `is_txn_abortable` set on an error returned by `rd_kafka_abort_transaction`? if this can indeed be set here, I think it probably needs some additional commentary.

mhowlett:
LGTM still (with some minor commentary)

src-cpp/rdkafkacpp.h
rnpridgeon: +1 for always fatal

src-cpp/rdkafkacpp.h
edenhill: Retriable, abortable and fatal errors are explained in rdkafka.h (to which we point from rdkafkacpp.h, to avoid duplication).

src-cpp/rdkafkacpp.h
edenhill: You're right, abort_transaction() should not return an txn_abortable() error, but rather retriable().

src-cpp/rdkafkacpp.h
edenhill: I updated the transactional API intro documentation

edenhill: I updated the transactional API documentation to include a recommendation how to handle errors that have neither retriable, abortable or fatal flag set, namely by treating them as fatal errors.

I'm hesitant to flag all these errors as fatal by default since the fatal errors are currently well defined in KIP-98 and the Java API .

src-cpp/rdkafkacpp.h
edenhill: I'm not sure I agree on the wording, "makes sense" is a bit  too informal for the current tone of the docs, and the return value is invalid, should not be used, for errors from non-transactional APIs.

PR #2744, opened 2020-03-04
Added consumer_group_metadata serdes to avoid bindings to hang on to object

edenhill:


src/rdkafka_cgrp.c
```diff
@@ -3657,3 +3661,137 @@ rd_kafka_consumer_group_metadata_dup (
 }
 
 
+/*
+ * Consumer group metadata serialization format v1:
+ *  "CGMDv1:"<group_id>"\0"
+ * Where <group_id> is the group_id string.
+ */
+static const char rd_kafka_consumer_group_metadata_magic[7] = "CGMDv1:";
```
mhowlett: not leaving anything to chance

mhowlett:
LGTM. the magic thing seems a bit too paranoid, but i guess that fits with the general theme of the PR.

thanks!

edenhill:
Paranoid? We need to identify the input as valid somehow.

mhowlett:
> Paranoid? We need to identify the input as valid somehow.

it's externally visible, but effectively an internal detail, is where i was coming from

edenhill:
The magic is defined in an internal source file, not in the public API.

PR #2749, opened 2020-03-09
Txn producer must bump epoch through broker, not locally.

edenhill:
Bump the epoch locally, which is okay for the idempotent producer, is not allowed for the transactional producer since the broker needs to track state of each pid+epoch.
However, this is not yet enforced by the broker which leads to inconsistent state and transactional.id states that go into a blocked unusable state.


src/rdkafka_buf.h
```diff
@@ -164,9 +164,12 @@ rd_tmpabuf_write_str0 (const char *func, int line,
 			rd_kafka_assert(NULL, rkbuf->rkbuf_rkb);	\
                         rd_rkb_log(rkbuf->rkbuf_rkb, log_decode_errors, \
                                    "PROTOERR",                          \
-                                   "Protocol parse failure "            \
+                                   "Protocol parse failure for %s v%hd " \
```
rnpridgeon: 👍

PR #2753, opened 2020-03-10
Add FNV-1a partitioner (#2724)

### LINKED ISSUE #2724: Support FNV-1a hash for partitioning (for Sarama compatibility)
This issue is more of a feature request derived from this confluent-kafka-go issue: https://github.com/confluentinc/confluent-kafka-go/issues/16#issuecomment-585678319
Raising here for awareness, I'll prob have a go at a PR to add this.
Please let me know if this needs to be in a specific format!

Description
===========
Sarama uses the FNV-1a hashing algorithm as its default partitioner. This means that anyone migrating from Sarama to confluent-kafka-go will either have to implement the partitioner themselves (making an extra metadata round trip call to get number of partitions) or change algorithms to one of the librdkafka supported algorithms (potentially causing short term out of order processing by downstream consumers).
I propose adding the FNV-1a hash as a supported partitioner in librdkafka, similar to how Murmur2 was added for Java compatibility.

Manicben:
Adds a new partitioner using the FNV-1a hashing algorithm, with some
tweaks to match Sarama's default hashing partitioner behaviour.
Main use case is for users switching from Sarama to librdkafka
(or confluent-kafka-go) and wanting to maintain ordering guarantees.

Test hashes were generated in the Go Playground using Go's `hash/fnv` lib.

Resolves #2724.

Didn't know whether it was OK to mention Sarama at all. Switching from it is the main reason for this, and the absolute value tweak to the hashing is required to be compatible with it.

Also, apologies for the whitespace changes, my IDE went ahead and trimmed all extra whitespace.


src/rdfnv1a.c
```diff
@@ -0,0 +1,110 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2012-2015, Magnus Edenhill
```
edenhill: 2020

src/rdfnv1a.c
```diff
@@ -0,0 +1,110 @@
...
+uint32_t rd_fnv1a (const void *key, size_t len) {
+        const uint32_t prime  = 0x01000193; // 16777619
+        const uint32_t offset = 0x811C9DC5; // 2166136261
+        size_t i;
+        int32_t h = offset;
+
+        const unsigned char *data = (const unsigned char *)key;
+
+        for (i = 0; i < len; i++) {
+            h ^= data[i];
```
edenhill: code-style is 8-space indent for scopes.

src/rdkafka_conf.c
```diff
@@ -1288,7 +1290,9 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           "`consistent` - CRC32 hash of key (Empty and NULL keys are mapped to single partition), "
           "`consistent_random` - CRC32 hash of key (Empty and NULL keys are randomly partitioned), "
           "`murmur2` - Java Producer compatible Murmur2 hash of key (NULL keys are mapped to single partition), "
-          "`murmur2_random` - Java Producer compatible Murmur2 hash of key (NULL keys are randomly partitioned. This is functionally equivalent to the default partitioner in the Java Producer.).",
+          "`murmur2_random` - Java Producer compatible Murmur2 hash of key (NULL keys are randomly partitioned. This is functionally equivalent to the default partitioner in the Java Producer.), "
```
edenhill: Break source line < 80 columns.

CONFIGURATION.md
```diff
@@ -1,7 +1,7 @@
 # Configuration properties
 ## Global configuration properties
 
-Property                                 | C/P | Range           |       Default | Importance | Description              
+Property                                 | C/P | Range           |       Default | Importance | Description
```
edenhill: Please avoid white-space diffs.

src/rdkafka.h
```diff
@@ -1331,7 +1331,7 @@ void rd_kafka_message_destroy(rd_kafka_message_t *rkmessage);
  * @remark This function MUST NOT be used with the producer.
  */
 static RD_INLINE const char *
-RD_UNUSED 
+RD_UNUSED
```
edenhill: Please void white-space diffs. (throughout)

CONFIGURATION.md
```diff
@@ -145,7 +145,7 @@ message.timeout.ms                       |  P  | 0 .. 2147483647 |        300000
...
+partitioner                              |  P  |                 | consistent_random | high       | Partitioner: `random` - random distribution, `consistent` - CRC32 hash of key (Empty and NULL keys are mapped to single partition), `consistent_random` - CRC32 hash of key (Empty and NULL keys are randomly partitioned), `murmur2` - Java Producer compatible Murmur2 hash of key (NULL keys are mapped to single partition), `murmur2_random` - Java Producer compatible Murmur2 hash of key (NULL keys are randomly partitioned. This is functionally equivalent to the default partitioner in the Java Producer.), `fnv1a` - FNV-1a hash of key (NULL keys are mapped to single partition), `fnv1a_random` - FNV-1a hash of key (NULL keys are randomly partitioned). <br>*Type: string*
```
Manicben: @edenhill Do you think it's worth mentioning here that `fnv1a_random` is `functionally equivalent to the default partitioner in the Sarama Producer` or along those lines?

edenhill: I think it is sufficient to just have the fnv1a name.

src/rdfnv1a.c
```diff
@@ -0,0 +1,110 @@
...
+
+        const unsigned char *data = (const unsigned char *)key;
+
+        for (i = 0; i < len; i++) {
+                h ^= data[i];
+                h *= prime;
+        }
+
+        /* Take absolute value to match the Sarama implementation */
+        if (h < 0) {
```
edenhill: Looking at https://github.com/Shopify/sarama/blob/master/partitioner.go#L203 it seems like this is Sarama's legacy approach, and that future versions will use the bitmasked Java-compatible approach instead.

I'm not sure which way to go here, but looking at Sarama's examples it seems like the legacy version is what is used.

Make a note in the comment above on line 38+ that this implementation is compatible with NewHashPartitioner and not NewReferenceHashPartitioner in sarama.

Manicben:
Would this by any chance make it into 1.4.0? Would be amazing if it did, but I understand that it may be too late.

edenhill:
It is an isolated feature that has good testing, so the plan is to include it in v1.4.0.

edenhill:
Thank you for a perfect PR!

PR #2754, opened 2020-03-10
Correct statistics names in docs

TimWSpence:
This isn't necessarily the correct solution - the current format is inconsistent, using `rxbytes` for brokers and partitions but `rx_bytes` at the top-level. But at the very least this will keep the documentation in line with the behaviour


edenhill:
Thank you!

TimWSpence:
You're welcome! :)

PR #2761, opened 2020-03-12
Re-enable static member test, and try to make KIP-511 test more reliable

edenhill:


src/rdkafka_sasl_scram.c
```diff
@@ -934,7 +934,7 @@ static int unittest_scram_nonce (void) {
         rd_kafka_sasl_scram_generate_nonce(&out2);
         RD_UT_ASSERT(out1.size == 32, "Wrong size %d", (int)out2.size);
 
-        RD_UT_ASSERT(memcmp(out1.ptr, out2.ptr, out1.size == 32) != 0,
```
mhowlett: you should give a more strongly typed language a go. can give a recommendation if you want ;-).

tests/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ set(
     0012-produce_consume.c
     0013-null-msgs.c
     0014-reconsume-191.c
+    0016-client_swname.c
```
mhowlett: tip: if you use your toes as well as your fingers, you can count up this high.

mhowlett: (buf if you're too lazy too, 16 comes after 15)

src/rdkafka_sasl_scram.c
edenhill: hehe, you said "go".

Newer clang actually warn about this :+1:

tests/CMakeLists.txt
edenhill: damnit

PR #2763, opened 2020-03-12
Fix lintian-detected misspellings

cbarcenas:
Just a minor fixup :) These misspellings generate [warnings](https://lintian.debian.org/tags/spelling-error-in-binary.html) in `lintian` (Debian's package linter).

```
I: librdkafka1: spelling-error-in-binary usr/lib/x86_64-linux-gnu/librdkafka.so.1 offets offsets
I: librdkafka1: spelling-error-in-binary usr/lib/x86_64-linux-gnu/librdkafka.so.1 succesful successful
```


edenhill:
Thank you!

PR #2766, opened 2020-03-14
Fix lintian-detected misspellings

litao3rd:
Just a minor fixup :) These misspellings generate warnings [1]
in lintian (Debian's package linter).

[1] https://lintian.debian.org/tags/spelling-error-in-binary.html


edenhill:
Thank you!

PR #2778, opened 2020-03-23
Remove deprecated compression.type

jakesyl:
Screen Shot 2020-03-23 at 12 15 22 PM


edenhill:
`compression.type` works fine and must not be removed (since it would break existing applications):
```
 examples/rdkafka_example -b $BROKERS -P -t test -X compression.type=lz4
% Type stuff and hit enter to send
foo
% Sent 3 bytes to topic test partition -1
% Message delivered (3 bytes, offset 0, partition 0): foo
```

PR #2792, opened 2020-04-01
configure: OpenSSL URLs changed, and propagate http error code when c…

edenhill:
…url fails


mklove/modules/configure.zlib
```diff
@@ -38,7 +38,7 @@ function install_source {
 
     echo "### Installing $name $ver from source to $destdir"
     if [[ ! -f Makefile ]]; then
-        curl -sL https://zlib.net/zlib-${ver}.tar.gz | \
+        curl -fL https://zlib.net/zlib-${ver}.tar.gz | \
```
rnpridgeon, line 41: :+1:

Issue #2284, opened 2019-04-18
Feature request: Support pluggable partition assignment algorithms for consumer


jeffwidman:
Please add support for pluggable partition assignment algorithms for the consumers.

 I was in a design review today where we will need this for a handful of upcoming services.

This was originally requested back in https://github.com/edenhill/librdkafka/issues/1135.

Using this ticket to track progress on this feature.

chris-zen:
I would like to implement a service like [Kafka Streams Interactive Queries](https://docs.confluent.io/current/streams/developer-guide/interactive-queries.html) on top of `librdkafka`. After looking at the source code of both projects, the only critical part that I miss in `librdkafka` is the ability to provide a `custom` assignor to be able to manage [standby replicas](https://tech.transferwise.com/achieving-high-availability-with-kafka-streams/#standbyreplicas) of local states.

According to @edenhill [here](https://github.com/edenhill/librdkafka/issues/1622#issuecomment-355529302), [here](https://github.com/edenhill/librdkafka/issues/1135#issuecomment-288488181) and the code [here](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_assignor.c#L420), it is possible to configure `custom` assignors internally, but it is not exposed at the API level, and only the builtin `range` and `roundrobin` assignors [are allowed](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_assignor.c#L513).

I was wondering if there is any plan to implement it, how difficult would it be, or what are the chances to get mentoring for contributing.

chris-zen:
Some more context on why is so relevant to be able to customise the assignor. This is the case for Kafka Streams, where there are many consumers with state, which makes rebalances quite expensive when not done properly. Currently Kafka Streams uses one assignor called Sticky that minimises state migration between instances (and also accounts for standby replicas).

There is a recent [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol) to make the protocol more incremental, that gives an interesting view on the current state and the relevance of the matter.

edenhill:
Let's look at making the assignor API public when we implement KIP-429 (et.al) to make sure all use-cases are covered.

Buttered:
Is there any update on this feature?

rhoban13:
[...]

PR #2800, opened 2020-04-05
Simplify group assignor init / config

mhowlett:
some things here seem unnecessarily complicated / general. am i missing anything important?

haven't tested this - if it looks right to you, will do that.


edenhill:
The group protocol is generic and not specific to consumers (see kafka connect), the code reflects this and is prepared for supporting future protocol types as well as user-pluggable ones, which might have been a premature generalization, but I'm not sure what the benefit is of removing this generalization?

Also, make sure the tests pass locally, there are compilation warnings and test failures due to this change.

mhowlett:
> The group protocol is generic and not specific to consumers (see kafka connect), the code reflects this and is prepared for supporting future protocol types as well as user-pluggable ones, which might have been a premature generalization, but I'm not sure what the benefit is of removing this generalization?

yep - reverted this. The main aspiration of this PR was to remove `rkas_enabled` and related logic (thoughts on that?), which caused me a not insignificant amount of wasted time working out why this code was the way it was. I then got carried away depreciating the config property. you're right, removing it won't meaningfully help anyone.

the only way assignors can be set (afaict) is via this config property, and they're always enabled, so I think the stuff i've removed should just be removed.

apologies for opening quickly - wanted your high level thoughts quickly and couldn't be bothered doing any more yesterday. I've now fixed the warnings and bugs and run the integration tests.

mhowlett:
[note: deleted previous comment that demonstrated i didn't read yours]

yeah, agree that having the group.protocol.type config isn't a big deal and it's better not to remove it. I added some clarification to the docs in the current state of this PR instead of removing it.

the user-pluggable related code caused me some amount of wasted time, so i still think it's an improvement to either remove it made clearer (current PR state removes it).

edenhill:
I don't really understand the reasoning behind these changes.

The code is prepared for (but does not yet expose) user-pluggable assignors, I do appreciate that unused abstractions can be removed to reduce complexity, but that should not be a reason in itself to change existing code that does not have issues, since any code change risks introducing bugs.
Also, the suggested patch breaks abstraction of an isolated part of the code (assignors_init()..).

I'd like to understand the reasoning for this change to understand why it is deemed warranted, what does it enable?

edenhill:
See #2284

mhowlett:
My reasoning was just reduce complexity in something that caused me to waste some time + make it clear in the user facing config that a property was currently pointless. Thanks for the context, makes sense!

Issue #2804, opened 2020-04-07
Pass debug context to log callback


wolfchimneyrock:
Description
===========
We would like to be able to control the verbosity level per debug contexts on the fly in our logging infrastructure.  This would require the ability to pass the context into the log callback, which currently only takes verbosity `level` as an integer and sub-context `fac` as a c-string.

I can think of a couple of ways to do this while maintaining backwards compatibility:

1. keep existing behavior of `log_cb` by default,  and add an additional config `log.callback.expects.context` (=false by default).  If the new config is set to true, then the context bitwise combined with the level and sent in the level parameter.  something like `(ctx << 3) | level`.
2. add a new callback definition `log_cb2` which has a parameter `debug_context` which is a c-string containing the concatenation of context(s).

Method one would be easier but would leak the internal enum values of the debug contexts, whereas method two would add additional methods to set the callback, etc.

If this sounds like something reasonable (or some other way to do it) I would be happy to submit a PR.

PR #2833, opened 2020-04-17
add rd_kafka_event_debug_contexts() for log events

wolfchimneyrock:
this additional callback will allow the client to dynamically
control which debug contexts are logged when using debug=all.

relates to #2804 

create the config `log_detailed_cb` which is set using new function `rd_kafka_conf_set_log_detailed_cb`.  If the config is not set, then the callback set in `log_cb` will be used by default.  If the config is set, then `log_detailed_cb` will be called instead of any `log_cb` that is set.

Additionally, create new builtin handlers:
  - `rd_kafka_log_detailed_print`
  - `rd_kafka_log_detailed_syslog`

Which behave analogously to the existing `log_cb` builtin handlers.

Logging macros are modified to transparently handle this extra parameter.

A new member is appended to the `log` union of `rd_kafka_op_s` which contains an array of strings representing the debug contexts of the log event.


edenhill:
Thank you for your efforts on this.

I'm not sure this is the direction we want to go though, to keep the API surface area down we try to craft more generic interfaces that can be extended over time to as need arise, especially so for the niche use cases like this one.

Taking this as an example it would be more appropriate to expose the debug contexts in the rd_kafka_event_t (RD_KAFKA_EVENT_LOG) interface  (and the corresponding C++ Event class).

wolfchimneyrock:
> Thank you for your efforts on this.
> 
> I'm not sure this is the direction we want to go though, to keep the API surface area down we try to craft more generic interfaces that can be extended over time to as need arise, especially so for the niche use cases like this one.
> 
> Taking this as an example it would be more appropriate to expose the debug contexts in the rd_kafka_event_t (RD_KAFKA_EVENT_LOG) interface (and the corresponding C++ Event class).

OK, i take it then I would leave `rd_kafka_event_log()` as-is and add something like `rd_kafka_event_log_context()` or `rd_kafka_event_log_detail()` ?

also, would you see in value in attaching the original rdkafka source file and line that logged via `__FILE__` and `__LINE__` macros?

edenhill:
Yeah, I'd call it `rd_kafka_event_debug_contexts()` and I'd do the flags2str conversion in that method rather than in the op-creation.

I don't think file and line are needed, the debug lines are easy to track down through their format.

src/rdkafka.c
```diff
@@ -227,7 +227,7 @@ int rd_kafka_wait_destroyed (int timeout_ms) {
 }
 
 static void rd_kafka_log_buf (const rd_kafka_conf_t *conf,
-                              const rd_kafka_t *rk, int level, const char *fac,
+                              const rd_kafka_t *rk, int level, int ctx, const char *fac,
```
edenhill, line 230: Keep below 80 columns.

src/rdkafka.h
```diff
@@ -4678,6 +4678,23 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev,
 			const char **fac, const char **str, int *level);
 
 
+/**
+ * @brief Extract log debug context from event.
+ *
+ * Event types:
+ *  - RD_KAFKA_EVENT_LOG
+ *
+ *  @param dst destination string for comma separated list, or NULL if using ctx_names.
```
edenhill, line 4687: I think we should only provide one of these output methods, it is simple enough for the application to convert from one or the other.
Since most debugs are with a single context I think the CSV string will be the most useful.

src/rdkafka.h
```diff
@@ -4678,6 +4678,23 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev,
...
+ *  @param ctx_names destination array for contexts, or NULL if using dst.
+ *  @param dstsize size of provided dst buffer or ctx_names array.
+ *  @returns 0 if successful using dst or number of contexts if using ctx_names,
+ *           or -1 if unsupported event type.
+ */
+RD_EXPORT
+int rd_kafka_event_debug_contexts (rd_kafka_event_t *rkev,
+            const char *dst, const char **ctx_names, size_t dstsize);
```
edenhill, line 4695: destination vars can't be const.

src/rdkafka_event.c
```diff
@@ -224,6 +224,55 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev, const char **fac,
 	return 0;
 }
 
+int rd_kafka_event_debug_contexts (rd_kafka_event_t *rkev,
+            const char *dst, const char **ctx_names, size_t dstsize) {
+    if (unlikely(rkev->rko_evtype != RD_KAFKA_EVENT_LOG))
+            return -1;
+
+    static const char* names[] = {
```
edenhill, line 232: '*' goes on the variable, not type.

src/rdkafka_event.c
```diff
@@ -224,6 +224,55 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev, const char **fac,
...
+            "cgrp",
+            "security",
+            "fetch",
+            "interceptor",
+            "plugin",
+            "consumer",
+            "admin",
+            "eos",
+            "mock",
+            0
```
edenhill, line 250: NULL

src/rdkafka_event.c
```diff
@@ -224,6 +224,55 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev, const char **fac,
...
+    };
+
+    int ctx = rkev->rko_u.log.ctx;
+	if (dst) {
```
edenhill, line 254: use 8 white-space indent

src/rdkafka_int.h
```diff
@@ -810,42 +810,48 @@ const char *rd_kafka_purge_flags2str (int flags);
...
 void rd_kafka_log0(const rd_kafka_conf_t *conf,
                    const rd_kafka_t *rk, const char *extra, int level,
+                   int ctx,
                    const char *fac, const char *fmt, ...) RD_FORMAT(printf,
-                                                                    6, 7);
+                                                                    7, 8);
 
 #define rd_kafka_log(rk,level,fac,...) \
-        rd_kafka_log0(&rk->rk_conf, rk, NULL, level, fac, __VA_ARGS__)
+        rd_kafka_log0(&rk->rk_conf, rk, NULL, level, RD_KAFKA_DBG_NONE, fac, __VA_ARGS__)
```
edenhill, line 821: < 80 cols

tests/0039-event.c
```diff
@@ -164,7 +164,56 @@ int main_0039_event_dr (int argc, char **argv) {
...
+
+        while (waitevent) {
+                rd_kafka_event_t *rkev;
+                rkev = rd_kafka_queue_poll(eventq, 1000);
+                switch (rd_kafka_event_type(rkev))
+                {
+                case RD_KAFKA_EVENT_LOG:
+                        rd_kafka_event_log(rkev, &fac, &msg, &level);
+                        rd_kafka_event_debug_contexts(rkev, ctx, NULL, 64);
+                        TEST_SAY("Got log  event: level: %d ctx: %s fac: %s: msg: %s\n",
```
edenhill, line 199: There is no verification of the string itself.
Suggest initializing ctx to all 'A', and then checking here that there are no 'A's in the string and its strlen is less than sizeof

src/rdkafka.h
```diff
@@ -4683,6 +4683,21 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev,
...
 
 
+/**
+ * @brief Extract log debug context from event.
+ *
+ * Event types:
+ *  - RD_KAFKA_EVENT_LOG
+ *
+ *  @param dst destination string for comma separated list.
+ *  @param dstsize size of provided dst buffer or ctx_names array.
```
edenhill, line 4693: references the old parameter

src/rdkafka_event.c
```diff
@@ -224,6 +224,37 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev, const char **fac,
 	return 0;
 }
 
+int rd_kafka_event_debug_contexts (rd_kafka_event_t *rkev,
+            char *dst, size_t dstsize) {
+        if (unlikely(rkev->rko_evtype != RD_KAFKA_EVENT_LOG))
+                return -1;
+
+        static const char *names[] = {
```
edenhill, line 232: variable definitions need to go at the top of the scope, above any code (the if-statement).

src/rdkafka_event.c
```diff
@@ -250,6 +247,9 @@ int rd_kafka_event_debug_contexts (rd_kafka_event_t *rkev,
                 NULL
         };
 
+        if (unlikely(rkev->rko_evtype != RD_KAFKA_EVENT_LOG))
+                return -1;
+
         int ctx = rkev->rko_u.log.ctx;
```
edenhill, line 253: This variable def needs to go above the if- code too.

Older MSVC compilers will not accept inline variable definitions.

edenhill: There is no need for this variable, btw, just reference the union directly below.

wolfchimneyrock: ok, thats likely why the win32 test was failing

src/rdkafka_event.c
```diff
@@ -224,6 +224,37 @@ int rd_kafka_event_log (rd_kafka_event_t *rkev, const char **fac,
...
+                "fetch",
+                "interceptor",
+                "plugin",
+                "consumer",
+                "admin",
+                "eos",
+                "mock",
+                NULL
+        };
+        int ctx = rkev->rko_u.log.ctx;
```
edenhill, line 249: This is not safe to do prior to checking the type.
Just use `rkev->rko_u.log.ctx` directly when calling flags2str.

edenhill:
Thank you!

PR #2836, opened 2020-04-19
Ensure CMake sets disabled defines to zero on Windows

benesch:
This patch fixes CMake builds on Windows, which have been broken since
v1.4.0.

The CMake build for Windows was setting options by defining or not
defining them, whereas for all other platforms options are defined to 1
or 0 to indicate set or not set.

In most code, options are checked with a preprocessor test like `#if
OPTION`, where it doesn't matter if OPTION is undefined or set to zero;
both approaches omit the code within the `#if`. But c426b31e2 introduced
code that requires OPTION to actually be defined to an integral value.

This patch ensures that the CMake configuration sets options to 1/0 when
targeting Windows. Technically only ENABLE_DEVEL needs to have a 1/0
value, but it seemed more future-proof to apply the change proactively
to all options.


edenhill:
Thank you!

PR #2839, opened 2020-04-21
Propagate consumer errors for unavailable subscribed topics 

edenhill:


INTRODUCTION.md
```diff
@@ -1435,10 +1436,41 @@ To read more about static group membership, see [KIP-345](https://cwiki.apache.o
...
+referenced by the client (by produce to, or consuming from, the topic, etc)
+the broker will automatically create the topic (with default partition counts
+and replication factor) if the broker configuration property
+`auto.create.topics.enable=true` is set.
+
+*Note*: A topic that is undergoing automatic creation may be reported as
+unavailable, with e.g., `RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART`, during the
+time the topic is being created and partition leaders are elected.
+
+While topic auto creation is useful for producer applications, it is not
```
mhowlett, line 1464: "may be useful" is possibly better

CONFIGURATION.md
```diff
@@ -114,6 +114,7 @@ rebalance_cb                             |  C  |                 |
...
 enable.partition.eof                     |  C  | true, false     |         false | low        | Emit RD_KAFKA_RESP_ERR__PARTITION_EOF event whenever the consumer reaches the end of a partition. <br>*Type: boolean*
 check.crcs                               |  C  | true, false     |         false | medium     | Verify CRC32 of consumed messages, ensuring no on-the-wire or on-disk corruption to the messages occurred. This check comes at slightly increased CPU usage. <br>*Type: boolean*
+allow.auto.create.topics                 |  C  | true, false     |         false | low        | Allow automatic topic creation on the broker when subscribing to or assigning non-existent topics. The broker must also be configured with `auto.create.topics.enable=true` for this configuraiton to take effect. Note: The default value (false) is different from the Java consumer (true). Requires broker version >= 0.11.0.0, for older broker versions only the broker configuration applies. <br>*Type: boolean*
```
mhowlett, line 117: technically requires a major version bump... but probably not important.

edenhill: Theoretically yes, but in practice there is no difference between having a consumer auto-create a topic which is later produced to by a producer, or the consumer subscribing to a non-existent topic until the producer creates it (automatically) when producing to it.
In short, auto topic creation doesn't make much sense in the consumer since the topic will be empty anyway.

This is also the reason we have a different default than the Java consumer. I asked Ismael about it and he agreed that our default here makes more sense.

INTRODUCTION.md
```diff
@@ -1435,10 +1436,41 @@ To read more about static group membership, see [KIP-345](https://cwiki.apache.o
 
 ### Topics
 
+#### Unknown or unauthorized topics
+
+If a consumer application subscribes to non-existent or unauthorized topics
+a consumer error will be propagated for each unavailable topic with the
+error code set to either `RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART` or a
+broker-specific error code, such as
+`RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED`.
```
mhowlett, line 1445: 99% of the time, users will be subscribing to a topic they know to exist, and they will want their application to fail if it doesn't. in C# for example, I just want to do `await consumer.SubscribeAsync("my-topic");`, and get an exception if there's one of the two problem listed above. setting an error handler is much less ergonomic. remind me again why we can't provide an API for the common case that behaves in this way? There could be an optional flag that makes it work in the way documented here for the 1% who need it.

mhowlett: (not something for this PR, but i'll take any opportunity to get up on that soap box)

src/rdkafka.c
```diff
@@ -2531,15 +2531,15 @@ int rd_kafka_consume_start0 (rd_kafka_topic_t *rkt, int32_t partition,
 
 int rd_kafka_consume_start (rd_kafka_topic_t *app_rkt, int32_t partition,
```
mhowlett, line 2532: it's noted in the `rd_kafka_new` docstring that this method is for operating in the simple legacy consumer mode. This isn't noted in the docstring for `rd_kafka_consume_start`. I think some more commenting aimed at distinguishing the two would probably be worthwhile (probably other places too).

src/rdkafka_msg.c
```diff
@@ -71,7 +71,7 @@ static RD_INLINE rd_kafka_resp_err_t rd_kafka_check_produce (rd_kafka_t *rk) {
 
 
 void rd_kafka_msg_destroy (rd_kafka_t *rk, rd_kafka_msg_t *rkm) {
-
+//FIXME
```
mhowlett, line 74: fixme

src/rdkafka_op.c
```diff
@@ -409,6 +409,33 @@ void rd_kafka_q_op_err (rd_kafka_q_t *rkq, rd_kafka_op_type_t optype,
...
+	va_start(ap, fmt);
+	rd_vsnprintf(buf, sizeof(buf), fmt, ap);
+	va_end(ap);
+
+	rko = rd_kafka_op_new(optype);
+	rko->rko_version = version;
+	rko->rko_err = err;
+	rko->rko_u.err.errstr = rd_strdup(buf);
+        rko->rko_u.err.rkm.rkm_rkmessage.rkt =
+                (rd_kafka_topic_t *)rd_kafka_lwtopic_new(rkq->rkq_rk, topic);
```
mhowlett, line 433: did you consider any alternatives to the lw topics? could you just provide the topic name instead? what use cases does providing the lw topic object enable that a name doesn't?

pros: nice API (may be worth it if the use cases are compelling enough)
cons: poor encapsulation creating a burden anywhere you might want to handle a lw topic masquerading as a topic.

INTRODUCTION.md
edenhill: The problem is that subscribe() is a one-time thing (for most applications), while these type of topic errors can happen at any time (even with a synchronous subscribe), which means that the user would have to have error handling for unavailable topics in two places: around the subscribe() call, and during consumption.
I don't think this makes the user program or experience less complex, it is better to educate and make the streaming event consumption as straight-forward as possible, since that is the business we're in.

src/rdkafka_op.c
edenhill: Changing to a string would be a breaking change to the public APIs, and providing a fully-blown rd_kafka_topic_t object to report topic-does-not-exist has problems since librdkafka never lets go of the proper topic objects until the client instance is terminated, and changing that would be a far more intrusive change.

edenhill: as for the cons: that is true, but only the legacy public APIs take a topic_t object, all new ones take a topic name.

INTRODUCTION.md
mhowlett: My three responses to that are: 

1. In practice, both of the two examples cited above are going to happen infrequently enough that people aren't going to bother explicitly handling them in the error handler (and I don't think we should recommend that they do). In practice, they will rely on just logging the problem along with other errors on looking at the log when something breaks.

2. Having a synchronous subscribe method throw an exception almost gives you error handling for free, since you'll usually be able to rely on an outer exception handler that you already have in place. It will also be more quickly evident that something when wrong and what the problem was via this mechanism.

3. These problems are most likely to occur on subscribe, and we absolutely add value if we help people work out more easily what a problem was in the majority of cases, even if it doesn't cover them all.

The same sort of argument applies for why we should have an option for synchronous client creation.

in summary, it is better to educate and make the streaming event consumption as straight-forward as possible, since that is the business we're in.

src/rdkafka_op.c
mhowlett: thanks, sounds good to me.

note: i hadn't got to the point of completely understanding everything, just noting out something that looked possibly off. thanks for bearing with me :-)

src/rdkafka_cgrp.c
```diff
@@ -1064,8 +1067,17 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
                         rd_kafka_cgrp_assignor_handle_Metadata_op);
                 rd_kafka_op_set_replyq(rko, rkcg->rkcg_ops, NULL);
 
-                rd_kafka_MetadataRequest(rkb, &topics,
-                                         "partition assignor", rko);
+                rd_kafka_MetadataRequest(
+                        rkb, &topics,
+                        "partition assignor",
+                        /* cgrp_update=false:
+                         * Since this Metadata request may not be identical
```
mhowlett, line 1074: I was confused around here for a while. This comment is fine... but if it had read "Since the subscription list across all members may not be identical to this consumer's ...", i probably would have spent less time confused.

src/rdkafka_metadata.c
```diff
@@ -584,8 +583,12 @@ rd_kafka_parse_Metadata (rd_kafka_broker_t *rkb,
 
         /* Check if cgrp effective subscription is affected by
          * new topic metadata.
-         * Ignore if this was a broker-only refresh (no topics) */
-        if ((requested_topics || all_topics) && rkb->rkb_rk->rk_cgrp)
+         * Ignore if this was a broker-only refresh (no topics), or
```
mhowlett, line 586: tangential to this PR (and i may have mentioned it before, recall thinking it before), it might be worth factoring anything that isn't parsing out of a method labelled "parse..". though, the method isn't really doing any harm as is.

PR #2840, opened 2020-04-21
 Fix produce/consume hang after partition goes away and comes back

edenhill:
This is a regression from fetch-from-follower.

And a bunch of other fixes.


src/rdkafka_topic.c
```diff
@@ -545,6 +545,8 @@ static int rd_kafka_toppar_leader_update (rd_kafka_topic_t *rkt,
         rd_kafka_toppar_lock(rktp);
 
         if (leader != NULL &&
+            rktp->rktp_broker != NULL &&
+            rktp->rktp_broker->rkb_source != RD_KAFKA_INTERNAL &&
```
mhowlett, line 549: there is so much detail to be aware of!

PR #2851, opened 2020-04-27
Improve connection error propagation, remove ApiVersion heuristics, and probe for CA cert path

edenhill:


CONFIGURATION.md
```diff
@@ -66,7 +66,7 @@ ssl_key                                  |  *  |                 |
...
 ssl_certificate                          |  *  |                 |               | low        | Client's public key as set by rd_kafka_conf_set_ssl_cert() <br>*Type: see dedicated API*
-ssl.ca.location                          |  *  |                 |               | low        | File or directory path to CA certificate(s) for verifying the broker's key. <br>*Type: string*
+ssl.ca.location                          |  *  |                 |               | low        | File or directory path to CA certificate(s) for verifying the broker's key. Defaults: On Windows the system's CA certificates are automatically looked up in the Windows Root certificate store. On Mac OSX it is recommended to install openssl using Homebrew, to provide CA certificates. On Linux install the distribution's ca-certificates package. If OpenSSL is statically linked or `ssl.ca.location` is set to `probe` a list of standard paths will be probed and the first one found will be used as the default CA certificate location path. If OpenSSL is dynamically linked the OpenSSL library's default path will be used (see `OPENSSLDIR` in `openssl version -a`). <br>*Type: string*
```
rnpridgeon, line 69: Are the homebrew instructions always necessary or only when openssl is statically linked.

src/rdkafka_broker.c
```diff
@@ -351,33 +348,170 @@ void rd_kafka_broker_set_state (rd_kafka_broker_t *rkb, int state) {
...
+        if (of > sizeof(errstr))
+                of = sizeof(errstr);
+
+        /* Provide more meaningful error messages in certain cases */
+        if (err == RD_KAFKA_RESP_ERR__TRANSPORT &&
+            !strcmp(errstr, "Disconnected")) {
+                if (rkb->rkb_state == RD_KAFKA_BROKER_STATE_APIVERSION_QUERY) {
+                        /* A disconnect while requesting ApiVersion typically
+                         * means we're connecting to a SSL-listener as
+                         * PLAINTEXT, but may also be caused by connecting to
```
rnpridgeon, line 383: 👍

src/rdkafka_broker.c
```diff
@@ -351,33 +348,170 @@ void rd_kafka_broker_set_state (rd_kafka_broker_t *rkb, int state) {
...
+                           state_duration_ms < 2000/*2s*/ &&
+                           rkb->rkb_rk->rk_conf.security_protocol !=
+                           RD_KAFKA_PROTO_SASL_SSL &&
+                           rkb->rkb_rk->rk_conf.security_protocol !=
+                           RD_KAFKA_PROTO_SASL_PLAINTEXT) {
+                        /* If disconnected shortly after transitioning to UP
+                         * state it typically means the broker listener is
+                         * configured for SASL authentication but the client
+                         * is not. */
+                        rd_kafka_broker_set_error(
```
rnpridgeon, line 417: 👍 

The broker might require SASL authentication or it may be misconfgured.

src/rdkafka_broker.c
```diff
@@ -351,33 +348,170 @@ void rd_kafka_broker_set_state (rd_kafka_broker_t *rkb, int state) {
...
+                            rkb->rkb_proto != RD_KAFKA_PROTO_SASL_SSL)
+                                rd_kafka_broker_set_error(
+                                        rkb, level, err,
+                                        "Disconnected while requesting "
+                                        "ApiVersion: "
+                                        "might be caused by incorrect "
+                                        "security.protocol configuration or "
+                                        "broker version is < 0.10 "
+                                        "(see api.version.request)",
+                                        ap/*ignored*/);
```
rnpridgeon, line 396: Since API requests are handled prior to performing any sort of SASL authentication it may be better to just explicitly state that this is likely a TLS issue. This could be a handshake failure, protocol mismatch etc. Just stating a possible `security.protocol` mismatch may send people chasing SASL issues as they believe TLS to be configured correctly. In my experience people almost always assume its SASL and not TLS when both are in use.

src/rdkafka_broker.c
```diff
@@ -2194,14 +2268,14 @@ rd_kafka_broker_handle_ApiVersion (rd_kafka_t *rk,
...
-                                     "security.protocol "
-                                     "or broker version is < 0.10 "
-                                     "(see api.version.request)",
-				     rd_kafka_err2str(err));
+                if (rkb->rkb_transport)
+                        rd_kafka_broker_fail(
+                                rkb, LOG_WARNING,
+                                RD_KAFKA_RESP_ERR__TRANSPORT,
+                                "ApiVersionRequest failed: %s: "
+                                "probably due to broker version < 0.10 "
```
rnpridgeon, line 2276: Should we mention the possibility of a protocol mismatch here as well?

CONFIGURATION.md
edenhill: Always, OSX does not provide any ca cert fels, it has its keystore, but when installing openssl thru brew I believe it either downloads proper ca-certs, or extracts certs from the keystore and puts them on disk.

src/rdkafka_broker.c
edenhill: Okay, will suggest checking SSL.

src/rdkafka_broker.c
edenhill: We should not arrive here on Disconnect, so those cases are handled separately.

CONFIGURATION.md
rnpridgeon: Surely OSX ships with some CA certs otherwise using Safari would be quite unpleasent

edenhill: They are in the keystore, not as files.

src/rdkafka.c
```diff
@@ -3495,24 +3499,11 @@ rd_kafka_poll_cb (rd_kafka_t *rk, rd_kafka_q_t *rkq, rd_kafka_op_t *rko,
...
-                                             rk->rk_name,
-                                             rko->rko_u.err.errstr);
-                        else
-                                rd_kafka_log(rk, LOG_ERR, "ERROR",
-                                             "%s: %s: %s",
-                                             rk->rk_name,
-                                             rko->rko_u.err.errstr,
-                                             rd_kafka_err2str(rko->rko_err));
-                }
+                else
```
mhowlett, line 3502: I think error events should be logged regardless of whether there is an `error_cb` since the behavior of `rd_kafka_log` is exactly what i'd want most of the time in response to one of these 'error' events, and the way the API is forces me to explicitly write code to do this if I set an `error_cb`.

Also, my memory is failing me on the specifics, but is `LOG_ERR` really appropriate for all these events? (don't have a good suggestion on what to do).

src/rdkafka_broker.c
```diff
@@ -351,33 +348,171 @@ void rd_kafka_broker_set_state (rd_kafka_broker_t *rkb, int state) {
...
+        else
+                rkb->rkb_last_err.cnt++;
+
+        rd_rkb_dbg(rkb, BROKER, "FAIL", "%s (%s)%s%s",
+                   errstr, rd_kafka_err2name(err),
+                   identical ? ": identical to last error" : "",
+                   suppress ? ": error log suppressed" : "");
+
+        if (level != LOG_DEBUG && (level <= LOG_CRIT || !suppress)) {
+                /* Don't log if an error callback is registered,
```
mhowlett, line 482: why?

src/rdkafka_broker.c
```diff
@@ -351,33 +348,171 @@ void rd_kafka_broker_set_state (rd_kafka_broker_t *rkb, int state) {
...
 
 /**
- * Failure propagation to application.
- * Will tear down connection to broker and trigger a reconnect.
+ * @brief Set, log and propagate broker fail error.
+ *
+ * The caller may omit the format if it thinks this is a recurring
+ * failure, in which case the following things are omitted:
+ *  - log message
+ *  - application OP_ERR
```
mhowlett, line 356: nit: document params?

src/rdkafka_broker.c
```diff
@@ -351,33 +348,171 @@ void rd_kafka_broker_set_state (rd_kafka_broker_t *rkb, int state) {
...
+            !strcmp(errstr, "Disconnected")) {
+                if (rkb->rkb_state == RD_KAFKA_BROKER_STATE_APIVERSION_QUERY) {
+                        /* A disconnect while requesting ApiVersion typically
+                         * means we're connecting to a SSL-listener as
+                         * PLAINTEXT, but may also be caused by connecting to
+                         * a broker that does not support ApiVersion (<0.10). */
+
+                        if (rkb->rkb_proto != RD_KAFKA_PROTO_SSL &&
+                            rkb->rkb_proto != RD_KAFKA_PROTO_SASL_SSL)
+                                rd_kafka_broker_set_error(
```
mhowlett, line 388: don't really like this. instead, I think I would have made a separate method to construct the error string, and invoke that at each call site. it is slightly longer, but the simplicity adds value imo.

src/rdkafka.c
edenhill: I think the point of an error callback is for the application to handle the error in any way it seems fit, perhaps by logging it - in which case we would have duplicate logs.
If there is no error handler we log it instead so the error is not silently ignored.
It was evident in the examples/ programs that had error_cb that error+log was not great.

We can adjust this later if needed.

src/rdkafka_broker.c
edenhill: Disconnects can happen in a myriad of places (read error, write error, poll-hup, ssl_io_update) and I don't want to put state logic in all those places, I think it makes more sense to have the additional error heuristics which checks error type and current state in one single place.

src/rdkafka_broker.c
edenhill: I think that suggesting that the broker is misconfigured is outside the scope of the client.

PR #2852, opened 2020-04-27
Clarify reuse restriction on arguments to rd_kafka_produce()

andrewthad:
Resolves https://github.com/edenhill/librdkafka/issues/2843

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3920,7 +3920,12 @@ rd_kafka_consumer_group_metadata_read (
  *                               partition, either set manually or by the
  *                               configured partitioner.
  *
- *    .._F_FREE and .._F_COPY are mutually exclusive.
+ *    .._F_FREE and .._F_COPY are mutually exclusive. If neither of these are
+ *    set, the caller must ensure that the memory backing \p payload remains
+ *    valid and is not modified or reused until the delivery callback is
+ *    invoked. Other buffers passed to `rd_kafka_produce()` don't have this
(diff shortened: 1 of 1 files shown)


src/rdkafka.h
```diff
@@ -3920,7 +3920,12 @@ rd_kafka_consumer_group_metadata_read (
  *                               partition, either set manually or by the
  *                               configured partitioner.
  *
- *    .._F_FREE and .._F_COPY are mutually exclusive.
+ *    .._F_FREE and .._F_COPY are mutually exclusive. If neither of these are
+ *    set, the caller must ensure that the memory backing \p payload is not
+ *    reused until the delivery callback is called. Other buffers passed to
```
andrewthad, line 3925: Perhaps "delivery callback is invoked" instead of "delivery callback is called"?

edenhill: Yeah, can we also make it even more clear with something like:
"..backing \p payload remains valid and is not modified or reused until .."

src/rdkafka.h
```diff
@@ -3920,7 +3920,12 @@ rd_kafka_consumer_group_metadata_read (
...
+ *    `rd_kafka_produce()` never have this restriction on reuse, i.e. the
```
edenhill, line 3926: never/don't

andrewthad:
Amended the commit to address the issues discussed in the inline comments.

edenhill:
Thank you!

Issue #2858, opened 2020-04-29
Producer: Wait for newly created topics to become visible


edenhill:
When creating topics with the adminAPI CreateTopics() there is a time window between a succesful response and the new topic being propagated and known throughout the cluster.
If a producer queries the cluster for topic metadata during this window it will seem as if the topic does not exist and if auto topic creation is disabled the producer will flag the topic as non-existent and fail further produce() calls to this topic as well as purge any messages enqueued.

The proposed solution is to not flag the topic as non-existent until we've queried its Metadata for at least X seconds (e.g., 30), allowing enough time for the topic to be propagated throughout the cluster.
This value should be configurable.

PR #2862, opened 2020-04-30
Add `topic.metadata.propagation.max.ms` to circumvent topic creation race (#2858)

edenhill:
cc @theduderog


theduderog:
Thank you!  This will be so nice. 🙏

CONFIGURATION.md
```diff
@@ -18,6 +18,7 @@ metadata.max.age.ms                      |  *  | 1 .. 86400000   |        900000
...
 topic.metadata.refresh.fast.cnt          |  *  | 0 .. 1000       |            10 | low        | **DEPRECATED** No longer used. <br>*Type: integer*
 topic.metadata.refresh.sparse            |  *  | true, false     |          true | low        | Sparse metadata requests (consumes less network bandwidth) <br>*Type: boolean*
+topic.metadata.propagation.max.ms        |  *  | 0 .. 3600000    |         30000 | low        | Apache Kafka topic creation is asynchronous and it takes some time for a new topic to propagate throughout the cluster to all brokers. If a client requests topic metadata after manual topic creation but before the topic has been fully propagated to the broker the client is requesting metadata from, the topic will seem to be non-existent and the client will mark the topic as such, failing queued produced messages with `ERR__UNKNOWN_TOPIC`. This setting delays marking a topic as non-existent until the configured propagation max time has passed. The maximum propagation time is calculated from the time the topic is first referenced in the client, e.g., on produce(). <br>*Type: integer*
```
mhowlett, line 21: does the java client have this configuration property, and if not, how does it avoid having it?

INTRODUCTION.md
```diff
@@ -1470,6 +1470,34 @@ As the topic metadata is refreshed every `topic.metadata.refresh.interval.ms`
...
+#### Topic metadata propagation for newly created topics
+
+Due to the asynchronous nature of topic creation in Apache Kafka it may
+take some time for a newly created topic to be known by all brokers in the
+cluster.
+If a client tries to use a topic after topic creation but before the topic
+has been fully propagated in the cluster it will seem as if the topic does not
+exist which would raise `RD_KAFKA_RESP_ERR__UNKNOWN_TOPIC` (et.al)
+errors to the application.
+To avoid these temporary errors to be raised the client will not flag
```
mhowlett, line 1482: "... errors being raised, the client ..."

INTRODUCTION.md
```diff
@@ -1470,6 +1470,34 @@ As the topic metadata is refreshed every `topic.metadata.refresh.interval.ms`
...
+a topic as non-existent until a propagation time has elapsed, this propagation
+defaults to 30 seconds and can be configured with
+`topic.metadata.propagation.max.ms`.
+The per-topic max propagation time starts ticking as soon as the topic is
+referenced (e.g., by produce()).
+
+It is perfectly fine to produce messages to unknown topics during the
```
mhowlett, line 1489: "perfectly fine" feels a bit colloquial to me. maybe just "If messages are produced to unknown topics ..

src/rdkafka_topic.c
```diff
@@ -924,36 +926,73 @@ static void rd_kafka_topic_assign_uas (rd_kafka_topic_t *rkt,
...
- * Received metadata request contained no information about topic 'rkt'
- * and thus indicates the topic is not available in the cluster.
+ * @brief Mark topic as non-existent, unless metadata propagation configuration
+ *        disallows it.
+ *
+ * @param err Propagate non-existent topic using this error code.
+ *            If \p err is RD_KAFKA_RESP_ERR_TOPIC_EXCEPTION it means the
+ *            topic is invalid and no propagation delay will be used.
+ *
+ * @locks topic_wrlock() MUST be held.
```
mhowlett, line 936: document return value?

src/rdkafka_topic.c
```diff
@@ -924,36 +926,73 @@ static void rd_kafka_topic_assign_uas (rd_kafka_topic_t *rkt,
...
+                             "for metadata propagation before marking topic "
+                             "as non-existent",
+                             RD_KAFKAP_STR_PR(rkt->rkt_topic),
+                             (int)(remains_us / 1000));
+                return rd_false;
+        }
 
-	rkt->rkt_ts_metadata = rd_clock();
+        if (!err)
+                err = RD_KAFKA_RESP_ERR__UNKNOWN_TOPIC;
```
mhowlett, line 967: this seems like it should be a pre-condition check (assert). a bit dodgy to have it as an allowed path?

src/rdkafka_topic.c
```diff
@@ -924,36 +926,73 @@ static void rd_kafka_topic_assign_uas (rd_kafka_topic_t *rkt,
...
 
-	rd_kafka_topic_wrunlock(rkt);
+        return rd_true;
+}
+
+/**
+ * Received metadata request contained no information about topic 'rkt'
+ * and thus indicates the topic is not available in the cluster.
+ */
+void rd_kafka_topic_metadata_none (rd_kafka_topic_t *rkt) {
```
mhowlett, line 989: is it really necessary to keep this function? it's only used once and doesn't do anything important. just means you need to bounce around the code one more time to follow what's going on.

src/rdkafka_topic.c
```diff
@@ -1012,11 +1051,11 @@ rd_kafka_topic_metadata_update (rd_kafka_topic_t *rkt,
...
 
-	/* Set topic state.
+        /* Set topic state.
          * UNKNOWN_TOPIC_OR_PART may indicate that auto.create.topics failed */
-	if (mdt->err == RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART ||
-            mdt->err == RD_KAFKA_RESP_ERR_TOPIC_EXCEPTION/*invalid topic*/)
-                rd_kafka_topic_set_state(rkt, RD_KAFKA_TOPIC_S_NOTEXISTS);
+        if (mdt->err == RD_KAFKA_RESP_ERR_TOPIC_EXCEPTION/*invalid topic*/ ||
+            mdt->err == RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART)
+                rd_kafka_topic_set_notexists(rkt, mdt->err);
```
mhowlett, line 1058: what about `rd_kafka_topic_handle_notexists`? - the word set makes me more likely think it's being set unconditionally.

CONFIGURATION.md
edenhill: The Java producer send() will block on initial metadata fetch and has a max.block.ms that controls how long non-blocking APIs (like send()) may block.

This is not the path we want to go in librdkafka, asynchronous APIs should be 100% asynchronous., and we already have asynchronous signaling in place (dr_cb).

src/rdkafka_topic.c
edenhill: Removed it

edenhill:
Please review.

CONFIGURATION.md
mhowlett: 👍 absolutely can not block otherwise async calls, that'll totally screw up the api.

Issue #2121, opened 2018-11-27
Segfault in round-robin assignor


ankon:
Description
===========

We're using the 'roundrobin' partition assignment strategy, and sometimes I get segfaults in rdk:main. I looked at one of these in more detail:
~~~~
#0  0x00007fc0f32d6fd3 in rd_kafka_group_member_find_subscription (rk=rk@entry=0x7fc0d4002e50, rkgm=rkgm@entry=0x7fc0b806a380, topic=0x7fc0b8076de0 "ml3-commands") at rdkafka_assignor.c:81
#1  0x00007fc0f32d7fb4 in rd_kafka_roundrobin_assignor_assign_cb (rk=0x7fc0d4002e50, member_id=<optimized out>, protocol_name=<optimized out>, metadata=<optimized out>, members=0x7fc0b806a2e0, 
    member_cnt=<optimized out>, eligible_topics=0x7fc0b805bfa0, eligible_topic_cnt=12, errstr=0x7fc0d97f9a00 "", errstr_size=8, opaque=0x0) at rdkafka_roundrobin_assignor.c:80
#2  0x00007fc0f32d74d8 in rd_kafka_assignor_run (rkcg=rkcg@entry=0x7fc0d4003b20, protocol_name=protocol_name@entry=0x6b8ec40 "roundrobin", metadata=0x7fc0b80769f0, members=members@entry=0x7fc0b806a2e0, 
    member_cnt=member_cnt@entry=2, errstr=errstr@entry=0x7fc0d97f9a00 "", errstr_size=512) at rdkafka_assignor.c:331
#3  0x00007fc0f32be5f7 in rd_kafka_cgrp_assignor_run (rkcg=0x7fc0d4003b20, protocol_name=0x6b8ec40 "roundrobin", err=<optimized out>, metadata=<optimized out>, members=0x7fc0b806a2e0, member_cnt=2)
    at rdkafka_cgrp.c:671
#4  0x00007fc0f32be7cf in rd_kafka_cgrp_assignor_handle_Metadata_op (rk=<optimized out>, rkq=<optimized out>, rko=<optimized out>) at rdkafka_cgrp.c:733
#5  0x00007fc0f32ab5fc in rd_kafka_op_call (rk=<optimized out>, rkq=<optimized out>, rko=0x7fc0b802d090) at rdkafka_op.c:519
[...]

PR #2867, opened 2020-05-04
Fix crash in roundrobin assignor for asymmetrical subscriptions (#2121)

edenhill:
This also adds declarative unit tests of the assignors.

### DIFF
--- a/src/rdkafka_assignor.c
+++ b/src/rdkafka_assignor.c
@@ -209,7 +209,7 @@ int rd_kafka_assignor_topic_cmp (const void *_a, const void *_b) {
         const rd_kafka_assignor_topic_t *b =
                 *(const rd_kafka_assignor_topic_t * const *)_b;
 
-        return !strcmp(a->metadata->topic, b->metadata->topic);
+        return strcmp(a->metadata->topic, b->metadata->topic);
 }
 
 /**
--- a/src/rdkafka_assignor.c
+++ b/src/rdkafka_assignor.c
@@ -27,6 +27,7 @@
  */
 #include "rdkafka_int.h"
 #include "rdkafka_assignor.h"
+#include "rdunittest.h"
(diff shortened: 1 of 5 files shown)


src/rdkafka_assignor.c
```diff
@@ -558,3 +559,403 @@ int rd_kafka_assignors_init (rd_kafka_t *rk, char *errstr, size_t errstr_size) {
...
+                        char errstr[256];
+
+                        RD_UT_SAY("Test case %s: %s assignor",
+                                  tests[i].name,
+                                  tests[i].expect[ie].protocol_name);
+
+                        /* Run assignor */
+                        err = rd_kafka_assignor_run(
+                                rk->rk_cgrp,
+                                tests[i].expect[ie].protocol_name,
```
mhowlett, line 866: heads up that i've changed this to be a pointer to the assignor in the change for stateful assignors. also, `rk->rk_cgrp` maintains this state so it's kind of superfluous to specify it separately. anyway fothcoming change.

src/rdkafka_assignor.c
```diff
@@ -208,7 +209,7 @@ int rd_kafka_assignor_topic_cmp (const void *_a, const void *_b) {
         const rd_kafka_assignor_topic_t *b =
                 *(const rd_kafka_assignor_topic_t * const *)_b;
 
-        return !strcmp(a->metadata->topic, b->metadata->topic);
+        return strcmp(a->metadata->topic, b->metadata->topic);
```
edenhill, line 212: Found this issue (which existed prior to this PR).

mhowlett: LGTM. we need better testing!

edenhill: It was found by the tests added in this PR ;)

PR #2873, opened 2020-05-06
Prefer least-idle connection

edenhill:


src/rdkafka_broker.c
```diff
@@ -1283,6 +1285,113 @@ rd_kafka_broker_random0 (const char *func, int line,
...
+        }
+
+        return good;
+}
+
+/**
+ * @brief Weighing function to select a usable broker connections,
+ *        promoting connections according to the scoring below.
+ *
+ * Scoring:
```
mhowlett, line 1348: would be better to write out as a strictly correct equation - as is i'm guessing or need to read the code, so comment is not helping.

edenhill: That's why you are the Math.
Feel free to take a stab at it.

edenhill: Alternatively we could just change the comment to a priority list and not duplicate the actual scoring in the comment.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ librdkafka.
    hints on the cause of the problem.
  * Consumer: Propagate errors when subscribing to unavailable topics (#1540)
  * Producer: Add `batch.size` producer configuration property (#638)
+ * Prefer least idle connection for periodic metadata refreshes, et.al.,
+   to allow truly idle connections to time out and to avoid load-balancer-killed
+   idle connection errors (#2845)
```
mhowlett, line 14: love it

src/rdkafka_broker.c
```diff
@@ -1283,6 +1285,113 @@ rd_kafka_broker_random0 (const char *func, int line,
...
+ * @locks_required rkb
+ */
+static int rd_kafka_broker_weight_usable (rd_kafka_broker_t *rkb) {
+        int weight = 0;
+
+        if (!rd_kafka_broker_state_is_up(rkb->rkb_state))
+                return 0;
+
+        weight += 2000 * (rkb->rkb_nodeid != -1); /* is not a bootstrap */
+        weight += 10 * !RD_KAFKA_BROKER_IS_LOGICAL(rkb);
```
mhowlett, line 1372: should you return 0 for logical brokers since they are just an indirection on another broker connection, which will be considered by `rd_kafka_broker_weighted` anyway?

src/rdkafka_broker.c
mhowlett: yeah. i think you can probably just remove it.

the heuristic looks good to me (except not understanding why logical brokers are included).

src/rdkafka_broker.c
edenhill: Logical brokers are dedicated real connections, they're not an indirection to another broker_t, but rather an indirection to a variable hostname/address which is updated when e.g. the group coordinator is changing.
This is to say they are fully functional standard connections and the only reason not to give them as high of a weigh as standard connections is that the logical connections are dedicated for a certain purpose and should preferably be left unused for other means, but it is better to use an idle logical connection for a metadata query than to create a new connection on a disconnected broker thread.

PR #2884, opened 2020-05-13
KIP-429: Incremental rebalancing

mhowlett:
This PR is just a start. It includes:
- changes to allow for stateful assignors.
- changes for KIP-54 except the actual assignor implementation.
- owned partitions on the wire (not used by kip-54, but ready for kip-429).

I've explicitly considered/checked compatibility with the Java consumer as appropriate.


src/rdkafka_assignor.h
```diff
@@ -129,9 +142,9 @@ void rd_kafka_group_member_clear (rd_kafka_group_member_t *rkgm);
  * rd_kafka_range_assignor.c
  */
 rd_kafka_resp_err_t
-rd_kafka_range_assignor_assign_cb (rd_kafka_t *rk,
+rd_kafka_range_assignor_assign_cb (rd_kafka_assignor_t *rkas,
```
edenhill, line 145: rk is generally the first argument.

Can we make rkas const?

src/rdkafka_cgrp.c
```diff
@@ -968,6 +979,25 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
                  * a matching assignor so error out early. */
                 ErrorCode = RD_KAFKA_RESP_ERR__BAD_MSG;
         }
+	else if (!ErrorCode) {
```
edenhill, line 982: else goes on same line as end-brace: `} else if ( ..

src/rdkafka_cgrp.c
```diff
@@ -968,6 +979,25 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
...
+                char *protocol_name = RD_KAFKAP_STR_DUP(&Protocol);
```
edenhill, line 983: Since this is only used in this stack frame use RD_KAFKAP_STR_DUPA and drop the free below.

src/rdkafka_cgrp.c
```diff
@@ -968,6 +979,25 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
...
+                                "Unsupported assignment strategy \"%s\"",
+                                protocol_name);
+                        if (rkcg->rkcg_assignor) {
+                                if (rkcg->rkcg_assignor->rkas_destroy_state)
+                                        rkcg->rkcg_assignor->rkas_destroy_state(
+                                                rkcg->rkcg_assignor_state);
+                                rkcg->rkcg_assignor_state = NULL;
+                                rkcg->rkcg_assignor = NULL;
+                        }
+                        ErrorCode = RD_KAFKA_RESP_ERR__UNKNOWN_PROTOCOL;
```
edenhill, line 997: There is no explicit handling of this error code down in the err: label, what is the expected behaviour when this occurs?

src/rdkafka_cgrp.c
```diff
@@ -992,6 +1022,14 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
                 goto err;
         }
 
+        if (rkcg->rkcg_assignor && rkcg->rkcg_assignor != rkas) {
```
edenhill, line 1025: Should assignor state be scrapped if the subscription is changed through a new subscribe/unsubscribe() call?

src/rdkafka_cgrp.c
```diff
@@ -3665,35 +3696,24 @@ void rd_kafka_cgrp_handle_SyncGroup (rd_kafka_cgrp_t *rkcg,
 		rkbuf->rkbuf_rkb = rd_kafka_broker_internal(rkcg->rkcg_rk);
 
         rd_kafka_buf_read_i16(rkbuf, &Version);
-        rd_kafka_buf_read_i32(rkbuf, &TopicCnt);
-
-        if (TopicCnt > 10000) {
-                err = RD_KAFKA_RESP_ERR__BAD_MSG;
+        if (!(assignment = rd_kafka_buf_read_topic_partitions(rkbuf, 0, rd_false)))
```
edenhill, line 3699: :+1:

src/rdkafka_cgrp.c
```diff
@@ -3730,25 +3750,34 @@ void rd_kafka_cgrp_handle_SyncGroup (rd_kafka_cgrp_t *rkcg,
...
 rd_kafka_consumer_group_metadata (rd_kafka_t *rk) {
+        int32_t generation_id = -1;
+
         if (rk->rk_type != RD_KAFKA_CONSUMER ||
             !rk->rk_conf.group_id_str)
                 return NULL;
 
-        return rd_kafka_consumer_group_metadata_new(rk->rk_conf.group_id_str);
+        if (rk->rk_cgrp)
+                generation_id = rk->rk_cgrp->rkcg_generation_id;
```
edenhill, line 3776: rkcg_generation_id needs to be initialized to -1 in cgrp_init  (maybe it is, havent gotten there yet)

src/rdkafka_cgrp.c
```diff
@@ -3785,13 +3815,17 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_write (
...
+        size_t generationid_len = sizeof(int32_t);
 
-        size = magic_len + groupid_len;
+        size = magic_len + groupid_len + generationid_len;
         buf = rd_malloc(size);
 
         memcpy(buf, rd_kafka_consumer_group_metadata_magic, magic_len);
         of += magic_len;
 
+        *(int32_t *)(buf+of) = htole32(cgmd->generation_id);
```
edenhill, line 3826: Unaligned casts are not allowed, do a memcpy instead.

edenhill: Not sure we need the endian swap since the serialized metadata needs to be used with the same client instance anyway.

src/rdkafka_cgrp.c
```diff
@@ -3820,7 +3856,9 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_read (
                         "Input buffer is not a serialized "
                         "consumer group metadata object");
 
-        group_id = buf + magic_len;
+        generation_id = le32toh(*(int32_t *)(buf+magic_len));
```
edenhill, line 3859: Unaligned casts not allowed, do memcpy.
And skip the endian swap.

src/rdkafka_cgrp.c
```diff
@@ -3835,8 +3873,8 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_read (
                         RD_KAFKA_RESP_ERR__BAD_MSG,
                         "Input buffer has invalid stop byte");
 
-        /* We now know that group_id is printable-safe and is nul-terminated. */
-        *cgmdp = rd_kafka_consumer_group_metadata_new(group_id);
+        /* We now know that group_id is printable-safe and is null-terminated. */
```
edenhill, line 3876: Please revert.
Ascii `\0` is called Nul, not Null.

This nomenclature helps disambiguating between NULL pointers (4-8 bytes) and NUL (1 byte).

src/rdkafka_cgrp.h
```diff
@@ -183,8 +182,10 @@ typedef struct rd_kafka_cgrp_s {
 
         int32_t            rkcg_generation_id;      /* Current generation id */
 
-        rd_kafka_assignor_t *rkcg_assignor;         /* Selected partition
-                                                     * assignor strategy. */
+        rd_kafka_assignor_t *rkcg_assignor;         /**< The current partition
+                                                     *   assignor. */
+        void              *rkcg_assignor_state;     /**< current partition
```
edenhill, line 187: Is the assignor state needed when not the group leader? If not, put this in rkcg_group_leader

src/rdkafka_request.c
```diff
@@ -238,6 +245,55 @@ rd_kafka_buf_read_topic_partitions (rd_kafka_buf_t *rkbuf,
...
 }
 
+
+/**
+ * @brief Write a list of topic+partitions to \p rkbuf
+ *
+ * @remark The \p assignment list MUST be sorted.
+ */
+void
+rd_kafka_buf_write_assignment (rd_kafka_buf_t *rkbuf,
```
edenhill, line 255: There is a rd_kafka_buf_write_topic_partitions() that can be used

tests/0105-transactions_mock.c
```diff
@@ -177,7 +177,7 @@ static void do_test_txn_recoverable_errors (void) {
                 RD_KAFKA_RESP_ERR_NOT_COORDINATOR,
                 RD_KAFKA_RESP_ERR_CONCURRENT_TRANSACTIONS);
 
-        cgmetadata = rd_kafka_consumer_group_metadata_new("mygroupid");
+        cgmetadata = rd_kafka_consumer_group_metadata_new("mygroupid", 0);
```
edenhill, line 180: No changes to public API, only extensions allowed.

tests/test.c
```diff
@@ -401,6 +402,7 @@ struct test tests[] = {
         _TEST(0110_batch_size, 0),
         _TEST(0111_delay_create_topics, 0, TEST_BRKVER_TOPIC_ADMINAPI,
               .scenario = "noautocreate"),
+        _TEST(0112_cooperative_rebalance, 0),
```
edenhill, line 405: Needs a minimum broker version

src/rdkafka_assignor.c
```diff
@@ -115,22 +120,35 @@ rd_kafka_consumer_protocol_member_metadata_new (
...
+         *   OwnedPartitions => [Topic Partitions] // added in v1
+         *     Topic => string
+         *     Partitions => [int32]
          */
 
         rkbuf = rd_kafka_buf_new(1, 100 + (topic_cnt * 100) + userdata_size);
 
-        rd_kafka_buf_write_i16(rkbuf, 0);
+        /* Version */
+        rd_kafka_buf_write_i16(rkbuf, 1);
```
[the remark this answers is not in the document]
mhowlett: i believe it is (i spent some time explicitly checking all aspects of this).

It's what the java client does: 
1. https://github.com/apache/kafka/blob/2.5/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java#L560
2. https://github.com/apache/kafka/blob/2.5/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L218
3. https://github.com/apache/kafka/blob/2.5/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerProtocol.java#L161

the existing librdkafka impl. ignores the version when reading, and does not care there is additional data:
https://github.com/edenhill/librdkafka/blob/v1.4.2/src/rdkafka_cgrp.c#L865

src/rdkafka_cgrp.c
mhowlett: cool. the allocation did seem unfortunate.

src/rdkafka_assignor.h
mhowlett: `const`: yep. assignors are effectively a constant part of the configuration. there should be const in a lot more places (have made the changes).

`rk`: it generally makes sense for it to be the first argument because it's the focus of many methods (plays the role of a `self`). does this apply here? `rk` is used by the existing assignors only to enable logging of debug messages. Given the java assignor classes actually have no dependency on the client instance, that is good to see. I think it's a bit unfortunate it's a parameter at all, and possibly even misleading to make it the first parameter. that said, it looks weird to me too as is and i'll change it (but see prev comments). the ideal interface would have a `rd_logger_t` or something i think (and that would be one of the later parameters). that seems not-practical/ideal to add at this point. commentary?

mhowlett: i put the `rkas` first because the method is most closely related to this. the relationship isn't as strong as if `rkas` managed dynamic state, but it's still pretty strong. anyway, the root cause of the difficult decision here is that `rk` doesn't really belong.

src/rdkafka.h
```diff
@@ -3762,7 +3762,8 @@ rd_kafka_consumer_group_metadata (rd_kafka_t *rk);
  *         rd_kafka_consumer_group_metadata_destroy().
  */
 RD_EXPORT rd_kafka_consumer_group_metadata_t *
-rd_kafka_consumer_group_metadata_new (const char *group_id);
+rd_kafka_consumer_group_metadata_new (const char *group_id,
```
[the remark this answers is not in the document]
mhowlett: oh, right. is this public because of testing? i was thinking it was private because it's not necessary to expose. we're going to have .._new3 and ..new4 before we're done with this...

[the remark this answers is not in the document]
mhowlett: oh, you say it's typically used for writing tests in the comment. don't mind me.

src/rdkafka_cgrp.c
mhowlett: good question. i took the view that the behavior could be the same as if the protocol isn't set, which results in the error being `RD_KAFKA_RESP_ERR__BAD_MSG`, which is also not explicitly considered. what happens there, afaict, is back off `rkcg_join_intvl` then join the group again. perhaps not catastrophic enough, but i don't think that's too bad. opinion?

src/rdkafka_cgrp.c
mhowlett: i don't think so - it's still relevant. maybe the new subscription overlaps, in which case the old assignment is useful information to the assignor.

also, the java consumer doesn't do this afaict.

src/rdkafka_cgrp.c
mhowlett: you mean `rd_kafka_cgrp_new`? yes.

src/rdkafka_cgrp.c
mhowlett: memcpy: yeah, I shouldn't really need to be told this...
endian swap: yeah, it didn't work initially and i went overboard amending it.

src/rdkafka_cgrp.c
mhowlett: didn't fully appreciate that, reverting. that said, according to wikipedia, i think null is probably more correct in written english.

src/rdkafka_cgrp.h
mhowlett: it's not needed by the leader, only the members.

src/rdkafka_request.c
mhowlett: very good, missed that.

mhowlett:
this is ready for re-review. just tacked on supported protocols field as well.

src/rdkafka_assignor.c
```diff
@@ -431,16 +447,30 @@ rd_kafka_assignor_add (rd_kafka_t *rk,
...
+                               const rd_list_t *topics,
+                               const rd_kafka_topic_partition_list_t
+                               *owned_partitions),
+                       void (*on_assignment_cb) (
+                               const struct rd_kafka_assignor_s *rkas,
+                               void **assignor_state,
+                               const rd_kafka_topic_partition_list_t *assignment,
+                               const rd_kafkap_bytes_t *userdata,
+                               const rd_kafka_consumer_group_metadata_t *rkcgm),
+                       void (*destroy_state) (void *assignor_state),
```
edenhill, line 473: this should have the _cb suffix to be consistent with the other callbacks here.

src/rdkafka_assignor.c
```diff
@@ -855,15 +900,24 @@ int unittest_assignors (void) {
...
+                        rd_kafka_assignor_t *rkas;
 
                         RD_UT_SAY("Test case %s: %s assignor",
                                   tests[i].name,
                                   tests[i].expect[ie].protocol_name);
 
+                        if (!(rkas = rd_kafka_assignor_find(rk,
+                                        tests[i].expect[ie].protocol_name))) {
+                                RD_UT_ASSERT(!err,
+                                             "Assignor case %s for %s failed: "
```
edenhill, line 912: nit: "Assignor test case "

src/rdkafka_assignor.h
```diff
@@ -66,15 +69,14 @@ typedef struct rd_kafka_assignor_s {
         rd_kafkap_str_t   *rkas_protocol_type;
         rd_kafkap_str_t   *rkas_protocol_name;
 
-        const void        *rkas_userdata;
-        size_t             rkas_userdata_size;
-
 	int                rkas_enabled;
 
+        int                rkas_supported_protocols;
```
edenhill, line 74: Add a doc-string saying what the expected flags are, e.g.:
`int           rkas_supported_porotocols;  /**< RD_KAFKA_ASSIGNOR_PROTOCOL_... */`

src/rdkafka_cgrp.c
```diff
@@ -823,7 +829,7 @@ rd_kafka_cgrp_assignor_handle_Metadata_op (rd_kafka_t *rk,
         if (rkcg->rkcg_join_state != RD_KAFKA_CGRP_JOIN_STATE_WAIT_METADATA)
                 return RD_KAFKA_OP_RES_HANDLED; /* From outdated state */
 
-        if (!rkcg->rkcg_group_leader.protocol) {
+        if (!rkcg->rkcg_group_leader.members) {
```
edenhill, line 832: Looks like it should perhaps check rkcg_assignor instead?

src/rdkafka_cgrp.c
```diff
@@ -889,6 +894,11 @@ rd_kafka_group_MemberMetadata_consumer_read (
         rd_kafka_buf_read_bytes(rkbuf, &UserData);
         rkgm->rkgm_userdata = rd_kafkap_bytes_copy(&UserData);
 
+        if (Version >= 1)
```
edenhill, line 897: Could do with a `&&` to avoid the nested ifs.

src/rdkafka_cgrp.c
edenhill: The backoff and retry sounds like the right approach to support rolling upgrades of a group :+1:

src/rdkafka_cgrp.c
```diff
@@ -3665,35 +3695,24 @@ void rd_kafka_cgrp_handle_SyncGroup (rd_kafka_cgrp_t *rkcg,
 		rkbuf->rkbuf_rkb = rd_kafka_broker_internal(rkcg->rkcg_rk);
 
         rd_kafka_buf_read_i16(rkbuf, &Version);
-        rd_kafka_buf_read_i32(rkbuf, &TopicCnt);
-
-        if (TopicCnt > 10000) {
-                err = RD_KAFKA_RESP_ERR__BAD_MSG;
+        if (!(assignment = rd_kafka_buf_read_topic_partitions(rkbuf, 0, rd_false)))
                 goto err;
```
edenhill, line 3699: It needs to go to `err_parse` so that the rkbuf->err (set in buf_read_topic..) will be used, otherwise we'll come to the error label without an error code set.

src/rdkafka_cgrp.c
```diff
@@ -3738,17 +3756,40 @@ rd_kafka_consumer_group_metadata_new (const char *group_id) {
 
         cgmetadata = rd_calloc(1, sizeof(*cgmetadata));
         cgmetadata->group_id = rd_strdup(group_id);
+        cgmetadata->generation_id = -1;
+
+        return cgmetadata;
+}
+
+rd_kafka_consumer_group_metadata_t *
+rd_kafka_consumer_group_metadata_new2 (const char *group_id,
```
edenhill, line 3765: Suggest changing group_metadata_new() to simply call this function to avoid code duplication.

src/rdkafka_cgrp.c
```diff
@@ -3785,13 +3827,17 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_write (
         size_t of = 0;
         size_t magic_len = sizeof(rd_kafka_consumer_group_metadata_magic);
         size_t groupid_len = strlen(cgmd->group_id) + 1;
+        size_t generationid_len = sizeof(int32_t);
```
edenhill, line 3830: `sizeof(cgmd->generation_id)`

src/rdkafka_cgrp.c
```diff
@@ -3805,10 +3851,12 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_read (
         rd_kafka_consumer_group_metadata_t **cgmdp,
         const void *buffer, size_t size) {
         size_t magic_len = sizeof(rd_kafka_consumer_group_metadata_magic);
+        size_t generationid_len = sizeof(int32_t);
```
edenhill, line 3854: `sizeof(generation_id)`

src/rdkafka_cgrp.c
```diff
@@ -3820,7 +3868,9 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_read (
                         "Input buffer is not a serialized "
                         "consumer group metadata object");
 
-        group_id = buf + magic_len;
+        memcpy(&generation_id, buf+magic_len, sizeof(generation_id));
```
edenhill, line 3871: This copy is unsafe, you haven't checked the length yet.
And why is it using sizeof(generation_id) instead of generationid_len? Pick one.

src/rdkafka_request.c
```diff
@@ -1173,44 +1184,18 @@ static void rd_kafka_group_MemberState_consumer_write (
...
-
-        if (of_PartCnt != -1)
-                rd_kafka_buf_update_i32(rkbuf, of_PartCnt, PartCnt);
-        rd_kafka_buf_update_i32(rkbuf, of_TopicCnt, TopicCnt);
-
+        rd_assert(rkgm->rkgm_assignment);
+        rd_kafka_buf_write_topic_partitions(
+                rkbuf,
+                rkgm->rkgm_assignment,
+                rd_false /*skip invalid offsets*/,
```
edenhill, line 1195: dito: let the comment say what it is doing rather than what it is disabling.

src/rdkafka_sticky_assignor.c
```diff
@@ -0,0 +1,194 @@
+/*
```
edenhill, line 1: Skipping review of this file

tests/0112-cooperative_rebalance.cpp
```diff
@@ -0,0 +1,115 @@
+/*
```
edenhill, line 1: Skipping review for now

tests/test.c
edenhill: Not addressed

src/rdkafka.h
```diff
@@ -3765,6 +3765,18 @@ RD_EXPORT rd_kafka_consumer_group_metadata_t *
...
 
+/**
+ * @brief Create a new consumer group metadata object.
+ *        This is typically only used for writing tests.
+ *
+ * @remark The returned pointer must be freed by the application using
+ *         rd_kafka_consumer_group_metadata_destroy().
+ */
+RD_EXPORT rd_kafka_consumer_group_metadata_t *
+rd_kafka_consumer_group_metadata_new2 (const char *group_id,
```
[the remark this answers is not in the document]
mhowlett: sounds good. i wasn't too concerned because this shouldn't be used by users generally.

src/rdkafka_assignor.c
```diff
@@ -115,22 +120,41 @@ rd_kafka_consumer_protocol_member_metadata_new (
...
+        /* Following data is ignored by v0 consumers */
+        if (!owned_partitions)
+                /* If there are no owned partitions, this is specified as an
+                 * empty array, not NULL. */
+                rd_kafka_buf_write_i32(rkbuf, 0); /* Topic count */
+        else
+                rd_kafka_buf_write_topic_partitions(
+                        rkbuf,
+                        owned_partitions,
+                        rd_false /*skip invalid offsets*/,
```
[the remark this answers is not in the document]
mhowlett: yep, that's consistent with what you do elsewhere too. changing.

src/rdkafka_assignor.c
mhowlett: yep, also `rkas_destroy_state_cb`

src/rdkafka_cgrp.c
mhowlett: this will also be set in the normal members (which are what requires the state), so no. i will add a comment to `rkcg_assignor` in `rdkafka_cgrp.h` to note this.

src/rdkafka_cgrp.c
mhowlett: haha, i forgot i wasn't in a constructor

src/rdkafka_cgrp.c
mhowlett: whoops-a-daisy

src/rdkafka_cgrp.c
```diff
@@ -3762,15 +3762,11 @@ rd_kafka_consumer_group_metadata_new (const char *group_id) {
...
+rd_kafka_consumer_group_metadata_new_with_genid (const char *group_id,
+                                                 int32_t generation_id) {
         rd_kafka_consumer_group_metadata_t *cgmetadata;
 
-        if (!group_id)
-                return NULL;
-
-        cgmetadata = rd_calloc(1, sizeof(*cgmetadata));
-        cgmetadata->group_id = rd_strdup(group_id);
+        cgmetadata = rd_kafka_consumer_group_metadata_new(group_id);
```
edenhill, line 3769: nit: I would flip this the other way, have metadata_new() call the richer new_with_genid(), so that there is only a single real constructor.

tests/test.c
```diff
@@ -402,7 +402,7 @@ struct test tests[] = {
         _TEST(0110_batch_size, 0),
         _TEST(0111_delay_create_topics, 0, TEST_BRKVER_TOPIC_ADMINAPI,
               .scenario = "noautocreate"),
-        _TEST(0112_cooperative_rebalance, 0),
+        _TEST(0112_cooperative_rebalance, 0, TEST_F_LOCAL, TEST_BRKVER(2,4,0,0)),
```
edenhill, line 405: TEST_F_LOCAL means it does not need a cluster, which conflicts with specifying a minimum broker version with TEST_BRKVER. which is it? :)

src/rdkafka_cgrp.c
mhowlett: whoops. i was struggling with a 35 degree day in an oven with no aircon. round 2 today.

edenhill: +1 for C!

edenhill:
There is a build failure:
https://travis-ci.org/github/edenhill/librdkafka/jobs/691794391#L725

edenhill:
This PR will need to be squashed due to unclean commit history, but that should be fine since it is basically one change

Issue #2885, opened 2020-05-13
`rdkafka_mock.h` does not check for c++ compiler


mpribble:
Description
===========
`rdkafka_mock.h` does not check if it is being compiled with a c++ compiler.

How to reproduce
================
Include `rdkafka_mock.h` in a c++ program. In the program, use one or more the mock functions (in my case, `rd_kafka_handle_mock_cluster()` and `rd_kafka_mock_cluster_bootstraps()`). Compile the program using a c++ compiler (I used clang), linking against `librdkafka`

This will result in undefined reference link errors. This is because the c++ compiler mangles the names of the symbols it included from `rdkafka_mock.h`, but the c compiler that built `librdkafka` does not.

A workaround (which I am using), is to  wrap the include of `rdkafka_mock.h` in `extern "C" {...}`. I.e.,

`extern "C" {  #include "rdkafka_mock.h" }`

This works of course, but I think it would be better if `rdkafka_mock.h` does this conditionally based on compiler in the same way `rdkafka.h` does.

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `1.4.2`
 - [x] Apache Kafka version: N/A
 - [x] librdkafka client configuration: N/A
 - [x] Operating system: N/A
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka N/A
 - [x] Provide broker log excerpts N/A
 - [ ] Critical issue: No - this is low priority

edenhill:
Aight, care to file a PR that copies what rdkafka.h does?

mpribble:
yeah sure, shouldn't be a problem.

mpribble:
Well, that was actually my first ever PR on github. Hopefully I did it right.

edenhill:
:tada: you did!

PR #2887, opened 2020-05-14
mock: Make rdkafka_mock.h c++ compatible (#2885)

mpribble:
Allow rdkafka_mock.h to be included from c and c++ programs.

### DIFF
--- a/src/rdkafka_mock.h
+++ b/src/rdkafka_mock.h
@@ -33,6 +33,12 @@
 #error "rdkafka_mock.h must be included after rdkafka.h"
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+#if 0
+} /* Restore indent */
+#endif
+#endif
 
 
 /**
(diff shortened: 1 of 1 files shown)


src/rdkafka_mock.h
```diff
@@ -33,6 +33,9 @@
 #error "rdkafka_mock.h must be included after rdkafka.h"
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+#endif
```
edenhill, line 39: Can you do the indent trick to keep the IDE from indenting all code to the extern level?
Like this:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L53

mpribble: Done. I went ahead and squashed the change into the original commit. I **think** that's what your contributor instructions say.

edenhill: You're acing this PR thing! 👍

mpribble: Thanks the vote of confidence. Luckily I use gitlab on a regular basis, and it's basically a copy/paste of github. Just happy to finally be making my first open source contribution, even a tiny one like this...feels good.

edenhill:
Thank you!

PR #2891, opened 2020-05-19
Add simple generic and strictly typed hash maps

edenhill:
This will be used by the sticky assignor, but thought I'd easen up that future PR by filing this one separately.

### DIFF
--- a/src/rdmap.c
+++ b/src/rdmap.c
@@ -1,0 +1,476 @@
+/*
+ * librdkafka - The Apache Kafka C/C++ library
+ *
+ * Copyright (c) 2020 Magnus Edenhill
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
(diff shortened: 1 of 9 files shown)


edenhill:
A proper review would be good, it is written from scratch.
But we can postpone until the sticky PR if you want.

src/rdmap.c
```diff
@@ -0,0 +1,487 @@
...
+ */
+
+#include "rd.h"
+#include "rdsysqueue.h"
+#include "rdstring.h"
+#include "rdmap.h"
+
+
+static RD_INLINE
+int rd_map_elem_cmp (const void *_a, const void *_b, const rd_map_t *rmap) {
```
mhowlett, line 36: why can't the parameters be `rd_map_elem_t *`?

src/rdmap.c
```diff
@@ -0,0 +1,487 @@
...
+rd_map_elem_t *rd_map_get_or_create (rd_map_t *rmap, const void *key) {
+        const rd_map_elem_t skel = { .key = (void *)key,
+                                     .hash = rmap->rmap_hash(key) };
+        rd_map_elem_t *elem;
+        int bkt;
+
+        if ((elem = rd_map_find(rmap, &bkt, &skel)))
+                return elem;
+
+        return rd_map_insert(rmap, bkt, &skel);
```
mhowlett, line 131: this takes ownership of `key`, but `rd_map_find` does not, so the caller of `rd_map_get_or_create` can't know if they need to free `key` or not, right?

src/rdmap.c
```diff
@@ -0,0 +1,487 @@
...
+                236897,
+                -1
+        };
+        struct rd_map_buckets buckets = RD_ZERO_INIT;
+        int i;
+
+        if (!expected_cnt) {
+                buckets.cnt = 199;
+        } else {
+                /* Strive for a maximum depth of 15 elements per bucket, but
```
mhowlett, line 191: maximum -> average (at capacity)

src/rdmap.c
```diff
@@ -0,0 +1,487 @@
...
+                 * limit the maximum bucket size.
```
mhowlett, line 192: "maximum bucket count"

src/rdmap.c
```diff
@@ -0,0 +1,487 @@
...
+
+
+/**
+ * @brief Test typed hash map with pre-defined type.
+ */
+static int unittest_typed_map (void) {
+        ut_my_typed_map_t rmap = RD_MAP_INITIALIZER(0,
+                                                    mykey_cmp, mykey_hash,
+                                                    NULL, NULL);
+        struct mykey k1 = { 1 };
```
mhowlett, line 297: maybe better to explicitly specify a value for `something_else`. side note: c language seems generally screwed up to me 
on struct initialization.

src/rdmap.c
```diff
@@ -0,0 +1,487 @@
...
+        return 0;
+}
+
+
+/**
+ * @brief Untyped hash map.
+ *
+ * This is a more thorough test of the underlying hash map implementation.
+ */
+static int unittest_untyped_map (void) {
```
mhowlett, line 385: maybe test `rd_map_get_or_create`

mhowlett:
lgtm after comments addressed to your satisfaction.

src/rdmap.c
edenhill: good spotting!

src/rdmap.c
edenhill: Good point, since this method isn't used I'll remove it.

src/rdmap.c
edenhill: Since something_else isn't used the auto initialization to zero should be fine.
What's the screwy part?

src/rdmap.c
edenhill: or remove it, much easier!

PR #2899, opened 2020-05-22
the configure script does not work correctly when configuration is  like "--LDFLAGS='-Wl,--build-id=none'"

sky92zwq:
`./configure --LDFLAGS='-Wl,--build-id=none'`

there are two '=' in a configration, then ` name="${opt%=*}"` will Intercept `LDFLAGS=-Wl,--build-id` as `name`. 

so ` name="${opt%%=*}"`may be a good choice to Intercept LDFLAGS out.

### DIFF
--- a/configure
+++ b/configure
@@ -81,7 +81,7 @@ while [[ ! -z $@ ]]; do
     shift
 
     if [[ $opt = *=* ]]; then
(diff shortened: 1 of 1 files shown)


edenhill:
Thank you!

Issue #2895, opened 2020-05-19
producev without varargs


andrewthad:
This is not an issue so much as a request for an additional function in the API. Currently, the most general and flexible way to create a single message with headers is `rd_kafka_producev`. This is, for example, the only way to create a message with headers. However, this function uses varargs, which makes it difficult to bind to in most high-level languages. The varargs are used to build a `rd_kafka_msg_t`, which is then fed into `rd_kafka_msg_partitioner`. It would make it easier to interface with librdkafka from high-level languages if the user could construct a `rd_kafka_msg_t` directly and then feed it into some `produce*` variant. I'm not terribly well versed in C library conventions, but I assume that this was not done because adding any field to `rd_kafka_msg_t` would be a breaking change in the API. I'm not sure if there is a way around that issue or not.

edenhill:
We could make a produce variant that takes an array of RD_KAFKA_V_..s.
E.g.:
```c

typedef union {
  rd_kafka_topic_t *rkt;
  const char *const_str;
  char *str;
  int i;
  int32 i32;
  int64 i64;
 ...
} rd_kafka_v_t ;

rd_kafka_error_t rd_kafka_produceva(rd_kafka_t *rk, rd_kafka_v_t *vs, size_t v_cnt);
```

What do you think?

andrewthad:
Since C unions are not tagged, I think that you would need to tag them explicitly with something like:

    typedef union { ... } rd_kafka_v_t; // As in your example
    enum rd_field {topic, flags, opaque, ...};
    typedef union {
       rd_field field;
       rd_kafka_v_t val;
    } rd_kafka_vp_t;
    rd_kafka_error_t *rd_kafka_produceva(rd_kafka_t *rk, rd_kafka_vp_t *vs, size_t v_cnt);

But aside from that quibble, the union-style approach would work fine for my case.

edenhill:
Please review this PR: https://github.com/edenhill/librdkafka/pull/2902

PR #2902, opened 2020-05-25
Add rd_kafka_produceva() using a vtype array as an alternative to va-args

edenhill:
It is problematic to construct va-arg lists for calling C in some high-level languages that wrap librdkafka, erlang in the case of #2895. 
This new method allows constructing an array of structs instead.

### DIFF
--- a/src/rdkafka_msg.c
+++ b/src/rdkafka_msg.c
@@ -35,6 +35,7 @@
 #include "rdkafka_header.h"
 #include "rdkafka_idempotence.h"
 #include "rdkafka_txnmgr.h"
+#include "rdkafka_error.h"
 #include "rdcrc32.h"
 #include "rdfnv1a.h"
 #include "rdmurmur2.h"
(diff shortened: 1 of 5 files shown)


andrewthad:
This looks like exactly what I was hoping for. Thanks. It's probably worth getting feedback from maintainers of librdkafka bindings in other languages. The bindings I'm working on are for Haskell, and I can confirm that this strategy works well for Haskell's foreign function interface and FFI-related tooling.

src/rdkafka_msg.c
```diff
@@ -330,6 +354,188 @@ int rd_kafka_msg_new (rd_kafka_topic_t *rkt, int32_t force_partition,
...
+        if (unlikely(err)) {
+                error = rd_kafka_error_new(
+                        err,
+                        "Failed to produce message: %s",
+                        rd_kafka_err2str(err));
+                goto err;
+        }
+
+        /* Partition the message */
+        err = rd_kafka_msg_partitioner(rkt, rkm, 1);
```
mhowlett, line 491: surprised to see this here, probably some unnecessary duplication of stuff, but not a big deal.

tests/0074-producev.c
```diff
@@ -57,6 +58,26 @@ static void do_test_srkt_leak (void) {
...
+        vus[0].u.cstr = "test";
+        vus[1].vtype = RD_KAFKA_VTYPE_VALUE;
+        vus[1].u.mem.ptr = buf;
+        vus[1].u.mem.size = sizeof(buf);
+        vus[2].vtype = RD_KAFKA_VTYPE_HEADER;
+        vus[2].u.header.name = "testheader";
+        vus[2].u.header.val = "test value";
+        vus[2].u.header.size = -1;
+
+        error = rd_kafka_produceva(rk, vus, 3);
```
mhowlett, line 71: test coverage is a bit light on

src/rdkafka_msg.c
edenhill: The produceva() code is a direct copy of producev().

PR #2905, opened 2020-05-27
Only trigger consumer group rejoin on complete Metadata responses

edenhill:
..and some other fixes

### DIFF
--- a/src/crc32c.c
+++ b/src/crc32c.c
@@ -400,8 +400,6 @@ int unittest_crc32c (void) {
         uint32_t crc;
         const char *how;
 
-        crc32c_global_init();
-
 #if WITH_CRC32C_HW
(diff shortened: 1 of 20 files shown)


src/crc32c.c
```diff
@@ -400,8 +400,6 @@ int unittest_crc32c (void) {
         uint32_t crc;
         const char *how;
 
-        crc32c_global_init();
```
mhowlett, line 403: doesn't seem right to remove this?

edenhill: https://github.com/edenhill/librdkafka/commit/3237ed9020f477edd480d4dac07b91f9ca72f3ff

mhowlett: oh, i see you do a global init.

PR #2907, opened 2020-05-28
Test 0097 was always skipped, and added new test case.

edenhill:


tests/0097-ssl_verify.cpp
```diff
@@ -34,6 +34,43 @@
...
+   /* [RdKafka::CERT_CA] = */
+   {
+    "RDK_SSL_pkcs",
+    "RDK_SSL_ca_der",
+    "RDK_SSL_ca_pem",
+   }
+  };
+
+
+static std::vector<char> read_file (const std::string path) {
```
mhowlett, line 60: is this generally useful enough to put somewhere common?

edenhill: rule of nineteen applies here.

PR #2908, opened 2020-05-28
lost partition notification for EAGER rebalance protocol

mhowlett:
ok, let's see what you think of this.

many less lines of code and much cleaner than adding `RD_KAFKA_RESP_ERR__LOST_PARTITIONS`, which is a good indicator it's better.

### DIFF
--- a/src-cpp/KafkaConsumerImpl.cpp
+++ b/src-cpp/KafkaConsumerImpl.cpp
@@ -144,7 +144,7 @@ RdKafka::KafkaConsumerImpl::assignment (std::vector<RdKafka::TopicPartition*> &p
 
 bool
 RdKafka::KafkaConsumerImpl::assignment_lost () {
-  return rd_kafka_assignment_lost(rk_);
+  return rd_kafka_assignment_lost(rk_) ? true : false;
 }
 
 
--- a/src-cpp/KafkaConsumerImpl.cpp
+++ b/src-cpp/KafkaConsumerImpl.cpp
@@ -141,6 +141,14 @@ RdKafka::KafkaConsumerImpl::assignment (std::vector<RdKafka::TopicPartition*> &p
 }
 
 
+
+bool
+RdKafka::KafkaConsumerImpl::assignment_lost () {
+  return rd_kafka_assignment_lost(rk_);
+}
+
+
+
 RdKafka::ErrorCode
 RdKafka::KafkaConsumerImpl::subscription (std::vector<std::string> &topics) {
   rd_kafka_topic_partition_list_t *c_topics;
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -2365,21 +2365,6 @@ public:
   virtual ErrorCode assignment (std::vector<RdKafka::TopicPartition*> &partitions) = 0;
 
 
-  /** @brief Check whether the consumer considers the current assignment to
-   *         have been lost involuntarily. This method is only applicable for
-   *         use with a subscribing consumer. Assignments are revoked
-   *         immediately when determined to have been lost, so this method is
-   *         only useful within a rebalance callback.
-   *
-   * @remark Calling rd_kafka_assign(), rd_kafka_incremental_assign() or
-   *         rd_kafka_incremental_unassign() resets this flag.
-   *
-   * @returns Returns true if the current partition assignment is considered
-   *          lost, false otherwise.
-   */
-  virtual bool assignment_lost () = 0;
-
-
   /** @brief Returns the current subscription as set by
    *         RdKafka::KafkaConsumer::subscribe() */
   virtual ErrorCode subscription (std::vector<std::string> &topics) = 0;
@@ -2655,6 +2640,22 @@ public:
    */
   virtual ConsumerGroupMetadata *groupMetadata () = 0;
 
+
+  /** @brief Check whether the consumer considers the current assignment to
+   *         have been lost involuntarily. This method is only applicable for
+   *         use with a subscribing consumer. Assignments are revoked
+   *         immediately when determined to have been lost, so this method is
+   *         only useful within a rebalance callback. Partitions that have
+   *         been lost may already be owned by other members in the group and
+   *         therefore commiting offsets, for example, may fail.
+   *
+   * @remark Calling rd_kafka_assign(), rd_kafka_incremental_assign() or
+   *         rd_kafka_incremental_unassign() resets this flag.
+   *
+   * @returns Returns true if the current partition assignment is considered
+   *          lost, false otherwise.
+   */
+  virtual bool assignment_lost () = 0;
 
 };
 
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -2364,6 +2364,22 @@ public:
    *         RdKafka::KafkaConsumer::assign() */
   virtual ErrorCode assignment (std::vector<RdKafka::TopicPartition*> &partitions) = 0;
 
+
+  /** @brief Check whether the consumer considers the current assignment to
+   *         have been lost involuntarily. This method is only applicable for
+   *         use with a subscribing consumer. Assignments are revoked
+   *         immediately when determined to have been lost, so this method is
+   *         only useful within a rebalance callback.
+   *
+   * @remark Calling rd_kafka_assign(), rd_kafka_incremental_assign() or
+   *         rd_kafka_incremental_unassign() resets this flag.
+   *
+   * @returns Returns true if the current partition assignment is considered
+   *          lost, false otherwise.
+   */
+  virtual bool assignment_lost () = 0;
+
+
   /** @brief Returns the current subscription as set by
    *         RdKafka::KafkaConsumer::subscribe() */
   virtual ErrorCode subscription (std::vector<std::string> &topics) = 0;
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -2364,7 +2364,6 @@ public:
    *         RdKafka::KafkaConsumer::assign() */
   virtual ErrorCode assignment (std::vector<RdKafka::TopicPartition*> &partitions) = 0;
 
-
   /** @brief Returns the current subscription as set by
    *         RdKafka::KafkaConsumer::subscribe() */
   virtual ErrorCode subscription (std::vector<std::string> &topics) = 0;
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -2649,8 +2649,8 @@ public:
    *         been lost may already be owned by other members in the group and
    *         therefore commiting offsets, for example, may fail.
    *
-   * @remark Calling rd_kafka_assign(), rd_kafka_incremental_assign() or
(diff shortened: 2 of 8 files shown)


src/rdkafka_cgrp.c
```diff
@@ -1409,7 +1410,7 @@ static void rd_kafka_cgrp_rejoin (rd_kafka_cgrp_t *rkcg) {
                      rd_kafka_cgrp_join_state_names[rkcg->rkcg_join_state],
                      rkcg->rkcg_assignment ? "" : "out");
 
-        rd_kafka_cgrp_rebalance(rkcg, "group rejoin");
+        rd_kafka_cgrp_rebalance(rkcg, rd_false/*assignment not lost*/, "group rejoin");
```
edenhill, line 1413: 80 cols

src/rdkafka_cgrp.c
```diff
@@ -2742,26 +2747,30 @@ static void rd_kafka_cgrp_group_leader_reset (rd_kafka_cgrp_t *rkcg,
 
 
 /**
- * @brief Group is rebalancing, trigger rebalance callback to application,
+ * @brief Initiate a rebalance. Trigger rebalance callback to application,
```
edenhill, line 2750: Technically it is the group coordinator that initiates a rebalance, not the client, so I think the former wording is more correct.

src/rdkafka_cgrp.c
```diff
@@ -2742,26 +2747,30 @@ static void rd_kafka_cgrp_group_leader_reset (rd_kafka_cgrp_t *rkcg,
...
  *        and transition to INIT state for (eventual) rejoin.
  */
 static void rd_kafka_cgrp_rebalance (rd_kafka_cgrp_t *rkcg,
+                                     rd_bool_t assignment_lost,
                                      const char *reason) {
 
         rd_kafka_dbg(rkcg->rkcg_rk, CONSUMER|RD_KAFKA_DBG_CGRP, "REBALANCE",
-                     "Group \"%.*s\" is rebalancing in "
-                     "state %s (join-state %s) %s assignment: %s",
+                     "Group \"%.*s\" rebalance initiated in "
```
edenhill, line 2758: dito

src/rdkafka_cgrp.c
```diff
@@ -2742,26 +2747,30 @@ static void rd_kafka_cgrp_group_leader_reset (rd_kafka_cgrp_t *rkcg,
...
                      rd_kafka_cgrp_state_names[rkcg->rkcg_state],
                      rd_kafka_cgrp_join_state_names[rkcg->rkcg_join_state],
                      rkcg->rkcg_assignment ? "with" : "without",
+                     assignment_lost ? " (lost)" : "",
                      reason);
 
         rd_snprintf(rkcg->rkcg_c.rebalance_reason,
                     sizeof(rkcg->rkcg_c.rebalance_reason), "%s", reason);
 
+        rkcg->rkcg_assignment_lost = rd_true;
```
edenhill, line 2770: should be = assignment_lost;, right?

src/rdkafka_cgrp.c
```diff
@@ -2866,7 +2876,8 @@ rd_kafka_cgrp_unsubscribe (rd_kafka_cgrp_t *rkcg, int leave_group) {
 	if (leave_group)
 		rkcg->rkcg_flags |= RD_KAFKA_CGRP_F_LEAVE_ON_UNASSIGN;
 
-        rd_kafka_cgrp_rebalance(rkcg, "unsubscribe");
+        rd_kafka_cgrp_rebalance(rkcg, rd_false/*assignment not lost*/,
```
edenhill, line 2879: Isn't an unsubscribe technically a loss of the asignment too?

src/rdkafka_subscription.c
```diff
@@ -112,6 +112,16 @@ rd_kafka_assign (rd_kafka_t *rk,
...
 
 
+int
+rd_kafka_assignment_lost (rd_kafka_t *rk) {
+        rd_kafka_cgrp_t *rkcg;
+
+        if (!(rkcg = rd_kafka_cgrp_get(rk)))
+                return 0;
+
+        return rkcg->rkcg_assignment_lost ? 1 : 0;
```
edenhill, line 122: All rkcg_assignment_lost accesses needs locking, or changing it to an atomic32 (which is easier), since this is called from an application thread, while rkcg_assignment_lost is set from the rdkafka main thread,

src/rdkafka_cgrp.c
mhowlett: yes, good point. what i was trying to capture is that we have the client initiating a revoke. I think the wording "group ... is rebalancing in state" is not precise / informative enough - if i read that in the logs, i'm going to wonder exactly what it means. if i read that the client is initiating a revoke, i have a very good idea what that means. i will change the wording and see what you think.

mhowlett: ... the client may be acting under instruction from the group coordinator, but it also may not (unsubscribe, or assignment lost).

mhowlett: i've reverted this because i'm struggling to work out what to do and the COOPERATIVE protocol is going to put this in a different perspective again, so will defer doing anything until making those changes.

mhowlett:
> Since assignment_lost is cleared in the poll callback now (but not in the event_t handlers) we're half way there.

I think you mean via `background_event_cb`. I believe I follow the path of how all this works. I think this is a good example of the primary disadvantage of the queuing / async architecture of librdkafka - at the point the event is produced, there is no easy way to directly trace through the follow on effects. By contrast, with synchronous code, that is easy. To be able to work on librdkafka, I feel i need a pretty good mental model of how everything fits together to have reasonable confidence i'm not missing anything. Of course async message architecture has some advantages too.

mhowlett:
> My gut feeling is that we want the assignment_lost to live on the OP_REBALANCE so/since it is associated with a specific rebalance, which would also void the need for locking the rkcg_assignment_lost.

I think that's definitely the best way to be modeling this requirement, but I don't see how you can expose that via a `rd_kafka_assignment_lost (rd_kafka_t *rk)` method. It's what I was effectively doing before when I had separate REVOKE and LOST event types.

`rd_kafka_assignment_lost` is conceptually checking the state of rk (cgrp), so the impl. matches that, so i think it's good.

I think this come down to deciding for / against the additional rebalance err type. I'm still liking the current method, despite it not matching up to the problem perfectly (but nor does a separate LOST event type i think, since I think that should be probably more a property of a REVOKE event).

src/rdkafka.h
```diff
@@ -3674,6 +3674,17 @@ rd_kafka_assignment (rd_kafka_t *rk,
                      rd_kafka_topic_partition_list_t **partitions);
 
 
+/**
+ * @brief Check whether the consumer considers the current assignment to
+ *        have been lost involuntarily. This method is only useful when
+ *        called from within a group rebalance callback (rebalance_cb)
```
edenhill, line 3680: "only useful on a subscribing high-level consumer instance when .."

`@remark Calling rd_kafka_assign() resets this flag`  - or something along those lines.
This is important to state so that people dont do:
```c
    assign(NULL);
    if (assignment_lost())
        yada;
```

edenhill: Also: "a group rebalance callback... or RD_KAFKA_EVENT_REBALANCE .."

src/rdkafka.h
```diff
@@ -3674,6 +3674,17 @@ rd_kafka_assignment (rd_kafka_t *rk,
...
+ *        because lost assignments are revoked immediately.
+ *
+ * @returns Returns 1 if the current partition assignment is considered
+ *          lost, 0 otherwise.
+ */
+RD_EXPORT int
```
edenhill, line 3686: Also add to rdkafkacpp.h

src/rdkafka.h
mhowlett: good points.

src-cpp/KafkaConsumerImpl.cpp
```diff
@@ -141,6 +141,14 @@ RdKafka::KafkaConsumerImpl::assignment (std::vector<RdKafka::TopicPartition*> &p
 }
 
 
+
+bool
+RdKafka::KafkaConsumerImpl::assignment_lost () {
+  return rd_kafka_assignment_lost(rk_);
```
edenhill, line 147: This needs an ternary check to avoid warnings on msvc:
` ..lost(rk_) ? true : false`

src-cpp/rdkafkacpp.h
```diff
@@ -2357,6 +2357,22 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
    *         RdKafka::KafkaConsumer::assign() */
   virtual ErrorCode assignment (std::vector<RdKafka::TopicPartition*> &partitions) = 0;
 
+
+  /** @brief Check whether the consumer considers the current assignment to
+   *         have been lost involuntarily. This method is only applicable for
+   *         use with a subscribing consumer. Assignments are revoked
+   *         immediately when determined to have been lost, so this method is
+   *         only useful within a rebalance callback.
```
edenhill, line 2365: Should also say what this means for the application, like not able to commit offsets, right?

src-cpp/rdkafkacpp.h
```diff
@@ -2357,6 +2357,22 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
...
+   *
+   * @remark Calling rd_kafka_assign(), rd_kafka_incremental_assign() or
+   *         rd_kafka_incremental_unassign() resets this flag.
+   *
+   * @returns Returns true if the current partition assignment is considered
+   *          lost, false otherwise.
+   */
+  virtual bool assignment_lost () = 0;
```
edenhill, line 2373: New methods must be added to the end of the virtual class to not break ABI compat.

mhowlett: so many rules!

src-cpp/KafkaConsumerImpl.cpp
mhowlett: this is so much easier in languages that give you squiggly red lines in your ide and are consistent across platforms.

src-cpp/rdkafkacpp.h
mhowlett: good idea

mhowlett:
shall we get this merged? the incr assign pr is pretty much ready to open now i think.

src-cpp/rdkafkacpp.h
```diff
@@ -2649,6 +2634,22 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
...
 
+  /** @brief Check whether the consumer considers the current assignment to
+   *         have been lost involuntarily. This method is only applicable for
+   *         use with a subscribing consumer. Assignments are revoked
+   *         immediately when determined to have been lost, so this method is
+   *         only useful within a rebalance callback. Partitions that have
+   *         been lost may already be owned by other members in the group and
+   *         therefore commiting offsets, for example, may fail.
+   *
+   * @remark Calling rd_kafka_assign(), rd_kafka_incremental_assign() or
```
edenhill, line 2645: This should reference the C++ API names (and make sure to run `make docs` to verify that Doxygen is able to pick them up correctly (it will warn otherwise I think)).

PR #2911, opened 2020-05-31
Add some error message when failed in decompressing kafka message in gzip

wbb1975:
In AWS Lambda enviroment, sometimes rd_kafka_msgset_reader_decompress will fail in decompressing kafka messages in gzip because of no enough memory. But from existing logs, you couldn't find any clue about it--instead the system will tell you an invalid compressed data.

### DIFF
--- a/src/rdgz.c
+++ b/src/rdgz.c
@@ -32,7 +32,6 @@
 #include "rdgz.h"
 
 #include <zlib.h>
-#include "rdbuf.h"
 
 #define RD_GZ_CHUNK  262144
 
--- a/src/rdgz.c
+++ b/src/rdgz.c
@@ -26,15 +26,17 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "rdkafka_int.h"
+
 #include "rd.h"
 #include "rdgz.h"
 
 #include <zlib.h>
-
+#include "rdbuf.h"
 
 #define RD_GZ_CHUNK  262144
 
-void *rd_gz_decompress (const void *compressed, int compressed_len,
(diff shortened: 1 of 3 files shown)


src/rdgz.c
```diff
@@ -26,15 +26,16 @@
...
 #include "rd.h"
 #include "rdgz.h"
 
 #include <zlib.h>
 
-
 #define RD_GZ_CHUNK  262144
 
-void *rd_gz_decompress (const void *compressed, int compressed_len,
+void *rd_gz_decompress (rd_kafka_broker_t *rkb, const void *compressed, int compressed_len,
```
edenhill, line 38: This is intertwining abstractions.
Better to pass an errstr and errstr_size to rd_gz_decompress where it can write an error string which can then be logged from the caller.

src/rdgz.c
```diff
@@ -57,13 +58,17 @@ void *rd_gz_decompress (const void *compressed, int compressed_len,
 		int r;
 		
 		if ((r = inflateInit2(&strm, 15+32)) != Z_OK)
+		{
```
edenhill, line 61: Code style is open-brace on same line as if-statement.

src/rdgz.c
```diff
@@ -57,13 +58,17 @@ void *rd_gz_decompress (const void *compressed, int compressed_len,
...
+			rd_rkb_log(rkb, LOG_ERR, "GZIP", "inflateInit2 failed %d",  r);
```
edenhill, line 62: This error message is not very useful for the use which would need to check the Z_ERROR_.. defines to know what the error was.
Translate the error code to a string.

PR #2913, opened 2020-06-02
Txn fixes: crash on error, and setting retriable flag

edenhill:


edenhill:
Added tests, additional fixes.
Please re-review as soon as possible.

CHANGELOG.md
```diff
@@ -74,6 +76,26 @@ librdkafka.
 
 ### Producer fixes
 
+ * Transactional producer could crash on request timeout due to dereferencing
```
mhowlett, line 79: should these be in the v1.5.0 section given they actually arrived in v1.4.4 according to the section below?

CHANGELOG.md
```diff
@@ -74,6 +76,26 @@ librdkafka.
...
+
+# librdkafka v1.4.4
+
+v1.4.4 is a maintenance release with the following fixes and enhancements:
+
+ * Transactional producer could crash on request timeout due to dereferencing
+   NULL pointer of non-existent response object (ESC-3009)
+ * Mark `rd_kafka_send_offsets_to_transaction()` CONCURRENT_TRANSACTION (et.al)
+   errors as retriable (ESC-2998)
+ * Fix theoeretical crash on coord request failure.
```
mhowlett, line 94: what is a theoretical crash?

src/rdkafka_buf.c
```diff
@@ -407,9 +408,12 @@ void rd_kafka_buf_handle_op (rd_kafka_op_t *rko, rd_kafka_resp_err_t err) {
         response = request->rkbuf_response; /* May be NULL */
         request->rkbuf_response = NULL;
 
-        rd_kafka_buf_callback(request->rkbuf_rkb->rkb_rk,
-			      request->rkbuf_rkb, err,
-                              response, request);
+        if (!(rk = rko->rko_rk)) {
```
mhowlett, line 411: haven't looked through any of the details, but the need for this kind of logic seems unfortunate.

CHANGELOG.md
edenhill: Good point!

CHANGELOG.md
edenhill: I effed something else up which triggered this crash, but we havent seen any reports of it in real life.

src/rdkafka_buf.c
edenhill: Yeah, we should probably always set rko_rk on ops, but that's a bigger change for later.

src/rdkafka_broker.c
```diff
@@ -5104,18 +5108,17 @@ static int rd_kafka_broker_thread_main (void *arg) {
...
-                                    rkb->rkb_rsal->rsal_curr + 1 ==
-                                    rkb->rkb_rsal->rsal_cnt)
-                                        rd_kafka_broker_serve(
-                                                rkb, rd_kafka_max_block_ms);
-			}
+                        /* Connect failure.
+                         * Try the next resolve result until we've
+                         * tried them all, in which case we back off the next
+                         * connection attempt to avoid busy looping. */
+			if (rkb->rkb_state == RD_KAFKA_BROKER_STATE_DOWN &&
```
mhowlett, line 5115: fwiw i like this written the old way, breaks it into steps. but inconsequential.

src/rdkafka_coord.c
```diff
@@ -235,19 +235,35 @@ void rd_kafka_coord_req (rd_kafka_t *rk,
         creq->creq_replyq = replyq;
         creq->creq_resp_cb = resp_cb;
         creq->creq_reply_opaque = reply_opaque;
+        creq->creq_refcnt = 1;
```
mhowlett, line 238: side note: on my todo list is to try and think through whether a refactor is possible anywhere that gets rid of refcounting (-> effectively multiple owners) in favor a single owner, which is simpler => more desirable if possible (even if there are some downsides, they may not be a big deal). probably not, but thought i'm mention the thought in case it gives you any ideas.

src/rdkafka_broker.c
edenhill: I think abstracting the address roundrobin check to its own function will improve readability

PR #2916, opened 2020-06-03
Ensure known toppars do not wind up in desired partitions list

### LINKED ISSUE #2915: Incorrect unknown partition error when assigning newly-created partitions
Consider the following sequence of events for a Kafka consumer (not using consumer groups):

  1. Assign consumer partition 0.
  2. Create partition 1.
  3. Assign consumer partition 1.
  4. Refresh metadata.

The metadata refresh will incorrectly generate an error that partition 0 is unknown. The issue seems to be a glitch into how the desired partition list (`rkt->rkt_deps`) is maintained.

This reproduces on the latest librdkafka etc. I'm going to submit a PR with a test case and a fix momentarily, but wanted an issue number to put in the test.

benesch:
rd_kafka_cgrp_assign calls rd_kafka_toppar_desired_add0 rather than its
wrapper rd_kafka_toppar_desired_add. The "add" wrapper preserves the
invariant that a known topic should never get added to the desired
partitions queue, while the "add0" function does not.

Maintaining this invariant is important for
rd_kafka_topic_partition_cnt_update, which assumes that a toppar is in
either the list of known partitions or the list of desired partitions,
but not both. Violating this invariant results in the situation
described in #2915, where updating assignments can trigger incorrect
"unknown partition" errors.

This patch rearranges rd_kafka_toppar_desired_add/add0 so that add0, in
addition to add, will avoid adding known partitions to the desired
partition list. The enclosed test correctly fails if run against the
current master (for the reasons described above).

Fix #2915.

### DIFF
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -569,8 +569,8 @@ void rd_kafka_toppar_desired_unlink (rd_kafka_toppar_t *rktp) {
 
 /**
  * @brief If rktp is not already desired:
- *  - mark as DESIRED|UNKNOWN
- *  - add to desired list
+ *  - mark as DESIRED|~REMOVE
+ *  - add to desired list if unknown
  *
  * @remark toppar_lock() MUST be held
  */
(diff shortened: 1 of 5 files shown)


edenhill:
Fantastic, thank you!

PR #2924, opened 2020-06-06
Fix incorrect cmake Module path for submodule use

omartijn:
When librdkafka is added to a project using `add_subdirectory`, the
include path for CMake modules that is added is incorrect, since it uses
`CMAKE_SOURCE_DIR` - which is the source directory of the root of the
project instead of `CMAKE_CURRENT_SOURCE_DIR`.

This matters not when librdkafka is built standalone (since the two
directories are the same), but when built as a submodule, it prevents
CMake from finding the bundled modules, like FindLibLZ4.cmake.


edenhill:
Thank you!

PR #2926, opened 2020-06-07
Const correctness

omartijn:
Improve const correctness by allowing Conf pointers to be const-qualified when constructing Handles. fatal_error member functions are now also const-qualified.


edenhill:
Thank you!!

PR #2929, opened 2020-06-09
Don't rely on LastStableOffset for old brokers

pf-qiu:
Old kafka brokers(<2.1.0) always reports -1 as LastStableOffset.
This prevents librdkafka from emitting EOF messages. Use
HighwaterMarkOffset instead as before.


pf-qiu:
This should fix https://github.com/edenhill/librdkafka/issues/2928

edenhill:
Thank you @pf-qiu !

PR #2939, opened 2020-06-17
KIP-54: Sticky assignor (part 1)

edenhill:
This is functionally complete and passes its unit tests, but needs another spin (in a follow-up PR) to:
 * update to latest AK changes
 * clean up FIXMEs
 * add integration tests


CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+
+ * Added `assignor` debug context for troubleshooting consumer partition
+   assignments.
+
+
+## Fixes
+
+### Consumer fixes
+
+ * The consumer assignors could ignore some topics if there were more subscribed
```
edenhill, line 18: This should be backported to 1.5.0.

src/rdkafka_assignor.c
```diff
@@ -364,14 +369,17 @@ rd_kafka_assignor_run (rd_kafka_cgrp_t *rkcg,
                                    rkas->rkas_opaque);
 
         if (err) {
-                rd_kafka_dbg(rkcg->rkcg_rk, CGRP, "ASSIGN",
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_ASSIGNOR,
```
mhowlett, line 372: not suggesting changes, but musing it would better if there was the concept of debug context groupings. i.e. you just specify ASSIGNOR here and that would be a child of CGRP. As a user, this gives me more certainty/clarity on what i'm getting debug messages for.

src/rdkafka_metadata.c
```diff
@@ -1205,3 +1205,111 @@ void rd_kafka_metadata_fast_leader_query (rd_kafka_t *rk) {
                                      NULL);
         }
 }
+
+
+
+
+/**
+ * @brief Create mock Metadata (for testing) based on the
+ *        var-arg tuples of (const char *topic, int partition_cnt).
```
mhowlett, line 1214: params are not varargs

src/rdkafka_partition.c
```diff
@@ -2848,39 +2925,76 @@ static int rd_kafka_topic_partition_cmp (const void *_a, const void *_b,
...
 
+/** @brief Compare only the topic */
+int rd_kafka_topic_partition_cmp_topic (const void *_a, const void *_b) {
+        const rd_kafka_topic_partition_t *a = _a;
+        const rd_kafka_topic_partition_t *b = _b;
+        return strcmp(a->topic, b->topic);
+}
+
+static int rd_kafka_topic_partition_cmp_opaque (const void *_a, const void *_b,
+                                                void *opaque) {
```
mhowlett, line 2936: why is `opaque` here if it's not used?

mhowlett: (it seems like it should be encapsulated more, the detail not exposed as high up).

src/rdmap.h
```diff
@@ -121,6 +123,26 @@ void *rd_map_get (const rd_map_t *rmap, const void *key);
...
+typedef void *(rd_map_copy_t) (const void *key_or_value);
+
+
+/**
+ * @brief Copy all elements from \p src to \p dst.
+ *        \p dst must be initialized and compatible with \p src.
+ *
+ * @param dst Destination map to copy to.
+ * @param src Source map to copy from.
+ * @param key_copy Key copy callback. If NULL the \p dst key will just
```
mhowlett, line 136: scary

src/rdkafka_partition.c
```diff
@@ -2816,6 +2856,44 @@ rd_kafka_topic_partition_list_copy (const rd_kafka_topic_partition_list_t *src){
         return dst;
 }
 
+
+/**
+ * @brief Compare two partition lists using partition comparator \p cmp.
+ *
+ * @warning This is an O(Na*Nb) operation.
```
mhowlett, line 2863: warning: i'm going to take this as a license to make a similar short-cut in a toppar list fn i need..

src/rdkafka_assignor.c
edenhill: The idea here is to be able to debug only the assignors and not the cgrp stuff, but this debug should be shown for either.

src/rdkafka_sticky_assignor.c
```diff
@@ -25,170 +25,3465 @@
...
+                             sort_by_map_elem_val_toppar_list_cnt);
+                return;
+        }
+}
+
+/**
+ * @returns true if the partition has two or more potential consumers.
+ */
+static RD_INLINE rd_bool_t
+partitionCanParticipateInReassignment (
```
mhowlett, line 455: `partitionCanParticipateInReassignment`: it's like you're writing java. I guess this was on purpose, and there's an argument for not modifying the names, but when translating between languages, i've noticed it's typical to change the naming to be idiomatic (same comment many other places).

src/rdkafka_sticky_assignor.c
```diff
@@ -25,170 +25,3465 @@
...
+
+
+
+
+/**
+ * @brief Populate subscriptions, current and previous assignments based on the
+ *        \p members assignments.
+ */
+static void
+preopulateCurrentAssignments (
```
mhowlett, line 1189: `preopulateCurrentAssignments`  prepopulate, not preopulate

src/rdkafka_sticky_assignor.c
```diff
@@ -25,170 +25,3465 @@
...
+        const rd_kafka_topic_partition_t *partition;
+        rd_list_t *consumers;
+        int i;
+
+        /* For each partition that is currently assigned to the group members
+         * add the member and its generation to
+         * sortedPartitionConsumersByGeneration (which is sorted afterwards)
+         * indexed by the partition. */
+        for (i = 0 ; i < (int)member_cnt ; i++) {
+                rd_kafka_group_member_t *consumer = &members[i];
```
mhowlett, line 1226: `member` seems a better name here than `consumer`, since that's what they're typically referred to in this context. i think consumer is less clear.

src/rdkafka_assignor.h
```diff
@@ -41,6 +41,7 @@ typedef struct rd_kafka_group_member_s {
         rd_kafkap_str_t                 *rkgm_group_instance_id;
         rd_kafkap_bytes_t               *rkgm_userdata;
         rd_kafkap_bytes_t               *rkgm_member_metadata;
+        int                              rkgm_generation;
```
mhowlett, line 44: this field doesn't really belong here, because it's specific to the sticky assignor. i expect this is best deserialized in `rd_kafka_sticky_assignor_assign_cb` (or helper) and passed as a parameter where needed.

also, this afaict, it's not being set (will always be -1).

also, you don't seem to be parsing the userdata at all - probably better if the owned partitions come from the user data in the EAGER case, even though it won't have any practical downside in the librdkafka case because the sticky partitioner is being introduced alongside the kip429 updates to the rebalance protocol.

src/rdkafka_partition.c
edenhill: It's a leaky abstraction of qsort_r() which takes an opaque arg.
Otherwise we'd have to jump through some hoops to trampoline a call to a users opaque-less cmp-arator.

src/rdkafka_sticky_assignor.c
edenhill: I AM writing Java!

src/rdkafka_assignor.h
mhowlett: we could probably skip providing a 'sticky' assignor and only do 'cooperative-sticky'. i doubt anyone will want the former (and it'll be easy to add if requested). this is probably what you thinking i'm now guessing. except `rkgm_generation` should still be just be a constant, as noted above.

INTRODUCTION.md
```diff
@@ -1749,7 +1751,7 @@ The [Apache Kafka Implementation Proposals (KIPs)](https://cwiki.apache.org/conf
 | KIP-42 - Producer and Consumer interceptors                              | 0.10.0.0                                  | Supported (not supported by Go, .NET and Python)                                              |
 | KIP-43 - SASL PLAIN and handshake                                        | 0.10.0.0                                  | Supported                                                                                     |
 | KIP-48 - Delegation tokens                                               | 1.1.0                                     | Not supported                                                                                 |
-| KIP-54 - Sticky partition assignment strategy                            | 0.11.0.0                                  | Not supported                                                                                 |
+| KIP-54 - Sticky partition assignment strategy                            | 0.11.0.0                                  | Supported                                                                                     |
```
mhowlett, line 1754: if we decide to only do cooperative-sticky, this would be not supported.

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+# librdkafka v1.6.0
+
+The v1.6.0 release is a feature release adding support for Incremental
```
mhowlett, line 3: Incremental -> incremental

INTRODUCTION.md
```diff
@@ -1795,7 +1797,7 @@ The [Apache Kafka Implementation Proposals (KIPs)](https://cwiki.apache.org/conf
 | KIP-320 - Consumer: handle log truncation                                | 2.1.0 (partially implemented) / 2.2.0     | Not supported                                                                                 |
 | KIP-322 - DeleteTopics disabled error code                               | 2.1.0                                     | Supported                                                                                     |
 | KIP-339 - AdminAPI: incrementalAlterConfigs                              | 2.3.0                                     | Not supported                                                                                 |
-| KIP-341 - Update Sticky partition assignment data                        | 2.3.0                                     | Not supported                                                                                 |
+| KIP-341 - Update Sticky partition assignment data                        | 2.3.0                                     | Supported                                                                                     |
```
mhowlett, line 1800: if we decide to only do cooperative-sticky this would be not supported.

maybe a good reason to do 'sticky' as well as 'cooperative-sticky' is we get to cross off 2 more kips.

mhowlett: we seem to have general agreement that skipping 'sticky' and going straight for 'cooperative-sticky' is fine.

src/rdkafka_assignor.h
edenhill: I wish to keep this to make our sticky assignor implementation as identical as possible to AbstractStickyAssignor.java, I think it is a bigger risk to factor out the generation parts of our implementation than to simply set this to its default value of -1.
It also makes a future demand for the non-cooperative sticky assignor easier to implement.

src/rdkafka_sticky_assignor.c
edenhill: This is using the Java implementation variable names to make code maintenance easier.

member and consumer is interchangable in th is context.

Issue #2952, opened 2020-06-30
Get broker information in the message delivery report callback


geo-matheou:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
It would be nice to get broker information in the message delivery report callback using either rd_kafka_t or rd_kafka_message_t structs. This can be helpful in situations where we get timeout issues and we would like to know which broker is responsible for these issues. 

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): v1.4.4
 - [ ] Apache Kafka version: `<REPLACE with e.g., 0.10.2.3>`
 - [ ] librdkafka client configuration: `<REPLACE with e.g., message.timeout.ms=123, auto.reset.offset=earliest, ..>`
 - [ ] Operating system: `<REPLACE with e.g., Centos 5 (x64)>`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
Is the broker id sufficient?

geo-matheou:
If the broker id contains the ip addreess / url of the broker would be great!

edenhill:
The broker id is the unique integer id for each broker, it is what identifies a single broker.

We don't want to store the connection address on each message object, it will either increase memory pressure or complexity, or both.

geo-matheou:
ok then it will be fine. 

Thank you for the update.

PR #2953, opened 2020-06-30
Added rd_kafka_message_broker_id() (#2952)

edenhill:
This also refactors and cleans up OP_ERR and OP_CONSUMER_ERR calls.

The broker_id is also useful in our tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ librdkafka.
  * Added `rd_kafka_produceva()` which takes an array of produce arguments
    for situations where the existing `rd_kafka_producev()` va-arg approach
    can't be used.
+ * Added `rd_kafka_message_broker_id()` to see the broker that a message
+   was produced or fetched from, or an error was associated with.
  * Added RTT/delay simulation to mock brokers.
 
 
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -487,8 +487,7 @@ static void rd_kafka_broker_set_error (rd_kafka_broker_t *rkb, int level,
                              "%s: %s", rkb->rkb_name, errstr);
 
                 /* Send ERR op to application for processing. */
-                rd_kafka_q_op_err(rkb->rkb_rk->rk_rep, RD_KAFKA_OP_ERR,
-                                  err, 0, NULL, 0, "%s: %s",
+                rd_kafka_q_op_err(rkb->rkb_rk->rk_rep, err, "%s: %s",
                                   rkb->rkb_name, errstr);
         }
 }
@@ -4447,14 +4446,14 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
 				case RD_KAFKA_RESP_ERR_MSG_SIZE_TOO_LARGE:
 				default: /* and all other errors */
 					rd_dassert(tver->version > 0);
-					rd_kafka_q_op_err(
-						rktp->rktp_fetchq,
-						RD_KAFKA_OP_CONSUMER_ERR,
-						hdr.ErrorCode, tver->version,
-						rktp,
-						rktp->rktp_offsets.fetch_offset,
-						"Fetch failed: %s",
-						rd_kafka_err2str(hdr.ErrorCode));
+                                        rd_kafka_consumer_err(
+                                                rktp->rktp_fetchq,
+                                                rd_kafka_broker_id(rkb),
+                                                hdr.ErrorCode, tver->version,
+                                                NULL, rktp,
+                                                rktp->rktp_offsets.fetch_offset,
+                                                "Fetch failed: %s",
+                                                rd_kafka_err2str(hdr.ErrorCode));
 					break;
 				}
 
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -5914,6 +5914,31 @@ rd_kafka_broker_update (rd_kafka_t *rk, rd_kafka_secproto_t proto,
 
 
 /**
+ * @returns the broker id, or RD_KAFKA_NODEID_UA if \p rkb is NULL.
+ *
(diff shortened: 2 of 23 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ librdkafka.
  * Added `rd_kafka_produceva()` which takes an array of produce arguments
    for situations where the existing `rd_kafka_producev()` va-arg approach
    can't be used.
+ * Added `rd_kafka_message_broker_id()` to see the broker that a message
+   was produced or fetched from, or an error was associated with.
```
mhowlett, line 29: nice!

src/rdkafka_broker.c
```diff
@@ -5913,6 +5912,31 @@ rd_kafka_broker_update (rd_kafka_t *rk, rd_kafka_secproto_t proto,
...
+ * @locks_required none
+ * @locks_acquired rkb_lock
+ */
+int32_t rd_kafka_broker_id (rd_kafka_broker_t *rkb) {
+        int32_t broker_id;
+
+        if (unlikely(!rkb))
+                return RD_KAFKA_NODEID_UA;
+
+        /* Avoid locking if already on the broker thread */
```
mhowlett, line 5928: first instinct is locking may be better outside (simplicity over terseness). didn't think about it much.

src/rdkafka_broker.c
```diff
@@ -4447,14 +4446,14 @@ rd_kafka_fetch_reply_handle (rd_kafka_broker_t *rkb,
...
 					rd_dassert(tver->version > 0);
-					rd_kafka_q_op_err(
-						rktp->rktp_fetchq,
-						RD_KAFKA_OP_CONSUMER_ERR,
-						hdr.ErrorCode, tver->version,
-						rktp,
-						rktp->rktp_offsets.fetch_offset,
-						"Fetch failed: %s",
-						rd_kafka_err2str(hdr.ErrorCode));
+                                        rd_kafka_consumer_err(
```
mhowlett, line 4449: unconvinced this is the right place to be sending notifications of errors we're not expecting here. why not the standard error path? if we send them there, the user will probably have default logging set up which is an appropriate action. also, that's in line with what our story is for what error notifications are about. this muddles things up. i understand you are worried about maintaining ordering, but in what way does that help here?

src/rdkafka_cgrp.c
```diff
@@ -490,13 +490,13 @@ static void rd_kafka_cgrp_handle_FindCoordinator (rd_kafka_t *rk,
...
                 rd_kafka_cgrp_coord_update(rkcg, -1);
-	else {
+        else {
                 if (rkcg->rkcg_last_err != ErrorCode) {
-                        rd_kafka_q_op_err(rkcg->rkcg_q,
-                                          RD_KAFKA_OP_CONSUMER_ERR,
-                                          ErrorCode, 0, NULL, 0,
-                                          "FindCoordinator response error: %s",
-                                          errstr);
+                        rd_kafka_consumer_err(
```
mhowlett, line 495: can you give a precise scenario where the user would be worse off if this was sent to the error handler instead?

i need an example, i'm failing to get the benefit.

src/rdkafka_cgrp.c
```diff
@@ -1131,11 +1131,12 @@ static void rd_kafka_cgrp_handle_JoinGroup (rd_kafka_t *rk,
                         ErrorCode = RD_KAFKA_RESP_ERR__FATAL;
 
                 } else if (actions & RD_KAFKA_ERR_ACTION_PERMANENT)
-                        rd_kafka_q_op_err(rkcg->rkcg_q,
-                                          RD_KAFKA_OP_CONSUMER_ERR,
-                                          ErrorCode, 0, NULL, 0,
-                                          "JoinGroup failed: %s",
-                                          rd_kafka_err2str(ErrorCode));
+                        rd_kafka_consumer_err(rkcg->rkcg_q,
```
mhowlett, line 1134: ditto...

exposing this inline with the messages implies to me there is a scenario where i should handle this explicitly in some way. I thought/hope our abstractions make this not a concern of the user. maybe there is something to do with committing?

i think this would be a great time to be documenting how to do correct consumer error handling.

src/rdkafka_cgrp.c
```diff
@@ -2768,11 +2770,12 @@ rd_kafka_cgrp_max_poll_interval_check_tmr_cb (rd_kafka_timers_t *rkts,
                      "leaving group",
                      rk->rk_conf.max_poll_interval_ms, exceeded);
 
-        rd_kafka_q_op_err(rkcg->rkcg_q, RD_KAFKA_OP_CONSUMER_ERR,
-                          RD_KAFKA_RESP_ERR__MAX_POLL_EXCEEDED, 0, NULL, 0,
-                          "Application maximum poll interval (%dms) "
-                          "exceeded by %dms",
-                          rk->rk_conf.max_poll_interval_ms, exceeded);
+        rd_kafka_consumer_err(rkcg->rkcg_q, RD_KAFKA_NODEID_UA,
```
mhowlett, line 2773: ok, this one makes sense to me to have inline :-).

mhowlett:
seeking clarity on errors before hitting approve

src/rdkafka_broker.c
edenhill: The entire point of this method is to abstract that detail away from the caller :)

src/rdkafka_broker.c
edenhill: Apart from the Replica fix this PR simply refactors the consumer error code but does not change the current behaviour (except for one place),   so let's have the error-raising discussion outside the PR.

src/rdkafka_cgrp.c
edenhill: I can't! Let's revise all error propagation at a later time.

PR #2974, opened 2020-07-08
Change `linger.ms` default from 0.5 to 5ms to improve efficiency

edenhill:
Increased batching leads to less protocol overhead, less CPU usage
on client and brokers, increased throughput and in some cases also
improved latency.


CHANGELOG.md
```diff
@@ -55,6 +55,11 @@ librdkafka.
    Users on Brokers 0.9.x or older should already be configuring
    `api.version.request=false` and `broker.version.fallback=...` so there
    should be no functional change.
+ * The default producer batch accumulation time, `linger.ms`, has been changed
+   from 0.5ms to 5ms to improve batch sizes and throughput while reducing
+   the per-message protocol overhead.
+   Applications that require lower produce latency than 5ms will need to
```
mhowlett, line 61: pedantically, this is of course more complicated, at higher throughputs they'll get better latency with 5ms linger.ms than 0.5ms. but doesn't matter...

src/rdkafka_conf.c
```diff
@@ -3645,15 +3645,20 @@ const char *rd_kafka_topic_conf_finalize (rd_kafka_type_t cltype,
 
 
         if (cltype == RD_KAFKA_PRODUCER) {
+                if (tconf->message_timeout_ms != 0 &&
+                    (double)tconf->message_timeout_ms <=
+                    conf->buffering_max_ms_dbl) {
```
mhowlett, line 3650: if `buffering_max_ms_dbl` was named `linger_ms_dbl` i would have read this quite a bit more quickly. thought the thought, so thought i'd mention it.

PR #2980, opened 2020-07-08
Incremental assign and unassign

mhowlett:
Ok, let's see what you think of this.

Previously opened and commented on here: https://github.com/mhowlett/librdkafka/pull/1 . Unfortunately, the rebasing is a bit screwed up over there and i didn't bother to fix it properly. The contextual comments may still be useful to refer back to though.

### DIFF
--- a/src-cpp/KafkaConsumerImpl.cpp
+++ b/src-cpp/KafkaConsumerImpl.cpp
@@ -187,6 +187,38 @@ RdKafka::KafkaConsumerImpl::unassign () {
 }
 
 
+RdKafka::Error *
+RdKafka::KafkaConsumerImpl::incremental_assign (const std::vector<TopicPartition*> &partitions) {
+  rd_kafka_topic_partition_list_t *c_parts;
+  rd_kafka_error_t *c_error;
+
+  c_parts = partitions_to_c_parts(partitions);
+  c_error = rd_kafka_incremental_assign(rk_, c_parts);
+  rd_kafka_topic_partition_list_destroy(c_parts);
+
+  if (c_error)
+    return new ErrorImpl(c_error);
+
+  return NULL;
+}
+
+
+RdKafka::Error *
+RdKafka::KafkaConsumerImpl::incremental_unassign (const std::vector<TopicPartition*> &partitions) {
+  rd_kafka_topic_partition_list_t *c_parts;
+  rd_kafka_error_t *c_error;
+
+  c_parts = partitions_to_c_parts(partitions);
+  c_error = rd_kafka_incremental_unassign(rk_, c_parts);
+  rd_kafka_topic_partition_list_destroy(c_parts);
+
+  if (c_error)
+    return new ErrorImpl(c_error);
+
+  return NULL;
+}
+
+
 RdKafka::ErrorCode
 RdKafka::KafkaConsumerImpl::committed (std::vector<RdKafka::TopicPartition*> &partitions, int timeout_ms) {
   rd_kafka_topic_partition_list_t *c_parts;
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -2421,6 +2421,40 @@ public:
   virtual ErrorCode unassign () = 0;
 
   /**
+   * @brief Incrementally add \p partitions to the current assignment.
+   *
+   * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
+   * this method should be used in a rebalance callback to adjust the current
+   * assignment appropriately in the case where the rebalance type is
+   * ERR__ASSIGN_PARTITIONS. The application must pass the partition list
+   * passed to the callback (or a copy of it), even if the list is empty.
+   * This method may also be used outside the context of a rebalance callback.
+   *
+   * @returns NULL on success, or an error object if the operation was
+   *          unsuccessful.
+   *
+   * @remark The returned object must be deleted by the application.
+   */
+  virtual Error *incremental_assign (const std::vector<TopicPartition*> &partitions) = 0;
+
+  /**
+   * @brief Incrementally remove \p partitions from the current assignment.
+   *
+   * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
+   * this method should be used in a rebalance callback to adjust the current
+   * assignment appropriately in the case where the rebalance type is
+   * ERR__REVOKE_PARTITIONS. The application must pass the partition list
+   * passed to the callback (or a copy of it), even if the list is empty.
+   * This method may also be used outside the context of a rebalance callback.
+   *
+   * @returns NULL on success, or an error object if the operation was
+   *          unsuccessful.
+   *
+   * @remark The returned object must be deleted by the application.
+   */
+  virtual Error *incremental_unassign (const std::vector<TopicPartition*> &partitions) = 0;
+
+  /**
    * @brief Consume message or get error event, triggers callbacks.
    *
    * Will automatically call registered callbacks for any such queued events,
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -2421,40 +2421,6 @@ public:
   virtual ErrorCode unassign () = 0;
 
   /**
-   * @brief Incrementally add \p partitions to the current assignment.
-   *
-   * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
-   * this method should be used in a rebalance callback to adjust the current
-   * assignment appropriately in the case where the rebalance type is
-   * ERR__ASSIGN_PARTITIONS. The application must pass the partition list
-   * passed to the callback (or a copy of it), even if the list is empty.
-   * This method may also be used outside the context of a rebalance callback.
-   *
-   * @returns NULL on success, or an error object if the operation was
-   *          unsuccessful.
-   *
-   * @remark The returned object must be deleted by the application.
-   */
-  virtual Error *incremental_assign (const std::vector<TopicPartition*> &partitions) = 0;
-
-  /**
-   * @brief Incrementally remove \p partitions from the current assignment.
-   *
-   * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
-   * this method should be used in a rebalance callback to adjust the current
-   * assignment appropriately in the case where the rebalance type is
-   * ERR__REVOKE_PARTITIONS. The application must pass the partition list
-   * passed to the callback (or a copy of it), even if the list is empty.
-   * This method may also be used outside the context of a rebalance callback.
-   *
-   * @returns NULL on success, or an error object if the operation was
-   *          unsuccessful.
-   *
-   * @remark The returned object must be deleted by the application.
-   */
-  virtual Error *incremental_unassign (const std::vector<TopicPartition*> &partitions) = 0;
-
-  /**
    * @brief Consume message or get error event, triggers callbacks.
    *
    * Will automatically call registered callbacks for any such queued events,
@@ -2689,6 +2655,42 @@ public:
    *          lost, false otherwise.
    */
   virtual bool assignment_lost () = 0;
+
+
+  /**
+   * @brief Incrementally add \p partitions to the current assignment.
+   *
+   * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
+   * this method should be used in a rebalance callback to adjust the current
+   * assignment appropriately in the case where the rebalance type is
+   * ERR__ASSIGN_PARTITIONS. The application must pass the partition list
+   * passed to the callback (or a copy of it), even if the list is empty.
+   * This method may also be used outside the context of a rebalance callback.
+   *
+   * @returns NULL on success, or an error object if the operation was
+   *          unsuccessful.
+   *
+   * @remark The returned object must be deleted by the application.
+   */
+  virtual Error *incremental_assign (const std::vector<TopicPartition*> &partitions) = 0;
+
+
+  /**
+   * @brief Incrementally remove \p partitions from the current assignment.
+   *
+   * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
+   * this method should be used in a rebalance callback to adjust the current
+   * assignment appropriately in the case where the rebalance type is
+   * ERR__REVOKE_PARTITIONS. The application must pass the partition list
+   * passed to the callback (or a copy of it), even if the list is empty.
+   * This method may also be used outside the context of a rebalance callback.
+   *
+   * @returns NULL on success, or an error object if the operation was
+   *          unsuccessful.
+   *
+   * @remark The returned object must be deleted by the application.
+   */
+  virtual Error *incremental_unassign (const std::vector<TopicPartition*> &partitions) = 0;
 
 };
 
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3715,10 +3715,12 @@ rd_kafka_incremental_unassign (rd_kafka_t *rk,
  * albeit empty assignment, and maintain internal state, while a \c NULL
  * value for \p partitions will reset and clear the internal state.
  *
- * When used from a rebalance callback, the application shall pass the
+ * When used from a rebalance callback, the application should pass the
  * partition list passed to the callback (or a copy of it) even if the list
  * is empty (i.e. should not pass NULL in this case) so as to maintain
- * internal join state.
+ * internal join state. This is not strictly required - the application
+ * may adjust the assignment provided by the group. However, this is rarely
+ * useful in practice.
  *
  * @returns An error code indicating if the new assignment was applied or not.
  *          RD_KAFKA_RESP_ERR__FATAL is returned if the consumer has raised
@@ -3902,6 +3904,14 @@ rd_kafka_consumer_group_metadata_new (const char *group_id);
 /**
  * @brief Create a new consumer group metadata object.
  *        This is typically only used for writing tests.
+ *
+ * @param group_id The group id (must not be NULL).
+ *
+ * @param generation_id The group generation id.
+ *
+ * @param member_id The group member id (must not be NULL).
+ *
+ * @param group_instance_id The group instance id (may be NULL).
  *
  * @remark The returned pointer must be freed by the application using
  *         rd_kafka_consumer_group_metadata_destroy().
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3660,19 +3660,65 @@ RD_EXPORT
 rd_kafka_resp_err_t rd_kafka_consumer_close (rd_kafka_t *rk);
 
 
+/**
+ * @brief Incrementally add \p partitions to the current assignment.
+ *
+ * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
+ * this method should be used in a rebalance callback to adjust the current
+ * assignment appropriately in the case where the rebalance type is
+ * RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS. The application must pass the
+ * partition list passed to the callback (or a copy of it), even if the
+ * list is empty. \p partitions must not be NULL. This method may also be
+ * used outside the context of a rebalance callback.
+ *
+ * @returns NULL on success, or an error object if the operation was
+ *          unsuccessful.
+ *
+ * @remark The returned error object (if not NULL) must be destroyed with
+ *         rd_kafka_error_destroy().
+ */
+RD_EXPORT rd_kafka_error_t *
+rd_kafka_incremental_assign (rd_kafka_t *rk,
+                             const rd_kafka_topic_partition_list_t
+                             *partitions);
+
+
+/**
+ * @brief Incrementally remove \p partitions from the current assignment.
+ *
+ * If a COOPERATIVE assignor (i.e. incremental rebalancing) is being used,
+ * this method should be used in a rebalance callback to adjust the current
+ * assignment appropriately in the case where the rebalance type is
+ * RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS. The application must pass the
+ * partition list passed to the callback (or a copy of it), even if the
+ * list is empty. \p partitions must not be NULL. This method may also be
+ * used outside the context of a rebalance callback.
+ *
+ * @returns NULL on success, or an error object if the operation was
+ *          unsuccessful.
+ *
+ * @remark The returned error object (if not NULL) must be destroyed with
+ *         rd_kafka_error_destroy().
+ */
+RD_EXPORT rd_kafka_error_t *
+rd_kafka_incremental_unassign (rd_kafka_t *rk,
+                               const rd_kafka_topic_partition_list_t
+                               *partitions);
+
 
 /**
  * @brief Atomic assignment of partitions to consume.
  *
  * The new \p partitions will replace the existing assignment.
  *
- * When used from a rebalance callback the application shall pass the
- * partition list passed to the callback (or a copy of it) (even if the list
- * is empty) rather than NULL to maintain internal join state.
-
  * A zero-length \p partitions will treat the partitions as a valid,
- * albeit empty, assignment, and maintain internal state, while a \c NULL
+ * albeit empty assignment, and maintain internal state, while a \c NULL
  * value for \p partitions will reset and clear the internal state.
+ *
+ * When used from a rebalance callback, the application shall pass the
+ * partition list passed to the callback (or a copy of it) even if the list
+ * is empty (i.e. should not pass NULL in this case) so as to maintain
+ * internal join state.
  *
  * @returns An error code indicating if the new assignment was applied or not.
  *          RD_KAFKA_RESP_ERR__FATAL is returned if the consumer has raised
@@ -3862,7 +3908,10 @@ rd_kafka_consumer_group_metadata_new (const char *group_id);
  */
 RD_EXPORT rd_kafka_consumer_group_metadata_t *
 rd_kafka_consumer_group_metadata_new_with_genid (const char *group_id,
-                                                 int32_t generation_id);
+                                                 int32_t generation_id,
+                                                 const char *member_id,
+                                                 const char
+                                                 *group_instance_id);
 
 
 /**
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3905,12 +3905,9 @@ rd_kafka_consumer_group_metadata_new (const char *group_id);
  * @brief Create a new consumer group metadata object.
  *        This is typically only used for writing tests.
  *
- * @param group_id The group id (must not be NULL).
- *
+ * @param group_id The group id.
  * @param generation_id The group generation id.
- *
- * @param member_id The group member id (must not be NULL).
- *
+ * @param member_id The group member id.
  * @param group_instance_id The group instance id (may be NULL).
  *
  * @remark The returned pointer must be freed by the application using
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -2452,25 +2452,34 @@ rd_kafka_cgrp_incremental_assign (rd_kafka_cgrp_t *rkcg,
 
         if (rd_kafka_fatal_error_code(rkcg->rkcg_rk)) {
                 rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
-                             "ASSIGN", "Group \"%s\": consumer is in "
-                             "a failed state, treating incremental "
+                             "ASSIGN", "Group \"%s\": consumer has raised "
+                             "a fatal error, treating incremental "
                              "assign of %d partitions as unassign of all"
-                             "%d partitions in the current assignment",
+                             "%d partition(s) in the current assignment",
                              rkcg->rkcg_group_id->str, partitions->cnt,
                              !rkcg->rkcg_assignment ? 0
                              : rkcg->rkcg_assignment->cnt);
                 err = rd_kafka_cgrp_assign(rkcg, NULL);
-                rd_assert(err);
-                return rd_kafka_error_new(err, rd_kafka_err2str(err));
+                return rd_kafka_error_new(err, "%s", rd_kafka_err2str(err));
         }
 
         if (rd_atomic32_get(&rkcg->rkcg_assignment_lost))
                 return rd_kafka_error_new(RD_KAFKA_RESP_ERR__STATE,
-                     "Partitions can not be added to a current assignment "
-                     "that is lost");
+                     "Partitions can not be added to a lost assignment");
+
+        if (partitions->cnt == 0) {
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                        "ASSIGN", "Group \"%s\": not adding empty partition "
+                        "list to existing assignment of %d partitions in join "
+                        "state %s (nothing to do)", rkcg->rkcg_group_id->str,
+                        !rkcg->rkcg_assignment ? 0
+                        : rkcg->rkcg_assignment->cnt,
+                        rd_kafka_cgrp_join_state_names[rkcg->rkcg_join_state]);
+                return NULL;
+        }
 
         /* Verify partitions do not exist in the current assignment before
-         * making any changes to it to so the operation doesn't partially
+         * making any changes to it so the operation doesn't partially
          * succeed. */
         for (i = 0; rkcg->rkcg_assignment && i < partitions->cnt; i++) {
                 const rd_kafka_topic_partition_t *rktpar = &partitions->elems[i];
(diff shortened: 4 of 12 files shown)


src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+        if (partitions->cnt == 0) {
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER, "UNASSIGN",
+                     "Group \"%s\": removing 0 of %d assigned partition(s) "
+                     "in join state %s (nothing to do)",
+                     rkcg->rkcg_group_id->str, rkcg->rkcg_assignment->cnt,
+                     rd_kafka_cgrp_join_state_names[rkcg->rkcg_join_state]);
+                return NULL;
+        }
+
+        rd_kafka_cgrp_version_new_barrier(rkcg);
```
mhowlett, line 2628: please consider from here down to the end of the method carefully. I'm not 100% confident I haven't missed something important.

edenhill:
0113 does not build on various platforms, use `RdKafka::ERR__..` instead of `RdKafka::ErrorCode::ERR__..`.
See CI failures.

src-cpp/KafkaConsumerImpl.cpp
```diff
@@ -187,6 +187,38 @@ RdKafka::KafkaConsumerImpl::unassign () {
 }
 
 
+RdKafka::Error *
+RdKafka::KafkaConsumerImpl::incremental_assign (const std::vector<TopicPartition*> &partitions) {
+  rd_kafka_topic_partition_list_t *c_parts;
```
edenhill, line 192: Since this is C++ and not ancient C89 we don't need to put variable definitions at the top of the scope but where they are used 🎉 :)

(no action required, just wanted to point it out)

src-cpp/rdkafkacpp.h
```diff
@@ -2420,6 +2420,40 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
    */
   virtual ErrorCode unassign () = 0;
 
+  /**
```
edenhill, line 2423: New methods must be added to the end of virtual classes as not to break the ABI.

src/rdkafka.h
```diff
@@ -3652,20 +3652,66 @@ RD_EXPORT
...
- * When used from a rebalance callback the application shall pass the
- * partition list passed to the callback (or a copy of it) (even if the list
- * is empty) rather than NULL to maintain internal join state.
-
  * A zero-length \p partitions will treat the partitions as a valid,
- * albeit empty, assignment, and maintain internal state, while a \c NULL
+ * albeit empty assignment, and maintain internal state, while a \c NULL
  * value for \p partitions will reset and clear the internal state.
  *
+ * When used from a rebalance callback, the application shall pass the
```
edenhill, line 3710: use `must` rather than `shall`, it is clearer.

src/rdkafka.h
```diff
@@ -3854,7 +3900,10 @@ rd_kafka_consumer_group_metadata_new (const char *group_id);
  */
 RD_EXPORT rd_kafka_consumer_group_metadata_t *
 rd_kafka_consumer_group_metadata_new_with_genid (const char *group_id,
-                                                 int32_t generation_id);
+                                                 int32_t generation_id,
```
edenhill, line 3903: Are any of these fields optional/NULLable? If so, should be reflected in doc string.

src/rdkafka_op.c
```diff
@@ -482,25 +491,55 @@ rd_kafka_op_t *rd_kafka_op_new_cb (rd_kafka_t *rk,
...
 
 	return rd_kafka_replyq_enq(&rko->rko_replyq, rko, 0);
 }
 
 
+/**
+ * @brief Reply to 'rko' re-using the same rko with rko_error specified
+ *        by \p error (may be NULL) and rko_err set to the corresponding
+ *        error code.
+ *
```
edenhill, line 521: Should say it assumes ownership of `\p error`

src/rdkafka_op.c
```diff
@@ -482,25 +491,55 @@ rd_kafka_op_t *rd_kafka_op_new_cb (rd_kafka_t *rk,
...
+ * If there is no replyq the rko is destroyed.
+ *
+ * @returns 1 if op was enqueued, else 0 and rko is destroyed.
+ */
+int rd_kafka_op_error_reply (rd_kafka_op_t *rko,
+                             rd_kafka_error_t *error) {
+
+        if (!rko->rko_replyq.q) {
+                rd_kafka_op_destroy(rko);
```
edenhill, line 530: error is leaked here.

src/rdkafka_op.c
```diff
@@ -570,6 +610,23 @@ rd_kafka_resp_err_t rd_kafka_op_err_destroy (rd_kafka_op_t *rko) {
...
+ */
+rd_kafka_error_t *rd_kafka_op_error_destroy (rd_kafka_op_t *rko) {
+        if (rko) {
+                rd_kafka_error_t *error = rko->rko_error;
+                rko->rko_error = NULL;
+                rd_kafka_op_destroy(rko);
+                return error;
+        }
+
+        return rd_kafka_error_new(
```
edenhill, line 624: error_new() takes a format string

src/rdkafka_subscription.c
```diff
@@ -92,25 +92,75 @@ rd_kafka_subscribe (rd_kafka_t *rk,
...
+rd_kafka_assign0 (rd_kafka_t *rk,
+                  rd_kafka_assign_method_t assign_method,
+                  const rd_kafka_topic_partition_list_t *partitions) {
         rd_kafka_op_t *rko;
         rd_kafka_cgrp_t *rkcg;
 
         if (!(rkcg = rd_kafka_cgrp_get(rk)))
-                return RD_KAFKA_RESP_ERR__UNKNOWN_GROUP;
+                return rd_kafka_error_new(RD_KAFKA_RESP_ERR__UNKNOWN_GROUP,
+                        rd_kafka_err2str(RD_KAFKA_RESP_ERR__UNKNOWN_GROUP));
```
edenhill, line 104: error_new() takes a format string.

And the error string should preferably be something more interesting than err2str, in this case:
"Requires a consumer with group.id configured"

tests/0113-cooperative_rebalance.cpp
```diff
@@ -30,83 +30,306 @@
 #include <map>
 #include <cstring>
 #include <cstdlib>
+#include <assert.h>
 #include "testcpp.h"
 
 /**
- * The beginnings of an integration test for cooperative rebalancing.
- * MH: what i'm currently using to debug with.
+ * MH: what i'm currently using to debug with. Not finished.
```
edenhill, line 37: not reviewing this file

tests/0113-cooperative_rebalance.cpp
```diff
@@ -30,83 +30,306 @@
...
+                                 std::vector<RdKafka::TopicPartition *> toppars2) {
+  std::vector<RdKafka::TopicPartition *> assignment;
+
+  consumer->incremental_assign(toppars1);
+  consumer->assignment(assignment);
+  test_assert(assignment.size() == 1, "Expecting current assignment to have size 1");
+  delete assignment[0];
+  assignment.clear();
+  RdKafka::Message *m = consumer->consume(5000);
+  test_assert(m->err() == RdKafka::ErrorCode::ERR_NO_ERROR, "Expecting a consumed message.");
```
edenhill, line 163: `RdKafka::ERR_NO_ERROR`, for whatever reason.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                             !rkcg->rkcg_assignment ? 0
+                             : rkcg->rkcg_assignment->cnt);
+                err = rd_kafka_cgrp_assign(rkcg, NULL);
+                rd_assert(err);
+                return rd_kafka_error_new(err, rd_kafka_err2str(err));
+        }
+
+        if (rd_atomic32_get(&rkcg->rkcg_assignment_lost))
+                return rd_kafka_error_new(RD_KAFKA_RESP_ERR__STATE,
+                     "Partitions can not be added to a current assignment "
```
edenhill, line 2469: I think "to a lost assignment" is equally correct and has less loaded words.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                     "that is lost");
+
+        /* Verify partitions do not exist in the current assignment before
+         * making any changes to it to so the operation doesn't partially
```
edenhill, line 2473: "to so"

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+         * succeed. */
+        for (i = 0; rkcg->rkcg_assignment && i < partitions->cnt; i++) {
+                const rd_kafka_topic_partition_t *rktpar = &partitions->elems[i];
+                if (rd_kafka_topic_partition_list_find(rkcg->rkcg_assignment,
+                                                       rktpar->topic,
+                                                       rktpar->partition)) {
+                        return rd_kafka_error_new(
+                                RD_KAFKA_RESP_ERR__INVALID_ARG,
```
edenhill, line 2481: I think CONFLICT or STATE is more appropriate

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                                "%s [%"PRId32"] is already part of the "
+                                "current assignment",
+                                rktpar->topic, rktpar->partition);
+                }
+        }
+
+        if (partitions->cnt == 0) {
```
edenhill, line 2488: Move this check before the traversal.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                        "ASSIGN", "Group \"%s\": adding 0 partitions to "
```
edenhill, line 2490: "not adding empty partition list .." is more correct, since nothing is being added, and reads better.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                        !rkcg->rkcg_assignment ? 0
+                        : rkcg->rkcg_assignment->cnt,
+                        rd_kafka_cgrp_join_state_names[rkcg->rkcg_join_state]);
+                return NULL;
+        }
+
+        rd_kafka_cgrp_version_new_barrier(rkcg);
+
+        rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER, "ASSIGN",
+                "Group \"%s\": adding %d partitions to existing assignment "
```
edenhill, line 2502: partition(s)

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                "of %d partitions in join state %s (v%"PRId32")",
```
edenhill, line 2503: dito

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                rkcg->rkcg_group_id->str, partitions->cnt,
+                !rkcg->rkcg_assignment ? 0 : rkcg->rkcg_assignment->cnt,
+                rd_kafka_cgrp_join_state_names[rkcg->rkcg_join_state],
+                rkcg->rkcg_version);
+
+        if (!rkcg->rkcg_assignment)
```
edenhill, line 2509: Move this to before the debug to avoid `rkcg->rkcg_assignment ?` ternary

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                rktp = rd_kafka_toppar_get2(rkcg->rkcg_rk,
+                                            rktpar->topic,
+                                            rktpar->partition,
+                                            0/*no-ua*/, 1/*create-on-miss*/);
+
+                if (!rktp)
+                        /* LOG_ERR already emitted by rd_kafka_toppar_get2 */
+                        continue;
+
+                rd_kafka_toppar_keep(rktp);
```
edenhill, line 2529: Don't think this is needed since get2 increases the refcount for the returned rktp.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+
+                partitions->elems[i]._private = rktp;
+                if (!rd_kafka_topic_partition_list_add0(
```
edenhill, line 2532: This function can't return NULL.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                          rktpar->topic,
+                          rktpar->partition,
+                          rktp)) {
+                        rd_assert(
+                                !*"rd_kafka_topic_partition_list_add0 failed");
+                }
+        }
+
+        rd_kafka_wrlock(rkcg->rkcg_rk);
+        rkcg->rkcg_c.assignment_size += partitions->cnt;
```
edenhill, line 2543: ` .. = rkcg_assignment->cnt` is less error prone.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                             "UNASSIGN", "Group \"%s\": consumer is "
+                             "in a failed state, treating incremental "
+                             "unassign of %d partitions as unassign of all "
+                             "%d partitions in the current assignment",
+                             rkcg->rkcg_group_id->str,
+                             partitions->cnt, !rkcg->rkcg_assignment ?
+                             0 : rkcg->rkcg_assignment->cnt);
+                err = rd_kafka_cgrp_assign(rkcg, NULL);
+                rd_assert(err);
```
edenhill, line 2586: dito

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                return rd_kafka_error_new(err, rd_kafka_err2str(err));
```
edenhill, line 2587: dito

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+static rd_kafka_error_t *
+rd_kafka_cgrp_incremental_unassign (rd_kafka_cgrp_t *rkcg,
+                                    rd_kafka_topic_partition_list_t
+                                    *partitions) {
+        int i;
+        rd_kafka_resp_err_t err;
+        int cur_assignment_cnt =
+                (rkcg->rkcg_assignment ? rkcg->rkcg_assignment->cnt : 0);
+
+        rd_assert(partitions);
```
edenhill, line 2572: There's no need to assert on this unless you expect the callers to be buggy, but since the number of callers is miniscule that is unlikely.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+         * assignment */
+        if (rd_kafka_fatal_error_code(rkcg->rkcg_rk)) {
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                             "UNASSIGN", "Group \"%s\": consumer is "
+                             "in a failed state, treating incremental "
+                             "unassign of %d partitions as unassign of all "
+                             "%d partitions in the current assignment",
+                             rkcg->rkcg_group_id->str,
+                             partitions->cnt, !rkcg->rkcg_assignment ?
+                             0 : rkcg->rkcg_assignment->cnt);
```
edenhill, line 2584: you have cur_assignmnt_cnt

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                                rktpar->topic, rktpar->partition);
+                }
+        }
+
+        /* Whether or not it was before, current assignment is now not lost. */
+        rd_atomic32_set(&rkcg->rkcg_assignment_lost, rd_false);
+
+        if (partitions->cnt == 0) {
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER, "UNASSIGN",
+                     "Group \"%s\": removing 0 of %d assigned partition(s) "
```
edenhill, line 2621: wording implies something is done (then says nothing is done), prefer we get down to business right away and say that the empty list is ignored.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                             rkcg->rkcg_group_id->str,
+                             partitions->cnt, !rkcg->rkcg_assignment ?
+                             0 : rkcg->rkcg_assignment->cnt);
+                err = rd_kafka_cgrp_assign(rkcg, NULL);
+                rd_assert(err);
+                return rd_kafka_error_new(err, rd_kafka_err2str(err));
+        }
+
+        if (rd_atomic32_get(&rkcg->rkcg_assignment_lost) &&
+            partitions->cnt < cur_assignment_cnt) {
```
edenhill, line 2591: Need to unassign all when assignment is lost, so suggest `partitions->cnt != cur_assignment_cnt`

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                if (!rd_kafka_topic_partition_list_del(rkcg->rkcg_assignment,
+                                                       rktpar->topic,
+                                                       rktpar->partition))
+                        rd_assert(
+                                !*"rd_kafka_topic_partition_list_del failed");
+        }
+
+        /* Update statistics */
+        rd_kafka_wrlock(rkcg->rkcg_rk);
+        rkcg->rkcg_c.assignment_size -= partitions->cnt;
```
edenhill, line 2657: ` = rkcg->rkcg_assignment_cnt` is less error prone.

src/rdkafka_cgrp.c
```diff
@@ -2658,6 +2940,11 @@ rd_kafka_cgrp_assign (rd_kafka_cgrp_t *rkcg,
         /* If the consumer has raised a fatal error we treat all
          * assigns as unassigns */
         if (rd_kafka_fatal_error_code(rkcg->rkcg_rk)) {
+                if (assignment)
+                        rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                                     "ASSIGN", "Group \"%s\": Consumer is in "
+                                     "a failed state, treating assign as "
```
edenhill, line 2946: "failed state" introduces new wording for an existing concept, please have it read "Consumer is in a fatal state/raised a fatal error/or something that says fatal and not failed".
The same goes for previous occurences.

src/rdkafka_cgrp.c
```diff
@@ -3152,6 +3442,19 @@ rd_kafka_cgrp_op_serve (rd_kafka_t *rk, rd_kafka_q_t *rkq,
                 rko = NULL;
                 break;
 
+        case RD_KAFKA_OP_CG_METADATA:
+                /* Return the current consumer group metadata. */
+                rko->rko_u.cg_metadata = rkcg->rkcg_member_id
+                        ? rd_kafka_consumer_group_metadata_new_with_genid(
+                                rkcg->rkcg_rk->rk_conf.group_id_str,
+                                rkcg->rkcg_generation_id,
+                                RD_KAFKAP_STR_DUP(rkcg->rkcg_member_id),
```
edenhill, line 3451: The dupped member_id is leaked here, right? (did this pass asan/valgrind?)

edenhill: Suggest `rkcg->rkcg_member_id->str` as rkcg_member_id is safe to use since it is new():ed

src/rdkafka_cgrp.c
```diff
@@ -3241,20 +3546,50 @@ rd_kafka_cgrp_op_serve (rd_kafka_t *rk, rd_kafka_q_t *rkq,
...
                          * when terminating. */
                         rd_kafka_cgrp_unassign(rkcg);
+
                         if (rko->rko_u.assign.partitions)
                                 err = RD_KAFKA_RESP_ERR__DESTROY;
+
+                        error = !err ? NULL
+                                     : rd_kafka_error_new(
+                                                err,
+                                                rd_kafka_err2str(err));
```
edenhill, line 3562: format string, and provide something more meaningful, e.g., "Consumer is terminating".

src/rdkafka_cgrp.c
```diff
@@ -3241,20 +3546,50 @@ rd_kafka_cgrp_op_serve (rd_kafka_t *rk, rd_kafka_q_t *rkq,
...
+                                                rkcg,
+                                                rko->rko_u.assign.partitions);
+                                break;
+                        case RD_KAFKA_ASSIGN_METHOD_INCR_UNASSIGN:
+                                error = rd_kafka_cgrp_incremental_unassign(
+                                                rkcg,
+                                                rko->rko_u.assign.partitions);
+                                break;
+                        default:
+                                rd_assert(0);
```
edenhill, line 3587: RD_NOTREACHED();

src/rdkafka_cgrp.c
```diff
@@ -3865,99 +4239,164 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_write (
...
+                                                        generation_id,
+                                                        member_id,
+                                                        group_instance_id);
 
         return NULL;
 }
 
 
-static int unittest_consumer_group_metadata (void) {
+static int unittest_iteration(const char *group_id,
```
edenhill, line 4317: unittest_consumer_group_metadata_iteration() or at least something that gives this a contextual name.
And ut_ is fine for static unittest funciton prefixes.

src-cpp/KafkaConsumerImpl.cpp
mhowlett: yeah, this is primarily due to just following the pattern set elsewhere in this class.

src-cpp/rdkafkacpp.h
mhowlett: oh yeah, should have remembered that.

src/rdkafka.h
mhowlett: it would be clearer, but it would also be incorrect :-)

these are your words, I just moved the paragraph because the different order made better sense. I also didn't like this word, however I was too lazy to reword. I'll do it now though. Changing to 'should' and adding additional explanation.

src/rdkafka.h
mhowlett: good idea, `group_instance_id` is nullable.

src/rdkafka_op.c
mhowlett: it's not, `rd_afka_op_destroy` `RD_IF_FREE`'s the error.

src/rdkafka_op.c
mhowlett: I don't understand why this is wrong, since this is just a format string without formatting? I will do something more verbose which i hope is what you are expecting, but i don't get why I'm doing it.

tests/0113-cooperative_rebalance.cpp
mhowlett: good, otherwise the number of comments would have at least doubled.

tests/0113-cooperative_rebalance.cpp
mhowlett: oh yeah, C++'s notion of an enum is a bit stupid.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                             "ASSIGN", "Group \"%s\": consumer is in "
+                             "a failed state, treating incremental "
+                             "assign of %d partitions as unassign of all"
+                             "%d partitions in the current assignment",
+                             rkcg->rkcg_group_id->str, partitions->cnt,
+                             !rkcg->rkcg_assignment ? 0
+                             : rkcg->rkcg_assignment->cnt);
+                err = rd_kafka_cgrp_assign(rkcg, NULL);
+                rd_assert(err);
```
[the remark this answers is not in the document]
mhowlett: it should always return an error because the consumer is in a failed state. i have now made the assert more specific to check that err is fatal. this purpose is just fail fast in the case of developer error, and it adds a bit to code clarity. but will do as you direct if you want it out.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,280 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                return rd_kafka_error_new(err, rd_kafka_err2str(err));
```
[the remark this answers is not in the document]
mhowlett: still not understanding, but will do what i think you want.

src/rdkafka_cgrp.c
mhowlett: yeah, makes more sense.

src/rdkafka_cgrp.c
mhowlett: ok, any of the three make sense to me. going with CONFLICT.

src/rdkafka_cgrp.c
mhowlett: equally correct, but reads better.

src/rdkafka_cgrp.c
mhowlett: well spotted.

src/rdkafka_cgrp.c
mhowlett: i can assure you it is necessary based on refcount problems that occur if it's not here... i arrived  at this basically by starting with cgrp_assign/unassign methods and changing them for incremental updates. I will try to understand clearly why this is necessary and write a further note.

src/rdkafka_cgrp.c
mhowlett: how about that. i have no idea why i was checking for that.

src/rdkafka_cgrp.c
mhowlett: yeah, better change it in case some people have fuzzy integer ALUs in their CPUs.

src/rdkafka_cgrp.c
mhowlett: did same as prev comment. will remove if you want, but kind of like it here.

src/rdkafka_cgrp.c
mhowlett: ditto

src/rdkafka_cgrp.c
mhowlett: ok, i'll remove it, it provides only tiny value. still, i don't think it hurts to have these things littered about. if you have 100 of them, 1 of them will probably end up paying off some time.

src/rdkafka_cgrp.c
mhowlett: probably didn't when i wrote the dbg.

src/rdkafka_cgrp.c
mhowlett: i prefer to get down to business right away as well.

src/rdkafka_cgrp.c
mhowlett: actually more error prone to change given it's there, because there's a chance i'll accidentally leave the +=.

src/rdkafka_cgrp.c
mhowlett: yes, i'll update the text as well. my thinking was the other side of the inequality is captured by the for loop below, but your suggestion leads to a better error.

src/rdkafka_cgrp.c
mhowlett: this is me being dyslexic.

src/rdkafka_cgrp.c
mhowlett: done

src/rdkafka_cgrp.c
mhowlett: apologies then for > 80 col

src/rdkafka_cgrp.c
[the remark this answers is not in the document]
mhowlett: changing my mind, removing the assert because i think it's more likely you'll accept it.

src/rdkafka.h
```diff
@@ -3849,12 +3897,23 @@ rd_kafka_consumer_group_metadata_new (const char *group_id);
  * @brief Create a new consumer group metadata object.
  *        This is typically only used for writing tests.
  *
+ * @param group_id The group id (must not be NULL).
+ *
```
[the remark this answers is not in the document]
edenhill: It is usually not needed to say that a pointer argument must not be NULL, that's the standard assumption, and as long as we say which ones are optional and can be NULL I think it is less ambigious in the overall picture if required fields are not explicitly called out as required, otherwise we would need to update all doc strings everywhere.

src/rdkafka_cgrp.c
edenhill: Ah, right, one refcount (from get2()) goes on partitions, and the other (from keep()) goes on rkcg_assignment.
I think it is better to change add0() to keep() the rktp if it is specified and remove the keep from here.
Also update the one place where add0 is called with a keep().

src/rdkafka_cgrp.c
edenhill: not addressed

src/rdkafka_cgrp.c
```diff
@@ -3774,46 +4102,60 @@ rd_kafka_consumer_group_metadata_new (const char *group_id) {
...
-                                                 int32_t generation_id) {
+                                                 int32_t generation_id,
+                                                 const char *member_id,
+                                                 const char
+                                                 *group_instance_id) {
         rd_kafka_consumer_group_metadata_t *cgmetadata;
 
-        if (!group_id)
-                return NULL;
+        rd_assert(group_id);
```
edenhill, line 4119: We typically don't check the validity of input arguments unless we need it to catch bugs, but this will crash three lines down anyway if any of these are null, and the calling conventions are clear, so not really necessary and just causes confusion: "Why is this checked here when other places dont check arguments?"

src/rdkafka_cgrp.c
```diff
@@ -3865,99 +4231,164 @@ rd_kafka_error_t *rd_kafka_consumer_group_metadata_write (
...
-                                RD_KAFKA_RESP_ERR__BAD_MSG,
-                                "Input buffer group id is not safe");
+        member_id = next;
+        next = str_is_printable(member_id, end);
+        if (!next)
+                return rd_kafka_error_new(RD_KAFKA_RESP_ERR__BAD_MSG,
+                                          "Input buffer member id is not "
+                                          "safe");
+
+        group_instance_id_is_null = (int8_t)*next++;
```
edenhill, line 4285: Use parens around the next deref and incr to show intent

src/rdkafka_op.c
edenhill: But it hasnt been assigned to rko_error yet.

src/rdkafka_op.c
edenhill: But you dont know that err2str() will not return something that can be interpreted as a format string, which makes it unsafe.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,272 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                             "ASSIGN", "Group \"%s\": consumer has raised "
+                             "a fatal error, treating incremental "
+                             "assign of %d partitions as unassign of all"
+                             "%d partition(s) in the current assignment",
+                             rkcg->rkcg_group_id->str, partitions->cnt,
+                             !rkcg->rkcg_assignment ? 0
+                             : rkcg->rkcg_assignment->cnt);
+                err = rd_kafka_cgrp_assign(rkcg, NULL);
+                return rd_kafka_error_new(err, "%s", rd_kafka_err2str(err));
```
edenhill, line 2463: Why are we using the error code from cgrp_assign() here (will it actually return error?).
The error is that the consumer has raised a fatal error, and that's what this error_new() should say.
So:
`rd_kafka_error_new(RD_KAFKA_RESP_ERR__FATAL, "Consumer has raised a fatal error");`

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,272 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+        if (rd_kafka_fatal_error_code(rkcg->rkcg_rk)) {
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP|RD_KAFKA_DBG_CONSUMER,
+                             "UNASSIGN", "Group \"%s\": consumer has "
+                             "raised a fatal error, treating incremental "
+                             "unassign of %d partitions as unassign of all "
+                             "%d partitions in the current assignment",
+                             rkcg->rkcg_group_id->str,
+                             partitions->cnt, cur_assignment_cnt);
+                err = rd_kafka_cgrp_assign(rkcg, NULL);
+                return rd_kafka_error_new(err, "%s", rd_kafka_err2str(err));
```
edenhill, line 2577: same as previous comment.

src/rdkafka_cgrp.c
```diff
@@ -2434,6 +2435,272 @@ static void rd_kafka_cgrp_offset_commit_tmr_cb (rd_kafka_timers_t *rkts,
...
+        }
+
+        if (rd_atomic32_get(&rkcg->rkcg_assignment_lost) &&
+            partitions->cnt != cur_assignment_cnt) {
+                return rd_kafka_error_new(
+                        RD_KAFKA_RESP_ERR__INVALID_ARG,
+                        "Current assignment is lost, but size of "
```
edenhill, line 2584: :clap:

src/rdkafka_cgrp.c
```diff
@@ -3241,20 +3538,50 @@ rd_kafka_cgrp_op_serve (rd_kafka_t *rk, rd_kafka_q_t *rkq,
...
-                                                   rko->rko_u.assign.partitions);
+                        switch (rko->rko_u.assign.method)
+                        {
+                        case RD_KAFKA_ASSIGN_METHOD_ASSIGN:
+                                /* New atomic assignment (payload != NULL),
+                                 * or unassignment (payload == NULL) */
+                                err = rd_kafka_cgrp_assign(rkcg,
+                                        rko->rko_u.assign.partitions);
+                                error = !err ? NULL
+                                             : rd_kafka_error_new(
```
edenhill, line 3564: needs format string

src/rdkafka_op.c
mhowlett: you're good at this!

src/rdkafka_cgrp.c
mhowlett: sounds good. but two places? `rd_kafka_toppar_offset_fetch` and `rd_kafka_topic_partition_add_copy`

src/rdkafka_cgrp.c
mhowlett: on the upside, i addressed it in two other places instead.

src/rdkafka.h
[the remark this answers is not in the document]
mhowlett: sounds good

src/rdkafka_cgrp.c
mhowlett: sounds good.

src/rdkafka_cgrp.c
mhowlett: yes it will return an error (a fatal error).
that's why i was using asserts before (which i updated to asserting that the return value was fatal.. before removing it because you were objecting to asserts), to make it clear to the reader.

anyway, your suggestion is objectively best, thanks.

src/rdkafka_cgrp.c
mhowlett: thanks, updated.

PR #3029, opened 2020-08-13
Document that rd_kafka_subscribe() copies its topic list argument

andrewthad:


src/rdkafka.h
```diff
@@ -3547,6 +3547,9 @@ rd_kafka_offsets_store (rd_kafka_t *rk,
  *
  * Regex and full topic names can be mixed in \p topics.
  *
+ * @remark The \p topics list is copied, not aliased. It is safe to free the
```
edenhill, line 3550: I think this is implied by the 'const'-ness of the parameter and no mention of the pointer being held on to after the call returns.
This is the standard calling convention for C functions in my experience (i.e., all of libc) and does not explicit documentation

andrewthad: Thanks. I write very little C, so I'm not familiar with some of the conventions. My understanding is now that if `librdkafka` does not explicitly document that a function aliases a pointer argument, it can be assumed that it does not alias it. Is that accurate?

edenhill: That's correct, here's an example: 
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L1567

PR #3031, opened 2020-08-14
fix producer header option

wolfchimneyrock:
Currently, the ability to test producing messages with a header in `rdkafka_performance` is broken because of a subtle `getopt` error.  Although the intent is to give the option `-H` an optional argument, what is happening is optarg is always discarding an option even if present since the format string looks like `...HH:` - it never parses the latter part.

There are a couple of considerations:
* [GNU optarg supports optional arguments with a double colon in the format string](https://www.gnu.org/software/libc/manual/html_node/Using-Getopt.html) (would use `H::`) but not all platforms use GNU.
* [the POSIX standard actually forbids GNU-style optional arguments](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02).

Thus IMHO the best approach is to just make them different options: `-H <argument>` for the producer, and `-h` for the consumer.

### DIFF
--- a/examples/rdkafka_performance.c
+++ b/examples/rdkafka_performance.c
@@ -863,7 +863,7 @@ int main (int argc, char **argv) {
 	while ((opt =
 		getopt(argc, argv,
 		       "PCG:t:p:b:s:k:c:fi:MDd:m:S:x:"
-                       "R:a:z:o:X:B:eT:Y:qvIur:lA:OwNHH:")) != -1) {
+                       "R:a:z:o:X:B:eT:Y:qvIur:lA:OwNhH:")) != -1) {
 		switch (opt) {
 		case 'G':
 			if (rd_kafka_conf_set(conf, "group.id", optarg,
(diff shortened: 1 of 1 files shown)


examples/rdkafka_performance.c
```diff
@@ -863,7 +863,7 @@ int main (int argc, char **argv) {
 	while ((opt =
 		getopt(argc, argv,
 		       "PCG:t:p:b:s:k:c:fi:MDd:m:S:x:"
-                       "R:a:z:o:X:B:eT:Y:qvIur:lA:OwNHH:")) != -1) {
+                       "R:a:z:o:X:B:eT:Y:qvIur:lA:OwNhH:")) != -1) {
```
edenhill, line 866: I'd like to (p)reserve -h for help.
Since the consumer-side -H header parsing is for performance measurements/comparisons only I think we can require an argument in the consumer mode too, e.g., `-H parse` and set if read_hdrs to 1 if optarg is "parse".

examples/rdkafka_performance.c
```diff
@@ -962,37 +962,41 @@ int main (int argc, char **argv) {
...
-                        val = strchr(name, '=');
-                        if (val) {
-                                name_sz = (size_t)(val-name);
-                                val++; /* past the '=' */
-                        }
+        case 'h':
+            read_hdrs = 1;
+            break;
+        case 'H':
+            if (mode == 'P') {
```
edenhill, line 966: This requires the mode to be specified prior to headers, which is not required by other options.

edenhill: Suggest just checking for "parse" and then later on after the argument check is complete verify that mode is consumer/group-consumer when read_hdrs are set.

edenhill (changes requested):

examples/rdkafka_performance.c
wolfchimneyrock: ok, will check after argument parsing but before client creation

edenhill (approved):
LGTM

edenhill:
Thank you!

Issue #3016, opened 2020-07-30
Configuration property offset.store.method is deprecated


Quuxplusone:
Description
===========

Since v1.0.0 (commit 7f0236091), rdkafka emits a warning message whenever the client sets `offset.store.method=broker`:

    Configuration property offset.store.method is deprecated: Offset commit store method: 'file' - DEPRECATED: local file store (offset.store.path, et.al), 'broker' - broker commit store (requires "group.id" to be configured and Apache Kafka 0.8.2 or later on the broker.)

Presumably the client should write their code in some way that avoids spammy warnings, right?

However, rdkafka's example code still shows an example setting `offset.store.method=broker`.
https://github.com/edenhill/librdkafka/blob/9ca73e4dcee57d38e287a8728a70f6eff24cc73a/examples/rdkafka_complex_consumer_example.c#L481-L483

OT1H, if we're supposed to set this config option as shown, then it shouldn't give this spammy warning.

OTOH, if the option is truly deprecated and should not be used, then rdkafka's example should be updated to no longer use it.

Checklist
=========

Some of this I don't have available yet, but I'll fill it in when I get it.

 - [x] librdkafka version (release number or git tag): `v1.4.0`
 - [ ] Apache Kafka version: `<REPLACE with e.g., 0.10.2.3>`
 - [ ] librdkafka client configuration: `<REPLACE with e.g., message.timeout.ms=123, auto.reset.offset=earliest, ..>`
 - [ ] Operating system: `<REPLACE with e.g., Centos 5 (x64)>`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
The examples should definitely be updated, there's plenty to be done in that area.
Feel free to open a PR where this setting is removed from the examples 💯

PR #3035, opened 2020-08-18
Remove offset.store.method=broker from the examples/ directory.

Quuxplusone:
This config option is deprecated and should no longer be used in examples.

The non-deprecated behavior is implicitly "offset.store.method=broker".
Trying to set "offset.store.method=broker" explicitly gives a warning, and
trying to set "offset.store.method" to anything else is deprecated AND gives
a warning.

However, "offset.store.method=file" is still supported in src/,
and therefore tested. I didn't touch tests/.

Addresses #3016.


edenhill:
Thank you!

PR #3036, opened 2020-08-18
Fix Typo.

dogac00:


src/tinycthread.h
```diff
@@ -186,7 +186,7 @@ int _tthread_timespec_get(struct timespec *ts, int base);
  * tinycthreads but the runtime libc also provides C11 threads.
  * The *BSD values are notably different. */
 #define thrd_success  0 /**< The requested operation succeeded */
-#define thrd_busy     1 /**< The requested operation failed because a tesource requested by a test and return function is already in use */
+#define thrd_busy     1 /**< The requested operation failed because the resource has been requested by a test and return function is already in use */
```
edenhill, line 189: This changes the meaning of the function (which isnt great to begin with), but I think the proper fix is to just fix the spelling error.
The proper fix is to use the official c11 error descriptions, but that makes us deviate from upstream tinycthread.

edenhill (changes requested):

PR #3042, opened 2020-08-21
Add dual shared and static cmake targets and enable cpack package generation

masariello:


.gitignore
```diff
@@ -30,3 +30,6 @@ test_report*.json
 cov-int
 gdbrun*.gdb
 TAGS
+build
+.vs
+out
```
edenhill, line 35: missing newline at end of file.

CMakeLists.txt
```diff
@@ -5,6 +5,19 @@ parseversion("src/rdkafka.h")
 
 project(RdKafka VERSION ${RDKAFKA_VERSION})
 
+set(CPACK_PACKAGE_VENDOR "Magnus Edenhill")
+set(CPACK_PACKAGE_CONTACT "info@edenhill.se")
+set(CPACK_PACKAGE_HOMEPAGE_URL https://github.com/edenhill/librdkafka)
```
edenhill, line 10: Should this have quotes around the URL?

packaging/cmake/README.md
```diff
@@ -33,6 +33,24 @@ Install library:
 
 If you use librdkafka as submodule in cmake project and want static link of librdkafka:
 
-      set(RDKAFKA_BUILD_STATIC ON CACHE BOOL "")
       add_subdirectory(librdkafka)
-      target_link_libraries(your_library_or_executable rdkafka)
+      target_link_libraries(your_library_or_executable rdkafka_static)
+
+# Build DEB/RPM packages on Linux
```
edenhill, line 39: 👏

packaging/cmake/rdkafka.targets
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
edenhill, line 1: Can we reuse packaging/nuget/templates/librdkafka.redist.targets?

src/CMakeLists.txt
```diff
@@ -155,100 +155,88 @@ if(WITHOUT_WIN32_CONFIG)
...
-  target_link_libraries(rdkafka PUBLIC m)
-endif()
+foreach(target rdkafka rdkafka_static)
 
-if(WITH_ZLIB)
-  find_package(ZLIB REQUIRED)
-  target_include_directories(rdkafka PUBLIC ${ZLIB_INCLUDE_DIRS})
-  target_link_libraries(rdkafka PUBLIC ZLIB::ZLIB)
-endif()
+  # Support '#include <rdkafka.h>'
```
edenhill, line 177: Why without the directory prefix? librdkafka provides multiple header files and as such should have its own include directory.

src/rdkafka.h
```diff
@@ -56,45 +56,48 @@ extern "C" {
...
 
 #ifdef _WIN32
-#include <basetsd.h>
-#ifndef WIN32_MEAN_AND_LEAN
-#define WIN32_MEAN_AND_LEAN
-#endif
-#include <winsock2.h>  /* for sockaddr, .. */
-#ifndef _SSIZE_T_DEFINED
-#define _SSIZE_T_DEFINED
+#  include <basetsd.h>
```
edenhill, line 59: Please avoid whitespace/reformat changes (here and in rdkafkacpp.h)

src-cpp/rdkafkacpp.h
```diff
@@ -56,27 +56,26 @@
...
+#      include <basetsd.h>
+#    endif
+#    ifndef _SSIZE_T_DEFINED
+#      define _SSIZE_T_DEFINED
 typedef SSIZE_T ssize_t;
+#    endif
+#  endif
+#  ifndef LIBRDKAFKA_STATICLIB
+#    ifdef LIBRDKAFKACPP_EXPORTS
+#      define RDCPP_EXPORT __declspec(dllexport)
```
edenhill, line 70: Why the change of define name?

tests/CMakeLists.txt
```diff
@@ -119,6 +119,10 @@ endif()
 add_executable(test-runner ${sources})
 target_link_libraries(test-runner PUBLIC rdkafka++)
 
+# we don't really need this but it's good to test static linking
```
edenhill, line 122: 👍

edenhill (changes requested):

edenhill:
@raulbocanegra @Oxymoron79 @benesch 
Can you please review these CMake changes?

packaging/cmake/rdkafka.targets
masariello: Unfortunately that .targets file assumes the standard nuget directory layout which is not honoured by the cpack nuget generator. I agree. It's not great. But that makes the 2 ways of generating nuget packages fairly disjoint, therefore re-using the .targets file would not make much sense.

src-cpp/rdkafkacpp.h
masariello: When building `rdkafkac++`, the C APIs in `rdkafka` (i.e. called by `rdkafkac++`) ought to be flagged as `dllimport`, rather than `dllexport`. If you use the same #define for both DLLs, then you cannot correctly switch them this way at build time.

src/rdkafka.h
masariello: Ah. sorry. I had a hard time matching the blocks, so added indentation to help readability. Is it OK to leave it in?

src/CMakeLists.txt
masariello: That comment was actually moved from line 175 on master

.gitignore
masariello: added

tests/CMakeLists.txt
masariello: Thanks

CMakeLists.txt
masariello: Not needed in cmake. But consistency is nice. Added

packaging/cmake/rdkafka.targets
edenhill: Makes sense.

src/rdkafka.h
edenhill: Please restore the original indent so this PR only changes what it actually aims to change.

We'll address style when we have a proper C style checker.

src-cpp/rdkafkacpp.h
edenhill: Okay, that makes sense, but call it RD_CPP_EXPORT.

src/rdkafka.h
masariello: done

src-cpp/CMakeLists.txt
```diff
@@ -51,7 +48,7 @@ set(PKG_CONFIG_CFLAGS
     "-I\${includedir}"
 )
 set(PKG_CONFIG_LIBS
-    "-L\${libdir} -lrdkafka++"
```
Oxymoron79, line 54: Don't remove `-lrdkafka++` here. The pkg-config file shall provide a valid link list for users of the (shared) library. Omitting  this parameter would lead to linking errors.

src/CMakeLists.txt
```diff
@@ -285,7 +273,7 @@ set(PKG_CONFIG_CFLAGS
     "-I\${includedir}"
 )
 set(PKG_CONFIG_LIBS
-    "-L\${libdir} -lrdkafka"
+    "-L\${libdir}"
```
Oxymoron79, line 276: Don't remove `-lrdkafka` here as well.

CMakeLists.txt
```diff
@@ -256,8 +268,28 @@ install(
     DESTINATION "share/licenses/librdkafka"
 )
 
+INSTALL(FILES packaging/cmake/rdkafka.targets
+    DESTINATION "build/native"
```
Oxymoron79, line 272: On Linux/Unix systems this installation destination (will be installed to `$PREFIX/build/native`) is not following the standard directory structure.
I'm not familiar with NuGet packages, but similar to Pkg-Config files or Cmake Modules, there should be a system-wide directory to install these files to.

src-cpp/CMakeLists.txt
masariello: done

src/CMakeLists.txt
masariello: done

CMakeLists.txt
masariello: Nuget files are not deployment packages. They are used only at build time, so this should not be an issues.

masariello: This has been solved. See comments in today's commits.

masariello:
Hi guys
I think I covered all the corners with the last changes. Can you take a look, please?
Thanks

edenhill:
There's similar work being done in #3130, would be good if these two efforts could be combined.

CMakeLists.txt
```diff
@@ -159,7 +182,6 @@ if(ENABLE_LZ4_EXT)
 endif()
 # }
 
-option(RDKAFKA_BUILD_STATIC "Build static rdkafka library" OFF)
```
neptoess, line 162: Is it absolutely required to build both static and dynamic for CPack? The reason I ask is that removing this option will conflict with changes in https://github.com/edenhill/librdkafka/pull/3130, particularly https://github.com/edenhill/librdkafka/pull/3130/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a

src/CMakeLists.txt
```diff
@@ -308,36 +301,33 @@ configure_file(
 install(FILES ${GENERATED_DIR}/rdkafka.pc
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig"
 )
-if(RDKAFKA_BUILD_STATIC)
```
neptoess, line 311: Is this left off because, in theory, rdkafka.pc == rdkafka-static.pc?

CMakeLists.txt
masariello: I guess not. CPacking will work as long as the static targets are named differently from the dynamic ones.

src/CMakeLists.txt
masariello: It's been a while. Not sure why I removed the section. I have now restored it and all spotted a few little mistakes. Now fixed.

neptoess (approved):

edenhill:
Please resolve the conflicts

PR #3048, opened 2020-08-26
Not using timespec_get on android api level < 29

yagnasrinath:


src/rdtime.h
```diff
@@ -79,7 +79,7 @@ BOOL rd_ut_QueryPerformanceCounter(_Out_ LARGE_INTEGER * lpPerformanceCount);
  */
 static RD_INLINE rd_ts_t rd_clock (void) RD_UNUSED;
 static RD_INLINE rd_ts_t rd_clock (void) {
-#ifdef __APPLE__
+#if defined(__APPLE__) || (defined(__ANDROID__) && __API_LEVEL__ < 29)
```
edenhill, line 82: I see `__ANDROID_API__` (e.g. 29) being mentioned in other contexts, e.g.:
https://android.googlesource.com/platform/bionic/+/master/libc/include/android/api-level.h#168

Which one should be used?

yagnasrinath: You are right. I got the wrong macro. Corrected it now. Thanks!

edenhill:
Thank you!

PR #3050, opened 2020-08-28
Fix misspelled join state field in STATISTICS.md

arhoads:
join_state is the spelling used in the outputted JSON


arhoads:
See https://github.com/edenhill/librdkafka/blob/329c97fe0f408a2f91471fa3d835861b2025d6ea/src/rdkafka.c#L1704

edenhill:
Thank you!

PR #3053, opened 2020-08-28
add recent additions to kafka broker error codes

gridaphobe:
See https://github.com/apache/kafka/blob/b937ec75677f8af13bf6fda686f07e9c62cdd20f/clients/src/main/java/org/apache/kafka/common/protocol/Errors.java#L317-L328 for the upstream error code names and descriptions. It looks like `THROTTLING_QUOTA_EXCEEDED ` and `PRODUCER_FENCED ` have not made it into a released broker yet, so we may want to hold off on them until they're released. But the others are documented at https://kafka.apache.org/protocol.html#protocol_error_codes already, and `INVALID_RECORD` in particular is being returned for messages with no keys that are produced to compacted topics.

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -486,11 +486,7 @@ enum ErrorCode {
         /** Broker failed to validate record */
         ERR_INVALID_RECORD = 87,
         /** There are unstable offsets that need to be cleared */
-        ERR_UNSTABLE_OFFSET_COMMIT = 88,
-        /** The throttling quota has been exceeded */
-        ERR_THROTTLING_QUOTA_EXCEEDED = 89,
-        /** Producer has been fenced by another producer with the same transaction.id */
-        ERR_PRODUCER_FENCED = 90
+        ERR_UNSTABLE_OFFSET_COMMIT = 88
 };
 
 
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -475,6 +475,22 @@ enum ErrorCode {
         /** Static consumer fenced by other consumer with same
          * group.instance.id. */
         ERR_FENCED_INSTANCE_ID = 82,
+        /** Eligible partition leaders are not available */
+        ERR_ELIGIBLE_LEADERS_NOT_AVAILABLE = 83,
+        /** Leader election not needed for topic partition */
+        ERR_ELECTION_NOT_NEEDED = 84,
+        /** No partition reassignment is in progress */
+        ERR_NO_REASSIGNMENT_IN_PROGRESS = 85,
+        /** Deleting offsets of a topic while the consumer group is subscribed to it */
+        ERR_GROUP_SUBSCRIBED_TO_TOPIC = 86,
+        /** Broker failed to validate record */
+        ERR_INVALID_RECORD = 87,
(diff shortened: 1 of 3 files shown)


src/rdkafka.c
```diff
@@ -676,6 +676,22 @@ static const struct rd_kafka_err_desc rd_kafka_err_descs[] = {
...
+        _ERR_DESC(RD_KAFKA_RESP_ERR_GROUP_SUBSCRIBED_TO_TOPIC,
+                  "Broker: Deleting offsets of a topic while the consumer group is subscribed to it"),
+        _ERR_DESC(RD_KAFKA_RESP_ERR_INVALID_RECORD,
+                  "Broker: Broker failed to validate record"),
+        _ERR_DESC(RD_KAFKA_RESP_ERR_UNSTABLE_OFFSET_COMMIT,
+                  "Broker: There are unstable offsets that need to be cleared"),
+        _ERR_DESC(RD_KAFKA_RESP_ERR_THROTTLING_QUOTA_EXCEEDED,
+                  "Broker: The throttling quota has been exceeded"),
+        _ERR_DESC(RD_KAFKA_RESP_ERR_PRODUCER_FENCED,
+                  "Broker: Producer has been fenced by another with the same transaction.id"),
```
edenhill, line 694: "by another producer" is clearer (here and in other places)

src/rdkafka.c
```diff
@@ -676,6 +676,22 @@ static const struct rd_kafka_err_desc rd_kafka_err_descs[] = {
         _ERR_DESC(RD_KAFKA_RESP_ERR_FENCED_INSTANCE_ID,
                   "Broker: Static consumer fenced by other consumer with same "
                   "group.instance.id"),
+        _ERR_DESC(RD_KAFKA_RESP_ERR_ELIGIBLE_LEADERS_NOT_AVAILABLE,
+                  "Broker: Eligible topic partition leaders are not available"),
```
edenhill, line 680: I think the "topic" part is a bit redundant, and inconsistent with existing properties, even though the Java code has them I suggest stripping it out and just having "partition".

edenhill (changes requested):
Good stuff. rdkafkacpp.h needs to be updated as well.

src-cpp/rdkafkacpp.h
```diff
@@ -475,6 +475,22 @@ enum ErrorCode {
...
+        /** No partition reassignment is in progress */
+        ERR_NO_REASSIGNMENT_IN_PROGRESS = 85,
+        /** Deleting offsets of a topic while the consumer group is subscribed to it */
+        ERR_GROUP_SUBSCRIBED_TO_TOPIC = 86,
+        /** Broker failed to validate record */
+        ERR_INVALID_RECORD = 87,
+        /** There are unstable offsets that need to be cleared */
+        ERR_UNSTABLE_OFFSET_COMMIT = 88,
+        /** The throttling quota has been exceeded */
+        ERR_THROTTLING_QUOTA_EXCEEDED = 89,
```
edenhill, line 491: We'll need to wait until these error code are in a proper AK release since it may change while on trunk head. 2.6.0's last error code is 88.

edenhill (changes requested):

edenhill (approved):
LGTM

edenhill:
Thank you!

PR #3055, opened 2020-08-31
Configuration property enhancements

edenhill:


edenhill:
This provides more meaningful error messages when a user tries to configure disabled-at-build-time features, such as Kerberos.

And.. CONFIGURATION.md will no longer be updated due to missing build features which will avoid dirty trees on clean builds.

edenhill:
The CI failures are unrelated

CHANGELOG.md
```diff
@@ -1,12 +1,25 @@
...
 
+## Enhancements
+
+ * The generated `CONFIGURATION.md` (through `rd_kafka_conf_properties_show())`)
+   now include all properties and values, regardless if they were included in
+   the build, and setting a disabled property or value through
+   `rd_kafka_conf_set()` now returns `RD_KAFKA_CONF_INVALID` and provides
+   a more useful error string saying why the property can't be set.
+ * Consumer configs on producers and vice versa will now be logged with
+   warning messages on client instantiation.
```
mhowlett, line 11: ideally, the client wouldn't even start up in this case. Maybe make this the behavior in 2.0?

CHANGELOG.md
```diff
@@ -1,12 +1,25 @@
...
+   All configuration warnings can be disabled by setting
+   `log.configuration.warnings=false`.
```
mhowlett, line 13: why would you ever want to do this?

src/rdkafka.c
```diff
@@ -987,6 +987,27 @@ static void rd_kafka_destroy_app (rd_kafka_t *rk, int flags) {
                      "(destroy flags %s (0x%x))",
                      flags ? flags_str : "none", flags);
 
+        /* If producer still has messages in queue the application
```
mhowlett, line 990: "If a producer still has messages queued, the .."

CHANGELOG.md
edenhill: Since this PR could be intrusive/noisy for apps that share the same config for producers and consumers we'll want to give people a way to opt out of these warnings as to prevent their monitoring systems to wreak havoc.

mhowlett (approved):
lgtm, except it seems unfortunate these are just warnings (i'm guessing you have your reasons), and i can't think why `log.configuration.warnings` has enough value to bother with.

edenhill:
> except it seems unfortunate these are just warnings (i'm guessing you have your reasons),

That would be a breaking change, requiring a semver bump.

> and i can't think why log.configuration.warnings has enough value to bother with.

Okay, then I'll remove it.

PR #3058, opened 2020-09-01
Update consumer.c

zhangzhanhong:
Line 246 may be rkm->payload


edenhill:
Thank you for the fix!

PR #3059, opened 2020-09-01
rdkafka_mock: fix message v2 overhead sanity check

gridaphobe:
The Message V2 API uses VARINTs, which means we have to use the
*minimal* overhead to sanity check the size of a ProduceRequest.

### DIFF
--- a/src/rdkafka_proto.h
+++ b/src/rdkafka_proto.h
@@ -505,7 +505,7 @@ typedef struct rd_kafka_buf_s rd_kafka_buf_t;
 /**
  * MsgVersion v2
  */
-#define RD_KAFKAP_MESSAGE_V2_OVERHEAD                                  \
+#define RD_KAFKAP_MESSAGE_V2_MAX_OVERHEAD                              \
         (                                                              \
         /* Length (varint) */                                          \
         RD_UVARINT_ENC_SIZEOF(int32_t) +                               \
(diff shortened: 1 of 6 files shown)


src/rdkafka_proto.h
```diff
@@ -523,6 +523,23 @@ typedef struct rd_kafka_buf_s rd_kafka_buf_t;
         RD_UVARINT_ENC_SIZEOF(int32_t)                                 \
         )
 
+#define RD_KAFKAP_MESSAGE_V2_MIN_OVERHEAD                              \
+        (                                                              \
+        /* Length (varint) */                                          \
+        RD_UVARINT_ENC_SIZE_0() +                                      \
```
edenhill, line 529: https://travis-ci.org/github/edenhill/librdkafka/jobs/723237176#L246

Suggest changing the ENC_SIZE_0() define to  `((int64_t)1)`

gridaphobe: Good catch, I used `size_t` instead to silence the signed-unsigned comparison warning.

edenhill (approved):
LGTM, great stuff!

edenhill:
How did this problem surface? In one of the existing tests or your own use of the mock broker?

gridaphobe:
We discovered it in our own use of the mock broker. We were sending tiny 5 byte messages in some unit tests. The mock broker worked fine as long as we didn’t allow batching, but once we let messages batch, they were all rejected with “invalid message size”.

edenhill:
Superb, thank you!

PR #3060, opened 2020-09-02
Fix destructor for KafkaConsumer

edenhill:


src-cpp/rdkafkacpp_int.h
```diff
@@ -1103,7 +1103,8 @@ class ConsumerGroupMetadataImpl : public ConsumerGroupMetadata {
 class KafkaConsumerImpl : virtual public KafkaConsumer, virtual public HandleImpl {
 public:
   ~KafkaConsumerImpl () {
-
+    if (rk_)
+      rd_kafka_destroy(rk_);
```
mhowlett, line 1107: maybe you should set `RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE`?

src-cpp/KafkaConsumerImpl.cpp
```diff
@@ -251,6 +251,7 @@ RdKafka::KafkaConsumerImpl::close () {
   while (rd_kafka_outq_len(rk_) > 0)
     rd_kafka_poll(rk_, 10);
   rd_kafka_destroy(rk_);
+  rk_ = NULL;
```
mhowlett, line 254: I'm guessing this will mean that any method call after `close` will segfault, and I also guess that maybe you get better failure behavior in this case if `rk_` weren't set to NULL.

src-cpp/rdkafkacpp_int.h
edenhill: Good call!

src-cpp/KafkaConsumerImpl.cpp
edenhill: Yeah, that's not good, changed it so it only calls close(), the destroy is done in the dtor as it is supposed.
Otherwise we would need to protect each object `rk_` access with a mutex, which would be costly and also add unnecessary concurrent constraints.

Thanks for being so switched on!

edenhill:
@mhowlett Please re-review

mhowlett (approved):

PR #3062, opened 2020-09-03
Add proper producer support for topic authorization failures

edenhill:
And fix a TxnMgr bug it surfaced :fire: .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@
    for certain combinations of members and partitions.
    This is a regression in v1.5.0. (#3024)
 
+### Producer fixes
+
+ * Topic authorization errors are now properly propagated for produced messages,
+   both through delivery reports and as `ERR_TOPIC_AUTHORIZATION_FAILED`
+   return value from `produce*()` (#2215)
+
 
 # librdkafka v1.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,10 @@
  * Topic authorization errors are now properly propagated for produced messages,
    both through delivery reports and as `ERR_TOPIC_AUTHORIZATION_FAILED`
    return value from `produce*()` (#2215)
+ * Treat cluster authentication failures as fatal in the transactional
+   producer (#2994).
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,15 @@
     refcounting issues when used on light-weight (error-only) topic objects
     such as consumer errors (#2693)
 
+### Producer fixes
+
+ * Topic authorization errors are now properly propagated for produced messages,
+   both through delivery reports and as `ERR_TOPIC_AUTHORIZATION_FAILED`
+   return value from `produce*()` (#2215)
+ * The transactional producer code did not properly reference-count partition
+   objects which could in very rare circumstances lead to a user-after-free bug
```
mhowlett, line 17: use-after-free

src/rdkafka_request.c
```diff
@@ -3001,12 +3005,20 @@ static int rd_kafka_handle_Produce_error (rd_kafka_broker_t *rkb,
...
+            perr->err == RD_KAFKA_RESP_ERR__TIMED_OUT_QUEUE) {
+                /* Translate request-level timeout error code
+                 * to message-level timeout error code. */
                 perr->err = RD_KAFKA_RESP_ERR__MSG_TIMED_OUT;
 
+        } else if (perr->err == RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED) {
+                /* If we're no longer authorized to access the topic mark
+                 * it as errored to deny further produce requests. */
+                rd_kafka_topic_wrlock(rktp->rktp_rkt);
+                rd_kafka_topic_set_error(rktp->rktp_rkt, perr->err);
```
mhowlett, line 3018: have you covered the case of auth flip-flopping from allowed to denied, then back to allowed again? (i don't see the topic error getting unset ever)

mhowlett: that case is probably of minuscule importance.

mhowlett (approved):

src/rdkafka_request.c
edenhill: The topic error will be reset on the next metadata refresh, given that the topic is now accessible.

PR #3066, opened 2020-09-10
Update consumer.c

Leo-Lin20:
I used this example in a recent project, but I found some problems.
`rd_kafka_conf_set(conf, "auto.offset.reset", "earliest", errstr, sizeof(errstr))`
This code will get "errstr"(No such configuration property: "auto.offset.reset").
so I think the configuration property is not in the "rd_kafka_conf_t", it maybe in the "rd_kafka_topic_conf_t"

Of course, I'm not sure what I said is right or wrong


edenhill:
No, it is fine set to topic configs on the global config object, in fact it is the proper way to do it unless you new topic-specific config.
My guess is that you're on a very old version of librdkafka that does not allow this.

The consumer.c example should be updated to only use the global config and not the topic config object.

PR #3076, opened 2020-09-17
Change default message retries to infinity

edenhill:
..since this is the recommendation.

The `retries` property was also reused for the default protocol request retry count, so changed that to a fixed value of 2 instead.


src/rdkafka_buf.c
```diff
@@ -134,7 +134,7 @@ rd_kafka_buf_t *rd_kafka_buf_new_request (rd_kafka_broker_t *rkb, int16_t ApiKey
         rd_kafka_broker_keep(rkb);
 
         rkbuf->rkbuf_rel_timeout = rkb->rkb_rk->rk_conf.socket_timeout_ms;
-        rkbuf->rkbuf_max_retries = rkb->rkb_rk->rk_conf.max_retries;
+        rkbuf->rkbuf_max_retries = RD_KAFKA_REQUEST_MAX_RETRIES;
```
mhowlett, line 137: didn't realize this was going on in general

mhowlett (approved):

src/rdkafka_buf.c
edenhill: Yeah this wasn't great.

PR #3100, opened 2020-10-05
Use reentrant version of rand() to avoid locking

azat:
rand() has a global lock [1], which is very very suboptiomal, so let's
consider the following example, suppose you have 1000 threads and you
need to generate 10e6 random values (if you have 10e6 messages with
random distribution you will need them):

  - rand:   0m2.336s
  - rand_r: 0m0.041s

  [1]: https://code.woboq.org/userspace/glibc/stdlib/random.c.html#__random

So rand() is ~60x times slower the rand_r(), for the reproducer see [2].

And also this means that with random partitioning you need to spend
extra 2.4 seconds to publish 10e6 message (while in real life, with
librdkafka performance will be even worse, since lots of threads will
contend for the lock, call futex and will slow down other threads too)

  [2]: https://gist.github.com/azat/690c3d63ec485555059c2a0ea7bff1c4

Plus here is top of syscalls (yes it is very inaccurate due to strace
overhead, but I think that for this particular usecase it is ok, and I
want to underline this anyway):

    $ strace -qq -f -c /tmp/rand_r |& head -n5
    % time     seconds  usecs/call     calls    errors syscall
    ------ ----------- ----------- --------- --------- ----------------
     46.13    0.064718          64      1000           clone
     13.93    0.019540           6      3001           rt_sigprocmask
     13.47    0.018891          18      1000           madvise
    $ strace -qq -f -c /tmp/rand |& head -n5
    % time     seconds  usecs/call     calls    errors syscall
    ------ ----------- ----------- --------- --------- ----------------
     86.76    4.615370          68     67369     26161 futex <-- **
      4.11    0.218796          72      3001           rt_sigprocmask
      3.17    0.168655         168      1000           clone

And since librdkafka uses threads massively this can become a bottleneck
(for example if you have consumer/producer per partition).

v2: introduce rd_rand_r() for win32 (that lacks of rand_r()).
v3: detect rand_r() at runtime use rand_s() on win32.
v4: add more information into the commit.


edenhill:
Windows build needs to be fixed.
Must not remove configuration properties without a semver bump, so if enable.random.seed is no longer used it should be marked with _RK_DEPRECATED, but I still think we need to use it on Windows (where there is no rand_r()) as rd_jitter() is called from the application thread.

azat:
>doozer/target/xenial-amd64

```
|[31m 0104_fetch_from_follower_mock            |     FAILED |  59.356s [0m|[31m test_consumer_poll():3804: Consume: consumer_poll() timeout (0/1 eof, 0/1000 msgs)[0m
```

flacky?

azat:
>Windows build needs to be fixed.

Added version of rand_r (rk_rand_r)

> Must not remove configuration properties without a semver bump, so if enable.random.seed is no longer used it should be marked with _RK_DEPRECATED

Fixed

> but I still think we need to use it on Windows (where there is no rand_r()) as rd_jitter() is called from the application thread.

After rk_rand_r(), looks like there is no point in this?

edenhill:
It seems rand_r() is deprecated and might be removed from future posix versions:
https://stackoverflow.com/a/29135420/1821055

This means there might be portability issues with relying on rand_r() always existing on Unix-like platforms.

What we could do is check for rand_r() in configure at build time, and if not available fall back on standard rand(). This means the srand config must not be deprecated.

But I'm curious, on what platform did you see performance issues with rand(), and do you have any numbers to show comparing librdkafka with rand() vs rand_r()?

azat:
>This means there might be portability issues with relying on rand_r() always existing on Unix-like platforms.

This PR does not contain rand_r already (rand_r inlined).

>But I'm curious, on what platform did you see performance issues with rand()

[`rand()` in glibc contains global lock](https://code.woboq.org/userspace/glibc/stdlib/random.c.html#__random) and it is very bad for performance

> and do you have any numbers to show comparing librdkafka with rand() vs rand_r()?

So here is the gist that will show the problem in numbers - https://gist.github.com/azat/690c3d63ec485555059c2a0ea7bff1c4

azat:
And here are some numbers for time spend in syscalls (yes it is very inaccurate due to strace overhead, but it shows the problem for this particular case)

```
$ strace -qq -f -c /tmp/rand_r |& head -n5
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 46.13    0.064718          64      1000           clone
 13.93    0.019540           6      3001           rt_sigprocmask
 13.47    0.018891          18      1000           madvise
```

```
$ strace -qq -f -c /tmp/rand |& head -n5
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 86.76    4.615370          68     67369     26161 futex
  4.11    0.218796          72      3001           rt_sigprocmask
  3.17    0.168655         168      1000           clone
```

edenhill:
Wow, that's quite a difference.

src/rdrand.c
```diff
@@ -28,8 +28,46 @@
 
 #include "rd.h"
 #include "rdrand.h"
+#include "rdtime.h"
+#include "tinycthread.h"
 
+/** Copied from glibc */
```
edenhill, line 34: We really don't want to use and maintain our own rand implementation, that's not an area of expertise for the librdkafka maintainers and is bound to be problematic sooner or later.

My suggestion is:
 * add a check for rand_r to configure.self and use it if available, else us standard rand() and enable.random.seed.
 * use rand_s() on win32
 * update the description for enable.random.seed and say that it is only used when rand_r() nor rand_s() is available.

edenhill (changes requested):

src/rdrand.c
azat: >We really don't want to use and maintain our own rand implementation, that's not an area of expertise for the librdkafka maintainers and is bound to be problematic sooner or later.

It is pretty simple and hadn't been changed since 1999, but okay will add runtime detection

> use rand_s() on win32

Actually I don't think that it should be used since

>The rand_s function writes a pseudorandom integer in the range 0 to UINT_MAX to the input pointer. The rand_s function uses the operating system to generate cryptographically secure random numbers. It does not use the seed generated by the srand function, nor does it affect the random number sequence used by rand.

So I guess it is even more heavy then rand() with global lock, and using it simply for random distribution is overkill

azat:
@edenhill friendly ping

src/rdrand.c
```diff
@@ -28,8 +28,39 @@
...
+ * operating system to generate cryptographically secure random numbers.
+ * And this is not what you need for simple random distribution.
+ */
+#ifndef _WIN32
+#include "../config.h"
+#endif
+
+int rd_jitter (int low, int high) {
+	int rand_num;
+#if defined(HAVE_RAND_R)
```
edenhill, line 47: Use `#if HAVE_RAND_R`

src/rdrand.h
```diff
@@ -33,12 +33,7 @@
 /**
  * Returns a random (using rand(3)) number between 'low'..'high' (inclusive).
  */
-static RD_INLINE int rd_jitter (int low, int high) RD_UNUSED;
-static RD_INLINE int rd_jitter (int low, int high) {
-	return (low + (rand() % ((high-low)+1)));
-	
-}
-
+int rd_jitter (int low, int high) RD_UNUSED;
```
edenhill, line 36: remove RD_UNUSED

edenhill (changes requested):
Almost there!

src/rdrand.c
```diff
@@ -28,8 +28,39 @@
...
 
+/*
+ * Config is only for rand_r(), there is no rand_r() on windows anyway
+ *
+ * Also note that rand_s() *SHOULD NOT BE* used here since it uses the
+ * operating system to generate cryptographically secure random numbers.
+ * And this is not what you need for simple random distribution.
+ */
+#ifndef _WIN32
+#include "../config.h"
```
edenhill, line 42: This is already included from rd.h

edenhill (changes requested):

edenhill:
This is great, thank you!

PR #3112, opened 2020-10-21
Incremental fixes to incremental rebalancing

edenhill:


examples/rdkafka_complex_consumer_example.c
```diff
@@ -190,33 +190,56 @@ static void print_partition_list (FILE *fp,
...
-		wait_eof = 0;
-		break;
+                if (!strcmp(rd_kafka_rebalance_protocol(rk), "COOPERATIVE"))
+                        error = rd_kafka_incremental_assign(rk, partitions);
+                else
+                        ret_err = rd_kafka_assign(rk, partitions);
+                wait_eof += partitions->cnt;
+                break;
+
+        case RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS:
```
mhowlett, line 214: probably should add a check for lost partitions (at least fprintf it).

would ideally show typical use as well with committing offsets, but adding that in this eg seems more optional.

examples/rdkafka_complex_consumer_example.cpp
```diff
@@ -151,14 +151,32 @@ class ExampleRebalanceCb : public RdKafka::RebalanceCb {
...
+      if (consumer->rebalance_protocol() == "COOPERATIVE") {
+        error = consumer->incremental_unassign(partitions);
+        partition_cnt -= (int)partitions.size();
+      } else {
+        ret_err = consumer->unassign();
+        partition_cnt = 0;
+      }
     }
-    eof_cnt = 0;
+    eof_cnt = 0; /* FIXME: Won't work well with COOPERATIVE */
```
mhowlett, line 172: fixme

mhowlett: maybe the easiest way to make this correct is to check if the assignor has been set to cooperative sticky on startup, and make this mutually exclusive with exit on eof being set.

examples/rdkafka_performance.c
```diff
@@ -339,33 +339,58 @@ static void msg_consume (rd_kafka_message_t *rkmessage, void *opaque) {
...
-	}
+                if (!strcmp(rd_kafka_rebalance_protocol(rk), "COOPERATIVE")) {
+                        error = rd_kafka_incremental_unassign(rk, partitions);
+                        partition_cnt -= partitions->cnt;
+                } else {
+                        ret_err = rd_kafka_assign(rk, NULL);
+                        partition_cnt = 0;
+                }
+
+                eof_cnt = 0; /* Not really correct for incremental case */
```
mhowlett, line 379: (not at all correct)

examples/transactions.c
```diff
@@ -364,6 +367,9 @@ consumer_group_rebalance_cb (rd_kafka_t *rk,
                              void *opaque) {
         int i;
 
+        if (!strcmp(rd_kafka_rebalance_protocol(rk), "COOPERATIVE"))
+            fatal("COOPERATIVE rebalance protocol not supported");
```
mhowlett, line 371: maybe "This example has not yet been updated to work with incremental rebalancing". I'm worried the proposed wording makes it seem like transactions don't work with incrreb.

mhowlett (approved):
some suggestions, lgtm after you action those to your satisfaction.

PR #3113, opened 2020-10-21
Add `dogstatsd.endpoint` configuration option

coignetp:
Hi!

I'd like to add a configuration option named `dogstatsd.endpoint`. The goal here is to allow the lib to emit to a DogStatsD endpoint (some documentation about DogStatsD here https://docs.datadoghq.com/developers/dogstatsd/, https://docs.datadoghq.com/developers/dogstatsd/datagram_shell). When `dogstatsd.endpoint` is set, some of the statistics are emitted to this endpoint every `statistics.interval.ms` ms.
This option provides an alternative to parsing the json stats and help people understand what's happening in their consumers/producers.

This pull request is still Work In Progress, since I have a few questions here. Also, I think I should add a `120_dogstatsd` test case.

Please tell me what you think about this change!


src/rdkafka_int.h
```diff
@@ -216,6 +216,9 @@ struct rd_kafka_s {
 	rd_kafka_q_t *rk_rep;   /* kafka -> application reply queue */
 	rd_kafka_q_t *rk_ops;   /* any -> rdkafka main thread ops */
 
+        int rk_dogstatsd_sockfd; /* TODO: put it in a better place */
+        rd_sockaddr_inx_t rk_dogstatsd_addr; /* TODO: better place */
```
coignetp, line 220: Maybe these should be nested in a struct somewhere else?

src/rdkafka_conf.h
```diff
@@ -156,7 +156,7 @@ typedef enum {
 
 /* Increase in steps of 64 as needed.
  * This must be larger than sizeof(rd_kafka_[topic_]conf_t) */
-#define RD_KAFKA_CONF_PROPS_IDX_MAX (64*27)
+#define RD_KAFKA_CONF_PROPS_IDX_MAX (64*28) // TODO
```
coignetp, line 159: I'm not sure I understood why this is needed

edenhill:
Hi Paul,

thanks for your work on this, it does seem like a useful feature to customers of Datadog.

However, I'm reluctant to add builtin support for a proprietary system, I think
this would be better as an add-on library that a user uses in conjunction with librdkafka, or as
a librdkafka plugin.

There's also asks to provide prometheus integration in librdkafka, but that too should reside outside of
the librdkafka code base.

What I think we want to do is add interceptors to the internal stats so that a (datadog, prometheus, ..) plugin
can intercept the metrics and bake it in a form that suits them.
I haven't thought this through in detail since I don't know the exact requirements, but due to the way that metrics
are collected inside librdkafka I want to avoid having interceptors for each update to a metric, but instead in the
metrics collector (that fires every statistics.interval.ms).
Maybe an interceptor that get's called for each emitted metric, as so:
```c
void on_metrics_emit (rd_kafka_t *rk, const char *metrics_name, const rd_kafka_metric_value_t *value, void *ic_opaque)
```
where `metrics_name` is `"brokers.3.rxbytes"` and `metric_value` is `{ .type = RD_KAFKA_METRIC_COUNTER_U64, .value.u64 = 12345 }`.

Let me know your thoughts on this overall approach.

/Magnus


P.S. There's also ongoing work to formalize a standard metrics interface for Kafka clients, a KIP will be posted shortly.

coignetp:
Hi,

Thank you for your detailed answer. I understand your concerns, and I’m trying to understand how can we make an interesting plugin or add-on. Would adding the support for statsd instead of dogstatsd be good for you? (https://github.com/statsd/statsd)
Also, in order to create a plugin, should we go here https://www.confluent.io/hub/ so it could be in the confluent-community package? Is there a documentation to do that? And if we create a `dogstatsd-interceptor` plugin and someone wants to use it, they would need to set `"plugin.library.paths"` configuration to `dogstatsd-interceptor`, and the plugin would implement an `on_metric_emit` interceptor or something, right? I just want to be sure I understand the process here.

About the standard metrics interface,if I understand correctly, it is for all Kafka client to expose their metrics? (`confluent-kafka-python` and `confluent-kafka-go` for example). How would this affect librdkafka metrics interface? It may be interesting to wait for it to be posted before creating the plugin, thanks for the information!


Concerning the addition of a new interceptor, I agree that having an interceptor at each metric update should be avoided, and I believe it can be a good idea to have `statistics.interval.ms` as a common configuration for stuff related to metrics update.

Just to be sure, this interceptor `on_metrics_emit` would be called for each emitted metric, meaning it would get called ~100 times every `statistics.interval.ms` with a different metric name and value? What about sending them by batch (like a `rd_avg_t` value type), or even sending them all at once? What would be the cons of having an interceptor that gets called once every `statistics.interval.ms` with all the metrics? Something like:

```C
void on_all_metrics_emit (rd_kafka_t *rk, const char **metrics_name, const rd_kafka_metric_value_t **values, const ssize_t size, void *ic_opaque)
```
where `metrics_name` is `{"brokers.2.tx",  "brokers.2.txbytes", "brokers.2.rx", "brokers.2.rxbytes"}` and `values` as a table of what you described.

Thanks again for all the information you provided!

Issue #614, opened 2016-04-15
poll wake?


rthalley:
We have a mode called RD_POLL_INFINITE_WAKEABLE that lets a producer do an infinite sleep in rk_kafka_poll() and yet allows any threads using that mode to be woken up with rd_kafka_poll_wake().

The advantage here is that you can do an indefinite sleep, thus not costing any effort when there is nothing to do, yet still be able to cause a timely program exit.  Prior to this change we had to have a short timeout in the poll so the thread could check if it needed to exit or not.

Interested?

edenhill:
Interesting. How did you wake up the polling thread?

rthalley:
We call cnd_broadcast(&rkq->rkq_cond).  The poll waiting code was modified to break out of the loop and return if the condition variable wait ended and we were in the RD_POLL_INFINITE_WAKEABLE mode.

edenhill:
Okay, there may be spurious wakeups on condvars though in which case the poll will not restart with INFINITE_WAKEABLE.

What  do you think of enqueueing an op on the queue instead letting the poller know it should bail out?

rthalley:
I wasn't worried about spurious wakeups since the caller has the poll wrapped in a loop, but I'm ok with an enqueue-an-op solution too.

edenhill:
This is usable in the language bindings as well.

edenhill:
I think the most useful approach is to provide this api on queue_t's, I'm not sure about the name but something like: `void rd_kafka_queue_cancel(rd_kafka_queue_t *rkqu);`

 * cancel
 * yield (might be confused with rd_kafka_yield())
 * wakeup
 * wake

Quuxplusone:
@edenhill: I see you closed this as "completed" on 2022-04-07, but can you put something here about _how_ it was completed? I'm just coming around to the conclusion that `rd_kafka_yield` does _not_ do what I want (in fact, some documentation claims it should never be called unless from within a callback handler!), which means either I need to revive #3114, or I need to use the feature you claim to have added in April. Can you elaborate on that feature, please?

PR #3114, opened 2020-10-21
Add rd_kafka_poll_wake() to wake up threads blocked in rd_kafka_poll().

Quuxplusone:
The idea is that if one thread is blocked in `rd_kafka_poll(rk, -1)`,
another thread can do `rd_kafka_poll_wake(rk)` to wake it up so
that we can cleanly join the polling thread and cleanly exit.

Addresses #614.

Upstreaming on the hope that if you take it, we don't have to maintain it as a patch anymore.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3711,6 +3711,9 @@ int rd_kafka_poll (rd_kafka_t *rk, int timeout_ms) {
         return r;
 }
 
+void rd_kafka_poll_wake (rd_kafka_t *rk) {
+        rd_kafka_q_wake(rk->rk_rep);
+}
 
 rd_kafka_event_t *rd_kafka_queue_poll (rd_kafka_queue_t *rkqu, int timeout_ms) {
         rd_kafka_op_t *rko;
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -2912,6 +2912,11 @@ void *rd_kafka_topic_opaque (const rd_kafka_topic_t *rkt);
 RD_EXPORT
 int rd_kafka_poll(rd_kafka_t *rk, int timeout_ms);
 
+ /**
+ * @brief Wake any threads blocked in rd_kafka_poll().
+ */
+RD_EXPORT
+void rd_kafka_poll_wake(rd_kafka_t *rk);
 
 /**
  * @brief Cancels the current callback dispatcher (rd_kafka_poll(),
--- a/src/rdkafka_queue.c
+++ b/src/rdkafka_queue.c
@@ -467,12 +467,14 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
 	}
 
         rd_timeout_init_timespec(&timeout_tspec, timeout_ms);
+        rkq->rkq_flags &= ~RD_KAFKA_Q_F_EXPLICIT_WAKE;
 
         /* Wait for op */
         while (!(rko = TAILQ_FIRST(&rkq->rkq_q)) &&
                !rd_kafka_q_check_yield(rkq) &&
                cnd_timedwait_abs(&rkq->rkq_cond, &rkq->rkq_lock,
-                                 &timeout_tspec) == thrd_success)
+                                 &timeout_tspec) == thrd_success &&
+               !(rkq->rkq_flags & ~RD_KAFKA_Q_F_EXPLICIT_WAKE))
                 ;
 
 	if (!rko) {
--- a/src/rdkafka_queue.h
+++ b/src/rdkafka_queue.h
@@ -65,6 +65,7 @@ struct rd_kafka_q_s {
                                       * by triggering the cond-var
                                       * but without having to enqueue
                                       * an op. */
+#define RD_KAFKA_Q_F_EXPLICIT_WAKE  0x10    /* Wake the queue immediately */
 
         rd_kafka_t   *rkq_rk;
 	struct rd_kafka_q_io *rkq_qio;   /* FD-based application signalling */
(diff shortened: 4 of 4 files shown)


src/rdkafka.c
```diff
@@ -3711,6 +3711,9 @@ int rd_kafka_poll (rd_kafka_t *rk, int timeout_ms) {
         return r;
 }
 
+void rd_kafka_poll_wake (rd_kafka_t *rk) {
+        rd_kafka_q_wake(rk->rk_rep);
```
edenhill, line 3715: There's already rd_kafka_q_yield() that does the same thing, could use that here instead.
Might be a good idea to change the cnd_signal() to cnd_broadcast() in q_yield().

edenhill (changes requested):

src/rdkafka.c
Quuxplusone: Bearing in mind that I have very little idea how any of this works ;) do you think it might make sense to provide _both_ 
- `rd_kafka_q_wake_one()`, formerly known as `rd_kafka_q_yield()`
- `rd_kafka_q_wake_all()`, formerly known as `rd_kafka_q_wake()`

so that we can use the latter here, without accidentally breaking or pessimizing any of the callers of the former?

edenhill: I think q_yield() should be using a broadcast instead of a signal, making q_yield and q_wake identical.

Quuxplusone: Well, I'm still attached to my idea, but as I don't actually know this code at all, I defer to your expertise. :) Do you want me to amend this patch along those lines, or are you planning to change the behavior of `rd_kafka_q_yield` in master regardless, or what's the deal at this point?

edenhill: Please go ahead and change q_yield() to use broadcast instead of signal.

Quuxplusone: Updated. I should perhaps have pointed out before updating that you said "making q_yield and q_wake identical" but that doesn't seem right at all; their code is VERY different.

edenhill: Funcionality wise they are identical, so I don't see the need for a q_wake() or EXPLICIT_WAKE, that's already what q_yield() does.

Quuxplusone: Hmm, then I don't understand what `rd_kafka_q_yield()` does. The intent behind this PR #3114 is that we have a caller roughly like this:

```
class Producer {
    std::thread t_;
    std::unique_ptr<rd_kafka_t, RdKafkaDeleter> rk_;
    std::atomic<bool> done_;
public:
    explicit Producer() : done_(false) {
        rk_ = ...snip...;
        t_ = std::thread([]() {
            while (!done_) {
                rd_kafka_poll(rk_.get(), -1);
            }
        });
    }
    void shutdown() {
        if (!done_.exchange(true)) {
            rd_kafka_poll_wake();  // Force rd_kafka_poll() to return to its caller
            t_.join();
        }
    }
    ~Producer() { shutdown(); }
};
```
My impression from looking at your code was that `rd_kafka_q_yield()` doesn't actually cause `rd_kafka_poll()` to return to its caller unless there's work to do; that's why we had to introduce the new `RD_KAFKA_Q_F_EXPLICIT_WAKE` flag, to say "If you woke up and there's no work to do, _still_ I need you to treat this particular wakeup as non-spurious and return to your caller, because the program explicitly requested this wakeup."

- What should we do in `Producer::shutdown()` instead of `rd_kafka_poll_wake()`? what one-liner should go there instead?

- Or is `class Producer` itself wrong-headed? Should it be _completely_ differently architected? and if so, how?

edenhill: Have you tried using q_yield()?

Quuxplusone: (Notice that I had at least one typo above: there should be an argument to `rd_kafka_poll_wake(...)`!)

Are you suggesting that the `class Producer` above would work fine if I replace the line `rd_kafka_poll_wake();` with `rd_kafka_q_yield(rk_.get());`? Is this how rdkafka is intended to be used?

abbccdda:
Any update on this thread?

Quuxplusone:
I have come to the conclusion that `rd_kafka_yield(rk);` does exactly what I wanted, and therefore I've started using that, and no longer think this "poll_wake" idea is necessary. I'm going to abandon this PR now; please feel free to reopen if you think it should still be open.

(Also note that I ended up using `rd_kafka_yield`, not `rd_kafka_q_yield`. https://github.com/edenhill/librdkafka/discussions/4021 is related.)

Quuxplusone:
We don't use this patch anymore. Now, for waking up a producer from `rd_kafka_poll`, we use:
```
void rd_kafka_poll_wake(rd_kafka_t *rk) {
    // Interrupt all `rd_kafka_poll` (in any other threads) that are
    // currently blocked on `rk`.
    // See https://github.com/confluentinc/librdkafka/issues/614
    // N.B. this uses `rd_kafka_queue_yield`, which is NOT the same thing
    // as `rd_kafka_yield(rk)`. The latter is a no-op unless
    // it has been called from inside a Kafka callback, and using it here
    // would be both insufficient and theoretically-UB.
    //
    auto q = Unique<rd_kafka_queue_t>(rd_kafka_queue_get_main(rk));
    rd_kafka_queue_yield(q.get());
}
```
(where `Unique<rd_kafka_queue_t>` is a `unique_ptr` that calls `rd_kafka_queue_destroy` in its destructor).
For waking up a group consumer from `rd_kafka_consumer_poll` we just call `rd_kafka_consumer_close(rk)`. For waking up a non-group consumer from `rd_kafka_consume_queue` we do nothing, just wait for its timeout argument to expire.

PR #3116, opened 2020-10-22
stdarg.h included for va_start macro use

amichelotti:
stdarg.h must included explicitly for va_start and va_end use in old compilers.

Resolve issue:
 [https://github.com/edenhill/librdkafka/issues/3115](url)


edenhill:
Oops, thank you!

PR #3124, opened 2020-10-27
Travis s390x support

shahidhs-ibm:
Travis CI [officially](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) supports s390x builds, adding support for same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,7 +86,16 @@ matrix:
     - sudo apt install -y gdb
     - ./configure --enable-devel --disable-lz4-ext --prefix="$PWD/dest"
     - ./packaging/tools/rdutcoverage.sh
-
+ - name: "Linux GCC s390x: +devel"
+   os: linux
+   arch: s390x   
+   dist: bionic
+   compiler: gcc
+   env: NO_ARTIFACTS=y
+   before_script:
+    - sudo apt update
+    - sudo apt install -y gdb
+    - ./configure --enable-devel --disable-lz4-ext --prefix="$PWD/dest"
 install:
  - ccache -s || echo "CCache is not available."
  - rm -rf artifacts dest


.travis.yml
```diff
@@ -25,6 +25,14 @@ matrix:
    env: ADDITIONAL_BUILDS="debian" LINKAGE=std
    before_script:
     - ./configure --CFLAGS="-std=c99" --CXXFLAGS="-std=c++98" --install-deps --disable-lz4-ext --disable-regex-ext --prefix="$PWD/dest"
+ - name: "Linux GCC s390x: +Debian packages +BuiltinRegex +Strict"
```
edenhill, line 28: Do we really need all of these? I suggest just adding one single s390 worker with no additional builds

shahidhs-ibm: @edenhill We simply just followed the way existing jobs are running on amd64. Could you please suggest which jobs are mandatory and we should keep and which ones are optional and OK to discard?

edenhill: Just copy the last job, "Linux GCC: +devel .." and remove the "+code-cov" and "rdutcoverage.sh" calls.

shahidhs-ibm: @edenhill Updated the `.travis.yml` file as per review comment. Please check.

.travis.yml
```diff
@@ -101,10 +110,10 @@ script:
 - if [[ $SKIP_MAKE != y && $RUN_INTEGRATION_TESTS != y ]]; then make -C tests run_local_quick ; fi
 - if [[ $SKIP_MAKE != y ]]; then make install ; fi
 - if [[ -z $NO_ARTIFACTS ]]; then (cd dest && tar cvzf ../artifacts/librdkafka-${CC}.tar.gz .) ; fi
-- for distro in $ADDITIONAL_BUILDS ; do packaging/tools/distro-build.sh $distro || exit 1 ; done
+- if [[ "$TRAVIS_ARCH" != "s390x" ]]; then for distro in $ADDITIONAL_BUILDS ; do packaging/tools/distro-build.sh $distro || exit 1 ; done; fi
```
edenhill, line 113: Don't believe this is needed since ADDITIONAL_BUILDS is not defined for the s390 worker

.travis.yml
```diff
@@ -101,10 +110,10 @@ script:
...
 - if [[ $COPYRIGHT_CHECK == y ]]; then make copyright-check ; fi
 - if [[ $DOC_CHECK == y ]]; then make docs ; fi
-- if [[ $RUN_INTEGRATION_TESTS == y ]]; then (cd tests && travis_retry ./interactive_broker_version.py -c "make quick" 2.4.0) ; fi
+- if [[ "$TRAVIS_ARCH" != "s390x" ]]; then if [[ $RUN_INTEGRATION_TESTS == y ]]; then (cd tests && travis_retry ./interactive_broker_version.py -c "make quick" 2.4.0) ; fi; fi
```
edenhill, line 116: dito

edenhill (changes requested):

.travis.yml
shahidhs-ibm: @edenhill Incorporated the review comments. Please check.

.travis.yml
shahidhs-ibm: @edenhill Incorporated the review comments. Please check.

.travis.yml
shahidhs-ibm: @edenhill Incorporated the review comments. Please check.

edenhill (approved):
LGTM

edenhill:
Thank you!

I changed the commit message as not to imply that s390 is supported, which it is not.

PR #3126, opened 2020-10-29
OpenSSL 1.1.1h and some other stuff

edenhill:


src/regexp.c
```diff
@@ -848,7 +848,6 @@ Reprog *re_regcomp(const char *pattern, int cflags, const char **errorp)
         g->prog = prog;
 	g->pstart = g->pend = rd_malloc(sizeof (Renode) * strlen(pattern) * 2);
 
-        memcpy(((char *)prog)+1, prog, 4);
```
mhowlett, line 851: i assume you know what you're doing.

mhowlett (approved):

src/regexp.c
edenhill: More importantly I have no idea what I was doing adding that line!

PR #3130, opened 2020-11-02
Build static librdkafka_windows.a with travis

neptoess:
Add to the MinGW travis worker to have it build a librdkafka static bundle, per https://github.com/confluentinc/confluent-kafka-go/pull/555#discussion_r515795421


neptoess:
@edenhill 
This should be ready to go. Let me know if it needs anything else.

neptoess:
Well, looks like one of the tests seg faulted in the latest Travis run, but the previous commit (1a589f6) ran that same test without issue, and all the latest commit changed was `packaging.py`

CMakeLists.txt
```diff
@@ -5,13 +5,14 @@ parseversion("src/rdkafka.h")
 
 project(RdKafka VERSION ${RDKAFKA_VERSION})
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/packaging/cmake/Modules/")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
```
edenhill, line 8: Avoid whitespace/indent changes unless necessary.

packaging/mingw-w64/configure-build-msys2-mingw.sh
```diff
@@ -10,15 +10,48 @@ cmake \
     -D RDKAFKA_BUILD_TESTS=ON \
     -D WITH_LIBDL=OFF \
     -D WITH_PLUGINS=OFF \
-    -D WITH_SASL=ON \
```
edenhill, line 13: Suggest creating a new file, configure-build-msys2-mingw-static.sh, since this file is supposed to build a dynamic lib.

packaging/mingw-w64/travis-before-install.sh
```diff
@@ -20,9 +20,16 @@ case $TRAVIS_OS_NAME in
         # to zstd instead of xz compression
         $msys2 pacman -Sy --noconfirm pacman
         choco upgrade --no-progress -y msys2
-        $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake mingw-w64-x86_64-openssl mingw-w64-x86_64-cyrus-sasl
 
         ## Install more MSYS2 packages from https://packages.msys2.org/base here
+        $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake mingw-w64-x86_64-openssl mingw-w64-x86_64-lz4 mingw-w64-x86_64-zstd
```
edenhill, line 25: add what you need (zstd?) but dont remove sasl

packaging/mingw-w64/configure-build-msys2-mingw.sh
```diff
@@ -10,15 +10,48 @@ cmake \
...
 
+# Bundle all the static dependencies with the static lib we just built
+mkdir mergescratch
+pushd mergescratch
+cp /C/tools/msys64/mingw64/lib/libzstd.a ./
+cp /C/tools/msys64/mingw64/lib/libcrypto.a ./
+cp /C/tools/msys64/mingw64/lib/liblz4.a ./
+cp /C/tools/msys64/mingw64/lib/libssl.a ./
+cp ../src/librdkafka.a ./
+cp ../src-cpp/librdkafka++.a ./librdkafkapp.a
```
edenhill, line 30: librdkafkacpp.a, but why the name change?

packaging/mingw-w64/configure-build-msys2-mingw.sh
```diff
@@ -10,15 +10,48 @@ cmake \
...
+ar -M << EOF
+create librdkafka_merge.a
```
edenhill, line 32: We call these bundled archives librdkafka-static.a in the *nix build, suggesting doing the same here.

edenhill (changes requested):

packaging/mingw-w64/configure-build-msys2-mingw.sh
neptoess: Done

CMakeLists.txt
neptoess: There was inconsistency (two spaces, four spaces, tab characters) between CMakeLists.txt files, and even within the same file in the case of `src/CMakeLists.txt`, so I ran a pretty printer, `cmake-format` on them. I went back and removed any changes that weren't related to indentation though (it wanted to enforce 80 column limits for example)

packaging/mingw-w64/travis-before-install.sh
neptoess: Cyrus SASL isn't used on Windows (https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_sasl.c#L363-L367), so there's no benefit in installing it, right?

packaging/mingw-w64/configure-build-msys2-mingw.sh
neptoess: `ar` can't work with files with `+` in their name. I added a comment in the script to call this out.

CMakeLists.txt
edenhill: That's okay, but let's do reformatting in a separate PR

packaging/mingw-w64/travis-before-install.sh
edenhill: Good point :+1:

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ endif()
 include(CheckLibraryExists)
 check_library_exists(m pow "" WITH_HDRHISTOGRAM)
 if(WITH_HDRHISTOGRAM)
-    list(APPEND BUILT_WITH "HDRHISTOGRAM")
```
edenhill, line 39: Still whitespace changes.

packaging/mingw-w64/configure-build-msys2-mingw-static.sh
```diff
@@ -0,0 +1,59 @@
...
+set -e
+
+cmake \
+    -G "MinGW Makefiles" \
+    -D CMAKE_INSTALL_PREFIX="$PWD/dest/" \
+    -D WITHOUT_WIN32_CONFIG=ON  \
+    -D RDKAFKA_BUILD_EXAMPLES=ON \
+    -D RDKAFKA_BUILD_TESTS=ON \
+    -D WITH_LIBDL=OFF \
+    -D WITH_PLUGINS=OFF \
```
edenhill, line 12: Why is this OFF? Plugins are supported on Windows

packaging/mingw-w64/configure-build-msys2-mingw-static.sh
```diff
@@ -0,0 +1,59 @@
...
+    -D WITH_SASL=ON \
+    -D WITH_SSL=ON \
+    -D WITH_ZLIB=OFF \
```
edenhill, line 15: Is there a reason not to enable zlib?

src-cpp/CMakeLists.txt
```diff
@@ -1,19 +1,19 @@
...
-    KafkaConsumerImpl.cpp
-    MessageImpl.cpp
-    MetadataImpl.cpp
-    ProducerImpl.cpp
-    QueueImpl.cpp
-    RdKafka.cpp
-    TopicImpl.cpp
-    TopicPartitionImpl.cpp
+  sources
+  ConfImpl.cpp
```
edenhill, line 5: whitespace..

edenhill (changes requested):

CMakeLists.txt
neptoess: Good point. I'll roll all this back. Makes the diff for this PR easier to read

packaging/mingw-w64/configure-build-msys2-mingw-static.sh
edenhill: We don't need LIBDL on windows, plugins are loaded with LoadLibrary.
So keep WITH_LIBDL=OFF but set WITH_PLUGINS=on

CMakeLists.txt
neptoess: I think I finally got all this

packaging/mingw-w64/configure-build-msys2-mingw-static.sh
neptoess: I was using the existing script as a starting point. Looking through the CMakeLists.txt file though, I noticed that I like the default for _most_ of the options that were being manually specified, so I just removed them from the script. Now building WITH_PLUGINS and WITH_ZLIB

packaging/mingw-w64/configure-build-msys2-mingw-static.sh
neptoess: Updated CMakeLists.txt to default WITH_PLUGINS ON on Windows. Also, there was a typo in src/CMakeLists.txt that caused WITH_PLUGINS to not work properly on Windows builds.

packaging/mingw-w64/travis-before-install.sh
```diff
@@ -20,9 +20,17 @@ case $TRAVIS_OS_NAME in
         # to zstd instead of xz compression
         $msys2 pacman -Sy --noconfirm pacman
         choco upgrade --no-progress -y msys2
-        $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake mingw-w64-x86_64-openssl mingw-w64-x86_64-cyrus-sasl
 
         ## Install more MSYS2 packages from https://packages.msys2.org/base here
+        $msys2 pacman --sync --noconfirm --needed mingw-w64-x86_64-gcc mingw-w64-x86_64-make mingw-w64-x86_64-cmake mingw-w64-x86_64-openssl mingw-w64-x86_64-lz4 mingw-w64-x86_64-zstd
+
+        # Removing the DLLs so cmake links against the static libs
```
edenhill, line 27: This is a bit destructive, not that it matters on Travis, but if someone was to run this locally it would mess up their mingw installation.

Is there any way to instruct CMake to skip these files when linking?

src/CMakeLists.txt
```diff
@@ -152,9 +152,9 @@ if(WITHOUT_WIN32_CONFIG)
     list(APPEND rdkafka_compile_definitions ENABLE_DEVEL=0)
   endif(ENABLE_DEVEL)
   if(WITH_PLUGINS)
-    list(APPEND rdkafka_compile_definitions WITH_PLUGIN=1)
+    list(APPEND rdkafka_compile_definitions WITH_PLUGINS=1)
```
edenhill, line 155: :+1:

src-cpp/CMakeLists.txt
```diff
@@ -37,49 +37,40 @@ if(NOT RDKAFKA_BUILD_STATIC)
...
+  install(
+    FILES ${GENERATED_DIR}/rdkafka++.pc
+    DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig"
   )
+else()
+  set(PKG_CONFIG_NAME "librdkafka++-static")
+  set(PKG_CONFIG_DESCRIPTION "The Apache Kafka C/C++ library (static)")
+  set(PKG_CONFIG_REQUIRES "")
+  set(PKG_CONFIG_CFLAGS "-I\${includedir} -DLIBRDKAFKA_STATICLIB")
+  if(WIN32)
```
edenhill, line 62: This is dependent on MINGW, not WIN32, in other places, why not here?

packaging/mingw-w64/travis-before-install.sh
neptoess: I didn't realize this was a thing, but after looking here https://github.com/Kitware/CMake/blob/master/Modules/Platform/Windows-GNU.cmake#L28, I came up with https://github.com/neptoess/librdkafka/blob/add-mingw-to-packaging/CMakeLists.txt#L27-L30. Works great, and I don't need to manually keep track of what `.dll.a` files to delete. Everyone wins

src-cpp/CMakeLists.txt
neptoess: The check in `src/CMakeLists.txt` also uses `WIN32` instead of `MINGW` (see https://github.com/neptoess/librdkafka/blob/add-mingw-to-packaging/src/CMakeLists.txt#L312-L319).

I can change these, but I'm not sure we're risking too much either way. `pkg-config` _does_ work for compilers other than MinGW, e.g. MSVC, clang, and these dependencies will still apply.

neptoess: You did make me realize there was a bug for static builds on non-Windows platforms, so I fixed that

neptoess:
@edenhill Is there anything you need from me to help move this along?

neptoess:
@edenhill 
Am I correct in thinking that the cmake changes needing review is the reason this hasn't been merged yet?

@masariello @omartijn @benesch @ed-alertedh @SpaceIm
I would greatly appreciate if you could look over this PR and provide any feedback

masariello:
Do mingw compilers have different ABIs for debug and optmized code builds? If so we should set CMAKE_DEBUG_POSTFIX and make sure that all the packaging logic flows with it.

neptoess:
@masariello 
I will investigate the ABIs of debug vs optimized builds.

As for having the logic in packaging follow, are you saying we should distribute both release and debug binaries of librdkafka?

masariello:
Where ABIs are different users do need both debug and optimized builds. So in that case they need to be distributed together, yes

neptoess:
@masariello 
I can find little to no good information regarding this (ABI compatibility issues between object files built with/without optimization) with regard to mingw-w64 gcc / g++

Is there some type of test we can do to verify this? I would imagine any issues could be detected during linking, right?

neptoess:
@masariello 
Upon further digging through the GCC docs, it looks like this ABI concern is only limited to C++ binaries. Since this PR is really just focused on getting a MinGW built static C library (to distribute with confluent-kafka-go), and I don't think @edenhill has any near term plans to drop vcpkg or NuGet distribution of Windows binaries for librdkafka, I think it's safe to say it shouldn't block this PR from being merged.

Agree?

masariello:
> @masariello
> Upon further digging through the GCC docs, it looks like this ABI concern is only limited to C++ binaries. Since this PR is really just focused on getting a MinGW built static C library (to distribute with confluent-kafka-go), and I don't think @edenhill has any near term plans to drop vcpkg or NuGet distribution of Windows binaries for librdkafka, I think it's safe to say it shouldn't block this PR from being merged.
> 
> Agree?

Yep. Agreed

neptoess:
@edenhill ,
I do hate to keep bothering you with this PR, but is there anything I can do to help move this along? There was another comment in https://github.com/confluentinc/confluent-kafka-go/issues/128 from someone struggling to use confluent-kafka-go on Windows, so I think there's still merit in getting this merged.

CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ list(APPEND BUILT_WITH "${CMAKE_C_COMPILER_ID}")
 list(APPEND BUILT_WITH "${CMAKE_CXX_COMPILER_ID}")
 # }
 
+if (MINGW AND RDKAFKA_BUILD_STATIC)
+  # ".dll.a" takes precedence over ".a" by default
```
edenhill, line 28: Aren't static libs using a `.lib` extension on Windows?

CMakeLists.txt
```diff
@@ -79,7 +84,7 @@ endif()
 # }
 
 # WITH_PLUGINS {
-if(WITH_LIBDL)
+if(WITH_LIBDL OR WIN32)
```
edenhill, line 87: Do we really want to use libdl and not LoadLibrary on cygwin win32?

src/CMakeLists.txt
```diff
@@ -253,80 +247,80 @@ if(WITH_LZ4_EXT)
...
+if (WIN32)
+  set(PKG_CONFIG_LIBS_PRIVATE "-lws2_32 -lsecur32 -lcrypt32")
+else()
+  set(PKG_CONFIG_LIBS_PRIVATE "-lpthread")
+  find_library(RT_LIBRARY rt)
+  if(RT_LIBRARY)
+    string(APPEND PKG_CONFIG_LIBS_PRIVATE " -lrt")
+  endif()
+
+  if(WITH_PLUGINS)
```
edenhill, line 268: WITH_PLUGINS is satisfied by Win32 (LoadLibrary) or libdl, so this should probably be WITH_LIBDL (but use LoadLibrary, not libdl on win32)

edenhill (changes requested):
Please also rebase on latest master

edenhill:
Oh, please also make sure the libraries are stripped (to reduce size).

CMakeLists.txt
neptoess: This is how MSYS2 distributes their MinGW-w64 static libraries. It looks like this is for historical reasons, as `.a` is what `ld` is looking for and `.lib` is what `LINK.exe` is looking for, but, at some point, someone added `.lib` support to the MinGW `ld`

edenhill: If the Go toolchain supports the .lib extension I think we should go with that.

CMakeLists.txt
neptoess: This line just turns the `WITH_PLUGINS` CMake flag on. All this does is add `rdkafka_plugin.c` to the build (see https://github.com/neptoess/librdkafka/blob/add-mingw-to-packaging/src/CMakeLists.txt#L82-L84). `rdkafka_plugin.c` relies on `rddl.c` for its functionality, and it looks like `rddl.c` properly detects `_WIN32` and handles plugin loading via `LoadLibrary`. If I'm misunderstanding this, let me know.

CMakeLists.txt
neptoess: Well, `CMAKE_FIND_LIBRARY_SUFFIXES` is used by CMake to find 3rd party dependencies, e.g. libssl. These are distributed by MSYS2 as `.a`, so this would have to stay.

As for confluent-kafka-go, we can change the bundled `librdkafka_windows.a` to `librdkafka_windows.lib`. This would be a bit of an oddity though, as `.lib` is only commonly used for MSVC static libraries, i.e. those built with `cl.exe`. If someone tries to use `librdkafka_windows.lib` in an MSVC project, or with `LINK.exe`, it will fail to link. This is probably why the convention is continue using `.a` for static libraries built with MinGW `gcc`. (While MinGW `ld` handles both the GNU `.a` format and the Microsoft `.lib` format, `LINK.exe` still only supports the Microsoft `.lib` format)

edenhill: You had me at oddity, let's go with `.a`.

neptoess:
@edenhill 
> Oh, please also make sure the libraries are stripped (to reduce size).

Not 100% positive on how to verify that, but I don't see any obvious debug symbols in the final static library.
```
$ objdump --syms librdkafka_windows.a | grep debug
[249](sec  1)(fl 0x00)(ty  20)(scl   2) (nx 0) 0x00000000000082d0 SSL_set_debug
[  9](sec  1)(fl 0x00)(ty  20)(scl   2) (nx 0) 0x0000000000000190 CRYPTO_set_mem_debug
[  0](sec  1)(fl 0x00)(ty  20)(scl   2) (nx 1) 0x0000000000000000 BIO_debug_callback
debug.c.obj:     file format pe-x86-64
[  8](sec  3)(fl 0x00)(ty   0)(scl   2) (nx 0) 0x0000000000000000 g_debuglevel
[ 16](sec  1)(fl 0x00)(ty  20)(scl   2) (nx 0) 0x0000000000000540 rd_kafka_event_debug_contexts
[155](sec  1)(fl 0x00)(ty  20)(scl   2) (nx 0) 0x0000000000015af0 rd_kafka_get_debug_contexts
```

edenhill:
You could jab a:
```
ls -la *.a
strip -g *.a
ls -la *a
```

in the build script (after ranlib) and see if there's any difference

neptoess:
> You could jab a:
> 
> ```
> ls -la *.a
> strip -g *.a
> ls -la *a
> ```
> 
> in the build script (after ranlib) and see if there's any difference

No difference. I guess that makes sense, since we're not compiling with debug symbols enabled. Should I throw the strip command in the build script anyway?

edenhill:
Yeah. Please do, can't hurt.

neptoess:
`strip -g` calls added

edenhill:
Great! Is this ready to be merged?

CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ list(APPEND BUILT_WITH "${CMAKE_C_COMPILER_ID}")
 list(APPEND BUILT_WITH "${CMAKE_CXX_COMPILER_ID}")
 # }
 
+if (MINGW AND RDKAFKA_BUILD_STATIC)
+  # ".dll.a" takes precedence over ".a" by default
+  set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+endif()
```
SpaceIm, line 30: Is it really useful? I would say that changing default behavior should be avoided to follow rule of least surprise.

src/CMakeLists.txt
```diff
@@ -257,80 +251,80 @@ if(WITH_LZ4_EXT)
...
 
-# Set up path to these sources for other sub-projects (tests, examples)
-# to be able to reach them.
-#set(rdkafka_SRC_DIR ${PROJECT_SOURCE_DIR}
-#    CACHE INTERNAL "${PROJECT_NAME} source dir" FORCE)
+if(MINGW)
+  message(STATUS "Detected MinGW build")
+  # MinGW ignores `#pragma comment(lib, ...)` so we need to specify link libs
+  target_link_libraries(rdkafka PUBLIC crypt32 ws2_32 secur32)
+endif(MINGW)
```
SpaceIm, line 258: I would just link crypt32 ws2_32 secur32 if WIN32, to make CMakeLists neat. By the way I think that crypt32 is not needed if WITH_SSL is OFF.

CMakeLists.txt
neptoess: It's required to make sure we link to the static libraries (`.a`) and not the dynamic libraries (`.dll.a`). the only other way to ensure this is to delete the `.dll.a` files from the `mingw64/lib` directory, which is slightly less than ideal.

SpaceIm: It's an environment consideration, CMakeLists shouldn't have this logic. It's the responsability of consumers to provide either static or shared libs. Build files of a library shouldn't force to link static or shared dependencies.

When librdkafka is static, nothing is linked to the static lib anyway, and there is no reason to force all static just because librdkafka is static.

edenhill: Actually for this librdkafka-static build we do want all dependencies to be available as static libraries so we can provide a self-contained library.

SpaceIm: Is there a reason for that?

SpaceIm: I mean in conan recipe of librdkafka, default is all static, but consumers can set any combination of shared/static for all the dependency graph, and I didn't notice any issue.

neptoess: Ah. Maybe some background would help. The primary purpose of this PR (MinGW static build) is to have Travis build a static library that can be bundled with confluent-kafka-go, so users of that library are not required to have librdkafka and/or its dependencies previously installed on their system. A C/C++ user of librdkafka would _very_ likely be using the NuGet librdkafka package or the vcpkg one. Both of those use dynamic linking, and the MinGW dynamically linked build also works well (thanks @edbordin).

SpaceIm: And if you pass `CMAKE_FIND_LIBRARY_SUFFIXES` value to cmake command line in CI files, instead of hardcoded value in CMakeLists?

packaging/mingw-w64/configure-build-msys2-mingw-static.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -e
+
+cmake \
+    -G "MinGW Makefiles" \
+	-D CMAKE_FIND_LIBRARY_SUFFIXES=".a"
```
SpaceIm, line 7:
```suggestion
    -D CMAKE_FIND_LIBRARY_SUFFIXES=".a" \
```

neptoess:
@edenhill 
Ready to merge

edenhill (approved):

edenhill:
Superb! Great work (and perseverence) on this!

edenhill:
Hey @neptoess , can you help me figure out why the Travis-CI mingw jobs are silently failing?

I've added some debugging and it seems like the test-runner.exe fails to execute, but I don't see an error anywhere (stderr and stdout seems to be weirdly mixed up).
Can you try to reproduce this issue locally?
https://app.travis-ci.com/github/edenhill/librdkafka/jobs/546274490#L1046

edbordin:
I had issues with GCC LTO being broken on MinGW/MSYS at some point in the past year - the resulting executables would just segfault. Is there any chance this is being built with `-flto` or related flags?

edenhill:
Thanks @edbordin, there doesn't seem to be an `-flto` flag in the make output though:
https://app.travis-ci.com/github/edenhill/librdkafka/jobs/546447298

neptoess:
Looking into this. I'm thinking this has to do with MSYS2 updating the MinGW compiler to GCC 11
https://packages.msys2.org/package/mingw-w64-x86_64-gcc

This happened 2021/10/21. When I did `pacman -Syu` on my existing MSYS2 install, it did not upgrade beyond GCC 10.3. Uninstalling and reinstalling MSYS2, and copying the `pacman` commands from the Travis script left me with 11.2. Will do some poking around, but I'm guessing there are some libraries being linked into our build that were built with GCC 10.

neptoess:
I did a fresh MSYS2 install, fresh git pull, and built the `tstwinb` branch locally. All tests passed. Going to focus on Travis now, since I think the code (both library and build scripts) is fine.
[test-run.log](https://github.com/edenhill/librdkafka/files/7467270/test-run.log)

edenhill:
Thanks Bill!
Going forward, is it possible to version pin the toolchain and dependencies on the mingw builder to avoid this in the future?

neptoess:
Well, because my local just worked (with a completely fresh MSYS2 install), I'm not confident this is a package version issue. I was going to try https://github.com/neptoess/librdkafka/commit/3ddeb537bdb83fd53e84afc3acd37adff7e76583 (I should have added a --no-confirm so it doesn't hang) to see if pacman reports any packages needing updated _after_ the `choco upgrade` command in the build script, but, due to the Travis pricing changes, I can't do any Travis builds at the moment (I have to email them for the usage based plan).

Would it be possible to manually trigger a build on a previous commit that worked?

edenhill:
Yeah, Travis costs are a sudden and steep barrier.
I think you can use my credits though if you open a PR and comment out all the other workers from .travis.yaml so you just have the single mingw one.

neptoess:
Good idea. Watching it now https://app.travis-ci.com/github/edenhill/librdkafka/builds/241104613

neptoess:
https://app.travis-ci.com/github/edenhill/librdkafka/builds/241106018
Even rebuilding v1.8.2 failed. Going to be tricky to debug, since a fresh install of MSYS2 on Win10 works fine.

edenhill:
These are the mingw toolchain version diffs between the proper 1.8.2  build and your build:
```patch
--- 543777511s  2021-11-03 19:25:32.823602012 +0100
+++ 241106018s  2021-11-03 19:24:25.243884678 +0100
@@ -3,15 +3,15 @@
  mingw-w64-x86_64-bzip2-1.0.8-2-any downloading...
  mingw-w64-x86_64-ca-certificates-20210119-1-any downloading...
  mingw-w64-x86_64-c-ares-1.17.2-1-any downloading...
- mingw-w64-x86_64-cmake-3.21.3-1-any downloading...
- mingw-w64-x86_64-crt-git-9.0.0.6316.acdc7adc9-1-any downloading...
+ mingw-w64-x86_64-cmake-3.21.4-1-any downloading...
+ mingw-w64-x86_64-crt-git-9.0.0.6327.f29c1101f-1-any downloading...
  mingw-w64-x86_64-curl-7.79.1-1-any downloading...
  mingw-w64-x86_64-expat-2.4.1-1-any downloading...
-mingw-w64-x86_64-gcc-10.3.0-8-any downloading...
- mingw-w64-x86_64-gcc-libs-10.3.0-8-any downloading...
+ mingw-w64-x86_64-gcc-11.2.0-1-any downloading...
+ mingw-w64-x86_64-gcc-libs-11.2.0-1-any downloading...
  mingw-w64-x86_64-gettext-0.19.8.1-10-any downloading...
  mingw-w64-x86_64-gmp-6.2.1-2-any downloading...
- mingw-w64-x86_64-headers-git-9.0.0.6316.acdc7adc9-1-any downloading...
+ mingw-w64-x86_64-headers-git-9.0.0.6327.f29c1101f-1-any downloading...
  mingw-w64-x86_64-isl-0.24-1-any downloading...
  mingw-w64-x86_64-jansson-2.14-1-any downloading...
  mingw-w64-x86_64-jemalloc-5.2.1-2-any downloading...
@@ -26,8 +26,8 @@
  mingw-w64-x86_64-libtasn1-4.17.0-1-any downloading...
  mingw-w64-x86_64-libtre-git-r128.6fb7206-2-any downloading...
  mingw-w64-x86_64-libunistring-0.9.10-4-any downloading...
- mingw-w64-x86_64-libuv-1.42.0-2-any downloading...
- mingw-w64-x86_64-libwinpthread-git-9.0.0.6316.acdc7adc9-1-any downloading...
+ mingw-w64-x86_64-libuv-1.42.0-3-any downloading...
+ mingw-w64-x86_64-libwinpthread-git-9.0.0.6327.f29c1101f-1-any downloading...
  mingw-w64-x86_64-libxml2-2.9.12-3-any downloading...
  mingw-w64-x86_64-lz4-1.9.3-1-any downloading...
  mingw-w64-x86_64-make-4.3-1-any downloading...
@@ -40,7 +40,7 @@
  mingw-w64-x86_64-pkgconf-1.8.0-2-any downloading...
  mingw-w64-x86_64-rhash-1.4.2-1-any downloading...
  mingw-w64-x86_64-windows-default-manifest-6.4-3-any downloading...
- mingw-w64-x86_64-winpthreads-git-9.0.0.6316.acdc7adc9-1-any downloading...
+ mingw-w64-x86_64-winpthreads-git-9.0.0.6327.f29c1101f-1-any downloading...
  mingw-w64-x86_64-xz-5.2.5-2-any downloading...
  mingw-w64-x86_64-zlib-1.2.11-9-any downloading...
  mingw-w64-x86_64-zstd-1.5.0-1-any downloading...

```

neptoess:
Hmm. These package versions all exactly match what I have locally. Is there any way to pull the `test-runner.exe` artifact out of Travis? I'd like to try running it locally, to rule out anything funny going on between Travis's bash shell and the new C runtime.

edenhill:
Not easily, no.

That shell invocation is a bit messed up output-wise, so maybe we're not seeing an error message for that reason.
Can we try to execute test-runner.exe in a later step and separate shell in travis?

neptoess:
Well, that's a little confusing. Breaking tests out into a separate step worked.
https://app.travis-ci.com/github/edenhill/librdkafka/jobs/546739494

edenhill:
That's good news!

Your new run-tests.sh script does not have the PATH update of the original script, maybe that's what messed up the original?
I have a vague memory that PATH also affects library loading on Windows, not sure if that is true for mingw though.

neptoess:
`%PATH%` does affect library loading on Windows, but I believe `%PATH%` is only checked if the DLL isn't found in the working directory first. This is why it's so common to see people modify, e.g. games, by making proxy DLLs and copying them to bin folder.

Also, looks like https://app.travis-ci.com/github/edenhill/librdkafka/builds/241213359 worked too, so we don't even need to use a different stage in Travis. I'm guessing just tearing down and making a new shell is the fix.

neptoess:
Opened #3607 to get the fix merged into master. Let me know if you want me to take a different approach

edenhill:
Thanks for so resolutely taking on this issue, @neptoess, much appreciated!

neptoess:
No problem. Happy to help

PR #3131, opened 2020-11-02
Add `ssl.ca.certificate.stores`

edenhill:


src/rdkafka_ssl.c
```diff
@@ -908,8 +988,20 @@ static int rd_kafka_ssl_set_certs (rd_kafka_t *rk, SSL_CTX *ctx,
...
+                r = rd_kafka_ssl_win_load_cert_stores(
+                        rk, ctx, rk->rk_conf.ssl.ca_cert_stores);
+                if (r == 0) {
+                        rd_kafka_log(rk, LOG_NOTICE, "CERTSTORE",
+                                "No CA certificates loaded from "
+                                "Windows certificate stores: "
+                                "falling back to default OpenSSL CA paths");
+                        r = -1;
+                } else if (r == -1)
+                        rd_kafka_log(rk, LOG_NOTICE, "CERTSTORE",
```
mhowlett, line 1001: should this be > LOG_NOTICE?

mhowlett (approved):

src/rdkafka_ssl.c
edenhill: Not sure if by > you mean more severe (lower log level), or less severe (higher log level), but I'm assuming the former.

The idea here is that it is possible that some stores might not be loadable, and that is only a problem if the connection to the cluster can't be made, in which case people will inspect their logs. But if the connection can be made, those load errors are insignificant and we dont want to trigger alerts.

PR #3135, opened 2020-11-05
main thread should keep polling while cgrp is alive (#3127)

### LINKED ISSUE #3127: Hang in KafkaConsumer shutdown when close() is not called

Description
===========
I've observed a new hang in the C++ `KafkaConsumer` in v1.5.2 when shutting down without calling `close()` explicitly. I believe this is related to the recent change to call `rd_kafka_destroy()` in the dtor. TSAN also occasionally detects a data race between `rd_kafka_toppar_op_serve` and `rd_kafka_fetch_reply_handle` on the `rktp_last_error` field of a toppar while shutting down this consumer.

How to reproduce
================
See https://gist.github.com/gridaphobe/a2f20726171eb957358bd2a91123015f for a small example, with debug logs, and the output of TSAN.

Checklist
=========

 - [x] librdkafka version (release number or git tag): v1.5.2 and 521f460e444e9bb9b8abdb9194fef0f126d37d06
 - [x] Apache Kafka version: mock broker
 - [x] librdkafka client configuration: https://gist.github.com/gridaphobe/a2f20726171eb957358bd2a91123015f#file-kafkaconsumer-unsubscribe-noclose-cpp-L93-L97
 - [x] Operating system: macOS 10.15 and ubuntu bionic
 - [ ] Critical issue

gridaphobe:
During termination, the main thread's loop continues as long
as there are ops left to poll, but is also responsible for
terminating the cgrp via `rd_kafka_cgrp_serve`. It is possible
for the loop to exit before the cgrp is terminated if you
unsubscribe and then immediately destroy the rdkafka handle,
which leaves some stray refcnts and hangs.

The solution is to keep the polling loop going as long as
there are remaining ops OR until the cgrp is terminated.

I also discovered a data race while investigating this issue.
TSAN should occasionally catch the race while running the
updated test 0116, but unfortunately it's pretty non-deterministic..

fixes #3127

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -1966,7 +1966,8 @@ static int rd_kafka_thread_main (void *arg) {
         mtx_unlock(&rk->rk_init_lock);
 
 	while (likely(!rd_kafka_terminating(rk) ||
-		      rd_kafka_q_len(rk->rk_ops))) {
+		      rd_kafka_q_len(rk->rk_ops) ||
+                      (rk->rk_cgrp && (rk->rk_cgrp->rkcg_state != RD_KAFKA_CGRP_STATE_TERM)))) {
                 rd_ts_t sleeptime = rd_kafka_timers_next(
                         &rk->rk_timers, 1000*1000/*1s*/, 1/*lock*/);
                 rd_kafka_q_serve(rk->rk_ops, (int)(sleeptime / 1000), 0,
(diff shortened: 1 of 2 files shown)


edenhill:
Eric, you're the best!

edenhill:
I think you can back out the rktp_last_error changes, I fixed that on master (those locations dont need to reset the last error at all).

edenhill:
Also, :100: for test-case!

gridaphobe:
> I think you can back out the rktp_last_error changes, I fixed that on master (those locations dont need to reset the last error at all).

Awesome, done!

edenhill (approved):
LGTM.

Great stuff!

PR #3151, opened 2020-11-17
Added a first fuzzer with purpose of integrating to OSS-Fuzz

DavidKorczynski:
Hi!

I have been doing some work on fuzzing Librdkafka and also integrating it into OSS-Fuzz (https://github.com/google/oss-fuzz). OSS-Fuzz is service from Google that performs continuous fuzzing of important open source projects. It's completely free and I would be happy to supply fuzzers to Librdkafka to help with the security of the project. By integrating into OSS-Fuzz you will receive bug reports, coverage reports and more from OSS-Fuzz, and all I need is an email(s) from you that will receive the reports. 

I started with a simple fuzzer for a portion of the regex code, but would be happy to extend with more fuzzers if you would like to integrate.

### DIFF
--- a/tests/fuzzers/README.md
+++ b/tests/fuzzers/README.md
@@ -1,5 +1,5 @@
 # Fuzzing
-Librdkafka supports fuzzing by way of Libfuzzer and OSS-Fuzz. This is ongoing work.
+librdkafka supports fuzzing by way of Libfuzzer and OSS-Fuzz. This is ongoing work.
 
 ## Launching the fuzzers
 The easiest way to launch the fuzzers are to go through OSS-Fuzz. The only prerequisite to this is having Docker installed.
@@ -9,8 +9,10 @@ With Docker installed, the following commands will build and run the fuzzers in 
 ```
 git clone https://github.com/google/oss-fuzz
 cd oss-fuzz
-sudo python3 infra/helper.py build_image librdkafka
-sudo python3 infra/helper.py build_fuzzers librdkafka
-sudo python3 infra/helper.py run_fuzzer librdkafka FUZZ_NAME
+python3 infra/helper.py build_image librdkafka
+python3 infra/helper.py build_fuzzers librdkafka
+python3 infra/helper.py run_fuzzer librdkafka FUZZ_NAME
 ```
 where FUZZ_NAME references the name of the fuzzer. Currently the only fuzzer we have is fuzz_regex
+
+Notice that the OSS-Fuzz `helper.py` script above will create a Docker image in which the code of librdkafka will be built. As such, depending on how you installed Docker, you may be asked to have root access (i.e. run with `sudo`).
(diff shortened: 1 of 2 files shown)


edenhill:
This is great! Could you add a README.md to tests/fuzzer that instructs how to run/trigger the fuzzer? E.g., point to OSS-Fuzzer github repo instructions or whatever is needed, and where to check for results?

For notifications, please use `rdkafka@edenhill.se`

DavidKorczynski:
Great! I added a README - once we merge this the commands in the README will be runnable. 

For reference, this is the PR in the OSS-Fuzz repository that contains the necessary logic for integrating librdkafka: https://github.com/google/oss-fuzz/pull/4652

tests/fuzzers/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Launching the fuzzers
+The easiest way to launch the fuzzers are to go through OSS-Fuzz. The only prerequisite to this is having Docker installed.
+
+With Docker installed, the following commands will build and run the fuzzers in this directory:
+
+```
+git clone https://github.com/google/oss-fuzz
+cd oss-fuzz
+sudo python3 infra/helper.py build_image librdkafka
```
edenhill, line 12: Why is sudo required?

tests/fuzzers/README.md
```diff
@@ -0,0 +1,16 @@
+# Fuzzing
+Librdkafka supports fuzzing by way of Libfuzzer and OSS-Fuzz. This is ongoing work.
```
edenhill, line 2: librdkafka, not capital L

tests/fuzzers/README.md
DavidKorczynski: I added explanation in the README now. The reason is that `helper.py` will create and run Docker images, and I think Docker is most often installed such that it requires sudo to create/run images. However, this is not strictly required e.g. https://github.com/sindresorhus/guides/blob/master/docker-without-sudo.md

edenhill: I'm not sure that's the norm, I run docker without sudo both on OSX and Ubuntu, so I suggest removing the sudo bits above.

DavidKorczynski: Fair and sounds good - removed!

edenhill:
Thank you!

edenhill:
I got a couple of fuzzer failure emails, but I'm not authorized to view the test results, how do I get added?

DavidKorczynski:
You should be able to access the errors from here: https://oss-fuzz.com/

Here https://github.com/google/oss-fuzz/blob/master/projects/librdkafka/project.yaml is the configuration file for the project. I added the email `rdkafka@edenhill.se` so you if log in to oss-fuzz.com with that email you are good to go. You may have to attach the `rdkafka@edenhill.se` email to a google account, but if you prefer to switch a `gmail` account you can tell me and I will update the configuration at https://github.com/google/oss-fuzz/blob/master/projects/librdkafka/project.yaml

edenhill:
Ah, so that's what that was for, then please switch it to `magnus@edenhill.se`, thanks!

DavidKorczynski:
Apologies for confusion! Fixed it up now, it should be merged soon. If you can't watch the bug reports due to adding the email after the bugs were found (I don't think this is the case but cannot remember entirely) then let me know and I will forward all of the existing reports to you.

PR #3156, opened 2020-11-20
Fix docs, change RD_KAFKA_CONF_RES_OK to RD_KAFKA_CONF_OK

jjhoughton:
Signed-off-by: Joshua Houghton <joshua.houghton@ripjar.com>

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -5249,10 +5249,10 @@ typedef rd_kafka_resp_err_t
  * @param errstr A human readable error string in case the interceptor fails.
  * @param errstr_size Maximum space (including \0) in \p errstr.
  *
- * @returns RD_KAFKA_CONF_RES_OK if the property was known and successfully
- *          handled by the interceptor, RD_KAFKA_CONF_RES_INVALID if the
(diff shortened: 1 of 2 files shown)


src/rdkafka.h
```diff
@@ -5249,7 +5249,7 @@ typedef rd_kafka_resp_err_t
  * @param errstr A human readable error string in case the interceptor fails.
  * @param errstr_size Maximum space (including \0) in \p errstr.
  *
- * @returns RD_KAFKA_CONF_RES_OK if the property was known and successfully
+ * @returns RD_KAFKA_CONF_OK if the property was known and successfully
  *          handled by the interceptor, RD_KAFKA_CONF_RES_INVALID if the
```
edenhill, line 5253: can you fix these two as well? RES_INVALID and RES_UNKNOWN?

edenhill (changes requested):

src/rdkafka.h
jjhoughton: Sure, this is just something i noticed while reading through the docs

edenhill:
Thank you!

PR #3160, opened 2020-11-24
RD_KAFKA_RESP_ERR__PARTITION_EOF is not a broker error

mhowlett:
i guess you've reasoned that what is here already is most correct, but i think there's a lot of value in differentiating whether an error is local or broker strictly on whether the error was generated on the broker or not. PR just a suggestion.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -396,7 +396,7 @@ static const struct rd_kafka_err_desc rd_kafka_err_descs[] = {
 	_ERR_DESC(RD_KAFKA_RESP_ERR__MSG_TIMED_OUT,
 		  "Local: Message timed out"),
 	_ERR_DESC(RD_KAFKA_RESP_ERR__PARTITION_EOF,
-		  "Broker: No more messages"),
+		  "Local: No more messages"),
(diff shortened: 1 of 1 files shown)


src/rdkafka.c
```diff
@@ -396,7 +396,7 @@ static const struct rd_kafka_err_desc rd_kafka_err_descs[] = {
 	_ERR_DESC(RD_KAFKA_RESP_ERR__MSG_TIMED_OUT,
 		  "Local: Message timed out"),
 	_ERR_DESC(RD_KAFKA_RESP_ERR__PARTITION_EOF,
-		  "Broker: No more messages"),
+		  "Local: No more messages"),
```
edenhill, line 399: Yeah, the reason this one differs is to help the user understand that there are no more messages on the broker, and not in any local queue.
So while the error string is technically incorrect I don't fixing it makes it easier on the user.

rolandyoung: Could it say `"Local: Broker has no more messages"` or `"Local: No more messages at broker"`?

edenhill: Good point :+1:

PR #3163, opened 2020-11-30
Incremental rebalancing fixes

mhowlett:
- partitions revoked events should never be emitted in the incremental rebalancing case when the assignment is empty.
- in the n_wildcard test it is valid for either one rebalance or two rebalances to be required for the consumers subscription to be updated.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -3886,9 +3886,8 @@ static void rd_kafka_cgrp_revoke_all_rejoin (rd_kafka_cgrp_t *rkcg,
                            "current assignment and rebalance");
         }
 
-
-        if (rkcg->rkcg_group_assignment) {
-
+        if (rkcg->rkcg_group_assignment &&
+            rkcg->rkcg_group_assignment->cnt > 0) {
                 if (assignment_lost)
                         rd_kafka_cgrp_assignment_set_lost(
                                 rkcg,
--- a/tests/0113-cooperative_rebalance.cpp
+++ b/tests/0113-cooperative_rebalance.cpp
@@ -2094,7 +2094,7 @@ static void poll_all_consumers (RdKafka::KafkaConsumer **consumers,
  * TODO: incorporate committing offsets.
  */
 
-static void u_complex (bool use_rebalance_cb, int subscription_variation) {
+static void u_multiple_subscription_changes (bool use_rebalance_cb, int subscription_variation) {
   const int N_CONSUMERS = 8;
   const int N_TOPICS = 2;
   const int N_PARTS_PER_TOPIC = N_CONSUMERS * N_TOPICS;
@@ -2792,8 +2792,8 @@ extern "C" {
       t_max_poll_interval_exceeded(i);
     /* Run all 2*3 variations of the u_.. test */
     for (i = 0 ; i < 3 ; i++) {
-      u_complex(true/*with rebalance_cb*/, i);
-      u_complex(false/*without rebalance_cb*/, i);
+      u_multiple_subscription_changes(true/*with rebalance_cb*/, i);
+      u_multiple_subscription_changes(false/*without rebalance_cb*/, i);
     }
 
     return 0;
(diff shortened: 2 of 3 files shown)


src/rdkafka_cgrp.c
```diff
@@ -3886,9 +3886,8 @@ static void rd_kafka_cgrp_revoke_all_rejoin (rd_kafka_cgrp_t *rkcg,
                            "current assignment and rebalance");
         }
 
-
-        if (rkcg->rkcg_group_assignment) {
-
+        if (rkcg->rkcg_group_assignment &&
```
edenhill, line 3889: So if a consumer joins a group and gets an empty assignment, it will never get a revoke rebalance callback?
I'm wondering if this change in signalling poses a problem for applications that needs to track whether it has an assignment or not, whether the group rebalances or not.

I think it'd be good to update the rdkafka.h docs to make it clear that there will be no empty revoke callbacks (even if there were empty assignments) in the cooperative case.

tests/0113-cooperative_rebalance.cpp
```diff
@@ -1711,25 +1709,45 @@ static void n_wildcard () {
...
+      if (rebalance_cb1.nonempty_assign_call_cnt == 1) {
+        /* just one rebalance was required */
+        TEST_ASSERT(rebalance_cb1.nonempty_assign_call_cnt == 1,
+                    "Expecting C_1's nonempty_assign_call_cnt to be 1 not %d ",
+                    rebalance_cb1.nonempty_assign_call_cnt);
+        TEST_ASSERT(rebalance_cb2.nonempty_assign_call_cnt == 1,
+                    "Expecting C_2's nonempty_assign_call_cnt to be 1 not %d ",
+                    rebalance_cb2.nonempty_assign_call_cnt);
+      } else {
+        /* two rebalances were required (occurs infrequently) */
```
edenhill, line 1736: Why not just TEST_ASSERT( call_cnt == 1 .. or 2) ?

tests/0113-cooperative_rebalance.cpp
```diff
@@ -2073,11 +2091,10 @@ static void poll_all_consumers (RdKafka::KafkaConsumer **consumers,
  *
  * @param subscription_variation 0..2
  *
- * FIXME: What's the stressy part?
  * TODO: incorporate committing offsets.
  */
 
-static void u_stress (bool use_rebalance_cb, int subscription_variation) {
+static void u_complex (bool use_rebalance_cb, int subscription_variation) {
```
edenhill, line 2097: Still a subjective value word that doesn't really reflect what is going on ;)
u_multiple_subscription_changes()

edenhill (changes requested):
Good stuff!

src/rdkafka_cgrp.c
mhowlett: > So if a consumer joins a group and gets an empty assignment, it will never get a revoke rebalance callback?

correct. this is outlined in the kip, and mirrored in the java docs (https://kafka.apache.org/26/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html):

> With cooperative rebalancing, onPartitionsRevoked(Collection) and onPartitionsLost(Collection) will only be triggered when there are non-empty partitions revoked or lost from this consumer member during a rebalance event.

on a related note: in java, for the EAGER case, revoke/assign symmetry is not retained in the case of lost partitions. In librdkafka, we do retain this symmetry. But we follow what the Java consumer does exactly for the COOPERATIVE case.

will update rdkafka.h

tests/0113-cooperative_rebalance.cpp
mhowlett: both C_1 and C_2 's count must either be 1 or both must be 2. i could write this differently, but i think spending any effort beyond whatever i first wrote out to articulate this is well below the bar of we-should-be-caring-about-it. it's easy to read, it's not in a key part of the code etc.

edenhill (approved):

PR #3173, opened 2020-12-09
move note about transaction.timeout.ms to begin_transaction

gridaphobe:
As far as I can tell, there is no broker-side requirement to perform transactional operations after calling `rd_kafka_init_transactions`, and the client in `examples/transactions.c` works just fine if I insert arbitrary delays between `init_transactions` and `begin_transaction`. It would also be strange if this requirement did exist, because `init_transactions` is meant to be called once at startup and we can't necessarily control the rate of outgoing messages.


edenhill (approved):
LGTM

PR #3177, opened 2020-12-11
Add arm64 travis job and linux-arm64 (ubuntu 18.04) build to NuGet package

edenhill:


mhowlett (approved):
yeah, if this is common enough, i guess so - how big is it?

edenhill:
14M, it needs to be stripped, will look into it.

PR #3178, opened 2020-12-11
Return added broker in rd_kafka_broker_update

mamapanda:
Based on the doxygen comment, a newly added broker should probably be returned via the `rkbp` parameter.


edenhill:
Thank you!

PR #3179, opened 2020-12-12
Fix possible corrupted decompression of snappy format

azat:
In incremental_copy_fast_path there is undefined behavior (and in some
other places too).

And under this circumstances gcc10 with -O1 -ftree-loop-vectorize (or
simply -O3), due to loop unroll, generates code that do copy by 16 bytes
at a time for the second loop (MOVDQU+MOVUPS), while this is not correct
since the memory may be overlapped and may be changed in the previous
iteration.

Fix this by eliminating those UB, by using memcpy over direct store/load
since these days direct store/loads looks redundant. Even on ARM [1].

  [1]: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67366

Upstream patch: https://github.com/andikleen/snappy-c/pull/22

*NOTE: clang is fine, and other older versions of gcc too*

This bug was found by [ClickHouse](https://github.com/ClickHouse/ClickHouse) test suite.

### DIFF
--- a/src/snappy.c
+++ b/src/snappy.c
@@ -67,6 +67,35 @@
 #define inline __inline
 #endif
 
+static inline u64 get_unaligned64(const void *b)
+{
+	u64 ret;
+	memcpy(&ret, b, sizeof(u64));
+	return ret;
+}
+static inline u32 get_unaligned32(const void *b)
+{
+	u32 ret;
+	memcpy(&ret, b, sizeof(u32));
+	return ret;
+}
+#define get_unaligned_le32(x) (le32toh(get_unaligned32((u32 *)(x))))
+
+static inline void put_unaligned64(u64 v, void *b)
+{
+	memcpy(b, &v, sizeof(v));
+}
+static inline void put_unaligned32(u32 v, void *b)
+{
+	memcpy(b, &v, sizeof(v));
+}
+static inline void put_unaligned16(u16 v, void *b)
+{
+	memcpy(b, &v, sizeof(v));
+}
+#define put_unaligned_le16(v,x) (put_unaligned16(htole16(v), (u16 *)(x)))
+
+
 #define CRASH_UNLESS(x) BUG_ON(!(x))
 #define CHECK(cond) CRASH_UNLESS(cond)
 #define CHECK_LE(a, b) CRASH_UNLESS((a) <= (b))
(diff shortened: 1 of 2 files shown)


azat:
>continuous-integration/appveyor/pr — AppVeyor build failed

gcc'ism is not supported (get_unaligned_memcpy/put_unaligned_memcpy)

azat:
>continuous-integration/travis-ci/pr — The Travis CI build failed


https://travis-ci.org/github/edenhill/librdkafka/jobs/749404551#L2915

```
| 0104_fetch_from_follower_mock            |     FAILED |  92.407s | test_consumer_poll():3864: Consume: consumer_poll() timeout (0/1 eof, 0/1000 msgs)
```

Is it flaky?

edenhill:
@azat Yes, that test is flaky on CI, but never locally, so a bit tricky to debug. You can ignore it.

azat:
@edenhill any feedback?

src/snappy.c
```diff
@@ -67,6 +67,35 @@
 #define inline __inline
 #endif
 
+static inline u64 get_unaligned64(const void *b)
```
edenhill, line 70: What's the reason this patch differs from the upstream patch?
For future maintenance's sake it'd be better if they were identical.

azat: > What's the reason this patch differs from the upstream patch?

MSVC does not support Statement Expressions [1], so they had been replaced with
new static inline functions.

  [1]: https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html#Statement-Exprs

>For future maintenance's sake it'd be better if they were identical.

Indeed, but it is already differs, and the reason is again MCSV/win32

azat: @edenhill friendly ping on this patch

src/snappy_compat.h
```diff
@@ -74,69 +74,6 @@ struct iovec {
...
-		_ret; })
-#define put_unaligned_memcpy(v,x) ({ \
-		typeof((v)) _v = (v); \
-		memcpy((x), &_v, sizeof(*(x))); })
-
-#define get_unaligned_direct(x) (*(x))
-#define put_unaligned_direct(v,x) (*(x) = (v))
-
-// Potentially unaligned loads and stores.
-// x86, PowerPC, and ARM64 can simply do these loads and stores native.
```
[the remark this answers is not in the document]
azat: Because this is UB.

And under this circumstances gcc10 with -O1 -ftree-loop-vectorize (or
simply -O3), due to loop unroll, generates code that do copy by 16 bytes
at a time for the second loop (MOVDQU+MOVUPS), while this is not correct
since the memory may be overlapped and may be changed in the previous
iteration.

edenhill:
Thank you!

PR #3180, opened 2020-12-15
Sketch of what a callback interface for TLS might look like

KJTsanaktsidis:
In https://github.com/edenhill/librdkafka/issues/2868, (and then again in https://github.com/edenhill/librdkafka/issues/3157), there was a discussion about how to update the SSL client certificate of a librdkafka client in a rotation scenario. Librdkafka already supports `rd_kafka_conf_set_ssl_cert_verify_cb` for delegating TLS CA validation checks to a client callback; in this PR, I added an analogous `rd_kafka_conf_set_ssl_cert_fetch_cb` option for setting a callback that provides a TLS certificate to librdkafka when a connection is opened.

The callback is called each time a TLS connection is opened, and it's expected that the application will fill the provided buffer with the TLS client certificate, TLS key, and any intermediates (up to 16, because I couldn't figure out a good way to make that number dynamic, but we can probably make something dynamic work I guess).

This is a bit of a WIP:
    * Obviously whatever feedback comes along :)
    * The test implementation is _very_ rough. In particular, I wonder if all of the TLS cert generation stuff should be moved into trivup? If so I can make a PR there.
    * I haven't quite figured out how to use this interface from golang - the whole point of it was I wanted to be able to plumb a golang `tls.Context` object into a librdkafka connection, so that we can integrate confluent-kafka-go with the golang stdlib TLS stack.
    * Looks like I have a lot of CI fun to deal with.


KJTsanaktsidis:
hum. There's a failed test that I assume is flaky, but also, the test case I added doesn't actually run:

```
[0m[36m[0121_ssl_cb                 /  0.000s] ================= Running test 0121_ssl_cb =================
[0m[36m[0121_ssl_cb                 /  0.000s] ==== Stats written to file stats_0121_ssl_cb_2700797783708785096.json ====
[0m[36m[0121_ssl_cb                 /  0.000s] Feature "ssl" is built-in
[0m[36m[0121_ssl_cb                 /  0.000s] Feature "ssl_client_cert_callback" is built-in
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_SSL_ca_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_SSL_pub_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_SSL_priv_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_UNTRUSTEDSSL_pub_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_UNTRUSTEDSSL_priv_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_INTERMEDIATESSL_pub_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_INTERMEDIATESSL_priv_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_INTERMEDIATESSL_intermediate_pub_pem env var
[0m[33m[0121_ssl_cb                 /  0.000s] WARN: SKIPPING TEST: Test requires RDK_INTERMEDIATESSL_intermediate_priv_pem env var
[0m[36m[0121_ssl_cb                 /  0.000s] 0121_ssl_cb: duration 0.074ms
[0m[36m[0121_ssl_cb                 /  0.000s] ================= Test 0121_ssl_cb SKIPPED =================
```

looks like trivup isn't run with the SSL stuff on travis?

KJTsanaktsidis:
Looks like i'm going to get conflicts every time someone else merges a new testcase, so I'll probably leave them for now until this is closer to being merged.

KJTsanaktsidis:
I also implemented this interface in confluent-kafka-go to allow passing a stdlib `tls.Config` object to handle TLS: https://github.com/confluentinc/confluent-kafka-go/compare/zendesk:ktsanaktsidis/ssl_cb_spike

edenhill:
Have not had time to look at this in detail, but will soon.

> * The test implementation is very rough. In particular, I wonder if all of the TLS cert generation stuff should be moved into trivup? If so I can make a PR there.
Sounds like a good plan. Trivup already has the SslApp that generates certs and stuff.

> * I haven't quite figured out how to use this interface from golang - the whole point of it was I wanted to be able to plumb a golang tls.Context object into a librdkafka connection, so that we can integrate confluent-kafka-go with the golang stdlib TLS stack.

Yeah, we can't really spontaneously call out (callbacks) from C to Go from threads not managed by the Go runtime, which would be the case here since the callbacks would be trigger by the rdkafka broker threads. This is why the go client makes use of rd_kafka_event_t's rather than callbacks, and those events are polled from queues in Go-maintained threads.

Making the SSL handshake enqueue an event, etc, to gather the required certificates would indeed be possible, albeit a bit complex. Before going down that route I'd like to understand the underlying requirements;

There is a requirement to allow certificate/key rotations, but that would not need to be per-broker.
Is there a use-case for per-broker certs/keys?


* Looks like I have a lot of CI fun to deal with.

Some of the tests are flaky in CI (despite vast efforts to reproduce these issues locally).

edenhill:
After some superficial pondering I'm leaning towards doing what we did with OAUTHBEARER token refreshes:
 - have a notification callback that simply says that something needs to be done, in the OAUTHBEARER case its just a refresh interval callback, in this case here it might be a refresh interval callback aswell (based on some ssl.certificate.refresh.interval.ms?) and/or a callback triggered (from the main loop, not broker thread) when a SSL connection fails due to cert verification failure (or whatever the error is).  E.g., ssl_certificate_refresh_cb
 -  add a `rd_kafka_ssl_certificate_set(rk, broker_id(or -1 for all), type(private,public,ca), encoding(PEM), buffer, buffer_size)`.  This call creates a new SSL_CTX for use by any new connections made after this point.

This solves multiple problems:
 * since the callback (or event) is triggered on the main thread it does not block the broker thread, nor does not have any of the Go problems.
 *  the callback is not required since the callback itself does not do anything, it is ssl_certificate_set() that performs the change, and this function can be called spontaneously by the application at any time if it has some other means of knowing when certs need to be rotated. Very flexible.

KJTsanaktsidis:
Thanks for having a look.

> Yeah, we can't really spontaneously call out (callbacks) from C to Go from threads not managed by the Go runtime

Actually you can - I got this working by using `//export`, see https://github.com/zendesk/confluent-kafka-go/blob/zendesk_release/kafka/tlscb_thunk.c and https://github.com/zendesk/confluent-kafka-go/blob/zendesk_release/kafka/tlscb.go. Maybe this is a bad idea for some reason I'm not aware of, but it certainly _seems_ to work.

> Is there a use-case for per-broker certs/keys?

Not that I can think of, no (maybe someone else has this use case, but we don't).

However, the interface for Golang's `tls.Config` object looks like this: `GetClientCertificate func(*CertificateRequestInfo) (*Certificate, error)`. Basically, a standard Golang `tls.Config` object expects to be called once per connection to return a certificate for that connection. So even though I doubt anybody using librdkafka really wants to be able to use a different client certificate per broker, the callback-based interface of the Golang TLS infrastructure kind of makes the assumption that it's possible. Hopefully that makes sense?

> rd_kafka_ssl_certificate_set

Would be backwards compared to how Go wants to handle TLS - that connections trigger a callback on a `tls.Config` object to fetch the right certificate.

> Making the SSL handshake enqueue an event, etc, to gather the required certificates would indeed be possible, albeit a bit complex

Seems to be the best solution for Golang (assuming the callbacks are bad for some reason), because it would allow confluent-kafka-go to call out to `GetClientCertificate` in response to the event and load the right cert. 


---------------------------------------------
Sidebar:

> This is why the go client makes use of rd_kafka_event_t's rather than callbacks, and those events are polled from queues in Go-maintained threads.

I wonder if it might be better to actually make the Golang client use a rebalance callback based on the method I described ☝️ , rather than polling for those events - that might make some of the incremental rebalance stuff a bit more straightfoward, perhaps (eg. https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer.go#L413 )

edenhill:
> Actually you can [call Go from unmanaged threads]..

This did not use to be the case. So, is this officially supported now?

..
> Seems to be the best solution for Golang

Right, but quite costly implementation and complexity-wise, and that's need to be balanced to the benefit of such an interface, which seems mostly to be Go tls-package compliant rather than functionally.

KJTsanaktsidis:
> Right, but quite costly implementation and complexity-wise, and that's need to be balanced to the benefit of such an interface, which seems mostly to be Go tls-package compliant rather than functionally.

You're absolutely right of course - I'm looking at this from a very golang-centric viewpoint but librdkafka powers a wide range of bindings and the needs of each need to be balanced with the complexity it would introduce to librdkafka.

> This did not use to be the case. So, is this officially supported now?

Took a bit of digging, but yes, afaict calling Go from C-created threads is officially supported:
* Original issue tracking the implementation of this: https://github.com/golang/go/issues/4435
* CL for the implementation: https://codereview.appspot.com/7304104
* There's a testcase in the tree for a regression that affected this functionality: https://golang.org/misc/cgo/test/issue8517_windows.c

So the question is, if the calling-Go-from-C is not an issue, what's the best interface for doing dynamic TLS in librdkafka? Should it be callback based with something like `rd_kafka_conf_set_ssl_cert_fetch_cb`, or application driven using a `rd_kafka_ssl_certificate_set` function?

 In our use-case, there are three reasons why I think we would prefer the `rd_kafka_conf_set_ssl_cert_fetch_cb` approach:
* As mentioned, it integrates with Golang's `tls.Config`
* It seems like a natural analogue to `rd_kafka_conf_set_ssl_cert_verify_cb`
* In production at Zendesk, we run most apps, including Kafka consumers, in Kubernetes. We use consul-template (https://github.com/hashicorp/consul-template) running in a sidecar container to fetch short-lived certificates from Hashicorp Vault, and store them in a tmpfs volume accessible to our application. It's not super-trivial to get _notified_ about this happening from the application side, because the sidecar is in a separate PID namespace and so can't do something obvious like sending SIGHUP to our app (there are of course other ways for this communication to happen e.g. over a named pipe or socket stored on the volume, or with inotify, but they all involve a bit of mucking around). With a callback approach, however, the application wouldn't _need_ to know that new TLS certificates are available; instead, it would simply just use the latest on-disk versions of that certificate each time a connection is initiated.

(Hopefully that made sense - me know if you want me to go into this use case in a bit more detail by the way).

However - just because this works for Go, I realise there might be other languages where using a C -> lang callback interface might be a problem. What are the important targets for you? Perhaps I could do some research?

edenhill:
> Took a bit of digging, but yes, afaict calling Go from C-created threads is officially supported:

That's great news!

> However - just because this works for Go, I realise there might be other languages where using a C -> lang callback

Spontaneous C callbacks works fairly well in both Python and .NET, so I think your proposed ..ssl_cert_fetch_cb will work fine.



Thanks for taking the time to dig into this problem.

KJTsanaktsidis:
Cool! If we're happy to go along with the callback approach, is there anything else in this PR I should redesign now? Or should I move on to tidying up the test/trivup stuff next?

src/rdkafka.h
```diff
@@ -2270,6 +2284,60 @@ typedef enum rd_kafka_cert_enc_t {
...
+ *
+ * @returns RD_KAFKA_CONF_OK if SSL is supported in this build, else
+ *          RD_KAFKA_CONF_INVALID.
+ *
+ * @warning This callback will be called from internal librdkafka threads.
+ * */
+RD_EXPORT
+rd_kafka_conf_res_t rd_kafka_conf_set_ssl_cert_fetch_cb(
+        rd_kafka_conf_t *conf,
+        int (*ssl_cert_fetch_cb) (rd_kafka_t *rk,
```
edenhill, line 2330: Should return the enum type, right?

src/rdkafka.h
```diff
@@ -2270,6 +2284,60 @@ typedef enum rd_kafka_cert_enc_t {
...
+                                  const char *broker_name,
+                                  int32_t broker_id,
+                                  char *buf, size_t *buf_size,
+                                  char **leaf_cert, size_t *leaf_cert_size,
+                                  char **pkey, size_t *pkey_size,
+                                  char *chain_certs[16], size_t chain_cert_sizes[16],
```
edenhill, line 2336: Why are these limited to 16?

src/rdkafka.h
```diff
@@ -2270,6 +2284,60 @@ typedef enum rd_kafka_cert_enc_t {
...
+                                  rd_kafka_cert_enc_t *format,
```
edenhill, line 2337: add a p suffix to all return-type pointers, makes them easier to differentiate in the code.
E.g.: `*buf_sizep`, `*formatp`

src/rdkafka.h
```diff
@@ -2270,6 +2284,60 @@ typedef enum rd_kafka_cert_enc_t {
...
+ *   - write the size of the certificate to certificate_buf_size,
+ *   - write a key to key_buf,
+ *   - write the format of that key to key_format
+ *   - write the size of the key to key_buf_size,
+ *
+ * Otherwise, if the provided buffers are too small (i.e *certificate_buf_size
+ * or *key_buf_size are too small), then the callback should:
+ *   - Return RD_KAFKA_CERT_FETCH_MORE_BUFFER,
+ *   - write the desired buffer size to *buf_size
+ * Note that the first time rd_kafka_conf_set_ssl_cert_fetch_cb is called, it is
```
edenhill, line 2315: Can we flip this around and simply have the callback allocate the memory (through rd_kafka_malloc) and have librdkafka free it when done?

src/rdkafka_conf.c
```diff
@@ -306,6 +306,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
                         },
                         { 0x400, "zstd", _UNSUPPORTED_ZSTD },
                         { 0x800, "sasl_oauthbearer", _UNSUPPORTED_SSL },
+                        { 0x1000, "ssl_client_cert_callback", _UNSUPPORTED_OPENSSL_1_0_2 },
```
edenhill, line 309: Don't think we need to add this as a feature flag, has little value for most users and the fetch_cb setter should just return NOT_SUPPORTED instead.

src/rdkafka_conf.c
```diff
@@ -847,7 +848,11 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           "Callback to verify the broker certificate chain.",
           _UNSUPPORTED_SSL
         },
-
+        { _RK_GLOBAL, "ssl.certificate.fetch_cb", _RK_C_PTR,
```
edenhill, line 851: Use all-underscore property names for properties that require programmatical setters, as to differentiate them from string properties.

edenhill: Oh, I see we missed this for verify_cb, damn. Okay, let's keep fetch_cb as is then.

src/rdkafka_conf.c
```diff
@@ -2810,6 +2815,27 @@ rd_kafka_conf_set_ssl_cert_verify_cb (
 #endif
 }
 
+rd_kafka_conf_res_t
+rd_kafka_conf_set_ssl_cert_fetch_cb(
```
edenhill, line 2819: style nit: space between function and open-paren

src/rdkafka_conf.c
```diff
@@ -2810,6 +2815,27 @@ rd_kafka_conf_set_ssl_cert_verify_cb (
...
+                                  int32_t broker_id,
+                                  char *buf, size_t *buf_size,
+                                  char **leaf_cert, size_t *leaf_cert_size,
+                                  char **pkey, size_t *pkey_size,
+                                  char *chain_cert[16], size_t chain_cert_size[16],
+                                  rd_kafka_cert_enc_t *format,
+                                  void *opaque)) {
+
+#if defined(WITH_SSL) && OPENSSL_VERSION_NUMBER >= 0x1000200fL
+    rd_kafka_anyconf_set_internal(_RK_GLOBAL, conf,"ssl.certificate.fetch_cb", ssl_cert_fetch_cb);
```
edenhill, line 2832: Keep below 80 cols.
Add space after ","

src/rdkafka_conf.c
```diff
@@ -2810,6 +2815,27 @@ rd_kafka_conf_set_ssl_cert_verify_cb (
...
+    return RD_KAFKA_CONF_OK;
+#else
+    return RD_KAFKA_CONF_INVALID;
```
edenhill, line 2835: :+1:

src/rdkafka_ssl.c
```diff
@@ -417,6 +417,123 @@ rd_kafka_transport_ssl_cert_verify_cb (int preverify_ok,
...
+        size_t pkey_size;
+        size_t chain_cert_sizes[16] = { 0 };
+        rd_kafka_cert_enc_t format;
+
+        char errstr[512] = { 0 };
+        rd_kafka_cert_t *leaf_cert = NULL;
+        rd_kafka_cert_t *pkey = NULL;
+
+        do {
+                if (buf) rd_kafka_mem_free(rk, buf);
```
edenhill, line 450: style:
```c
if (buf)
    rd_kafka_mem_free..
```

src/rdkafka_ssl.c
```diff
@@ -417,6 +417,123 @@ rd_kafka_transport_ssl_cert_verify_cb (int preverify_ok,
...
+        char *buf = NULL;
+        char *leaf_cert_ptr;
+        char *pkey_ptr;
+        char *chain_cert_ptrs[16] = { 0 };
+        size_t leaf_cert_size;
+        size_t pkey_size;
+        size_t chain_cert_sizes[16] = { 0 };
+        rd_kafka_cert_enc_t format;
+
+        char errstr[512] = { 0 };
```
edenhill, line 445: This clears all 512 bytes, we just need to clear the first.
Do:
`*errstr = '\0';` after the variable defs.

src/rdkafka_ssl.c
```diff
@@ -417,6 +417,123 @@ rd_kafka_transport_ssl_cert_verify_cb (int preverify_ok,
...
+static int rd_kafka_transport_ssl_cert_fetch_cb(SSL *ssl, void *arg) {
+        rd_kafka_transport_t *rktrans = rd_kafka_curr_transport;
+        rd_kafka_broker_t *rkb;
+        rd_kafka_t *rk;
+        int r;
+        int i;
+        int ret;
+        int num_chain_certs;
+
+        rd_assert(rktrans != NULL);
```
edenhill, line 430: All variable defs need to go at the top of the scope before any code, for compatibility with <c99 compilers.

src/rdkafka_ssl.c
```diff
@@ -417,6 +417,123 @@ rd_kafka_transport_ssl_cert_verify_cb (int preverify_ok,
...
+        } while (cb_res == RD_KAFKA_CERT_FETCH_MORE_BUFFER);
+
+        if (cb_res == RD_KAFKA_CERT_FETCH_NONE) {
+                ret = 1;
+                goto out;
+        } else if (cb_res == RD_KAFKA_CERT_FETCH_ERR) {
+                ret = 0;
+                goto out;
+        }
+
```
edenhill, line 468: keep below 80 cols

src/rdkafka_ssl.c
```diff
@@ -417,6 +417,123 @@ rd_kafka_transport_ssl_cert_verify_cb (int preverify_ok,
...
+                                                       &pkey_ptr, &pkey_size,
+                                                       chain_cert_ptrs, chain_cert_sizes,
+                                                       &format,
+                                                       rk->rk_conf.opaque);
+        } while (cb_res == RD_KAFKA_CERT_FETCH_MORE_BUFFER);
+
+        if (cb_res == RD_KAFKA_CERT_FETCH_NONE) {
+                ret = 1;
+                goto out;
+        } else if (cb_res == RD_KAFKA_CERT_FETCH_ERR) {
```
edenhill, line 464: I think this should be logged, and that we should pass errstr+errstr_size to the callback .

src/rdkafka_ssl.c
```diff
@@ -417,6 +417,123 @@ rd_kafka_transport_ssl_cert_verify_cb (int preverify_ok,
...
+                           "Certificate fetch callback private key loading failed: %s", errstr);
+                ret = 0;
+                goto out;
+        }
+
+        num_chain_certs = sizeof(chain_cert_ptrs)/sizeof(chain_cert_ptrs[0]);
+        for (i = 0; i < num_chain_certs; i++) {
+                if (!chain_cert_ptrs[i]) continue;
+
+                rd_kafka_cert_t *chain_cert = rd_kafka_cert_new(&rk->rk_conf, RD_KAFKA_CERT_PUBLIC_KEY, format,
```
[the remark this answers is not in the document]
edenhill: 80 cols

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
+/*
```
edenhill, line 1: 0122 is now taken, switch to 0124

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2020, Magnus Edenhill
```
edenhill, line 4: 2021

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+        const char *broker_name,
+        int32_t broker_id,
+        char *buf, size_t *buf_size,
+        char **leaf_cert, size_t *leaf_cert_size,
+        char **pkey, size_t *pkey_size,
+        char *chain_certs[16], size_t chain_cert_sizes[16],
+        rd_kafka_cert_enc_t *format,
+        void *opaque
+) {
+    struct test_context *ctx;
```
edenhill, line 67: Use 8 whitespace indent

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+    *leaf_cert_size = ctx->direct_signed_client_cert_pem_length;
+    *pkey = buf + ctx->direct_signed_client_cert_pem_length;
+    *pkey_size = ctx->direct_signed_client_key_pem_length;
+    *format = RD_KAFKA_CERT_ENC_PEM;
+    memcpy(*leaf_cert, ctx->direct_signed_client_cert_pem, ctx->direct_signed_client_cert_pem_length);
+    memcpy(*pkey, ctx->direct_signed_client_key_pem, ctx->direct_signed_client_key_pem_length);
+    return RD_KAFKA_CERT_FETCH_OK;
+}
+
+static int ssl_client_cert_cb_bad(
```
edenhill, line 88: space before paren

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+        TEST_FAIL_LATER("File %s could not be read: %m\n", path_val);
+        fclose(fh);
+        return;
+    }
+    fclose(fh);
+    return;
+}
+
+int main_0122_ssl_cb(int argc, char **argv) {
+    struct test_context ctx;
```
edenhill, line 239: = RD_ZERO_INIT;  and skip the memset

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+
+    memset(&ctx, 0, sizeof(struct test_context));
+
+    if (!test_check_builtin("ssl")) {
+        TEST_SKIP("Test requires SSL support\n");
+        goto cleanup;
```
edenhill, line 245: Nothing set up, so just return

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+    }
+    if (!test_check_builtin("ssl_client_cert_callback")) {
```
edenhill, line 247: You can assume we're always on 1.0.2 or later (it is EOL, so 1.0.1 is really dead now), so you can skip this test.

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+        TEST_SKIP("Test requires ssl_client_cert_callback support (i.e. built with openssl >= 1.0.2\n");
+        goto cleanup;
+    }
+
+
```
edenhill, line 252: 80 cols throughout

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+    read_ssl_file_from_env(
+            "RDK_INTERMEDIATESSL_priv_pem", &ctx.intermediate_signed_client_key_pem, &ctx.intermediate_signed_client_key_pem_length
+    );
+    read_ssl_file_from_env(
+            "RDK_INTERMEDIATESSL_intermediate_pub_pem", &ctx.intermediate_signed_intermediate_cert_pem, &ctx.intermediate_signed_intermediate_cert_pem_length
+    );
+    read_ssl_file_from_env(
+            "RDK_INTERMEDIATESSL_intermediate_priv_pem", &ctx.intermediate_signed_intermediate_key_pem, &ctx.intermediate_signed_intermediate_key_pem_length
+    );
+    if (test_curr->state == TEST_SKIPPED || test_curr->state == TEST_FAILED) {
```
edenhill, line 280: Test's shouldn't fiddle with this directly.
It is better to TEST_FAIL() immediately where the error is.

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+        goto cleanup;
+    }
+
+    do_test_ssl_cb(&ctx, 1, ssl_client_cert_cb_good_direct);
```
edenhill, line 284: This is not really a callback, so rename it to do_test_ssl_cert() and also pass it a `const char *what` string that says what the test is doing.
Then in do_test_ssl_cert() before any of the code you do:
```c
SUB_TEST_QUICK("%s", what);
....all the testing...
SUB_TEST_PASS();
}
```

This will make it easier to track down which test is failing, and allow you to run a specific subtest with:
`TESTS=0124 SUBTESTS=intermediate make`

tests/interactive_broker_version.py
```diff
@@ -122,6 +122,26 @@ def test_version (version, cmd=None, deploy=True, conf={}, debug=False, exec_cnt
             else:
                 cmd_env['RDK_SSL_{}'.format(k)] = v
 
+        # Also create some invalid certificates, that can be used by tests that want to ensure
```
edenhill, line 125: :+1:

tests/cluster_testing.py
```diff
@@ -140,6 +141,145 @@ def print_report_summary (fullreport):
           (pass_cnt, pass_clr, fail_cnt, fail_clr))
 
 
+def create_selfsigned_cert(ssl, cn):
```
edenhill, line 144: I'm thinking these are generally useful and should be in trivup, can you open a PR there instead - I'll merge it quickly and make a new release?

tests/0122-ssl_cb.c
```diff
@@ -0,0 +1,308 @@
...
+    if (test_curr->state == TEST_SKIPPED || test_curr->state == TEST_FAILED) {
```
edenhill, line 285: Remove these, we want to fail fast, so use TEST_FAIL instead of FAIL_LATER.

edenhill (changes requested):
Great stuff!

src/rdkafka.h
KJTsanaktsidis: Mostly because I wanted to be able to avoid dynamic memory allocation, and handing around ownership of dynamically allocated buffers, being part of the interface of the callback. But if we're happy to have the callback allocate memory with rd_kafka_malloc, and make it librdkafka's responsibility to subsequently free that (i.e. ownership of the buffers is being handed over as well), then the length can be dynamic too.

src/rdkafka.h
KJTsanaktsidis: Yeah I'm not really a C expert but I thought it would be "ugly" to transfer ownership of the buffers like that as part of the callback interface. It would work though and make the API probably simpler to use. So I'll make this change, and document it appropriately.

src/rdkafka_ssl.c
KJTsanaktsidis: Hah, didn't realise that's a thing that we still needed! Will fix this up.

tests/0122-ssl_cb.c
KJTsanaktsidis: Yup, i'll merge + tidy up the conflicts etc.

tests/0122-ssl_cb.c
KJTsanaktsidis: it was 2020 when I wrote it! 😂

KJTsanaktsidis:
Thanks for the detailed review - all of your comments make good sense so I'm happy to switch that stuff around.

Sounds like the first step is to extract the SSL testing stuff into trivup. I'm currently working through some feedback on one of my confluent-kafka-go PR's, when I'm done with that I'll tackle this.

After that, I'll give the tests a bit of love and attention as you suggested, change the memory-management aspects ofthe interface to be based on passing around rd_kafka_malloc'd blocks of memory, and go over the code style with a fine tooth comb.

Thanks again, really appreciate it.

edenhill:
No, Thank you! :100:

I'm not sure you're aware but there's also the OpenSSL engine support landing in the same release, which is somewhat related to your work:
https://github.com/edenhill/librdkafka/pull/3315

KJTsanaktsidis:
Actually the OpenSSL Engine stuff is interesting. I had a bit of a dig through the code, I think it would be possible in confluent-kafka-go to register an OpenSSL engine which provided client keys to librdkafka, and pass that as the engine_id parameter.

I couldn't see an obvious way to also have the engine overwrite the TLS server certificate verification routine, which is a bummer, but of course the existing callback can still be used for this.

Reckon that's an approach worth pursuing, or is the specific callback for loading a client cert still worth having?

edenhill:
I think it'll be cumbersome to write an engine and include that in the Go client (seeing how much work and fiddling it was to get librdkafka and all its dependencies bundled), so I think a standalone TLS interface is still warranted.

src/rdkafka.h
KJTsanaktsidis: Hum. So I went looking at this, there _is_ no `rd_kafka_malloc`. I'm going to define `rd_kafka_mem_alloc` as an analogue to `rd_kafka_mem_free` and have it just wrap `rd_malloc`.

tests/0122-ssl_cb.c
KJTsanaktsidis: So the reason I went to such effort to do it this way was because I wanted to make sure the test appropriately cleaned up after itself on fail. Memory and open file handles mostly. Seems like `TEST_FAIL` just calls `pthread_exit`, which seems rather violent :D

If the working assumption here is that we don't care about leaking in test cases, I can definitely go and just drop TEST_FAIL in all the right places.

src/rdkafka.h
KJTsanaktsidis: 🤦  so there _wasn't_ an allocator function on my branch... so I implemented one as `rd_kafka_mem_alloc`... then rebased onto master to fixup conflicts, and lo and behold, `rd_kafka_mem_malloc` appears. Now I feel daft.

KJTsanaktsidis:
@edenhill Think this is ready for another look. I've adjusted the direction of memory ownership so that the callbacks allocate buffers with `rd_kafka_mem_malloc` and pass ownership of them to librdkafka, which seems to simplify things a fair bit.

I've done my best to keep the formatting consistent, but if there's a tool or configuration of some kind in the tree that fixes up the formatting a-la-rubocop or something let me know and I'll feed it through that as well.

KJTsanaktsidis:
Hey, just a polite bump @edenhill to see if you've had a chance to look at this yet? :) 🙏

KJTsanaktsidis:
@edenhill just wondering if you had any further thoughts on this?

egerszu:
@edenhill , @KJTsanaktsidis Any update on this please? We have to recreate the kafka clients every 10 minutes due to the certificate rotation and an unstable network.

KJTsanaktsidis:
We’ve been using this patch in production for a couple of months now and it works great.

What language are you using librdkafka for? If Golang I can recommend the above patch for confluent-kafka-go too.

antonio-oliva:
Hi @KJTsanaktsidis and others,
I was wondering if the latest version of the confluent-kafka-go library, i.e., v2.3.0, supports certificates renewal when TLS is used for the broker-clients communication. 
The scenario I would like to implement is:
1. Setting up the TLS configuration with the proper certificate
2. Every time the old certificates are substituted with renewed ones (even before the expiration) broker and clients to automatically reload certificates, without having to restart them.

Is such a behavior implemented in confluent-kafka-go v2.3.0?

danhje:
Really hope this can be merged, and surfaced through confluent-kafka-python.

suleymanbyzt:
any progress on this? We're hitting the exact problem this PR solves in production 24h Vault PKI client certs, and every rotation forces a Kafka consumer rebuild that causes state loss bugs downstream.

PR #3187, opened 2020-12-21
KIP-447

edenhill:


CHANGELOG.md
```diff
@@ -1,26 +1,37 @@
...
-librdkafka v1.6.0 is a feature release adding support for
-[KIP-429 Incremental rebalancing](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol)
-and the sticky consumer group partition assignor.
+librdkafka v1.6.0 is feature release:
 
- * Incremental rebalancing and assignments - FIXME
- * Sticky consumer group partition assignor - FIXME
- * Sticky producer partitioning (`sticky.partitioning.linger.ms`) -
+ * [KIP-429 Incremental rebalancing](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol) with sticky
+   consumer group partition assignor (KIP-54) (by @mhowlett).
```
mhowlett, line 6: in fairness you could put KIP-54 on another line and not give me credit :-)

CHANGELOG.md
```diff
@@ -1,26 +1,37 @@
...
-   affects the distribution of randomly partitioned messages, where previously
-   these messages would be evenly distributed over the available partitions
-   they are now partitioned to a single partition for the duration of the
-   sticky time (10 milliseconds by default) before a new random sticky
-   partition is selected.
+ * Sticky producer partitioning (`sticky.partitioning.linger.ms`) is
+   enabled by default (10 milliseconds) which affects the distribution of
+   randomly partitioned messages, where previously these messages would be
+   evenly distributed over the available partitions they are now partitioned
+   to a single partition for the duration of the sticky tim
```
mhowlett, line 23: tim -> time

examples/transactions.c
```diff
@@ -27,7 +27,7 @@
  */
 
 /**
- * @name Transactions example
+ * @name Transactions example for Apache Kafka 2.5.0 (KIP-447) and later.
```
mhowlett, line 30: we really need to update the examples in the other repos as well.

examples/transactions.c
```diff
@@ -36,8 +36,9 @@
  * The transaction is committed every 5 seconds or 100 messages, whichever
  * comes first. As the transaction is committed a new transaction is started.
  *
- * @remark This example does not yet support incremental rebalancing and thus
- *         not the cooperative-sticky partition.assignment.strategy.
+ * This example makes use of incremental rebalancing (KIP-429) and the
+ * cooperative-sticky partition.assignment.strategy on the consumer, providing
+ * hitless rebalances.
```
mhowlett, line 41: 👊 🤜

mhowlett: hitless doesn't sound right.

maybe "optimized, non stop-the-world rebalances."? (not great either, but i think a step forward)

examples/transactions.c
```diff
@@ -184,135 +154,137 @@ create_transactional_producer (const rd_kafka_topic_partition_t *rktpar) {
...
-                rd_kafka_error_destroy(error);
+        if (offsets->cnt == 0) {
+                fprintf(stderr, "No current assignment to rewind\n");
+                rd_kafka_topic_partition_list_destroy(offsets);
+                return;
         }
 
-        rd_kafka_destroy(rk);
-}
+        /* Note: Timeout must be lower than max.poll.interval.ms */
```
mhowlett, line 186: is that enforced? (it doesn't seem to be directly in `rd_kafka_committed` but may be further in, didn't look).

examples/transactions.c
```diff
@@ -184,135 +154,137 @@ create_transactional_producer (const rd_kafka_topic_partition_t *rktpar) {
...
+         * committed offset is available. */
+        for (i = 0 ; i < offsets->cnt ; i++) {
+                /* No committed offset, start from beginning */
+                if (offsets->elems[0].offset < 0)
+                        offsets->elems[0].offset =
+                                RD_KAFKA_OFFSET_BEGINNING;
+        }
 
+        /* Perform seek */
+        error = rd_kafka_seek_partitions(consumer, offsets, -1);
```
mhowlett, line 202: what if this blocks longer than `max.poll.interval.ms`? should this also have a positive timeout?

src/rdkafka.h
```diff
@@ -3428,6 +3434,39 @@ rd_kafka_resp_err_t rd_kafka_seek (rd_kafka_topic_t *rkt,
...
+ * If \p timeout_ms is not 0 the call will wait this long for the
+ * seeks to be performed. If the timeout is reached the internal state
+ * will be unknown for the remaining partitions to seek and this function
+ * will return an error with the error code set to
+ * `RD_KAFKA_RESP_ERR__TIMED_OUT`.
+ *
+ * If \p timeout_ms is 0 it will initiate the seek but return
+ * immediately without any error reporting (e.g., async).
+ *
+ * This call triggers a fetch queue barrier flush.
```
mhowlett, line 3453: maybe add another sentence to clarify this in more layperson terms.

src/rdkafka.h
```diff
@@ -3428,6 +3434,39 @@ rd_kafka_resp_err_t rd_kafka_seek (rd_kafka_topic_t *rkt,
...
+ * @brief Seek consumer for partitions in \p partitions to the per-partition
+ *        offset in the \c .offset field of \p partitions.
+ *
+ * The offset may be either absolute (>= 0) or a logical offset.
+ *
+ * If \p timeout_ms is not 0 the call will wait this long for the
+ * seeks to be performed. If the timeout is reached the internal state
+ * will be unknown for the remaining partitions to seek and this function
+ * will return an error with the error code set to
+ * `RD_KAFKA_RESP_ERR__TIMED_OUT`.
```
mhowlett, line 3448: i think it would be good to add a short summary here about what happens when you call seek. enough so that the user can understand why it might block and that it currently flushes the message cache.

also in `rd_kafka_seek`

examples/transactions.c
mhowlett: what if neither `seek_partitions` or `committed` blocks for > 10s, but combined they do?

i guess you'll get a max poll interval error in the consume loop. i notice there you just log it though, whereas timeouts above are considered fatal.

examples/transactions.c
```diff
@@ -184,135 +154,137 @@ create_transactional_producer (const rd_kafka_topic_partition_t *rktpar) {
...
+                rd_kafka_topic_partition_list_destroy(offsets);
+                return;
         }
 
-        rd_kafka_destroy(rk);
-}
+        /* Note: Timeout must be lower than max.poll.interval.ms */
+        err = rd_kafka_committed(consumer, offsets, 10*1000);
+        if (err)
+                fatal("Failed to acquire committed offsets: %s",
```
mhowlett, line 189: you consider lots of things to be fatal. is this your recommendation for a real-world app?

it's what i'd do, reasoning: 1. these problems are unlikely 2. there are (probably?) some scenarios for which killing the app and have something restart it is the only thing you can do 3. you might as well just do this for a relatively wide range of issues to keep the code simple, which has the added benefit of making error handling logic errors less likely.

it would be good to get the error handling strategy / philosophy articulated well.

examples/transactions.c
edenhill: We could provide a `rd_kafka_get_remaing_time_until_i_need_to_poll_again(consumer, desired_max_time)` which can be used as a timeout, but we can't really change the timeout behaviour of existing calls.

But I doubt it matters in practice, if any of these (seek, commit, max.poll, ..) fail the application needs to perform some abortive and resetting action anyway, so I'd say it is probably simpler to assume seek, commit, etc, with infinite timeout will finish in time, and if they don't the max poll will kick in and trigger a rebalance - which the application does not necessarily need to handle, things will recover anyway.

examples/transactions.c
```diff
@@ -184,135 +154,137 @@ create_transactional_producer (const rd_kafka_topic_partition_t *rktpar) {
...
         rd_kafka_error_t *error;
 
-        fprintf(stdout,
-                "Aborting transaction and rewinding offset for %s [%d]\n",
-                state->rktpar->topic, state->rktpar->partition);
+        fprintf(stdout, "Aborting transaction and rewinding offsets\n");
 
         /* Abort the current transaction */
-        error = rd_kafka_abort_transaction(state->producer, -1);
+        error = rd_kafka_abort_transaction(producer, -1);
```
mhowlett, line 221: might as well throw in an unactionable complaint here about there being no mechanism for the application to abort an operation with infinite timeout.

examples/transactions.c
edenhill: Someone once said: "error out early", so I did.

The reason for this aggressive error handling is to keep the example simple, otherwise the error handling matrix would obscure the meat of what we're trying to demonstrate in this here example.

We could add comments in the actually-not-so-fatal-places why it happens, how it can be handled, etc, but ultimately we want rich error objects to tell this tale instead, which pretty much boils down to if an error is retriable or not.
This is not possible with the current API though.
A hacky alternative for the legacy APIs could be something like:
```c
        err = rd_kafka_committed(consumer, offsets, 10*1000);
        if (err) {
               if (rd_kafka_err_is_retriable(rd_kafka_committed, err))
                     goto retry;
                fatal("Failed to acquire committed offsets: %s",
̣             ..
         }

```

or something that creates a rich error_t* object from an API and an error code, like so:

```c
rd_kafka_error_t *rd_kafka_err2error (rd_kafka_api_function_ptr, rd_kafka_resp_err_t err);
```

edenhill: or a per-thread last-error object (we already have this for the legacy-legacy errno to resp_err_t conversion):
```c
rd_kafka_error_r *rd_kafka_get_last_error_object(rd_kafka_t *rk);
```

edenhill: All these alternatives are a bit funky in C, but the high-level bindings can hide it.

examples/transactions.c
edenhill: `rd_kafka_queue_yield()` is half-way there..

examples/transactions.c
```diff
@@ -184,135 +154,137 @@ create_transactional_producer (const rd_kafka_topic_partition_t *rktpar) {
...
         /* Send offsets to transaction coordinator */
-        error = rd_kafka_send_offsets_to_transaction(state->producer,
-                                                     offset, cgmd, -1);
+        error = rd_kafka_send_offsets_to_transaction(producer,
+                                                     offsets, cgmd, -1);
         rd_kafka_consumer_group_metadata_destroy(cgmd);
-        rd_kafka_topic_partition_list_destroy(offset);
+        rd_kafka_topic_partition_list_destroy(offsets);
         if (error) {
                 if (rd_kafka_error_txn_requires_abort(error)) {
```
mhowlett, line 289: it's mandatory to check an error returned by a transactional api to see whether it requires and abort (right?), and then abort if so. why don't the methods just do this, and return an error indicating that they've done it? would have lead to a simpler to user api. or am i missing a benefit?

edenhill: mimicing the Java API, afair

examples/transactions.c
```diff
@@ -321,112 +293,125 @@ static void commit_transaction_and_start_new (struct state *state) {
...
                              rd_kafka_topic_partition_list_t *partitions,
                              void *opaque) {
-        int i;
+        rd_kafka_t *producer = (rd_kafka_t *)opaque;
+        rd_kafka_error_t *error;
 
-        if (!strcmp(rd_kafka_rebalance_protocol(rk), "COOPERATIVE"))
-                fatal("This example has not yet been modified to work with "
-                      "cooperative incremental rebalancing "
+        if (strcmp(rd_kafka_rebalance_protocol(consumer), "COOPERATIVE"))
```
mhowlett, line 365: the incremental assign will fail in this isn't the case - i don't think checking this is necessary. i only added the `rd_kafka_rebalance_protocol` method because it was required by the C# binding. generally, I don't think users should use it.

examples/transactions.c
```diff
@@ -321,112 +293,125 @@ static void commit_transaction_and_start_new (struct state *state) {
...
-                                " %s [%d] with transactional producer %s\n",
-                                rktpar->topic, rktpar->partition,
-                                rd_kafka_name(state->producer));
-                }
+                fprintf(stdout,
+                        "Consumer group rebalanced: "
+                        "%d new partition(s) assigned\n",
+                        partitions->cnt);
 
                 /* Let the consumer know the rebalance has been handled
```
mhowlett, line 378: i'm not into the words "let the consumer know". i'd say what it does (assign the consumer to read from the additional partitions), and that it's mandatory.

examples/transactions.c
```diff
@@ -321,112 +293,125 @@ static void commit_transaction_and_start_new (struct state *state) {
...
-                assigned_partitions = NULL;
+                /* Begin new transaction */
+                error = rd_kafka_begin_transaction(producer);
+                if (error)
+                        fatal_error("Failed to begin transaction", error);
 
-                /* Let the consumer know the rebalance has been handled
-                 * and revoke the current assignment. */
-                rd_kafka_assign(rk, NULL);
+                /* Let the consumer know the rebalance has been handled */
```
mhowlett, line 411: again not into the wording "let the.."

examples/transactions.c
```diff
@@ -602,67 +606,70 @@ int main (int argc, char **argv) {
...
         fprintf(stdout, "Closing consumer\n");
         rd_kafka_consumer_close(consumer);
-
         rd_kafka_destroy(consumer);
 
+        fprintf(stdout, "Closing producer\n");
+        rd_kafka_destroy(producer);
+
         return 0;
 }
```
mhowlett, line 675: logic seem fine

the main thing I feel is i want more clarity on error handling / timeouts. i think the way to do it is try and compress a set of general ideas into some blurb somewhere, so as to create a mental model for thinking about this and have commentary in specific places linking back to that.

src/rdbuf.c
```diff
@@ -220,7 +220,7 @@ rd_buf_get_segment_at_offset (const rd_buf_t *rbuf, const rd_segment_t *hint,
                               size_t absof) {
         const rd_segment_t *seg = hint;
 
-        if (unlikely(absof > rbuf->rbuf_len))
+        if (unlikely(absof >= rbuf->rbuf_len))
```
mhowlett, line 223: seems important!

src/rdbuf.c
```diff
@@ -588,6 +588,81 @@ void rd_buf_push (rd_buf_t *rbuf, const void *payload, size_t size,
 
 
 
+/**
+ * @brief Erase \p size bytes at \p of from buffer.
```
mhowlett, line 592: of -> absof

src/rdbuf.c
```diff
@@ -1582,6 +1658,189 @@ static int do_unittest_write_iov (void) {
...
+                RD_UT_ASSERT(rd_slice_remains(&s) == 0,
+                             "expected no remaining bytes in slice, but got "
+                             "%"PRIusz" for i=%d",
+                             rd_slice_remains(&s), i);
+
+                rd_buf_destroy(&b);
+        }
+
+
+        RD_UT_PASS();
```
mhowlett, line 1841: test cases lgtm

src/rdbuf.h
```diff
@@ -89,6 +89,14 @@ typedef struct rd_buf_s {
...
         rd_segment_t     *rbuf_wpos;          /**< Current write position seg */
         size_t            rbuf_len;           /**< Current (written) length */
+        size_t            rbuf_erased;        /**< Total number of bytes
+                                               *   erased from segments.
+                                               *   This amount is taken into
+                                               *   account when checking for
+                                               *   writable space which is
+                                               *   always at the end of the
+                                               *   buffer and thus can't make
+                                               *   use of the erased parts. */
```
mhowlett, line 99: i haven't looked at this in depth, but want to note that it seems off to me that state related to deleting part of a buf should leak outside that scope of that operation.

CHANGELOG.md
```diff
@@ -49,6 +60,8 @@ and the sticky consumer group partition assignor.
  * Will now log a warning when multiple ClusterIds are seen, which is an
    indication that the client might be erroneously configured to connect to
    multiple clusters which is not supported.
+ * Added `rd_kafka_seek_partitions()` to seek multiple partitions to
```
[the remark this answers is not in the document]
mhowlett: it does. probably falls off the bottom of the priority list atm i think

src/rdbuf.h
edenhill: So maybe it is a good time to check what the implementation does and why this is necessary ;)

src/rdkafka_broker.c
```diff
@@ -1042,6 +1042,11 @@ static void rd_kafka_broker_buf_enq0 (rd_kafka_broker_t *rkb,
 static void rd_kafka_buf_finalize (rd_kafka_t *rk, rd_kafka_buf_t *rkbuf) {
         size_t totsize;
 
+        if (rkbuf->rkbuf_flags & RD_KAFKA_OP_F_FLEXVER) {
+                /* Empty struct tags */
+                rd_kafka_buf_write_i8(rkbuf, 0);
```
mhowlett, line 1047: don't understand why, assume you know what you're doing.

src/rdkafka_buf.c
```diff
@@ -148,9 +154,18 @@ rd_kafka_buf_t *rd_kafka_buf_new_request (rd_kafka_broker_t *rkb, int16_t ApiKey
...
         rd_kafka_buf_write_i32(rkbuf, 0);
 
-        /* ClientId: possibly updated later if a flexible version */
+        /* ClientId */
         rd_kafka_buf_write_kstr(rkbuf, rkb->rkb_rk->rk_client_id);
 
+        if (is_flexver) {
+                /* Must set flexver after writing the client id since
+                 * it is still a standard non-compact string. */
+                rkbuf->rkbuf_flags |= RD_KAFKA_OP_F_FLEXVER;
```
mhowlett, line 163: is this right? from the kip:

> The new RequestHeader version will be version 1, superseding version 0.  In this new version, the RequestHeader's ClientId string will be a COMPACT_STRING rather than  STRING

src/rdkafka_buf.h
```diff
@@ -951,11 +967,60 @@ static RD_INLINE void rd_kafka_buf_update_u32 (rd_kafka_buf_t *rkbuf,
...
+ * @brief Write the final array count to the position returned from
+ *        rd_kafka_buf_write_arraycnt_pos().
+ *
+ * Update int32_t in buffer at offset 'of' but serialize it as
+ * compact uvarint (that must not exceed 4 bytes storage)
+ * if the \p rkbuf is marked as FLEXVER, else just update it as
+ * as a standard update_i32().
+ *
+ * @remark For flexibleVersions this will shrink the buffer and move data
+ *         and may thus be costly.
```
mhowlett, line 987: ouch.

it looks like it's feasible to compute the size of these ahead of time (where required, currently anyway), and that would be more efficient.

or, maybe a better strategy to initially allocate a smaller number of bytes for arraycnt (whatever is most commonly expected, possibly have that configurable per usage), and insert bytes if the estimate isn't right.

but optimization. if it works i don't care, task for later PR.

edenhill: memmoves are pretty fast since everything is in cache.
Pre-calculating a multi-tiered list of topics and partition sizes is more work.

mhowlett (approved):
various comments to do what you like with. i'm kind of assuming you read the kip correctly in places.

CHANGELOG.md
edenhill: Better to be able to blame you!

examples/transactions.c
edenhill: Meh.

examples/transactions.c
edenhill: Nope, see next comment.

examples/transactions.c
edenhill: :+1:

src/rdkafka_buf.c
edenhill: Yeah I don't think that actually materialized, the header ClientId is still a 16-bit lengthed string.

edenhill:
Thank you for the review!

examples/transactions.c
```diff
@@ -146,26 +121,21 @@ static void dr_msg_cb (rd_kafka_t *rk,
...
-
-        snprintf(transactional_id, sizeof(transactional_id),
-                 "librdkafka_transactions_example_%s-%d",
-                 rktpar->topic, rktpar->partition);
 
         if (rd_kafka_conf_set(conf, "bootstrap.servers", brokers,
                               errstr, sizeof(errstr)) != RD_KAFKA_CONF_OK ||
-            rd_kafka_conf_set(conf, "transactional.id", transactional_id,
+            rd_kafka_conf_set(conf, "transactional.id",
+                              "librdkafka_transactions_example",
```
mhowlett, line 136: maybe have a clientid command line arg and construct the transactional.id from that? else you can't run more than one instance of the example app at once to spread the load.

examples/transactions.c
```diff
@@ -440,7 +419,8 @@ consumer_group_rebalance_cb (rd_kafka_t *rk,
  * @brief Create the input consumer.
  */
 static rd_kafka_t *create_input_consumer (const char *brokers,
```
mhowlett, line 421: almost tautological

examples/transactions.c
```diff
@@ -321,112 +293,119 @@ static void commit_transaction_and_start_new (struct state *state) {
...
+                                            error);
+                } else {
+                        fprintf(stdout,
+                                "Consumer group rebalanced: %d partition(s) "
+                                "revoked: committing current transaction\n",
+                                partitions->cnt);
 
-                        destroy_transactional_producer(state->producer);
-                        free(state);
+                        commit_transaction(consumer, producer);
```
mhowlett, line 396: don't you need to send offsets to transaction here as well?

examples/transactions.c
edenhill: this is a redundant comment

examples/transactions.c
edenhill: commit_transaction() does that

mhowlett: ahh, whoops sorry

PR #3197, opened 2021-01-04
Revert transaction.timeout.ms change

edenhill:
Turns out the KIP-447 recommendation to reduce transaction.timeout.ms is no longer true.


CHANGELOG.md
```diff
@@ -29,11 +29,6 @@ librdkafka v1.6.0 is feature release:
    supported on Apache Kafka 2.5 or later, on earlier releases you will
    need to use one producer per input partition for EOS. This limitation
    is not enforced by the producer or broker.
- * The default for the producer configuration property
```
mhowlett, line 32: i assume this change is with good reason, but i guess i should check what that is since it apparently deviates from the java implementation.

mhowlett (approved):

PR #3199, opened 2021-01-05
add FreeBSD variant of pthread_setname_np

snar:
FreeBSD uses a bit different variant of pthread_set_name_np, this patch
allows use of this function.


packaging/cmake/try_compile/pthread_setname_freebsd_test.c
```diff
@@ -0,0 +1,7 @@
+#include <pthread.h>
+#include <pthread_np.h>
+
+int main() {
+  pthread_set_name_np("abc");
```
edenhill, line 5: Seems like it is missing the threadid argument

edenhill (changes requested):

packaging/cmake/try_compile/pthread_setname_freebsd_test.c
snar: Corrected

edenhill (approved):

edenhill:
Thank you!

PR #3206, opened 2021-01-08
Include sys/wait.h for OpenBSD

rettakjak:
Like FreeBSD, OpenBSD needs to include `sys/wait.h` for macros like `WIFSIGNALED`.


rettakjak:
`sys/wait.h` needs to be included when on OpenBSD in order to build librdkafka.  I see that a CI build failed though.  Looking at the failure and at the failures in other PRs, I think this is a typical sort of failure.  Is that correct?

rettakjak:
@edenhill friendly ping

edenhill:
Thank you!

PR #3220, opened 2021-01-19
Check for rd_kafka_snappy_init_env_sg failure

mamapanda:
I saw this and thought it might be more robust to check `rd_kafka_snappy_init_env_sg` for errors here before proceeding.


edenhill:
Did you see this fail?
The only reason they would fail is the process has run out of virtual memory space, in which case it is unlikely that the log will be able to allocate its necessary memory and everything will crash.

mamapanda:
I did not. I happened to come across this for a different reason and thought it should've been checked. My bad if I was mistaken!

PR #3227, opened 2021-01-22
Transactional producer fixes

edenhill:


edenhill:
The commit log is probably useful to understand the changes.

CHANGELOG.md
```diff
@@ -29,6 +30,17 @@ librdkafka v1.6.0 is feature release:
...
+   the **Transactional Producer fixes** below for more information.
+
+
+## Known issues
+
+ * The Transactional Producer's API timeout handling is inconsistent with the
+   underlying protocol requests, it is therefore strongly recommended that
+   applications call `rd_kafka_commit_transaction()` and
+   `rd_kafka_abort_transaction()` with the `timeout_ms` parameter
+   set to `-1`, which will use the remaining transaction timeout.
```
mhowlett, line 43: when would someone ever want a different timeout?
is there are reason to give people a way to shoot themselves in the foot?
i guess backwards-compatibility?
maybe there is an upgrade path whereby in 1.6 you get a warning if you provide a timeout != -1 and then in 2.0 you break it.

src/rdkafka_request.c
```diff
@@ -4473,12 +4473,11 @@ rd_kafka_EndTxnRequest (rd_kafka_broker_t *rkb,
...
         /* Committed */
         rd_kafka_buf_write_bool(rkbuf, committed);
-
+        rkbuf->rkbuf_u.EndTxn.commit = committed;
 
         rd_kafka_buf_ApiVersion_set(rkbuf, ApiVersion, 0);
 
-        /* Let the handler perform retries */
-        rkbuf->rkbuf_max_retries = RD_KAFKA_REQUEST_NO_RETRIES;
+        rkbuf->rkbuf_max_retries = RD_KAFKA_REQUEST_MAX_RETRIES;
```
mhowlett, line 4480: a couple of other txn requests set this to 3, wondering if there's a good reason for that.

mhowlett (approved):
tests look good

README.md
```diff
@@ -13,6 +13,11 @@ the producer and 3 million msgs/second for the consumer.
 
 **librdkafka** is licensed under the 2-clause BSD license.
 
+KAFKA is a registered trademark of The Apache Software Foundation and
+has been licensed for use by librdkafka. librdkafka has no
+affiliation with and is not endorsed by The Apache Software Foundation.
```
mhowlett, line 18: as per my PRs on the bindings, i'd be inclined to make a new section "License" further down and put this text and the license in that. i think you want higher impact stuff at the top.

CHANGELOG.md
edenhill: These calls are supposed to be continuable on timeout so that an app that does not want to block for too long can call it repeatedly in a loop, that is not currently the case though and the transaction is set to an abortable state on timeout. We should fix the retriable/continuable thing which will also allow any timeout.

src/rdkafka_request.c
edenhill: I think they're all NO or MAX retries now, shouldn't be any 3:ers left, where you seeing them?

PR #3235, opened 2021-01-28
sticky_assignor: Fix incorrect access of map entry

hallfox:
The balance function accesses a key (typed as a `const char*`) instead of
the likely intended value, based on the type actually being casted. This
seems to mean that initializing is always set to false, and it is
causing a unaligned memory access on Solaris 64-bit builds.


edenhill (approved):
Good find, thanks!

Issue #3139, opened 2020-11-10
High CPU usage when the local socket used to wake the broker thread for produce is disconnected


ajbarb:
Description
===========
I was investigating a High CPU usage on the client running on windows. I found out that WSAPoll was returning immediately and the broker threads were continuously spinning. On looking at a heap dump I found out that the second socket on the fd list of rktrans_pfd has revents set to 3.

0:098> dx rktrans->rktrans_pfd[1]
rktrans->rktrans_pfd[1]                 [Type: pollfd]
    [+0x000] fd               : 0xbe0 [Type: unsigned __int64]
    [+0x008] events           : 768 [Type: short]
    [+0x00a] revents          : 3 [Type: short]

For some reason it seems like the local socket used for event based thread wake went bad. This causes WSAPoll to return immediately and the broker thread spins.

How to reproduce
================
1. Run the librdkafka_example project
2. Observe that CPU usage is low.
3. Use currports to kill the local tcp sockets.
4. CPU usage jumps.
![image](https://user-images.githubusercontent.com/1256612/98723487-4dcf1800-2347-11eb-8f24-ff2f46c4448a.png)

5. Before killing the connection, WSAPoll seems to be working fine but after killing the connection, WSAPoll return immediately.

Checklist
=========

Please provide the following information:

 - [x] librdkafka version: v1.4.4 (Tried it with the latest version too and hit the same issue)
 - [ ] Apache Kafka version: `<REPLACE with e.g., 0.10.2.3>`
 - [x] librdkafka client configuration: `queue.buffering.backpressure.threshold=1000000, queue.buffering.max.ms=1000, linger.ms=1000
 - [x] Operating system: Windows 10.0.19042.572
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [x] Critical issue

edenhill:
revents 0x3 is POLLIN|RDNORM on Windows, which would be raised if the socket was closed, but we're assuming the internal queue sockets are not closed.

Does this happen spontaneously or only when you forcefully kill these connections?

ajbarb:
This happens spontaneously in our deployments. I was able to capture a dump of this in action and then tried to reproduce this locally. I was able to get the same effect after forcefully killing the internal queue socket in my local repro.

edenhill:
Guess my Intellisense lied to me.  HUP|ERR does make more sense in this case.


[...]

PR #3236, opened 2021-01-29
Add loopback sock resiliency

ajbarb:
This change adds recovery action for rebuilding the loopback sockets if they get into bad state. Resolve issue #3139.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -272,6 +272,74 @@ int16_t rd_kafka_broker_ApiVersion_supported (rd_kafka_broker_t *rkb,
                 return maxver;
 }
 
+/**
+ * @brief Setup the wake fd for IO events
+ *
+ * @locality broker creation or reconnection
+ * @locks none
+ */
+static void rd_kafka_broker_setup_queue_wakeup_fd(rd_kafka_broker_t* rkb) {
+    int r;
+
+    /*
+     * Fd-based queue wake-ups using a non-blocking pipe.
+     * Writes are best effort, if the socket queue is full
+     * the write fails (silently) but this has no effect on latency
+     * since the POLLIN flag will already have been raised for fd.
+     */
+    rkb->rkb_wakeup_fd[0] = -1;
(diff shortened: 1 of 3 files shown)


ajbarb:
@edenhill, any chance this can be added to current milestone? I will be happy to address any changes needed asap.

edenhill:
Do we know why these localhost connections are being closed in the first place?

ajbarb:
> Do we know why these localhost connections are being closed in the first place?

No I tried looking at the event logs and couldn't find anything. One other way to debug this was to enable the TCP/IP ETW traces on a port level since they could be huge. Since the ports are dynamic here I couldn't pursue it. At this point the adding resiliency made more sense.

edenhill:
I don't agree, without finding the root cause we don't know if this is the best, or even proper solution to the problem.

As the fix involves state in the fast-path we'll want to be careful and not jump to a seemingly ok fix without knowing if that's really the case.

ajbarb:
> I don't agree, without finding the root cause we don't know if this is the best, or even proper solution to the problem.
> 
> As the fix involves state in the fast-path we'll want to be careful and not jump to a seemingly ok fix without knowing if that's really the case.

I see. Do you have any recommendation on if something could be done about the revents that are not being handled today:
rktrans->rktrans_pfd[1].revents  // POLLERR and POLLHUP

I feel it doesn't seem right to let the poll spin the way it is doing irrespective of root cause since this code is vulnerable to this missing events. What is an okay action here ignoring the recovery part that I tried.

edenhill:
I think we should close the socket on POLLHUP for sure, and log an error to help people troubleshoot it.
But I'd rather not see us add the re-setup code until we understand the issue in more detail.

ajbarb:
> I think we should close the socket on POLLHUP for sure, and log an error to help people troubleshoot it.
> But I'd rather not see us add the re-setup code until we understand the issue in more detail.

Sounds good. Let me explore closing the socket and logging an error. Will do that in a separate PR. Thank you!

Issue #3166, opened 2020-12-04
Exponential backoff not honored by fast leader query


ajbarb:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
Librdkafka doesn't enforce exponential back-off during fast leader query. At max the backoff is 2*topic.metadata.refresh.fast.interval.ms. The issue seems to be in the rd_kafka_metadata_leader_query_tmr_cb where the backoff value passed is rtmr_interval and rd_kafka_timer_backoff calls the rd_kafka_timer_schedule which updates the rtmr_next to rtmr_interval + rtmr_interval.

Shouldn't the new rtmr_interval get doubled every time we try to backoff in rd_kafka_metadata_leader_query_tmr_cb? Something like this:

                    rtmr->rtmr_interval = rtmr->rtmr_interval * 2;
                    rd_kafka_timer_backoff(rkts, rtmr,
                        (int)rtmr->rtmr_interval);

How to reproduce
================
* Fail the rd_kafka_fetch_reply_handle with error RD_KAFKA_RESP_ERR_NOT_LEADER_FOR_PARTITION continously
* Run the rdkafka_example with:
    conf->set("topic.metadata.refresh.fast.interval.ms", "2000", errstr);
    conf->set("topic.metadata.refresh.interval.ms", "32000", errstr);
* Notice that the max delay between fast fetch request is 4sec.

**IMPORTANT**: Always try to reproduce the issue on the latest released version (see https://github.com/edenhill/librdkafka/releases), if it can't be reproduced on the latest version the issue has been fixed.

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): 1.5.0
 - [ ] Apache Kafka version: `<REPLACE with e.g., 0.10.2.3>`
 - [x] librdkafka client configuration: topic.metadata.refresh.fast.interval.ms=2000, topic.metadata.refresh.interval.ms=32000
 - [x] Operating system: Windows 10
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

ajbarb:
@edenhill, is my understanding correct?

larry-cdn77:
I am also seeing this issue in version 1.8.2

The pro

PR #3237, opened 2021-01-29
Increment the backoff exponentially

ajbarb:
Address issue #3166

### DIFF
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -1234,7 +1234,7 @@ static void rd_kafka_metadata_leader_query_tmr_cb (rd_kafka_timers_t *rkts,
                     rk->rk_conf.metadata_refresh_interval_ms)
                         rd_kafka_timer_stop(rkts, rtmr, 1/*lock*/);
                 else
-                    rd_kafka_timer_exp_backoff(rkts, rtmr);
+                        rd_kafka_timer_exp_backoff(rkts, rtmr);
         }
 
         rd_list_destroy(&topics);
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -1233,11 +1233,8 @@ static void rd_kafka_metadata_leader_query_tmr_cb (rd_kafka_timers_t *rkts,
                     rtmr->rtmr_interval * 2 / 1000 >=
                     rk->rk_conf.metadata_refresh_interval_ms)
                         rd_kafka_timer_stop(rkts, rtmr, 1/*lock*/);
-                else {
-                        rtmr->rtmr_interval = rtmr->rtmr_interval * 2;
-                        rd_kafka_timer_backoff(rkts, rtmr,
-                                               (int)rtmr->rtmr_interval);
-                }
+                else
+                    rd_kafka_timer_exp_backoff(rkts, rtmr);
         }
 
         rd_list_destroy(&topics);
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -1234,8 +1234,11 @@ static void rd_kafka_metadata_leader_query_tmr_cb (rd_kafka_timers_t *rkts,
                     rk->rk_conf.metadata_refresh_interval_ms)
                         rd_kafka_timer_stop(rkts, rtmr, 1/*lock*/);
                 else
+                {
+                        rtmr->rtmr_interval = rtmr->rtmr_interval * 2;
                         rd_kafka_timer_backoff(rkts, rtmr,
                                                (int)rtmr->rtmr_interval);
+                }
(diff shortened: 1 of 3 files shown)


src/rdkafka_metadata.c
```diff
@@ -1234,8 +1234,11 @@ static void rd_kafka_metadata_leader_query_tmr_cb (rd_kafka_timers_t *rkts,
                     rk->rk_conf.metadata_refresh_interval_ms)
                         rd_kafka_timer_stop(rkts, rtmr, 1/*lock*/);
                 else
+                {
```
edenhill, line 1237: put open-brace on same line as else:
`else {`

edenhill (changes requested):

src/rdkafka_metadata.c
ajbarb: Done.

src/rdkafka_metadata.c
```diff
@@ -1233,9 +1233,11 @@ static void rd_kafka_metadata_leader_query_tmr_cb (rd_kafka_timers_t *rkts,
                     rtmr->rtmr_interval * 2 / 1000 >=
                     rk->rk_conf.metadata_refresh_interval_ms)
                         rd_kafka_timer_stop(rkts, rtmr, 1/*lock*/);
-                else
+                else {
+                        rtmr->rtmr_interval = rtmr->rtmr_interval * 2;
```
edenhill, line 1237: Sorry for missing this the last review round, but I think a better way is pass the interval to backoff:
`..backoff(rkts, rtmr, (int)rtmr->rtmr_interval * 2)`

(and then remove the added { .. } since they're no longer needed.

ajbarb: Will that work? Won't it end being the same problem were the backoff is only rtmr->rtmr_interval * 2 always? I don't see the rtmr_interval getting updated in rd_kafka_timer_backoff.

edenhill: You're right, what I think we'll want to do is:
 1. rename the function to rd_kafka_exp_backoff()
 2.  remove the backoff_us argument
 3. in exp_backoff() if the timer is scheduled update rtmr_interval as you suggested, e.g., something like:
```c
void rd_kafka_timer_exp_backoff (rd_kafka_timers_t *rkts, rd_kafka_timer_t *rtmr {
	rd_kafka_timers_lock(rkts);
	if (rd_kafka_timer_scheduled(rtmr)) {
                rtmr->rtmr_interval *= 2;
		rd_kafka_timer_unschedule(rkts, rtmr);
        }
	rd_kafka_timer_schedule(rkts, rtmr, rtmr->rtmr_interval);
	rd_kafka_timers_unlock(rkts);
}
```

ajbarb: Sounds good. I addressed the comments. Thanks!

src/rdkafka_metadata.c
```diff
@@ -1234,8 +1234,7 @@ static void rd_kafka_metadata_leader_query_tmr_cb (rd_kafka_timers_t *rkts,
                     rk->rk_conf.metadata_refresh_interval_ms)
                         rd_kafka_timer_stop(rkts, rtmr, 1/*lock*/);
                 else
-                        rd_kafka_timer_backoff(rkts, rtmr,
-                                               (int)rtmr->rtmr_interval);
+                    rd_kafka_timer_exp_backoff(rkts, rtmr);
```
edenhill, line 1237: Use 8 whitespace indent

src/rdkafka_timer.c
```diff
@@ -168,15 +168,17 @@ void rd_kafka_timer_start0 (rd_kafka_timers_t *rkts,
...
-void rd_kafka_timer_backoff (rd_kafka_timers_t *rkts,
-			     rd_kafka_timer_t *rtmr, int backoff_us) {
-	rd_kafka_timers_lock(rkts);
-	if (rd_kafka_timer_scheduled(rtmr))
-		rd_kafka_timer_unschedule(rkts, rtmr);
-	rd_kafka_timer_schedule(rkts, rtmr, backoff_us);
-	rd_kafka_timers_unlock(rkts);
+void rd_kafka_timer_exp_backoff (rd_kafka_timers_t *rkts,
+                 rd_kafka_timer_t *rtmr) {
+    rd_kafka_timers_lock(rkts);
```
edenhill, line 175: Use 8 whitespace indent

edenhill (changes requested):
Just some style changes needed.

src/rdkafka_metadata.c
ajbarb: Sorry missed that. Addressed it.

edenhill:
Thank you!

PR #3240, opened 2021-01-31
Fix PVS Studio (part-1)

pavel-pimenov:
https://github.com/edenhill/librdkafka/issues/3190


edenhill:
Most of these are false positives in the current code's calling conventions, and others were fixed in a recent PR.

PR #3249, opened 2021-02-07
Set WITH_STATIC_LIB_libcrypto appropriately in CMake build system

benesch:
This define is set by the mklove build system when linking against a
static copy of OpenSSL and controls some important SSL initialization
behavior. This commit teaches the CMake build system to do the same.


edenhill:
Thank you!

PR #3256, opened 2021-02-10
ESC-4410: Partitions stuck in PEND_TXN state if multiple AddPartitionsToTxn reqs are sent

edenhill:
This also fixes an issue with partition list not remaining sorted in these cases, which could affect subsequent AddPartitionsToTxn reqs.

### DIFF
--- a/src/rdkafka_metadata_cache.c
+++ b/src/rdkafka_metadata_cache.c
@@ -194,7 +194,7 @@ int rd_kafka_metadata_partition_id_cmp (const void *_a,
  *
  * This makes a copy of \p topic
  *
- * @locks rd_kafka_wrlock()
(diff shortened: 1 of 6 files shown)


src/rdkafka_metadata_cache.c
```diff
@@ -194,7 +194,7 @@ int rd_kafka_metadata_partition_id_cmp (const void *_a,
  *
  * This makes a copy of \p topic
  *
- * @locks rd_kafka_wrlock()
+ * @locks_required rd_kafka_wrlock()
```
mhowlett, line 197: it would probably be worth doing a PR that changed the terminology `@locks` to `@locks_required` everywhere. simple and makes it a little bit more readable.

mhowlett (approved):

PR #3259, opened 2021-02-11
Port to IBMi OS400

AlexeiBaranov:
Port to IBM i OS400

The proposed changes are required to be able to build, test and use librdkafka on the native IBM i operating system - OS400. 

All necessary changes are isolated using #ifndef/ifdef with predefined IBM i macro __OS400__


edenhill:
Interesting PR.

I will not have time to review this in detail yet, but from a short skim..:
 * Move the os400 directory to packaging/os400
 *  There's got to be a better way to add that char-pragma than to modify every single source file. Is there a compiler flag to icc? Or is it sufficient to add it to rdposix.h?
 *  Maintain the librdkafka code style, see https://github.com/edenhill/librdkafka/blob/master/CONTRIBUTING.md

This is still a niche platform so if the changes become too intrusive we'll probably hold off on merging in which case a fork might be a better option. But we'll see when we get closer to the finish line.

Also, this port will not be officially supported since we don't have access to, or knowledge of, IBMi OS400.

AlexeiBaranov:
- [x] Move the os400 directory to packaging/os400
**done**
- [ ] There's got to be a better way to add that char-pragma than to modify every single source file. Is there a compiler flag to icc? Or is it sufficient to add it to rdposix.h?
**Unfortunately this ugly way is the only one.** icc does not process correctly the encoding parameter from the command line, and can use the encoding that set at the very beginning of the *.c file. Even if you change the encoding in the included *.h file, this setting will be valid only until the end of this *.h file
 - [x] Maintain the librdkafka code style, see https://github.com/edenhill/librdkafka/blob/master/CONTRIBUTING.md
 **More or less done, updated sources are committed**

I can reduce _"ifdef hell"_ a bit. In some places, I had to move to #ifdef some code fragments, where it was enough to make slightly change, for example:
`TEST_ASSERT(rkev);` - icc can't compile this line - it think that it is required
`TEST_ASSERT(rkev, "");` - compiled by icc without error
Obviously, such a change will not affect the existing code. But for now, I put it in a separate #ifdef for clarity. 
The same thing with static char buffers - they could be extended in the master code to fit strings produced by OS400 functions like snprintf("...%p...")

> This is still a niche platform so if the changes become too intrusive we'll probably hold off on merging in which case a fork might be a better option. But we'll see when we get closer to the finish line.
> 
> Also, this port will not be officially supported since we don't have access to, or knowledge of, IBMi OS400.

I understand and agree.

Issue #3257, opened 2021-02-11
Minor bug fixes in transactions example


sanjay24:
Description
===========
Few issues with transactions code
- consume rebalance on partition revoke should check if the assignments are lost for ti to abort and commit on partition revocation, it does the opposite
- rewind consumer doesn't iterate over all the partitions correctly

How to reproduce
================
run transactions example

=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `1.6.0`
 - [x] Apache Kafka version: `2.6.0`
 - [ ] librdkafka client configuration: `<REPLACE with e.g., message.timeout.ms=123, auto.reset.offset=earliest, ..>`
 - [ ] Operating system: `<REPLACE with e.g., Centos 5 (x64)>`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

PR #3260, opened 2021-02-12
tansactions example: do not use consumer handle for txn api (#3257)

sanjay24:
rd_kafka_begin_transaction called with consumer handle, should be producer instead


edenhill:
Ouch, thanks!

PR #3269, opened 2021-02-18
Clean buffer after rebalancing for batch queue

jliunyu:
Summary: This is bug fix for https://github.com/confluentinc/confluent-kafka-python/issues/1013

Issue: Buffer is not cleaned after rebalance if messages are polled using the batch queue method, so the consumer will still get old messages.

Solution: when assign happens, a new op event with type RD_KAFKA_OP_BARRIER will be created, a new version is been created at mean time. If the consumer met this event, will clean the buffer by comparing the version of msgs and the new version just created.

Test passed in local test:
Testing:
[0122_buffer_cleaning_after_rebalance/ 62.283s] ================= Test 0122_buffer_cleaning_after_rebalance PASSED =================
[<MAIN>                      / 63.161s] ALL-TESTS: duration 63160.668ms
TEST 20210224223610 (bare, scenario default) SUMMARY
| <MAIN>                                   |     PASSED |  63.161s |
| 0122_buffer_cleaning_after_rebalance     |     PASSED |  62.283s |
[<MAIN>                      / 63.163s] 0 thread(s) in use by librdkafka
[<MAIN>                      / 63.163s] 
###
###

### DIFF
--- a/src/rdkafka_op.c
+++ b/src/rdkafka_op.c
@@ -238,7 +238,7 @@ rd_kafka_op_t *rd_kafka_op_new0 (const char *source, rd_kafka_op_type_t type) {
                 [RD_KAFKA_OP_GET_REBALANCE_PROTOCOL] =
                 sizeof(rko->rko_u.rebalance_protocol),
                 [RD_KAFKA_OP_LEADERS] = sizeof(rko->rko_u.leaders),
-                [RD_KAFKA_OP_BARRIER] = sizeof(rko->rko_u.barrier),
+                [RD_KAFKA_OP_BARRIER] = sizeof(rko->rko_version),
         };
         size_t tsize = op2size[type & ~RD_KAFKA_OP_FLAGMASK];
 
@@ -412,6 +412,7 @@ void rd_kafka_op_destroy (rd_kafka_op_t *rko) {
                 RD_IF_FREE(rko->rko_u.leaders.partitions,
                            rd_kafka_topic_partition_list_destroy);
                 break;
+
 	default:
 		break;
 	}
--- a/src/rdkafka_op.c
+++ b/src/rdkafka_op.c
@@ -238,7 +238,7 @@ rd_kafka_op_t *rd_kafka_op_new0 (const char *source, rd_kafka_op_type_t type) {
                 [RD_KAFKA_OP_GET_REBALANCE_PROTOCOL] =
                 sizeof(rko->rko_u.rebalance_protocol),
                 [RD_KAFKA_OP_LEADERS] = sizeof(rko->rko_u.leaders),
-                [RD_KAFKA_OP_BARRIER] = sizeof(_RD_KAFKA_OP_EMPTY),
+                [RD_KAFKA_OP_BARRIER] = _RD_KAFKA_OP_EMPTY,
         };
         size_t tsize = op2size[type & ~RD_KAFKA_OP_FLAGMASK];
 
--- a/src/rdkafka_op.c
+++ b/src/rdkafka_op.c
@@ -238,7 +238,7 @@ rd_kafka_op_t *rd_kafka_op_new0 (const char *source, rd_kafka_op_type_t type) {
                 [RD_KAFKA_OP_GET_REBALANCE_PROTOCOL] =
                 sizeof(rko->rko_u.rebalance_protocol),
                 [RD_KAFKA_OP_LEADERS] = sizeof(rko->rko_u.leaders),
-                [RD_KAFKA_OP_BARRIER] = sizeof(rko->rko_version),
+                [RD_KAFKA_OP_BARRIER] = sizeof(_RD_KAFKA_OP_EMPTY),
         };
         size_t tsize = op2size[type & ~RD_KAFKA_OP_FLAGMASK];
 
--- a/src/rdkafka_op.c
+++ b/src/rdkafka_op.c
@@ -413,7 +413,7 @@ void rd_kafka_op_destroy (rd_kafka_op_t *rko) {
                            rd_kafka_topic_partition_list_destroy);
                 break;
         case RD_KAFKA_OP_BARRIER:
-            break;
+                break;
 	default:
 		break;
 	}
--- a/src/rdkafka_op.c
+++ b/src/rdkafka_op.c
@@ -95,6 +95,7 @@ const char *rd_kafka_op2str (rd_kafka_op_type_t type) {
                 [RD_KAFKA_OP_GET_REBALANCE_PROTOCOL] =
                 "REPLY:GET_REBALANCE_PROTOCOL",
                 [RD_KAFKA_OP_LEADERS] = "REPLY:LEADERS",
+                [RD_KAFKA_OP_BARRIER] = "REPLY:BARRIER",
         };
 
         if (type & RD_KAFKA_OP_REPLY)
@@ -237,6 +238,7 @@ rd_kafka_op_t *rd_kafka_op_new0 (const char *source, rd_kafka_op_type_t type) {
                 [RD_KAFKA_OP_GET_REBALANCE_PROTOCOL] =
                 sizeof(rko->rko_u.rebalance_protocol),
                 [RD_KAFKA_OP_LEADERS] = sizeof(rko->rko_u.leaders),
+                [RD_KAFKA_OP_BARRIER] = sizeof(rko->rko_u.barrier),
         };
         size_t tsize = op2size[type & ~RD_KAFKA_OP_FLAGMASK];
 
@@ -410,7 +412,8 @@ void rd_kafka_op_destroy (rd_kafka_op_t *rko) {
                 RD_IF_FREE(rko->rko_u.leaders.partitions,
                            rd_kafka_topic_partition_list_destroy);
                 break;
-
+        case RD_KAFKA_OP_BARRIER:
+            break;
 	default:
 		break;
 	}
--- a/src/rdkafka_op.c
+++ b/src/rdkafka_op.c
@@ -412,8 +412,6 @@ void rd_kafka_op_destroy (rd_kafka_op_t *rko) {
                 RD_IF_FREE(rko->rko_u.leaders.partitions,
                            rd_kafka_topic_partition_list_destroy);
                 break;
-        case RD_KAFKA_OP_BARRIER:
-                break;
 	default:
 		break;
 	}
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -197,8 +197,8 @@ void rd_kafka_toppar_op_version_bump (rd_kafka_q_t *rktp_fetchq,
         rd_kafka_op_t *rko;
 
         rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
-	    rko->rko_version = version;
-	    rd_kafka_q_enq(rktp_fetchq, rko);
+        rko->rko_version = version;
+        rd_kafka_q_enq(rktp_fetchq, rko);
 }
 
 
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -188,7 +188,7 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
 
 /**
  * @brief Enqueue an RD_KAFKA_OP_BARRIER type of operation
- *         when the op_version is updated.
+ *        when the op_version is updated.
  *
  * @locality Toppar handler thread
  */
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -187,9 +187,11 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
 }
 
 /**
- * @brief Enqueue an RD_KAFKA_OP_BARRIER type of operation
+ * @brief Update rktp_op_version.
+ *        Enqueue an RD_KAFKA_OP_BARRIER type of operation
  *        when the op_version is updated.
  *
+ * @locks_required rd_kafka_toppar_lock() must be held.
  * @locality Toppar handler thread
  */
 void rd_kafka_toppar_op_version_bump (rd_kafka_toppar_t *rktp,
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -186,6 +186,21 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
 	rd_kafka_toppar_consumer_lag_req(rktp);
 }
 
+/**
+ * Enqueue an RD_KAFKA_OP_BARRIER type of operation
+ * when the op_version is updated.
+ *
+ * Locality: toppar handler thread
+ */
+void rd_kafka_toppar_op_version_bump (rd_kafka_q_t *rktp_fetchq,
+		                             int32_t version) {
+        rd_kafka_op_t *rko;
+
+        rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
+	    rko->rko_version = version;
+	    rd_kafka_q_enq(rktp_fetchq, rko);
+}
+
 
 /**
  * Add new partition to topic.
@@ -239,6 +254,7 @@ rd_kafka_toppar_t *rd_kafka_toppar_new0 (rd_kafka_topic_t *rkt,
         rktp->rktp_ops->rkq_opaque = rktp;
         rd_atomic32_init(&rktp->rktp_version, 1);
 	rktp->rktp_op_version = rd_atomic32_get(&rktp->rktp_version);
+        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, rktp->rktp_op_version);
 
         rd_atomic32_init(&rktp->rktp_msgs_inflight, 0);
         rd_kafka_pid_reset(&rktp->rktp_eos.pid);
@@ -1586,6 +1602,7 @@ static void rd_kafka_toppar_fetch_start (rd_kafka_toppar_t *rktp,
         }
 
 	rktp->rktp_op_version = version;
+        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, version);
 
         if (rkcg) {
                 rd_kafka_assert(rktp->rktp_rkt->rkt_rk, !rktp->rktp_cgrp);
@@ -1684,7 +1701,6 @@ void rd_kafka_toppar_fetch_stopped (rd_kafka_toppar_t *rktp,
  */
 void rd_kafka_toppar_fetch_stop (rd_kafka_toppar_t *rktp,
 				 rd_kafka_op_t *rko_orig) {
-		rd_kafka_op_t *rko;
         int32_t version = rko_orig->rko_version;
 
 	rd_kafka_toppar_lock(rktp);
@@ -1696,9 +1712,7 @@ void rd_kafka_toppar_fetch_stop (rd_kafka_toppar_t *rktp,
                      rd_kafka_fetch_states[rktp->rktp_fetch_state], version);
 
 	rktp->rktp_op_version = version;
-	rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
-	rko->rko_version = version;
-	rd_kafka_q_enq(rktp->rktp_fetchq, rko);
+        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, version);
 
 	/* Abort pending offset lookups. */
 	if (rktp->rktp_fetch_state == RD_KAFKA_TOPPAR_FETCH_OFFSET_QUERY)
@@ -1759,6 +1773,7 @@ void rd_kafka_toppar_seek (rd_kafka_toppar_t *rktp,
 	}
 
 	rktp->rktp_op_version = version;
+        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, version);
 
 	/* Abort pending offset lookups. */
 	if (rktp->rktp_fetch_state == RD_KAFKA_TOPPAR_FETCH_OFFSET_QUERY)
@@ -1814,6 +1829,7 @@ static void rd_kafka_toppar_pause_resume (rd_kafka_toppar_t *rktp,
 	rd_kafka_toppar_lock(rktp);
 
 	rktp->rktp_op_version = version;
+        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, version);
 
         if (!pause && (rktp->rktp_flags & flag) != flag) {
                 rd_kafka_dbg(rk, TOPIC, "RESUME",
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -1684,6 +1684,7 @@ void rd_kafka_toppar_fetch_stopped (rd_kafka_toppar_t *rktp,
  */
 void rd_kafka_toppar_fetch_stop (rd_kafka_toppar_t *rktp,
 				 rd_kafka_op_t *rko_orig) {
+		rd_kafka_op_t *rko;
         int32_t version = rko_orig->rko_version;
 
 	rd_kafka_toppar_lock(rktp);
@@ -1695,7 +1696,7 @@ void rd_kafka_toppar_fetch_stop (rd_kafka_toppar_t *rktp,
                      rd_kafka_fetch_states[rktp->rktp_fetch_state], version);
 
 	rktp->rktp_op_version = version;
-	rd_kafka_op_t *rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
+	rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
 	rko->rko_version = version;
 	rd_kafka_q_enq(rktp->rktp_fetchq, rko);
 
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -1695,6 +1695,9 @@ void rd_kafka_toppar_fetch_stop (rd_kafka_toppar_t *rktp,
                      rd_kafka_fetch_states[rktp->rktp_fetch_state], version);
 
 	rktp->rktp_op_version = version;
+	rd_kafka_op_t *rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
+	rko->rko_version = version;
+	rd_kafka_q_enq(rktp->rktp_fetchq, rko);
 
 	/* Abort pending offset lookups. */
 	if (rktp->rktp_fetch_state == RD_KAFKA_TOPPAR_FETCH_OFFSET_QUERY)
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -187,18 +187,19 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
 }
 
 /**
- * Enqueue an RD_KAFKA_OP_BARRIER type of operation
- * when the op_version is updated.
+ * @brief Enqueue an RD_KAFKA_OP_BARRIER type of operation
+ *         when the op_version is updated.
  *
- * Locality: toppar handler thread
+ * @locality Toppar handler thread
  */
-void rd_kafka_toppar_op_version_bump (rd_kafka_q_t *rktp_fetchq,
-		                             int32_t version) {
+void rd_kafka_toppar_op_version_bump (rd_kafka_toppar_t *rktp,
+                                      int32_t version) {
         rd_kafka_op_t *rko;
 
+        rktp->rktp_op_version = version;
         rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
         rko->rko_version = version;
-        rd_kafka_q_enq(rktp_fetchq, rko);
+        rd_kafka_q_enq(rktp->rktp_fetchq, rko);
 }
 
 
@@ -254,7 +255,6 @@ rd_kafka_toppar_t *rd_kafka_toppar_new0 (rd_kafka_topic_t *rkt,
         rktp->rktp_ops->rkq_opaque = rktp;
         rd_atomic32_init(&rktp->rktp_version, 1);
 	rktp->rktp_op_version = rd_atomic32_get(&rktp->rktp_version);
-        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, rktp->rktp_op_version);
 
         rd_atomic32_init(&rktp->rktp_msgs_inflight, 0);
         rd_kafka_pid_reset(&rktp->rktp_eos.pid);
@@ -1601,8 +1601,7 @@ static void rd_kafka_toppar_fetch_start (rd_kafka_toppar_t *rktp,
                 goto err_reply;
         }
 
-	rktp->rktp_op_version = version;
-        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, version);
+        rd_kafka_toppar_op_version_bump(rktp, version);
 
         if (rkcg) {
                 rd_kafka_assert(rktp->rktp_rkt->rkt_rk, !rktp->rktp_cgrp);
@@ -1711,8 +1710,7 @@ void rd_kafka_toppar_fetch_stop (rd_kafka_toppar_t *rktp,
                      rktp->rktp_partition,
                      rd_kafka_fetch_states[rktp->rktp_fetch_state], version);
 
-	rktp->rktp_op_version = version;
-        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, version);
+        rd_kafka_toppar_op_version_bump(rktp, version);
 
 	/* Abort pending offset lookups. */
 	if (rktp->rktp_fetch_state == RD_KAFKA_TOPPAR_FETCH_OFFSET_QUERY)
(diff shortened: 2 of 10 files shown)


src/rdkafka_partition.c
```diff
@@ -1695,6 +1695,9 @@ void rd_kafka_toppar_fetch_stop (rd_kafka_toppar_t *rktp,
                      rd_kafka_fetch_states[rktp->rktp_fetch_state], version);
 
 	rktp->rktp_op_version = version;
+	rd_kafka_op_t *rko = rd_kafka_op_new(RD_KAFKA_OP_BARRIER);
```
edenhill, line 1698: Variable defs must go at top of scope.

src/rdkafka_queue.c
```diff
@@ -517,7 +517,35 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
 }
 
 
+int invalid_all_old_msgs_in_array(rd_kafka_op_t **rkos,
```
edenhill, line 520: Make static, use size_t for counts, use rdkafka naming convention:
`rd_kafka_purge_outdated_messages()`

src/rdkafka_queue.c
```diff
@@ -517,7 +517,35 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
+                                  int32_t version,
+                                  rd_kafka_message_t **rkmessages,
+                                  int cnt,
+                                  size_t rkmessages_size) {
+    TAILQ_HEAD(, rd_kafka_op_s) tmpq = TAILQ_HEAD_INITIALIZER(tmpq);
+    rd_kafka_op_t *rko, *next;
+    rd_kafka_message_t *tmp[rkmessages_size];
+
+    int valid_count = 0;
+    for(int i = 0; i < cnt; i++) {
```
edenhill, line 530: int i needs to be declared at top of scope

src/rdkafka_queue.c
```diff
@@ -517,7 +517,35 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
 
 
+int invalid_all_old_msgs_in_array(rd_kafka_op_t **rkos,
+                                  int32_t version,
+                                  rd_kafka_message_t **rkmessages,
+                                  int cnt,
+                                  size_t rkmessages_size) {
+    TAILQ_HEAD(, rd_kafka_op_s) tmpq = TAILQ_HEAD_INITIALIZER(tmpq);
+    rd_kafka_op_t *rko, *next;
+    rd_kafka_message_t *tmp[rkmessages_size];
```
edenhill, line 527: Variable array sizes is not compatible with older compilers, and is also not stack-size safe, use rd_malloc.

src/rdkafka_queue.c
```diff
@@ -517,7 +517,35 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
+    int valid_count = 0;
+    for(int i = 0; i < cnt; i++) {
+        if(rd_kafka_op_version_outdated(rkos[i], version)) {
+            TAILQ_INSERT_TAIL(&tmpq, rkos[i], rko_link);
+        } else {
+            tmp[valid_count++] = rd_kafka_message_get(rkos[i]);
+        }
+    }
 
+    /* Discard non-desired and already handled ops */
```
edenhill, line 538: Why do we need two passes? Can't the op be destroyed in the loop above?

src/rdkafka_queue.c
```diff
@@ -537,6 +565,7 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
         rd_kafka_t *rk = rkq->rkq_rk;
         rd_kafka_q_t *fwdq;
         struct timespec timeout_tspec;
+        rd_kafka_op_t *rkos[rkmessages_size];
```
edenhill, line 568: Don't think we need to maintain a separate array for the rkos. The rkmessage is part of the rko allocation, and rkmessage->_private points back to the rko, so we can just use that in the purge function.

src/rdkafka_queue.c
```diff
@@ -587,6 +616,10 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
                                        RD_KAFKA_Q_CB_RETURN, NULL);
                 if (res == RD_KAFKA_OP_RES_KEEP ||
                     res == RD_KAFKA_OP_RES_HANDLED) {
+                    if(rko->rko_type == RD_KAFKA_OP_BARRIER) {
+                        // For EAGER rebalancing, clean the whole buffer
```
edenhill, line 620: Use /**/ comments.
The comment is not strictly related to EAGER rebalancing, but rather that the partition version changed.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,227 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test.h"
+#include <pthread.h>
```
edenhill, line 30: pthread is not available on windows, use tinycthread instead, see e.g. 0075-retry.c

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,227 @@
...
+static rd_kafka_t *c1;
+
+typedef struct rd_kafka_consumer_s {
+        char *what;
+        rd_kafka_queue_t *rkq;
+        int timeout_ms;
+        int consumemsgcnt;
+        rd_kafka_t *rk;
+        uint64_t testid;
+} rd_kafka_consumer_t;
```
edenhill, line 44: Avoid the rd_kafka_ prefix in tests, it makes its definition ambigious (is it a test or proper rdkafka thing?).
For private test files you can just use `consumer_t`.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,226 @@
...
+        char *what;
+        rd_kafka_queue_t *rkq;
+        int timeout_ms;
+        int consumemsgcnt;
+        rd_kafka_t *rk;
+        uint64_t testid;
+} consumer_t;
+
+static int test_consumer_batch_queue(void *arguments) {
+	    struct consumer_s *args = arguments;
```
edenhill, line 46: mix of tabs and spaces

edenhill (changes requested):
Good work!
See comments.

src/rdkafka_op.c
```diff
@@ -237,6 +238,7 @@ rd_kafka_op_t *rd_kafka_op_new0 (const char *source, rd_kafka_op_type_t type) {
                 [RD_KAFKA_OP_GET_REBALANCE_PROTOCOL] =
                 sizeof(rko->rko_u.rebalance_protocol),
                 [RD_KAFKA_OP_LEADERS] = sizeof(rko->rko_u.leaders),
+                [RD_KAFKA_OP_BARRIER] = sizeof(rko->rko_version),
```
edenhill, line 241: The size must be the size of its union, and if there is no union it should be RD_KAFKA_OP_EMPTY

src/rdkafka_partition.c
```diff
@@ -186,6 +186,21 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
 	rd_kafka_toppar_consumer_lag_req(rktp);
 }
 
+/**
+ * Enqueue an RD_KAFKA_OP_BARRIER type of operation
```
edenhill, line 190: Use doxygen tags:
```c
/**
 * @brief Enqueue ...
 *
 ...
 * @locality toppar handler thread
 */
```

src/rdkafka_partition.c
```diff
@@ -186,6 +186,21 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
...
+ * when the op_version is updated.
+ *
+ * Locality: toppar handler thread
+ */
+void rd_kafka_toppar_op_version_bump (rd_kafka_q_t *rktp_fetchq,
+		                             int32_t version) {
+        rd_kafka_op_t *rko;
+
```
edenhill, line 198: I think you should move the actual bumping of the op version into this function as well:
```
rktp->rktp_op_version = version;
```

src/rdkafka_partition.c
```diff
@@ -239,6 +254,7 @@ rd_kafka_toppar_t *rd_kafka_toppar_new0 (rd_kafka_topic_t *rkt,
         rktp->rktp_ops->rkq_opaque = rktp;
         rd_atomic32_init(&rktp->rktp_version, 1);
 	rktp->rktp_op_version = rd_atomic32_get(&rktp->rktp_version);
+        rd_kafka_toppar_op_version_bump(rktp->rktp_fetchq, rktp->rktp_op_version);
```
edenhill, line 257: I don't think there's any need for an OP_BARRIER for a newly created rktp, there is nothing to barrier.

src/rdkafka_partition.c
```diff
@@ -186,6 +186,21 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
...
 }
 
+/**
+ * Enqueue an RD_KAFKA_OP_BARRIER type of operation
+ * when the op_version is updated.
+ *
+ * Locality: toppar handler thread
+ */
+void rd_kafka_toppar_op_version_bump (rd_kafka_q_t *rktp_fetchq,
+		                             int32_t version) {
```
edenhill, line 196: indent looks funky

src/rdkafka_queue.c
```diff
@@ -516,8 +516,35 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
 	return cnt;
 }
 
+/**
+ * Filter out and destroy outdated messages
```
edenhill, line 520: `* @brief ..`

src/rdkafka_queue.c
```diff
@@ -516,8 +516,35 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
+static size_t rd_kafka_purge_outdated_messages (int32_t version,
+                                  rd_kafka_message_t **rkmessages,
+                                  int cnt) {
+        rd_kafka_op_t *rko;
+        size_t valid_count = 0;
+        int i;
+
+        for (i = 0; i < cnt; i++) {
+                rko = rkmessages[i]->_private;
+                if(rd_kafka_op_version_outdated(rko, version)) {
```
edenhill, line 535: style nit: `if (`

src/rdkafka_queue.c
```diff
@@ -516,8 +516,35 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
+                        /* This also destroys the corresponding rkmessage. */
+                        rd_kafka_op_destroy(rko);
+                } else {
+                        rkmessages[valid_count++] = rkmessages[i];
+                }
+        }
 
+        for (i = valid_count; i < cnt; i++)
+                rkmessages[i] = NULL;
```
edenhill, line 544: Is this really needed if we return the new valid count?

src/rdkafka_queue.c
```diff
@@ -587,6 +614,10 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
                                        RD_KAFKA_Q_CB_RETURN, NULL);
                 if (res == RD_KAFKA_OP_RES_KEEP ||
                     res == RD_KAFKA_OP_RES_HANDLED) {
+                        if (!rko && unlikely(rko->rko_type == RD_KAFKA_OP_BARRIER)) {
+                                /* Client outdated messages on version barrier bump */
+                                cnt = rd_kafka_purge_outdated_messages(rko->rko_version, rkmessages, cnt);
```
edenhill, line 619: style: keep below 80 cols

src/rdkafka_queue.c
```diff
@@ -587,6 +614,10 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
                                        RD_KAFKA_Q_CB_RETURN, NULL);
                 if (res == RD_KAFKA_OP_RES_KEEP ||
                     res == RD_KAFKA_OP_RES_HANDLED) {
+                        if (!rko && unlikely(rko->rko_type == RD_KAFKA_OP_BARRIER)) {
```
edenhill, line 617: This if-statement is wrong, both checking for NULL and then derferencing that null pointer.
But there is no case where `rko` is NULL here though, so this code block is never reached.
If the new tests pass with this (faulty) code it means the tests don't really cover what it is supposed to test.

src/rdkafka_queue.c
```diff
@@ -587,6 +614,10 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
                                        RD_KAFKA_Q_CB_RETURN, NULL);
                 if (res == RD_KAFKA_OP_RES_KEEP ||
                     res == RD_KAFKA_OP_RES_HANDLED) {
```
edenhill, line 616: The rko will have been destroyed if res == HANDLED, and the line below will thus dereference freed memory.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2012-2015, Magnus Edenhill
```
edenhill, line 4: Let Copyright be just 2021

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test.h"
+/* Typical include path would be <librdkafka/rdkafka.h>, but this program
+ * is built from within the librdkafka source tree and thus differs. */
+#include "rdkafka.h"  /* for Kafka driver */
+
+static const int produce_msg_cnt = 400;
```
edenhill, line 34: Suggest passing this in the consumer_t struct instead of using a global.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
...
+        uint64_t testid = args->testid;
+        const char *what = args->what;
+
+        rd_kafka_message_t **rkmessage = malloc(consume_msg_cnt * sizeof(*rkmessage));
+
+        test_msgver_init(&mv, testid);
+
+        TIMING_START(&t_cons, "CONSUME");
+
+        while ((msg_cnt = rd_kafka_consume_batch_queue(rkq, timeout_ms, rkmessage, consume_msg_cnt)) == 0)
```
edenhill, line 65: style: keep below 80 cols

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
...
+        rd_kafka_t *c2;
+        int p;
+        const int timeout_ms = 30000;
+        uint64_t testid;
+        const int consume_msg_cnt = 500;
+        rd_kafka_conf_t *conf;
+        struct consumer_s c1_args;
+        struct consumer_s c2_args;
+
+        SUB_TEST("partition.assignment.strategy = %s\n", strategy);
```
edenhill, line 102: Remove newline \n

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
...
+
+        test_conf_init(&conf, NULL, 60);
+        test_conf_set(conf, "enable.auto.commit", "false");
+        test_conf_set(conf, "auto.offset.reset", "earliest");
+
+        testid = test_id_generate();
+
+        /* Produce messages */
+        topic = test_mk_topic_name(__FUNCTION__, 1);
```
edenhill, line 111: We'll want the test number in the topic name, so suggest you just do `test_mk_topic_name("0122-buffer_cleaning")

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
...
+
+        for (p = 0 ; p < partition_cnt ; p++)
+                test_produce_msgs_easy(topic,
+                                       testid,
+                                       p,
+                                       produce_msg_cnt / partition_cnt);
+        /* Create simple consumer */
+        if (strcmp(strategy, "cooperative-sticky") == 0)
```
edenhill, line 119: style nit: ` if (!strcmp(...))`

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
...
+typedef struct consumer_s {
+        const char *what;
+        rd_kafka_queue_t *rkq;
+        int timeout_ms;
+        int consume_msg_cnt;
+        rd_kafka_t *rk;
+        uint64_t testid;
+} consumer_t;
+
+static int test_consumer_batch_queue (void *arguments) {
```
edenhill, line 45: Why is this taking a void* arg and now consumer_t*?

tests/test.c
```diff
@@ -432,6 +433,7 @@ struct test tests[] = {
         _TEST(0119_consumer_auth, 0, TEST_BRKVER(2,1,0,0)),
         _TEST(0120_asymmetric_subscription, TEST_F_LOCAL),
         _TEST(0121_clusterid, TEST_F_LOCAL),
+		_TEST(0122_buffer_cleaning_after_rebalance, TEST_BRKVER(2,4,0,0)),
```
edenhill, line 436: use whitespace indent

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,172 @@
...
+                if (test_msgver_add_msg(rk, &mv, rkmessage[i]) == 0)
+                        TEST_FAIL("The message is not from testid %"PRId64" \n", testid);
+        }
+
+        test_msgver_verify(what, &mv, TEST_MSGVER_ORDER|TEST_MSGVER_DUP, 0, produce_msg_cnt/2);
+        test_msgver_clear(&mv);
+
+        for (i = 0; i < consume_msg_cnt; i++)
+                rd_kafka_message_destroy(rkmessage[i]);
+
```
edenhill, line 78: rkmessage is being leaked here.
Make sure to run tests with valgrind or asan

edenhill (changes requested):
Almost there! Good work!

src/rdkafka_op.c
```diff
@@ -237,6 +238,7 @@ rd_kafka_op_t *rd_kafka_op_new0 (const char *source, rd_kafka_op_type_t type) {
                 [RD_KAFKA_OP_GET_REBALANCE_PROTOCOL] =
                 sizeof(rko->rko_u.rebalance_protocol),
                 [RD_KAFKA_OP_LEADERS] = sizeof(rko->rko_u.leaders),
+                [RD_KAFKA_OP_BARRIER] = sizeof(_RD_KAFKA_OP_EMPTY),
```
edenhill, line 241: no sizeof, see existing uses of _RD_KAFKA_OP_EMPTY

src/rdkafka_partition.c
```diff
@@ -186,6 +186,22 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
 	rd_kafka_toppar_consumer_lag_req(rktp);
 }
 
+/**
+ * @brief Enqueue an RD_KAFKA_OP_BARRIER type of operation
```
edenhill, line 190: Should also say that the rktp_op_version is updated

src/rdkafka_partition.c
```diff
@@ -186,6 +186,22 @@ static void rd_kafka_toppar_consumer_lag_tmr_cb (rd_kafka_timers_t *rkts,
...
+ *        when the op_version is updated.
+ *
+ * @locality Toppar handler thread
```
edenhill, line 193: Also specify `@locks_required` (since rktp_op_version is written I'm guessing we need rd_kafka_toppar_lock()?)

src/rdkafka_queue.c
```diff
@@ -516,8 +516,32 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
+ * @brief Filter out and destroy outdated messages.
+ *
+ * @returns Returns the number of valid messages.
+ *
+ * @locality Any thread.
+ */
+static size_t rd_kafka_purge_outdated_messages (int32_t version,
+                                  rd_kafka_message_t **rkmessages,
+                                  int cnt) {
+        rd_kafka_op_t *rko;
```
edenhill, line 529: move this to the scope/block where it is used

src/rdkafka_queue.c
```diff
@@ -516,8 +516,32 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
+        size_t valid_count = 0;
+        int i;
+
+        for (i = 0; i < cnt; i++) {
+                rko = rkmessages[i]->_private;
+                if (rd_kafka_op_version_outdated(rko, version)) {
+                        /* This also destroys the corresponding rkmessage. */
+                        rd_kafka_op_destroy(rko);
+                } else {
```
edenhill, line 538: Perhaps as an optimization to avoid a write shuffle:  `else if (i > valid_count++)`

src/rdkafka_queue.c
```diff
@@ -582,9 +606,17 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
                         continue;
                 }
 
+                if (unlikely(rko->rko_type == RD_KAFKA_OP_BARRIER)) {
+                        cnt = rd_kafka_purge_outdated_messages(
+                                rko->rko_version,
+                                rkmessages,
+                                cnt);
+                }
```
edenhill, line 614: Should be safe to destroy the rko and continue here, yeah?

src/rdkafka_queue.c
```diff
@@ -582,9 +606,17 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
...
+
                 /* Serve non-FETCH callbacks */
                 res = rd_kafka_poll_cb(rk, rkq, rko,
                                        RD_KAFKA_Q_CB_RETURN, NULL);
+
```
edenhill, line 619: nit: remove this inserted blank line

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,179 @@
...
+        rd_kafka_message_t **rkmessage = malloc(consume_msg_cnt * sizeof(*rkmessage));
+
+        test_msgver_init(&mv, testid);
+
+        TIMING_START(&t_cons, "CONSUME");
+
+        while ((msg_cnt = rd_kafka_consume_batch_queue(rkq,
+                timeout_ms, rkmessage, consume_msg_cnt)) == 0)
+                continue;
```
edenhill, line 66: The continue is not needed, a single ; is fine.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,179 @@
...
+        /* Produce messages */
+        topic = test_mk_topic_name("0122-buffer_cleaning", 1);
+
+        for (p = 0 ; p < partition_cnt ; p++)
+                test_produce_msgs_easy(topic,
+                                       testid,
+                                       p,
+                                       produce_msg_cnt / partition_cnt);
+        /* Create simple consumer */
+        if (!strcmp(strategy, "cooperative-sticky"))
```
edenhill, line 122: Suggest removing the strcmp and simply pass strategy to conf_set..

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,179 @@
...
+
+        /* Produce messages */
+        topic = test_mk_topic_name("0122-buffer_cleaning", 1);
+
+        for (p = 0 ; p < partition_cnt ; p++)
+                test_produce_msgs_easy(topic,
+                                       testid,
+                                       p,
+                                       produce_msg_cnt / partition_cnt);
+        /* Create simple consumer */
```
edenhill, line 121: A "simple consumer" is a low-level non-balanced consumer, but we're creating a high-level balanced consumer here so change to just `Create consumers`

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,226 @@
...
+        if(strcmp(strategy, "cooperative-sticky") == 0)
+                test_conf_set(conf, "partition.assignment.strategy", "cooperative-sticky");
+
+        rd_kafka_conf_set_rebalance_cb(conf, rebalance_cb);
+
+        c1 = test_create_consumer(topic, rebalance_cb,
+                                  rd_kafka_conf_dup(conf), NULL);
+        c2 = test_create_consumer(topic, NULL, conf, NULL);
+
+        test_consumer_subscribe(c1, topic);
```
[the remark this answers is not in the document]
edenhill: Even with the updated code I think there's a problem:
1. c1 subscribes
2. wait for c1 assignment
3. c2 subscribes
4. ..rebalance
5. wait for c2 assignment
6. then consume c1 and c2

Problem is that since no messages are beginning to be batched in a consume_batch..() call there is no messages to clear out when the barrier is received, those messages will already be purged by "outdated()" further up in the q_serve loop.

Did you attempt to run this test-case with a non-fixed version of librdkafka to verify that the test-case triggers the original bug?

edenhill (changes requested):
Good stuff!

I'm a bit concerned with the test, I don't think it actually triggers/verifies the buggy behaviour.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -41,7 +41,8 @@ typedef struct consumer_s {
         int produce_msg_cnt;
 } consumer_t;
 
-static int test_consumer_batch_queue (consumer_t *arguments) {
+static int test_consumer_batch_queue (void *arg) {
+        consumer_t *arguments = arg;
```
jliunyu, line 45: Convert void *arg to consumer_t *arguments to handle build warning

src/rdkafka_queue.c
jliunyu: Hi Magnus, thanks for pointing this out! I think copy should before valid_count + 1, so I change this part to the below logic:

if (rd_kafka_op_version_outdated(rko, version)) {
                        /* This also destroys the corresponding rkmessage. */
                        rd_kafka_op_destroy(rko);
} else if ((size_t)i > valid_count) {
                        rkmessages[valid_count++] = rkmessages[i];
} else {
                        valid_count++;
}

src/rdkafka_queue.c
```diff
@@ -516,8 +516,33 @@ int rd_kafka_q_serve (rd_kafka_q_t *rkq, int timeout_ms,
...
+/**
+ * @brief Filter out and destroy outdated messages.
+ *
+ * @returns Returns the number of valid messages.
+ *
+ * @locality Any thread.
+ */
+static size_t rd_kafka_purge_outdated_messages (int32_t version,
+                                  rd_kafka_message_t **rkmessages,
+                                  int cnt) {
```
edenhill, line 528: Mix of size_t and int, pick one.
Looks like the caller is using unsigned int, so size_t is preferable.

src/rdkafka_queue.c
```diff
@@ -582,6 +607,15 @@ int rd_kafka_q_serve_rkmessages (rd_kafka_q_t *rkq, int timeout_ms,
                         continue;
                 }
 
+                if (unlikely(rko->rko_type == RD_KAFKA_OP_BARRIER)) {
+                        cnt = rd_kafka_purge_outdated_messages(
```
edenhill, line 611: Cast back to unsigned int (I think MSVC will warn about these lossy auto casts otherwise)

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+        const char *what;
+        rd_kafka_queue_t *rkq;
+        int timeout_ms;
+        int consume_msg_cnt;
+        rd_kafka_t *rk;
+        uint64_t testid;
+        int produce_msg_cnt;
+} consumer_t;
+
+static int test_consumer_batch_queue (void *arg) {
```
edenhill, line 44: The test_ prefix is used by the test framework itself, rename this to `consume_batch_queue()`

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+        const char *what = arguments->what;
+        const int produce_msg_cnt = arguments->produce_msg_cnt;
+
+        rd_kafka_message_t **rkmessage = malloc(consume_msg_cnt * sizeof(*rkmessage));
+
+        test_msgver_init(&mv, testid);
+
+        TIMING_START(&t_cons, "CONSUME");
+
+        while ((msg_cnt = rd_kafka_consume_batch_queue(rkq,
```
edenhill, line 65: Should it really be called multiple times?
We want to verify that a single call to batch_queue() handles purging outdated stuff .

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+                timeout_ms, rkmessage, consume_msg_cnt)) == 0)
+                ;
+
+        for (i = 0; i < msg_cnt; i++) {
+                if (test_msgver_add_msg(rk, &mv, rkmessage[i]) == 0)
+                        TEST_FAIL("The message is not from testid "
+                                  "%"PRId64" \n", testid);
+        }
+        test_msgver_verify(what,
```
edenhill, line 74: I think we'll actually want to combine the messages from both consumers in the same mv and then do DUP check, as we're not really checking if they receive the same messages here.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+                           &mv,
+                           TEST_MSGVER_ORDER|TEST_MSGVER_DUP,
+                           0,
+                           produce_msg_cnt/2);
+        test_msgver_clear(&mv);
+
+        for (i = 0; i < msg_cnt; i++)
+                rd_kafka_message_destroy(rkmessage[i]);
```
edenhill, line 82: This can be done in the add_msg loop

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+        TIMING_STOP(&t_cons);
```
edenhill, line 83: This should only time the consume_batch_queue call itself

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+static void do_test_consume_batch (const char *strategy) {
+        const int partition_cnt = 4;
+        rd_kafka_queue_t *rkq1, *rkq2;
+        const char *topic;
+        rd_kafka_t *c1;
+        rd_kafka_t *c2;
+        int p;
+        const int timeout_ms = 30000;
+        uint64_t testid;
+        const int consume_msg_cnt = 500;
```
edenhill, line 101: A comment explaining why these are different would be good.
Or preferably a more thorough explanation of the test approach in the docstring to this function.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+        topic = test_mk_topic_name("0122-buffer_cleaning", 1);
+
+        for (p = 0 ; p < partition_cnt ; p++)
+                test_produce_msgs_easy(topic,
+                                       testid,
+                                       p,
+                                       produce_msg_cnt / partition_cnt);
+
+        /* Create consumers */
+        test_conf_set(conf, "partition.assignment.strategy", strategy);
```
edenhill, line 127: nit: Set this where the other config is set.

tests/0122-buffer_cleaning_after_rebalance.c
```diff
@@ -0,0 +1,187 @@
...
+        c1_args.rkq = rkq1;
+        c1_args.timeout_ms = timeout_ms;
+        c1_args.consume_msg_cnt = consume_msg_cnt;
+        c1_args.rk = c1;
+        c1_args.testid = testid;
+        c1_args.produce_msg_cnt = produce_msg_cnt;
+
+        if (thrd_create(&thread_id, test_consumer_batch_queue, &c1_args)
+            != thrd_success)
+                TEST_FAIL("Failed to verify batch queue messages for %s",
```
edenhill, line 149: Failed to create thread

edenhill (changes requested):
Really close now!

edenhill (approved):
This looks great!  Good work on this rather hairy bug, Jing!

PR #3274, opened 2021-02-24
Default ssl.ca.location to probe on OSX

edenhill:
Let's get this into 1.6.1


rnpridgeon (approved):
Nice!

edenhill:
Thanks The Ryan!

PR #3275, opened 2021-02-24
KIP-360: Make idemp fatal errors recoverable for txn producer

edenhill:


src/rdkafka_idempotence.c
```diff
@@ -145,38 +145,52 @@ rd_kafka_idemp_broker_any (rd_kafka_t *rk,
...
  *        raising a fatal error.
  *
+ * @param is_fatal if true, force fatal error regardless of error code.
+ *
  * @returns rd_true if a fatal error was triggered, else rd_false.
  *
  * @locks rd_kafka_wrlock() MUST be held
  * @locality rdkafka main thread
  */
 rd_bool_t rd_kafka_idemp_check_error (rd_kafka_t *rk,
```
mhowlett, line 155: the passing in of `is_fatal` as well as processing in this function to possibly update it is a bit conflated. maybe have a function that does only does the stuff out of `if (!is_fatal) return rd_false;` and if the stuff before is needed in multiple places, have that in another function.

mhowlett: i guess it makes the preface logic harder... whatever you like, i think the point holds though

src/rdkafka_idempotence.c
```diff
@@ -261,18 +277,45 @@ void rd_kafka_idemp_pid_fsm (rd_kafka_t *rk) {
...
+
+                        if (err == RD_KAFKA_RESP_ERR__UNSUPPORTED_FEATURE &&
+                            rd_kafka_pid_valid(rk->rk_eos.pid))
+                                is_fatal = rd_true;
+                } else {
+                        rd_rkb_dbg(rkb, EOS, "GETPID", "Acquiring ProducerId");
+
+                        err = rd_kafka_InitProducerIdRequest(
+                                rkb,
+                                rk->rk_conf.eos.transactional_id,
```
mhowlett, line 312: if this isn't a transactional producer, what is the transactional_id being passed for? i assume this is NULL or something in this case, so correct, but maybe write that explicitly.

src/rdkafka_mock_handlers.c
```diff
@@ -1473,10 +1463,74 @@ rd_kafka_mock_handle_SyncGroup (rd_kafka_mock_connection_t *mconn,
...
  * @brief Generate a unique ProducerID
  */
-static void rd_kafka_mock_pid_generate (rd_kafka_mock_cluster_t *mcluster,
-                                        rd_kafka_pid_t *pid) {
+static const rd_kafka_pid_t
+rd_kafka_mock_pid_new (rd_kafka_mock_cluster_t *mcluster) {
+        rd_kafka_pid_t *pid = rd_malloc(sizeof(*pid));
+        rd_kafka_pid_t ret;
+
         pid->id = rd_jitter(1, 900000) * 1000;
```
mhowlett, line 1471: (nit) doesn't seem like this guaranteed to be unique?

src/rdkafka_mock_handlers.c
```diff
@@ -1473,10 +1463,74 @@ rd_kafka_mock_handle_SyncGroup (rd_kafka_mock_connection_t *mconn,
...
+        rd_list_add(&mcluster->pids, pid);
+        ret = *pid;
+        mtx_unlock(&mcluster->lock);
+
+        return ret;
+}
+
+
+/**
+ * @brief Checks if the given pid is known, else returns an error.
```
mhowlett, line 1484: also checks epoch

mhowlett (approved):

Issue #3279, opened 2021-03-01
ThreadSanitizer: use of an invalid mutex (e.g. uninitialized or destroyed) (pid=7)


filimonov:
Description
===========

ThreadSanitizer in ClickHouse CI catched 'use of an invalid mutex' 
https://clickhouse-test-reports.s3.yandex.net/0/174f49a9f13f87e3ad326043c249dad06e9b0e9b/integration_tests_(thread).html#fail1 

How to reproduce
================

[Scenario](https://github.com/ClickHouse/ClickHouse/blob/0fb9bc2f648577ae14e351df54bd804f6b314914/tests/integration/test_storage_kafka/test.py#L2435-L2464) like following
1. Consumer get created (unused - no subscribe / assign / poll calls, the topic does not exists)
2. Producer for the same topic created (nothing get produced). 
3. Everything is closed. 

It seems the failure happened on p.2.

Thread sanitizer report:

```
WARNING: ThreadSanitizer: use of an invalid mutex (e.g. uninitialized or destroyed) (pid=7)
    #0 pthread_mutex_lock <null> (clickhouse+0x8a234b6)
    #1 mtx_lock obj-x86_64-linux-gnu/../contrib/librdkafka/src/tinycthread.c:136:10 (clickhouse+0x15955fd9)
    #2 rd_kafka_buf_destroy_final obj-x86_64-linux-gnu/../contrib/librdkafka/src/rdkafka_buf.c:47:25 (clickhouse+0x1583444e)
    #3 rd_kafka_buf_callback obj-x86_64-linux-gnu/../contrib/librdkafka/src/rdkafka_buf.c:483:9 (clickhouse+0x1583565a)
    #4 rd_kafka_bufq_purge obj-x86_64-linux-gnu/../contrib/librdkafka/src/rdkafka_buf.c:255:17 (clickhouse+0x15835369)
    #5 rd_kafka_broker_fail obj-x86_64-linux-gnu/../contrib/librdkafka/src/rdkafka_broker.c:577:2 (clickhouse+0x1580ee02)
    #6 rd_kafka_broker_op_serve obj-x86_64-linux-gnu/../contrib/librdkafka/src/rdkafka_broker.c:3263:17 (clickhouse+0x1581f945)
    #7 rd_kafka_broker_ops_serve obj-x86_64-linux-gnu/../contrib/librdkafka/src/rdkafka_broker.c:3338:24 (clickhouse+0x1581f945)
[...]

PR #3285, opened 2021-03-08
Don't destroy metadata cache until brokers are destroyed (#3279)

edenhill:
As ops on broker queues may have references to the metadata cache's mutex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# librdkafka NEXT
+
+## Fixes
+
+### General fixes
+
+ * Fix accesses to freed metadata cache mutexes on client termination (#3279)
+
+
 # librdkafka v1.6.1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# librdkafka NEXT
+
+## Fixes
+
+### General fixes
+
+ * Fix accesses to freed metadata cache mutexes on client termination (#3279)
```
mhowlett, line 7: would this be causing segfault's on termination?
dotnet people have been complaining about that and i haven't investigated yet because it can happen if they are using the client after dispose, and people often do.

mhowlett (approved):

CHANGELOG.md
edenhill: Possible but unlikely, this was only caught in ClickHouse's TSAN tests.

PR #3287, opened 2021-03-09
Make broker_update() blocking to assure state is synchronized

edenhill:
As reported in https://github.com/confluentinc/confluent-kafka-python/issues/1043

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@
 ### General fixes
 
  * Fix accesses to freed metadata cache mutexes on client termination (#3279)
+ * There was a race condition on receiving updated metadata where a broker id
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,13 @@
 ### General fixes
 
  * Fix accesses to freed metadata cache mutexes on client termination (#3279)
```
mhowlett, line 17: side note: in .net i'm up to 1.6.2 already, maybe you want to skip a version here, maybe not. i'm not sure which is better / worse.

edenhill: Right, this will most likely be v1.7.0 anyway.

mhowlett (approved):

PR #3289, opened 2021-03-09
Change connections.max.idle.ms default on Azure to <4 minutes

edenhill:
..since the load balancer default timeout is 4 minutes.


edenhill:
Turns out the default timeout is 4 minutes on azure:
https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-tcp-reset#configurable-tcp-idle-timeout

mhowlett (approved):
i did wonder, when you said 4 yesterday

edenhill:
```
WONDER NO MORE
   LIKE YOU DID BEFORE
 NOW IT IS FOUR
```

PR #3308, opened 2021-03-21
fix 'apis' pointer was used after the memory was releaase

pavel-pimenov:
V774 [CWE-416] The 'apis' pointer was used after the memory was released. rdkafka_broker.c 2450


edenhill:
False positive, since err is set the function will return and `apis` will not be used.

PR #3309, opened 2021-03-21
V796 [CWE-484] It is possible that 'break' statement

pavel-pimenov:
...is missing in switch statement. rdkafka_conf.c 3250


src/rdkafka_conf.c
```diff
@@ -3248,6 +3248,7 @@ void rd_kafka_conf_properties_show (FILE *fp) {
 		case _RK_C_STR:
                 case _RK_C_KSTR:
 			typeinfo = "string";
+			break;
```
edenhill, line 3251: It should fall through (eventually hitting line 3262 where it prints prop->sdef, if anything there should be a `/*FALLTHRU*/` comment here.

PR #3310, opened 2021-03-21
Fix uninitialized variable

pavel-pimenov:
- V614 [CWE-457] Potentially uninitialized variable 'CoordId' used. rdkafka_cgrp.c 589
 - V614 [CWE-457] Potentially uninitialized variable 'CoordPort' used. rdkafka_cgrp.c 591
 - V614 [CWE-457] Potentially uninitialized variable 'is_modified' used. rdkafka_conf.c 4207


edenhill:
Both of these are false positives, and while it might help silence some warnings it is a disservice to the code reader since the initialization suggests that these variables are used without being set first, which is false.

Issue #3306, opened 2021-03-19
Issues with "partition.assignment.strategy=cooperative-sticky"


shanson7:
Description
===========
We are making changes to support cooperative/incremental rebalance. We are seeing errors during rebalance and what seems to be sub-optimal rebalancing (both when using the default `rebalance_cb` and our custom one, which just logs the partitions that are passed in and calls `incremental_(un)assign`). 

During rebalance it is not uncommon to get errors like:

```
19MAR2021_12:25:56.621 289 139893795714816 INFO ipbkafka_consumer.cpp 387 IPBKAFKA.CONSUMER KAFKA LOG: severity = 4 fac = COMMITFAIL event = [thrd:main]: Offset commit (unassigned partitions) failed for 2/5 partition(s) in join-state wait-incr-unassign-to-complete: Broker: Specified group generation id is not valid: apdx[5]@581544535(Broker: Specified group generation id is not valid), apdx[11]@418573170(Broker: Specified group generation id is not valid)
```

Additionally, it seems when a consumer leaves the group, partitions are being shuffled between the remaining consumers (I would expect that cooperative rebalance would only assign partitions from the consumer that left without revoking any).

How to reproduce
================

Bring up a consumer and let it start consuming. Bring up a second and let it rebalance. Bring up a third. bring down the third.

When the third consumer is added, we see the above error. When the third consumer leaves, we see partitions shuffled unexpectedly. I can reproduce consistently, so I can upload load with whatever debug settings are relevant (at least "cgrp" I imagine).

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `1.6.1`
 - [x] Apache Kafka version: `kafka_2.13-2.6.0`
 - [x] librdkafka client configuration: `partition.assignment.strategy=cooperative-sticky`
 - [x] Operating system: `rhel7`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

mhowlett:
thanks @shanson7 - can reproduced both, investigating.

mhowlett:
i believe the shuffling is happening because the `prevAssignment` as initialized here:

https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_sticky_assignor.c#L1189

[...]

Issue #3226, opened 2021-01-22
Constantly facing "Timed out LeaveGroupRequest in flight" Error in kafka consumer


Udayaprasad:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========

We are running Kafka in dockerized containers with 3 zookeepers and 3 Kafka broker instances. below are the image details of each of the service.
`zookeeper:3.4.9` and `confluentinc/cp-kafka:5.5.3`

In the AI code, we have 8 producers, which will take the streaming videos as input (one producer, one camera) and send them to the consumers (one producer, one consumer - Message size will be more since it is processing image frames).

We are facing a Consumer timeout error after 3-4 hours of running AI models. Below is the Timeout error it occurs. I have also pasted the Kafka container configuration below, please verify.

It is constantly failing. We have tried with different configurations by increasing the timeout and reducing the message bytes, but no luck.

@edenhill Could you please suggest any alternatives to avoid these timeout errors?

Error Details

```
AI_Models | %3|1611286416.032|FAIL|rdkafka#consumer-12| [thrd:GroupCoordinator]: GroupCoordinator: Connect to ipv4#127.0.0.1:4702 failed: Connection refused (after 0ms in state CONNECT, 4 identical error(s) suppressed)
AI_Models | %3|1611286416.449|FAIL|rdkafka#consumer-15| [thrd:GroupCoordinator]: GroupCoordinator: Connect to ipv4#127.0.0.1:4702 failed: Connection refused (after 0ms in state CONNECT, 4 identical error(s) suppressed)
```

How to reproduce
================
Create 8 producers, consumers, and topics with the following Kafka configuration. 

```
KAFKA_ADVERTISED_LISTENERS: LISTENER_DOCKER_INTERNAL://kafka:19096,LISTENER_DOCKER_EXTERNAL://${DOCKER_HOST_IP:-127.0.0.1}:4701
KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: LISTENER_DOCKER_INTERNAL:PLAINTEXT,LISTENER_DOCKER_EXTERNAL:PLAINTEXT
KAFKA_LOG_SEGMENT_BYTES: 16777216
KAFKA_LOG_RETENTION_BYTES: 134217728
KAFKA_MESSAGE_MAX_BYTES: 20000000
KAFKA_REQUEST_TIMEOUT_MS: 800000
[...]

PR #3313, opened 2021-03-22
Prevent auto-commit during rebalance

mhowlett:
resolves the commit fail issue reported in #3306 by disallowing auto commits whilst a rebalance is in progress.
does not resolve the partition shuffling issue.


mhowlett:
on second thoughts, manual committing will need consideration as well. WIP.

mhowlett:
see #3226

pratikthakkar24:
hey @mhowlett  do you mean, we should not manually commit messages while rebalance is in progress ? (I have registered a rebalance callback).

pratikthakkar24:
> on second thoughts, manual committing will need consideration as well. WIP.

If we are using manual commit and in case of one of the three consumers shut down, how do we get to know in the other two consumers that a rebalance is in progress in the broker and we should stop manually committing our offsets of other partitions because there is a rebalance activity going on in the broker ?

PR #3315, opened 2021-03-23
Adding OpenSSL engine support in LibrdKafka for SSL connections

adinigam:
Currently LibrdKafka uses Openssl behind the scenes to establish SSL channel between broker and client.
To create the SSL channel, OpenSSL needs public and private keys. Today these keys are provided in byte* format to openssl. Byte* can either be generated by librdkafka or application itself. But ultimately to generate the byte* format, keys needs to be exported from where-ever it is residing.
In many scenarios exporting might not be an option since it might not be permitted by hardware/software, eg: if application is using non-exportable certificates in windows certificate store.

Good news is OpenSSL provides plugin model which is called engine. This engine can be used to interact with the storage and use the public and private keys for the certificate without exporting in byte* format. Such clients will have to write custom engines.

I am proposing to enhance LibrdKafka to support OpenSSL engine. Below are the changes I am taking up in this PR.

1) Add new parameter "ssl.engine.location" which will take the full path to the binary containing the openssl engine

2) This engine will be initialized if the parameter "ssl.engine.location" value is not null. Below is the initialization code in the method rd_kafka_ssl_ctx_init

3) Call engine method ENGINE_load_ssl_client_cert with appropriate params to fetch the pointers to the public and private keys. Key parameter is building and disposing the stack of ca-certs in rd_kafka_ssl_set_certs.

Note: this is a fresh Pr on https://github.com/edenhill/librdkafka/pull/2602


adinigam:
@edenhill For some reason I was unable to rebase to latest master, hence sending fresh PR on latest master.
But just curious to know even when all comments were resolved why was the merge being pushed further and further (1.5.2 -> 1.6 -> 1.7) even when all comments were resolved and all gates were green ?
Just asking this question if i could have done anything to speed up my checkins.

examples/openssl_engine_example.cpp
```diff
@@ -0,0 +1,315 @@
...
+    delete conf;
+
+    /*
+     * Wait for RdKafka to decommission.
+     * This is not strictly needed (when check outq_len() above), but
+     * allows RdKafka to clean up all its resources before the application
+     * exits so that memory profilers such as valgrind wont complain about
+     * memory leaks.
+     */
+    RdKafka::wait_destroyed(5000);
```
edenhill, line 312: This call is no longer needed.
But there should be a `delete producer;` call.

src-cpp/rdkafkacpp.h
```diff
@@ -1283,6 +1283,13 @@ class RD_EXPORT Conf {
                                          const void *buffer, size_t size,
                                          std::string &errstr) = 0;
 
+  /** @brief Set callback_data for ssl engine
+   *
+   * @returns CONF_OK on success, else CONF_INVALID.
+   */
+  virtual Conf::ConfResult set_engine_callback_data(void *value,
```
edenhill, line 1290: New virtual methods needs to be added to the end of the class

src/rdkafka.h
```diff
@@ -2311,6 +2311,21 @@ rd_kafka_conf_set_ssl_cert (rd_kafka_conf_t *conf,
...
+/**
+ * @brief Set callback_data for OpenSSL engine.
+ *
+ * @param conf Configuration object.
+ * @param callback_data Data being passed to engine i.e ENGINE_load_ssl_client_cert.
+ *
+ * @remark The /c ssl.engine.location configuration must be set for this to have affect.
+ */
+RD_EXPORT
+void rd_kafka_conf_set_engine_callback_data(rd_kafka_conf_t *conf,
```
edenhill, line 2325: style nit: functions should have a space between name and open-paren.

src/rdkafka_conf.c
```diff
@@ -829,6 +829,22 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           "Path to client's keystore (PKCS#12) used for authentication.",
           _UNSUPPORTED_SSL
         },
+#if OPENSSL_VERSION_NUMBER >= 0x10100000
```
edenhill, line 832: Needs to use _UNSUPPORTED_SSL_1_1 macro, see above.

edenhill (changes requested):
Seems like not all of the comments I made on the closed PR has been addressed. Let me know when this is ready for review.

src/rdkafka_conf.c
adinigam: @edenhill Can you explain how is going to compile since ssl.engine_location will no longer be available in no-ssl scenarios ?

edenhill:
The PR comments (here and in the closed PR) need to be addressed this week for this feature to be eligible for the upcoming v1.7.0 release.

src/rdkafka_conf.c
adinigam: putting the  #if OPENSSL_VERSION_NUMBER >= 0x10100000  back along with _UNSUPPORTED_SSL_1_1

adinigam:
> 
> 
> The PR comments (here and in the closed PR) need to be addressed this week for this feature to be eligible for the upcoming v1.7.0 release.

ajbarb:
> > The PR comments (here and in the closed PR) need to be addressed this week for this feature to be eligible for the upcoming v1.7.0 release.
> 
> @edenhill Please check now. I missed out on some comments in previous pr

@adinigam thanks for addressing the comments.
@edenhill thank you so much for the comments and please let us know if something is missing.

src/rdkafka_conf.h
```diff
@@ -32,6 +32,9 @@
 #include "rdlist.h"
 #include "rdkafka_cert.h"
 
+#if WITH_SSL
+#include <openssl/engine.h>
```
edenhill, line 36: Doesn't this need to be version-guarded to 1.1.0 too?

src/rdkafka_conf.h
```diff
@@ -239,6 +242,12 @@ struct rd_kafka_conf_s {
                 /** CSV list of Windows certificate stores */
                 char *ca_cert_stores;
                 char *crl_location;
+#if OPENSSL_VERSION_NUMBER >= 0x10100000
```
edenhill, line 245: and WITH_SSL

src/rdkafka_conf.h
```diff
@@ -239,6 +242,12 @@ struct rd_kafka_conf_s {
...
+                ENGINE *engine;
+                char   *engine_location;
```
edenhill, line 247: the char*s should be outside the if-guard

src/rdkafka_ssl.c
```diff
@@ -1288,6 +1358,11 @@ static int rd_kafka_ssl_set_certs (rd_kafka_t *rk, SSL_CTX *ctx,
  */
 void rd_kafka_ssl_ctx_term (rd_kafka_t *rk) {
         SSL_CTX_free(rk->rk_conf.ssl.ctx);
+
+#if OPENSSL_VERSION_NUMBER >= 0x10100000
+        ENGINE_free(rk->rk_conf.ssl.engine);
```
edenhill, line 1363: Add NULL check

tests/0124-openssl_invalid_engine.c
```diff
@@ -0,0 +1,55 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test.h"
+
+int main_0124_openssl_invalid_engine (int argc, char **argv) {
+#if WITH_SSL && OPENSSL_VERSION_NUMBER >= 0x10100000
```
edenhill, line 32: Don't think we can rely on the OPENSSL defines to be available here. I think a simpler alternative is to skip the test if
conf_set("ssl.engine.location") returns UNKNOWN. E.g.:
```c
 rd_kafka_conf_res_t res;
 ...
        res = rd_kafka_conf_set(conf, "ssl.engine.location", "invalid_path", 
                              errstr, sizeof(errstr));
        if (res != RD_KAFKA_CONF_UNKNOWN) {
                  rd_kafka_destroy(conf);
                  TEST_SKIP("%s\n", errstr);
                  return 0;
        } else if (res != RD_KAFKA_CONF_OK)
                 TEST_FAIL("%s", errstr);
....

```

tests/0124-openssl_invalid_engine.c
```diff
@@ -0,0 +1,55 @@
...
+        rd_kafka_conf_t *conf;
+        rd_kafka_t *rk;
+        char errstr[512];
+
+        conf = rd_kafka_conf_new();
```
edenhill, line 37: Use test_conf_init(&conf, ..) so we can pass TEST_DEBUG, etc.

tests/test.c
```diff
@@ -433,6 +434,7 @@ struct test tests[] = {
         _TEST(0120_asymmetric_subscription, TEST_F_LOCAL),
         _TEST(0121_clusterid, TEST_F_LOCAL),
         _TEST(0123_connections_max_idle, 0),
+        _TEST(0124_openssl_invalid_engine, 0),
```
edenhill, line 437: Mark as TEST_F_LOCAL since it doesn't require a broker.

edenhill (changes requested):

src/rdkafka_conf.h
adinigam: doesn't hurt either way since actual variable definition and usage are conditioned.
also engine.h was introduced way back but some methods which i am using were introduced in 1.1.0
but will add the condition for consistency

edenhill: Oh, if openssl/engine.h is available in 1.0.2 (our oldest supported version) it is fine to leave this as-is.

tests/0124-openssl_invalid_engine.c
adinigam: i think you mean
if (res **==** RD_KAFKA_CONF_UNKNOWN) {
                  rd_kafka_destroy(conf);
                  TEST_SKIP("%s\n", errstr);
                  return 0;

tests/0124-openssl_invalid_engine.c
```diff
@@ -0,0 +1,62 @@
...
+        rd_kafka_t *rk;
+        char errstr[512];
+        rd_kafka_conf_res_t res;
+
+        test_conf_init(&conf, NULL, 30);
+        if (rd_kafka_conf_set(conf, "security.protocol", "ssl", 
+                              errstr, sizeof(errstr)) != RD_KAFKA_CONF_OK)
+                TEST_FAIL("%s", errstr);
+
+        res = rd_kafka_conf_set(conf, "ssl.engine.location", "invalid_path",
```
edenhill, line 42: Move this above security.protocol since security.protocol will fail the test if ssl support is not available, while the ssl.engine.location check skips the test in that case.

edenhill (changes requested):
The test does not build:
```
/project/repo/checkout/tests/0124-openssl_invalid_engine.c:46:34: warning: passing argument 1 of 'rd_kafka_destroy' from incompatible pointer type [-Wincompatible-pointer-types]
                 rd_kafka_destroy(conf);
```

adinigam:
> 
> 
> The test does not build:
> 
> ```
> /project/repo/checkout/tests/0124-openssl_invalid_engine.c:46:34: warning: passing argument 1 of 'rd_kafka_destroy' from incompatible pointer type [-Wincompatible-pointer-types]
>                  rd_kafka_destroy(conf);
> ```

this errors are hard to catch on my local box since all compilation happens with ssl and engine.
fixed it

tests/0124-openssl_invalid_engine.c
adinigam: fixed.
this was hard to catch since my local environment has ssl and engine.

tests/0124-openssl_invalid_engine.c
```diff
@@ -35,22 +35,21 @@ int main_0124_openssl_invalid_engine (int argc, char **argv) {
...
         test_conf_init(&conf, NULL, 30);
-        if (rd_kafka_conf_set(conf, "security.protocol", "ssl", 
-                              errstr, sizeof(errstr)) != RD_KAFKA_CONF_OK)
-                TEST_FAIL("%s", errstr);
-
         res = rd_kafka_conf_set(conf, "ssl.engine.location", "invalid_path", 
                                 errstr, sizeof(errstr));
 
         if (res == RD_KAFKA_CONF_UNKNOWN) {
-                rd_kafka_destroy(conf);
```
edenhill: You need to destroy the conf or it will leak.

edenhill (changes requested):

tests/0124-openssl_invalid_engine.c
adinigam: sorry didn't realize there is rd_kafka_conf_destroy(conf);

src/rdkafka_ssl.c
```diff
@@ -1470,44 +1470,52 @@ int rd_kafka_ssl_ctx_init (rd_kafka_t *rk, char *errstr, size_t errstr_size) {
...
-                rk->rk_conf.ssl.engine = ENGINE_by_id(rk->rk_conf.ssl.engine_id);
+                rk->rk_conf.ssl.engine = 
+                        ENGINE_by_id(rk->rk_conf.ssl.engine_id);
                 if (!rk->rk_conf.ssl.engine) {
                         rk->rk_conf.ssl.engine = ENGINE_by_id("dynamic");
                         if (!rk->rk_conf.ssl.engine) {
                                 rd_snprintf(errstr, errstr_size, 
-                                            "Engine initialization failed in ENGINE_by_id : ");
+                                            "Engine initialization failed in"
+                                            " ENGINE_by_id : ");
```
edenhill, line 1485: remove space before :
`" ENGINE_by_id: "

for all these

edenhill (changes requested):

edenhill:
Thank's alot for contributing this functionality! :champagne:

edenhill:
## Heads up - deprecation
We will be updating to OpenSSL 3.0.x soon which deprecates the use of OpenSSL engines (in favour of providers) - this means that we'll also deprecate the use of engines, and that we might provide prebuilt librdkafka builds (for e.g, Python and Go clients) without engine support.

PR #3320, opened 2021-03-25
Stats: Fix consumer_lag 

edenhill:
This makes consumer_lag show the committed consumer lag, i.e., the same thing as what existing consumer_lag monitoring applications will see (but without querying the coordinator), and adds a consumer_lag_stored which is the consumer lag based on the stored offset (the to-be-committed offset) which is the lag from the application's point of view.

### DIFF
--- a/STATISTICS.md
+++ b/STATISTICS.md
@@ -183,7 +183,8 @@ eof_offset | int gauge | | Last PARTITION_EOF signaled offset
 lo_offset | int gauge | | Partition's low watermark offset on broker
 hi_offset | int gauge | | Partition's high watermark offset on broker
 ls_offset | int gauge | | Partition's last stable offset on broker, or same as hi_offset is broker version is less than 0.11.0.0.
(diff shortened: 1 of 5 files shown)


STATISTICS.md
```diff
@@ -183,7 +183,8 @@ eof_offset | int gauge | | Last PARTITION_EOF signaled offset
 lo_offset | int gauge | | Partition's low watermark offset on broker
 hi_offset | int gauge | | Partition's high watermark offset on broker
 ls_offset | int gauge | | Partition's last stable offset on broker, or same as hi_offset is broker version is less than 0.11.0.0.
-consumer_lag | int gauge | | Difference between (hi_offset or ls_offset) - max(app_offset, committed_offset). hi_offset is used when isolation.level=read_uncommitted, otherwise ls_offset.
+consumer_lag | int gauge | | Difference between (hi_offset or ls_offset) - committed_offset). hi_offset is used when isolation.level=read_uncommitted, otherwise ls_offset.
```
mhowlett, line 186: extra parenthesis after committed_offset.

also, `-` means difference, so better to replace that with 'and'. same comment for consumer_lag_stored.

mhowlett (approved):

PR #3326, opened 2021-04-01
Don't auto-commit during rebalancing

edenhill:


jliunyu (approved):

mhowlett (approved):
LGTM. The error is going to occur when people manually commit as well.. which is bad UX since it's possible to handle that automatically. But a fix for another PR.

PR #3329, opened 2021-04-05
cmake: Fix using absolute path in include directories

hmenn:
* External library include directories were binded as public dependency.
  That causes absolute path rependency and It produces
  absolute paths and that crashs build on yocto.

You can check this pr for additional information about crash: https://github.com/openembedded/meta-openembedded/pull/324 and this https://errors.yoctoproject.org/Errors/Details/575207/


edenhill:
Ping @benesch @ed-alertedh @neptoess for review

neptoess (commented):
Looks good to me. Good catch @hmenn .

edenhill:
Thanks!

PR #3336, opened 2021-04-07
Remove CMAKE_FIND_LIBRARY_SUFFIXES From MinGW Build Script

neptoess:
Remove CMAKE_FIND_LIBRARY_SUFFIXES from packaging/mingw-w64/configure-build-msys2-mingw-static.sh
It turns out that this will not cause any issues with the static bundle

Thanks to @SpaceIm for initially pointing it out.


edenhill:
Thanks :+1:

PR #3345, opened 2021-04-13
update php bindings

nick-zh:
I took the liberty of also removing phpkafka, it seems to me, that it is not maintained anymore


edenhill:
Thanks!

PR #3352, opened 2021-04-16
Fix qlen value in debug statement

shanson7:
Per [gitter convo](https://gitter.im/edenhill/librdkafka?at=6051d19d28e6153d7224c665) this debug statement prints the same value (`rd_kafka_q_len(&msetr->msetr_rkq)`) twice, when it seems one of these values should be the length of the parent queue.


edenhill:
Thanks!

PR #3354, opened 2021-04-16
CMake: Remove Libs.private entries from rdkafka-static.pc and rdkafka++-static.pc

neptoess:
Per @edenhill comment [here](https://github.com/confluentinc/confluent-kafka-go/pull/564#discussion_r614670205), when building static pkg-config files, add any additional linker flags for dynamic dependencies, e.g. `-ldl` on *nix, `-lws2_32` on Windows, to `Libs`, not `Libs.private`. As this only affects CMake static builds, the risk is minimal (nothing uses these builds yet, but the hope is that confluent-kafka-go will be able to use static MinGW-built artifacts to provide Windows support)


edenhill (approved):

edenhill:
Thanks!

PR #3360, opened 2021-04-19
Fixes for the sticky assignor balancing part

edenhill:


mhowlett (approved):
LGTM, let's get an RC and i'll give it an extensive manual workout as well.

PR #3362, opened 2021-04-20
Additional sticky fixes

edenhill:


src/rdkafka_sticky_assignor.c
```diff
@@ -2025,8 +2025,31 @@ static void rd_kafka_sticky_assignor_state_destroy (void *assignor_state) {
...
+ * its new assignment and including it in the next rebalance as its
+ * owned-partitions.
+ */
+static void ut_set_owned (rd_kafka_group_member_t *rkgm) {
+        if (rkgm->rkgm_owned)
+                rd_kafka_topic_partition_list_destroy(rkgm->rkgm_owned);
+
+        rkgm->rkgm_owned =
+                rd_kafka_topic_partition_list_copy(rkgm->rkgm_assignment);
+}
```
mhowlett, line 2042: without investigating why - there may be a good reason - it seems it may be unnecessarily unconstrained to allow rkgm_assignment to be used here.

mhowlett (approved):

PR #3368, opened 2021-04-27
fix #3367; Resolve conflict with prefix and libdir

### LINKED ISSUE #3367: Improve configure --libdir (etc) help texts

Description
===========
To link rdkafka with prebuilt openssl and sasl, we use `libdir` to specify the 3rparty library path.  To install the library with a specific path(Which is useful in docker) we will configure it with `prefix` option.

But then the library(librdkafka.so) will be copied into `libdir`, which is expected to into `prefix`, this is a bug.

How to reproduce
================

```
./configure --prefx=/tmp/rd_install --libdir=$3RDPARTY_LIB_DIR
make -j$(nproc)
make install
```

Then check `/tmp/rd_install`  and `$3RDPARTY_LIB_DIR` to find that the libraries in `$3RDPARTY_LIB_DIR`

Checklist
=========

 - [x] librdkafka version (release number or git tag): `<1.6.1>`
 - [x] Apache Kafka version: `<NULL>`
 - [x] librdkafka client configuration: `<NULL>`
 - [x] Operating system: `<Rhel 7>`
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka:NULL
 - [x] Provide broker log excerpts:NULL
 - [ ] Critical issue

theidexisted:


edenhill:
libdir is the install directory, not the search path.

PR #3369, opened 2021-04-27
Correctly handle empty aborted transactions

edenhill:
Seems like there is a case where aborted transactions will have an
ABORT ctrl message marker, but not be included in the aborted transactions
list.
This would cause the next aborted-transaction-offset to be popped from the
aborted transaction list even though it did not match the aborted transaction
marker. This in turn could lead to aborted messages for subsequent aborted
transactions in the same MessageSet to not be skipped and instead delivered
to the application.

The fix is to silently (unless debug) ignore these unsolicited abort markers
if they're not in the aborted transaction list and not pop off the next
non-matching offset from the aborted transaction list.

### DIFF
--- a/src/rdkafka_msgset_reader.c
+++ b/src/rdkafka_msgset_reader.c
@@ -81,7 +81,7 @@
 
 static RD_INLINE int64_t
 rd_kafka_aborted_txns_pop_offset (rd_kafka_aborted_txns_t *aborted_txns,
-                                  int64_t pid);
+                                  int64_t pid, int64_t max_offset);
 static RD_INLINE int64_t
 rd_kafka_aborted_txns_get_offset (const rd_kafka_aborted_txns_t *aborted_txns,
                                   int64_t pid);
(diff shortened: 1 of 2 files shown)


src/rdkafka_msgset_reader.c
```diff
@@ -1527,12 +1533,20 @@ rd_kafka_aborted_txns_offsets_for_pid (rd_kafka_aborted_txns_t *aborted_txns,
  * offset for the specified pid.
  *
  * @param increment_idx if true, the offset index will be incremented.
+ * @param max_offset If the next aborted offset is greater than \p max_offset
+ *                   then the index is not incremented (regardless of
+ *                   \p increment_idx) and the function returns -1.
+ *                   This may be the case for empty aborted transactions
+ *                   that have an ABORT marker but are not listed in the
+ *                   AbortedTxns list.
```
mhowlett, line 1541: seems good

mhowlett (approved):

jliunyu (approved):

PR #3372, opened 2021-04-30
KIP-735: increase default session.timeout.ms

edenhill:


CONFIGURATION.md
```diff
@@ -96,7 +96,7 @@ interceptors                             |  *  |                 |
...
+session.timeout.ms                       |  C  | 1 .. 3600000    |         45000 | high       | Client group session and failure detection timeout. The consumer sends periodic heartbeats (heartbeat.interval.ms) to indicate its liveness to the broker. If no hearts are received by the broker for a group member within the session timeout, the broker will remove the consumer from the group and trigger a rebalance. The allowed range is configured with the **broker** configuration properties `group.min.session.timeout.ms` and `group.max.session.timeout.ms`. Also see `max.poll.interval.ms`. <br>*Type: integer*
```
mhowlett, line 99: The kip talks about the interplay between `session.timeout.ms` and `request.timeout.ms` (30s). librdkafka's `socket.timeout.ms` defaults to 60s, which is still higher than 45s. Do similar considerations apply here?

Side note: I don't like the name `socket.timeout.ms`, because it sounds like a lower level thing than it is. Have you considered adding an alias `request.timeout.ms`?

edenhill: Very sharp eyed!

All the group-related requests have their own timeouts, namely:
OffsetCommit = 60s (socket.timeout.ms)
JoinGroup:
                /* Request timeout is max.poll.interval.ms + grace
                 * if the broker supports it, else
                 * session.timeout.ms + grace. */
                (ApiVersion >= 1 ?
                 rk->rk_conf.max_poll_interval_ms :
                 rk->rk_conf.group_session_timeout_ms) +
                3000/* 3s grace period*/,
SyncGroup: session.timeout.ms + 3s
LeaveGroup: 5s (hardcoded)


As for socket.timeout.ms - it is from the Scala days, and unfortunately we already have request.timeout.ms which controls the ProduceRequest.RequestTimeoutMs field which is the broker-side timeout for produce requests.

mhowlett: it may be worth considering `api.request.timeout.ms`, though it's clunky. i guess probably better to just leave it as you say.

mhowlett (approved):

INTRODUCTION.md
```diff
@@ -1906,6 +1906,7 @@ The [Apache Kafka Implementation Proposals (KIPs)](https://cwiki.apache.org/conf
 | KIP-602 - Use all resolved addresses by default                          | 2.6.0                       | Supported                                                                                     |
 | KIP-651 - Support PEM format for SSL certs and keys                      | 2.7.0                       | Supported                                                                                     |
 | KIP-654 - Aborted txns with non-flushed msgs should not be fatal         | 2.7.0                       | Supported                                                                                     |
+| KIP-735 - Increase default consumer session timeout                      | TBA                         | Supported                                                                                     |
```
ijuma, line 1909: We can say 3.0.0 here.

PR #3374, opened 2021-05-02
Clang cl compilation fixes

GregBowyer:
Clang when run as clang-cl on windows has subtle issues with SIMD intrinsics. Essentially it tries to parse its own headers but gets confused about the handling of none standard `__attribute__` markers. This leads it to confuse the meaning of certain SIMD types.

Fortunately librdkafka (and snappy as embedded) only need some basic operations that clang (even as clang-cl` support). As such we can use the builtins

### DIFF
--- a/src/rdendian.h
+++ b/src/rdendian.h
@@ -111,14 +111,24 @@
 #endif
 
 #elif defined(_WIN32)
-#include <intrin.h>
-
-#define be64toh(x) _byteswap_uint64(x)
-#define be32toh(x) _byteswap_ulong(x)
-#define be16toh(x) _byteswap_ushort(x)
 #define le16toh(x) (x)
 #define le32toh(x) (x)
 #define le64toh(x) (x)
+
+#if defined(__clang__)
+/* Clang on windows masquerades as MSVC
+ * but has strange behaviours with handling simd intrinsics
+ * fortunatly it understands bswap builtins which is what we want
+ */
+#define be64toh(x) __builtin_bswap64(x)
+#define be32toh(x) __builtin_bswap32(x)
+#define be16toh(x) __builtin_bswap15(x)
+#else
+#include <intrin.h>
+#define be64toh(x) _byteswap_uint64(x)
+#define be32toh(x) _byteswap_ulong(x)
+#define be16toh(x) _byteswap_ushort(x)
+#endif
 
 #elif defined _AIX      /* AIX is always big endian */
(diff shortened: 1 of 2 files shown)


src/rdendian.h
```diff
@@ -111,14 +111,24 @@
...
+#define le64toh(x) (x)
 
+#if defined(__clang__)
+/* Clang on windows masquerades as MSVC
+ * but has strange behaviours with handling simd intrinsics
+ * fortunatly it understands bswap builtins which is what we want
+ */
+#define be64toh(x) __builtin_bswap64(x)
+#define be32toh(x) __builtin_bswap32(x)
+#define be16toh(x) __builtin_bswap15(x)
```
edenhill, line 125: 15? or 16?

src/rdendian.h
```diff
@@ -111,14 +111,24 @@
...
 #elif defined(_WIN32)
-#include <intrin.h>
+#define le16toh(x) (x)
+#define le32toh(x) (x)
+#define le64toh(x) (x)
 
+#if defined(__clang__)
+/* Clang on windows masquerades as MSVC
+ * but has strange behaviours with handling simd intrinsics
+ * fortunatly it understands bswap builtins which is what we want
```
edenhill, line 121: fortunately

src/rdendian.h
```diff
@@ -111,14 +111,24 @@
...
 
 #elif defined(_WIN32)
-#include <intrin.h>
+#define le16toh(x) (x)
+#define le32toh(x) (x)
+#define le64toh(x) (x)
 
+#if defined(__clang__)
+/* Clang on windows masquerades as MSVC
+ * but has strange behaviours with handling simd intrinsics
```
edenhill, line 120: Do you know of any resource/link that provides more info on the strange behaviours?

edenhill (changes requested):

src/rdendian.h
GregBowyer: Whoops, I should have applied patch rather than redo it in github ui.

src/rdendian.h
GregBowyer: Not a lot unfortunately:

The best I found are some breadcrumbs around the internet such as:
https://lists.llvm.org/pipermail/llvm-dev/2014-October/077367.html

When I figured out it was byte swaps I stole this:
https://chromium.googlesource.com/chromium/src/+/master/base/sys_byteorder.h#42

PR #3376, opened 2021-05-03
Add support for buildling on illumos

filimonov:
Pushing the patch sent by @bnaecker in https://github.com/ClickHouse-Extras/librdkafka/pull/4 to the upstream.

> This is part of a larger set of changes for building ClickHouse on [illumos](https://illumos.org).
>
> illumos does not use the `DT_REG` and friends macros for determining the type of a directory entry. An extra `stat(2)` system call is required, followed by a similar query on the `struct stat` returned from that call.

More details: https://github.com/ClickHouse/ClickHouse/pull/23746


edenhill:
Thank you!

PR #3383, opened 2021-05-07
Improve compression codec error message

ellimist:
Currently the error message: `Not implemented` is pretty cryptic since it lacks any context. 

This PR addresses that, by specifying this is about a missing compression codec.

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -262,7 +262,7 @@ enum ErrorCode {
         ERR__STATE = -172,
 	/** Unknown protocol */
         ERR__UNKNOWN_PROTOCOL = -171,
-	/** Not implemented */
+	/** Compression Codec not implemented */
         ERR__NOT_IMPLEMENTED = -170,
 	/** Authentication failure*/
 	ERR__AUTHENTICATION = -169,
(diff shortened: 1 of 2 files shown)


dasch:
We've had several cases where users spent a long time debugging because this error message was so unclear. :100: to improving it!

src-cpp/rdkafkacpp.h
```diff
@@ -262,7 +262,7 @@ enum ErrorCode {
         ERR__STATE = -172,
 	/** Unknown protocol */
         ERR__UNKNOWN_PROTOCOL = -171,
-	/** Not implemented */
+	/** Compression Codec not implemented */
         ERR__NOT_IMPLEMENTED = -170,
```
edenhill, line 266: This error code is also used for other stuff that is not related to compression, so it cant be changed to say compression, needs to be more generic.

Can you explain where exactly you are seeing Not implemented errors?

ellimist: We had this "Not implemented" error showing up when trying to consume topics produced with `zstd` compression, while the consumer did not have support for it. It was quite challenging narrowing it down to that, and we've had multiple people lose time due to it 🤔 

This code comment made it sound like the assumption is that the Not implemented comes for the compression codecs only:

https://github.com/edenhill/librdkafka/blob/bd89239632e5d00200d05f65e32845420602f792/src/rdkafka_queue.c#L919-L921

Any advice on how to make that more visible? 🤔

PR #3431, opened 2021-06-16
Mark sasl ready if relogin disabled

mpekalski:
It fixes the issue https://github.com/edenhill/librdkafka/issues/3430

Basically if one has a cached kerberos ticket and disables relogin then sasl should be marked as ready and there should be no first attempt to refresh the ticket.


edenhill:
Thanks for this!

PR #3435, opened 2021-06-17
Keep session alive during long consumer group rebalances

hpk:
If a consumer group rebalance takes longer than `session.timeout.ms` (for example, one of the consumers is taking a long time to complete a batch of messages), then other consumers will fail their session timeout check after some time. This is because heartbeat responses from the broker with a `REBALANCE_IN_PROGRESS` error do not update the session timeout.

This PR simply updates the timer when `REBALANCE_IN_PROGRESS` is received in a heartbeat response.

There was a similar improvement in the Java client here: https://github.com/apache/kafka/pull/8834

That change modified the broker to send no error in response to heartbeats during a rebalance, but this PR should fix the issue for clients of older brokers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ librdkafka v1.7.1 is a maintenance release.
    `ERR_COORDINATOR_NOT_AVAILABLE`, and `ERR_NOT_COORDINATOR` (#3398).
    Offset commits will be retried twice.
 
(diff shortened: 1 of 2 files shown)


hpk:
@edenhill Any thoughts on this? This is a pretty low-risk change that potentially fixes #3450

edenhill:
Good fix, thanks!

Issue #3420, opened 2021-06-11
StoreOffset on consumer with Assign() no longer auto committed in 1.7.0


ulbodk:
Description
===========
We are using librdkafka via the Confluent .NET Api. Up until 1.6.3 (with librdkafka 1.6.1) we have successfully been using `StoreOffset` on consumers with `Assign(topicPartition)` (and not `Subscribe()`). With 1.7.0 the stored offsets are no longer committed in the background. If we add an explicit call to `Commit()` after `StoreOffset` then they get committed.

Note, *problem is not* with the `SetOffsetsCommittedHandler`, we have verified that no commit are made on our Kafka cluster with 1.7.0.

How to reproduce
================
The following test code works with 1.6.1 but not 1.7.0:

```csharp
ConsumerConfig cfg = // default values
cfg.EnableAutoCommit = true;
cfg.AutoCommitIntervalMs = 1000;

AutoResetEvent committed = new AutoResetEvent(false);
var builder = new ConsumerBuilder<string, byte[]>(cfg)
	.SetOffsetsCommittedHandler((c, o) => committed.Set());

using (IConsumer<string, byte[]> con = builder.Build())
{
	// IMPORTANT: problem only when using Assign, not when using Subscribe()
	con.Assign(new Confluent.Kafka.TopicPartition(topic, 0));

	ConsumeResult<string, byte[]> msg = con.Consume(5000);
	Assert.NotNull(msg);

	con.StoreOffset(msg);

	for (int i=0; i<100; i++)
	{
		if (committed.WaitOne(100))
			return;
		con.Consume(1000);
	}

	throw new Exception("No commit!");
}
```

edenhill:
Interesting, can you reproduce this with `Debug: "cgrp,topic"` and provide the logs?

ulbodk:
Yes, please see attached log.

[librdkafka_log.txt](https://github.com/edenhill/librdkafka/files/6647926/librdkafka_log.txt)

ulbodk:
Yes, please see attached file:

[librdkafka_log.txt](https://github.com/edenhill/librdkafka/files/6648009/librdkafka_log.txt)

hallfox:
We can also replicate this issue with 1.7.0 using just the C API.

hallfox:
[...]

PR #3436, opened 2021-06-17
Fix auto commit not running for non-subscription consumers

hallfox:
Taking a shot at addressing the issue in #3420. Something's amiss with the `rkcg_join_state` at this point so we attempt to avoid the check unless there is a current subscription.


edenhill:
Thanks for your contribution, but closing this in favour of this https://github.com/edenhill/librdkafka/pull/3454 which adds tests.

PR #3455, opened 2021-07-12
Allow only one simultaneous Join or Sync GroupRequest, and make auto.offset.reset more reliable

edenhill:


edenhill:
There should really be an initial.offset=beginning in addition to auto.offset.reset, that is:
if there is no committed offset (which is not an error), use initial.offset.  On error use auto.offset.reset.

CHANGELOG.md
```diff
@@ -20,6 +21,19 @@
...
+   updated the subscription) while there was an outstanding JoinGroup or
+   SyncGroup request, an additional request would sometimes be sent before
+   handling the response of the first. This in turn lead to internal state
+   issues that could cause a crash or malbehaviour.
+   The consumer will now wait for any outstanding JoinGroup or SyncGroup
+   responses before re-joining the group.
+ * `auto.offset.reset` could previously be triggered by temporary errors,
+   such as disconnects and timeouts (after the two retries are exhausted).
+   This is now fixed so that the auto offset reset policy is only triggered
+   for permanent errors.
```
mhowlett, line 34: 👏

src/rdkafka_cgrp.c
```diff
@@ -1462,6 +1525,200 @@ static void rd_kafka_cooperative_protocol_adjust_assignment (
...
+                                rkcg->rkcg_generation_id, member_id,
+                                rkcg->rkcg_rk->rk_conf.group_instance_id);
+                rkcg->rkcg_assignor->rkas_on_assignment_cb(
+                        rkcg->rkcg_assignor,
+                        &(rkcg->rkcg_assignor_state),
+                        assignment, &UserData, cgmd);
+                rd_kafka_consumer_group_metadata_destroy(cgmd);
+        }
+
+        // FIXME: Remove when we're done debugging.
```
mhowlett, line 1594: are we done debugging?

mhowlett (approved):

PR #3477, opened 2021-07-28
Fix DeleteRecords crash on empty partial response

chrisbeard:
Addresses the crash outlined in https://github.com/edenhill/librdkafka/issues/3476


edenhill:
This turned out to be a bit more complex as we want to propagate the error code from the failed request to the admin result.
Will file a PR for that, so closing this, but thanks anyway!

Issue #2934, opened 2020-06-15
Insecure version of zlib.dll used (1.2.8.0)


rkoelbel:
Description
===========
Unfortunately we cannot use librdkafka anymore because it uses an insecure version of zlib.dll. For details see https://www.cvedetails.com/vulnerability-list/vendor_id-72/product_id-1820/GNU-Zlib.html

How to reproduce
================
Check version of zlib.dll (1.2.8.0)

Checklist
=========

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `1.4.2`
 - [x] Apache Kafka version: `2.5.0`
 - [x] Operating system: `Windows 10 (x64)`
 - [x] Critical issue

faiz-usmani:
@edenhill this issue will certainly turn up as a matter of concern for most production set-ups using librdkafka. 
The issue is actually wider on the producer end, where the chances of a Windows machine is much higher, compared to a server side Consumer build. 
On that note, is this something that is surely being fixed for version 1.8.0, and if so what is the tentative date for version 1.8.0's release?

edenhill:
Yes this will be fixed in v1.8.0 which is scheduled for August.

CVEs:
 * CVE-2016-9840: undefined behaviour (compiler dependent) in inflate (decompression) code: this is used by the librdkafka consumer. Risk of successfully exploiting through consumed messages very low.
 * CVE-2016-9841: undefined behaviour (compiler dependent) in inflate code: this is used by the librdkafka consumer. Risk of successfully exploiting through consumed messages seems very low.
 * CVE-2016-9842: undefined behaviour in inflateMark(): this API is not used by librdkafka
 * CVE-2016-9843: issue in crc32_big() which is called from crc32_z(): this API is not used by librdkafka.

PR #3490, opened 2021-08-10
Use vcpkgs for dependencies on Windows

edenhill:
This addresses the zlib upgrade in #2934 and makes sure we'll have up-to-date dependencies going forward.

It also makes adding additional dependencies easier.

### DIFF
--- a/packaging/nuget/packaging.py
+++ b/packaging/nuget/packaging.py
@@ -327,19 +327,19 @@ class NugetPackage (Package):
                                               a.info.get('arch'),
                                               a.info.get('bldtype'))
             if 'toolset' not in a.info:
-                a.info['toolset'] = 'v120'
+                a.info['toolset'] = 'v140'
 
         mappings = [
-            [{'arch': 'x64', 'plat': 'linux', 'fname_glob': 'librdkafka-gcc.tar.gz'}, './include/librdkafka/rdkafka.h', 'build/native/include/librdkafka/rdkafka.h'],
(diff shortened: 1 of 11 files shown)


packaging/nuget/packaging.py
```diff
@@ -327,19 +327,19 @@ def build (self, buildtype):
                                               a.info.get('arch'),
                                               a.info.get('bldtype'))
             if 'toolset' not in a.info:
-                a.info['toolset'] = 'v120'
```
mhowlett, line 330: wondering if this affects OOTB compatibility anywhere important, but 2015 was a long time ago...

mhowlett (approved):

jliunyu (approved):

packaging/nuget/packaging.py
edenhill: Yeah, I'm not completely sure either. But we'll find out

src/rdkafka_ssl.c
```diff
@@ -39,6 +39,8 @@
 #ifdef _WIN32
 #include <wincrypt.h>
 #pragma comment (lib, "crypt32.lib")
+#pragma comment (lib, "libcrypto.lib")
+#pragma comment (lib, "libssl.lib")
```
mabrarov, line 43: The names of these libraries (the file names) depend on build configuration. For example, for debug build of OpenSSL with / for MSVC these libraries can be named as libcrypto**d**.lib and libssl**d**.lib.

The CMake project already uses [FindOpenSSL module](https://cmake.org/cmake/help/v3.20/module/FindOpenSSL.html) which finds correct libs itself:

https://github.com/confluentinc/librdkafka/blob/75c43142e1762779f913d60ccb9b9bec931176c7/CMakeLists.txt#L97

so correct libraries should be added into linker options by CMake.

mabrarov: Refer to #5334

Issue #3489, opened 2021-08-09
Request: producer flush ignores linger time


HenningRoigaard:
I would like to request, that the flush method on the producer forces an immediate flush, ignoring any remaining linger time. Currently, the method simply blocks until ongoing linger time has passed.

The primary use case relates to shutdown of services: 
We use a rather high linger of several minutes to achieve effective batching (due to low frequency produce environment). This works fine, and the producer flushes itself from time to time in the async manner that is intended. However, when the service shuts down, we need to ensure that any buffered messages are flushed from the producer. 
Naively, I used the Flush method to accomplish this. But due to blocking behavior, if any messages are buffered, we may have to wait many minutes. This provides are very bad experience for operations, and worse, the service may be terminated before production by an external coordinator (e.g., Kubernetes) which results in (unnecessary) message loss.

This issue is derived from issue [1670](https://github.com/confluentinc/confluent-kafka-dotnet/issues/1670)for the confluent-kafka-dotnet client. See that issue for reproduction and preliminary discussion.

PR #3494, opened 2021-08-11
Override linger.ms when flush()ing (#3489)

edenhill:


mhowlett (approved):
LGTM, nice bit of polish

Issue #3476, opened 2021-07-28
DeleteRecords can crash on empty partial response


chrisbeard:
Description
===========
It appears that DeleteRecords can crash when a broker shuts down mid-request.

We only observe this happening as of `v1.7.0` with an rdkafka client that tunnels connections to the Kafka cluster through an HTTP proxy. The rdkafka client tunnels through an HTTP proxy using the [CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) method via the connect callback. When brokers turn without the proxy, we see no issue. We'd previously been using `v1.6.0` and `v1.6.1` with no issue.

I fully admit, this setup is strange and may very well be to blame, or at least may be causing things to fail in a strange way that librdkafka would otherwise handle gracefully (without the HTTP proxy in the picture).

Here's the walkback from our coredump:
```
#0  rd_kafka_DeleteRecords_response_merge (rko_fanout=0x7fd528006ca0,  
    rko_partial=<optimized out>) at rdkafka_admin.c:3267 
#1  0x0000000000882b01 in rd_kafka_admin_fanout_worker (rk=0x7fd52800fd80,  
    rkq=<optimized out>, rko=0x7fd468009290) at rdkafka_admin.c:1268 
#2  0x0000000000814f44 in rd_kafka_op_call (rk=<optimized out>,  
    rkq=<optimized out>, rko=0x7fd468009290) at rdkafka_op.c:689 
#3  0x0000000000816268 in rd_kafka_op_handle (rk=0x7fd52800fd80,  
    rkq=0x7fd543ffb210, rko=0x7fd468009290, cb_type=RD_KAFKA_Q_CB_CALLBACK,  
    opaque=0x7fd52800fd80, callback=0x7d7110 <rd_kafka_poll_cb>) 
    at rdkafka_op.c:859 
#4  0x000000000080b455 in rd_kafka_q_serve (rkq=0x7fd528006830,  
    timeout_ms=<optimized out>, max_cnt=max_cnt@entry=0,  
    cb_type=cb_type@entry=RD_KAFKA_Q_CB_CALLBACK,  
    callback=callback@entry=0x0, opaque=opaque@entry=0x0) 
    at rdkafka_queue.c:507 
#5  0x00000000007d6884 in rd_kafka_thread_main (arg=arg@entry=0x7fd52800fd80) 
    at rdkafka.c:2079 
[...]

PR #3505, opened 2021-08-18
Fix crash in DeleteRecords on request error

edenhill:
#3476

### DIFF
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -320,7 +320,10 @@ rd_kafka_admin_fanout_worker (rd_kafka_t *rk, rd_kafka_q_t *rkq,
  */
 
 /**
- * @brief Create a new admin_result op based on the request op \p rko_req
+ * @brief Create a new admin_result op based on the request op \p rko_req.
+ *
+ * @remark This moves the rko_req's admin_request.args list from \p rko_req
+ *         to the returned rko. The \p rko_req args will be emptied.
  */
 static rd_kafka_op_t *rd_kafka_admin_result_new (rd_kafka_op_t *rko_req) {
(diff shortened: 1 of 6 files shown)


src/rdkafka_admin.c
```diff
@@ -3257,13 +3266,40 @@ rd_kafka_DeleteRecords_response_merge (rd_kafka_op_t *rko_fanout,
...
+                RD_KAFKA_TPLIST_FOREACH(reqpartition, reqpartitions) {
+                        rd_kafka_topic_partition_t *respart;
+
+                        /* Find result partition */
+                        respart = rd_kafka_topic_partition_list_find(
+                                respartitions,
+                                reqpartition->topic,
+                                reqpartition->partition);
+
+                        rd_assert(respart || !*"respart not found");
```
mhowlett, line 3292: if i'm reading this correctly, this would occur as a result of the broker returning an unexpected set of partitions. I don't think that should be an assert, it should be an error.

mhowlett (approved):
LGTM, except if my comment about the assert is right, you should definitely fix that.

src/rdkafka_admin.c
edenhill: There are alot of partition lists here, so the confusion is real.

respartitions = all partitions in the DeleteRecords() call. These are the result partitions that will be provided back to the app with per-partition errors.
reqpartitions = the partitions in this particular DeleteRecordsRequest. A (copied) subset of respartitions.
partitions = the partitions returned by the broker in the DeleteRecordsResponse.

If there's a mismatch between respartitions and reqpartitions there's a bug in the client, thus the assert.
If there's a mismatch between respartitions and partitions there's a bug in the broker, not an assert.

PR #3507, opened 2021-08-20
Decrease the amount of work being done on Travis-CI

edenhill:
.. to reduce the build minutes since they're now limited.


jliunyu (approved):

shahidhs-ibm:
@edenhill It seems that s390x Travis builds has reduced very much after this PR. Is this done because of Travis credits shortage? If yes, we can try raising request on Travis for increasing credits. Please let us know if you need any help.

PR #3533, opened 2021-09-08
mklove: make zlib test program compilable

chrisnovakovic:
The test program that is used at compile-time to detect whether zlib is available fails to compile due to `NULL` being undefined:

```
_mkltmpyos55w.c:5:20: error: use of undeclared identifier 'NULL'
     z_stream *p = NULL;
                   ^
1 error generated.
```

This means that zlib availability is only automatically detected when using pkg-config.

Import `stddef.h` (which defines `NULL`) in the test program, allowing zlib to be automatically detected via a compilation check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ librdkafka v1.8.0 is a security release:
 
 ### General fixes
 
+ * Correctly detect presence of zlib via compilation check. (Chris Novakovic)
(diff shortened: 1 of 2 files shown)


chrisnovakovic:
The sole failing test doesn't appear to be caused by this change.

edenhill:
Thank you!

PR #3534, opened 2021-09-08
Avoid unused variable warnings for return values used only in assertions

chrisnovakovic:
The return values of many functions are assigned to variables for the sole purpose of using them in assertion conditions, which causes unused variable warnings to be emitted at compile-time when `NDEBUG` is defined. Use some compiler trickery in the definition of `rd_assert()` to avoid unused variable warnings when variables in the assertion condition are only used for that purpose, and ensure that assertions whose conditions contain single-use variables use `rd_assert()` instead of `assert()`.


chrisnovakovic:
Rebased on master.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ librdkafka v2.0.2 is a bugfix release:
 * Fix OpenSSL version in Win32 nuget package (#4152).
 
 
+## Enhancements
```
chrisnovakovic, line 8: @edenhill I'm not sure if this matches what appears to be the new Changelog style post-2.0 - I can modify it if necessary.

PR #3560, opened 2021-09-27
Retrieve jwt token from token provider

jliunyu:
Retrieve jwt from token provider and forward it to the broker:  the example received from provider is listed as the json below, extract the token key with key work "access_token".
```
{
    “access_token”: “eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiIsImtpZCI6ImFiY2VkZmcifQ.eyJleHAiOjE2MzI2ODYwNjAsImlhdCI6MTYzMjY4NTc2MCwic3ViIjoic3ViIn0.yY0F9NUD-MJydlF0NzfNXQtl7gYKO6A-yKM-6_6RTcg”
}
```

Test:
1. The extract jwt from json is tested with unit test.
2. The HTTP(S) request to http server part is tested with the following function which is not added to the unit test part:
```
int unittest_sasl_oauthbearer_oidc (void) {
        rd_kafka_conf_res_t res;

        rd_kafka_conf_t *conf = NULL;
        rd_kafka_t *rk;
        char errstr[512];

        RD_UT_BEGIN();

        conf = rd_kafka_conf_new();
        res = rd_kafka_conf_set(conf, "sasl.oauthbearer.token.endpoint.url",
                                "http://localhost:8080/retrieve", NULL, 0);
        res = rd_kafka_conf_set(conf, "sasl.oauthbearer.client.id",
                                "123", NULL, 0);
        res = rd_kafka_conf_set(conf, "sasl.oauthbearer.client.secret",
                               "abc", NULL, 0);
        res = rd_kafka_conf_set(conf, "sasl.oauthbearer.scope",
                                "test-scope", NULL, 0);
        res = rd_kafka_conf_set(conf, "sasl.oauthbearer.extensions",
                                "grant_type=client_credentials,"
                                "scope=test-scope",
                                NULL, 0);
        rk = rd_kafka_new(RD_KAFKA_CONSUMER, conf, errstr, sizeof(errstr));
        rd_kafka_conf_set_oauthbearer_oidc_token_refresh_cb(rk, NULL, NULL);

        rd_kafka_destroy(rk);
        RD_UT_PASS();
}
```

3. Tested with OKTA as token provider and mock up broker.
```
[0126_oauthbearer_oidc       /  4.993s] [ do_test_produce_consumer_with_OIDC:48: Test producer and consumer with oidc configuration: PASS (4.99s) ]
[0126_oauthbearer_oidc       /  4.994s] [ do_test_produce_consumer_with_OIDC_should_fail:110: Test OAUTHBEARER/OIDC failing with invalid JWT ]
[0126_oauthbearer_oidc       /  4.995s] Created    kafka instance 0126_oauthbearer_oidc#consumer-3
[<MAIN>                      /  5.026s] 1 test(s) running: 0126_oauthbearer_oidc
[<MAIN>                      /  6.031s] 1 test(s) running: 0126_oauthbearer_oidc
%3|1635444868.772|FAIL|0126_oauthbearer_oidc#consumer-3| [thrd:sasl_plaintext://0.0.0.0:57804/bootstrap]: sasl_plaintext://0.0.0.0:57804/bootstrap: SASL authentication error: {"status":"invalid_token"} (after 313ms in state AUTH_REQ)
[0126_oauthbearer_oidc       /  6.314s] Expected error: Local: Authentication failure: sasl_plaintext://0.0.0.0:57804/bootstrap: SASL authentication error: {"status":"invalid_token"} (after 313ms in state AUTH_REQ)
[0126_oauthbearer_oidc       /  6.314s] Closing consumer 0126_oauthbearer_oidc#consumer-3
[0126_oauthbearer_oidc       /  6.314s] CONSUMER.CLOSE: duration 0.050ms
[0126_oauthbearer_oidc       /  6.315s] [ do_test_produce_consumer_with_OIDC_should_fail:110: Test OAUTHBEARER/OIDC failing with invalid JWT: PASS (1.32s) ]
[0126_oauthbearer_oidc       /  6.315s] 0126_oauthbearer_oidc: duration 6314.545ms
[0126_oauthbearer_oidc       /  6.315s] ================= Test 0126_oauthbearer_oidc PASSED =================
[<MAIN>                      /  7.036s] ALL-TESTS: duration 7035.850ms
TEST 20211028111429 (bare, scenario default) SUMMARY
| <MAIN>                                   |     PASSED |   7.037s |
| 0126_oauthbearer_oidc                    |     PASSED |   6.315s |
[<MAIN>                      /  7.041s] 0 thread(s) in use by librdkafka
[<MAIN>                      /  7.041s] 
###
###  ./test-runner in bare mode PASSED! ###
###
bash-3.2$ 
```

4. The build failure from Windows is not related to this PR:
```
[100%] Built target test-runner
Install the project...
-- Install configuration: ""
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/cmake/RdKafka/RdKafkaConfig.cmake
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/cmake/RdKafka/RdKafkaConfigVersion.cmake
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/cmake/RdKafka/FindLZ4.cmake
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/cmake/RdKafka/RdKafkaTargets.cmake
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/cmake/RdKafka/RdKafkaTargets-noconfig.cmake
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/share/licenses/librdkafka/LICENSES.txt
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/pkgconfig/rdkafka-static.pc
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/librdkafka.a
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/include/librdkafka/rdkafka.h
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/include/librdkafka/rdkafka_mock.h
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/pkgconfig/rdkafka++-static.pc
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/lib/librdkafka++.a
-- Installing: C:/Users/travis/build/edenhill/librdkafka/dest/include/librdkafka/rdkafkacpp.h
~/build/edenhill/librdkafka/mergescratch ~/build/edenhill/librdkafka
~/build/edenhill/librdkafka
```
5. Tested with mock up token provider.
```
[0126_oauthbearer_oidc       /  7.192s] 0126_oauthbearer_oidc: duration 7191.931ms
[0126_oauthbearer_oidc       /  7.192s] ================= Test 0126_oauthbearer_oidc PASSED =================
[<MAIN>                      /  8.018s] ALL-TESTS: duration 8018.276ms
TEST 20211115233334 (bare, scenario default) SUMMARY
| <MAIN>                                   |     PASSED |   8.019s |
| 0126_oauthbearer_oidc                    |     PASSED |   7.192s |
[<MAIN>                      /  8.021s] 0 thread(s) in use by librdkafka
[<MAIN>                      /  8.021s] 
###
###  ./test-runner in bare mode PASSED! ###
###
```

6. valgrind: TESTS=0126 ./run-test.sh valgrind --exit-on-first-error=no
Saw two leaks, but I don't think they are related to my change.
(1) related to rd_http_req_destroy(&hreq)
```==116526== HEAP SUMMARY:
==116526==     in use at exit: 80 bytes in 1 blocks
==116526==   total heap usage: 6,618 allocs, 6,617 frees, 433,555 bytes allocated
==116526== 
==116526== 80 bytes in 1 blocks are definitely lost in loss record 1 of 1
==116526==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==116526==    by 0x4A051CC: rd_malloc (rd.h:131)
==116526==    by 0x4A05F26: rd_buf_new (rdbuf.c:380)
==116526==    by 0x4A622C9: rd_http_req_init (rdhttp.c:139)
==116526==    by 0x4A65468: ut_sasl_oauthbearer_oidc_should_succeed (rdkafka_sasl_oauthbearer_oidc.c:410)
==116526==    by 0x4A6598B: unittest_sasl_oauthbearer_oidc (rdkafka_sasl_oauthbearer_oidc.c:504)
==116526==    by 0x4A116FF: rd_unittest (rdunittest.c:497)
==116526==    by 0x48DEFCC: rd_kafka_unittest (rdkafka.c:4860)
==116526==    by 0x1A5A3B: main_0000_unittests (0000-unittests.c:68)
==116526==    by 0x2884B7: run_test0 (test.c:1103)
==116526==    by 0x288ADD: run_test_from_thread (test.c:1167)
==116526==    by 0x4B686D9: start_thread (pthread_create.c:474)
==116526== 
==116526== 
```
Saw the same error for rdhttp.c.
```
==110197== 80 bytes in 1 blocks are definitely lost in loss record 1 of 2
==110197==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==110197==    by 0x4A051CC: rd_malloc (rd.h:131)
==110197==    by 0x4A05F26: rd_buf_new (rdbuf.c:380)
==110197==    by 0x4A622EF: rd_http_req_init (rdhttp.c:139)
==110197==    by 0x4A62B9E: rd_http_get_json (rdhttp.c:388)
==110197==    by 0x4A63009: unittest_http (rdhttp.c:475)
==110197==    by 0x4A11725: rd_unittest (rdunittest.c:497)
==110197==    by 0x48DEFCC: rd_kafka_unittest (rdkafka.c:4860)
==110197==    by 0x1A5A3B: main_0000_unittests (0000-unittests.c:68)
==110197==    by 0x2884B7: run_test0 (test.c:1103)
==110197==    by 0x288ADD: run_test_from_thread (test.c:1167)
==110197==    by 0x4B686D9: start_thread (pthread_create.c:474)
```
(2) Related to test_create_topic()
```
HEAP SUMMARY:
==121240==     in use at exit: 5,032 bytes in 70 blocks
==121240==   total heap usage: 6,830 allocs, 6,760 frees, 1,155,127 bytes allocated
==121240== 
==121240== 8 bytes in 1 blocks are indirectly lost in loss record 1 of 13
==121240==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==121240==    by 0x4999A2B: rd_calloc (rd.h:125)
==121240==    by 0x4999E88: rd_list_prealloc_elems (rdlist.c:108)
==121240==    by 0x4999D88: rd_list_init_copy (rdlist.c:69)
==121240==    by 0x49D3DB6: rd_kafka_NewTopic_copy (rdkafka_admin.c:1479)
==121240==    by 0x49D68D1: rd_kafka_CreateTopics (rdkafka_admin.c:1739)
==121240==    by 0x299822: test_admin_create_topic (test.c:4585)
==121240==    by 0x299C99: test_create_topic (test.c:4653)
==121240==    by 0x1B0B60: do_test_produce_consumer_with_OIDC (0126-oauthbearer_oidc.c:66)
==121240==    by 0x1B1876: main_0126_oauthbearer_oidc (0126-oauthbearer_oidc.c:206)
==121240==    by 0x2884C3: run_test0 (test.c:1103)
==121240==    by 0x288AE9: run_test_from_thread (test.c:1167)
==121240== 
==121240== 
==121240== Exit program on first error (--exit-on-first-error=yes)

```
Saw the same error for 0090-idempotence.c
```
HEAP SUMMARY:
==120515==     in use at exit: 340 bytes in 8 blocks
==120515==   total heap usage: 6,359 allocs, 6,351 frees, 2,160,748 bytes allocated
==120515== 
==120515== 8 bytes in 1 blocks are indirectly lost in loss record 1 of 7
==120515==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==120515==    by 0x4999A2B: rd_calloc (rd.h:125)
==120515==    by 0x4999E88: rd_list_prealloc_elems (rdlist.c:108)
==120515==    by 0x4999D88: rd_list_init_copy (rdlist.c:69)
==120515==    by 0x49D3DB6: rd_kafka_NewTopic_copy (rdkafka_admin.c:1479)
==120515==    by 0x49D68D1: rd_kafka_CreateTopics (rdkafka_admin.c:1739)
==120515==    by 0x299822: test_admin_create_topic (test.c:4585)
==120515==    by 0x299C99: test_create_topic (test.c:4653)
==120515==    by 0x176AB0: do_test_implicit_ack (0090-idempotence.c:133)
==120515==    by 0x177078: main_0090_idempotence (0090-idempotence.c:165)
==120515==    by 0x2884C3: run_test0 (test.c:1103)
==120515==    by 0x288AE9: run_test_from_thread (test.c:1167)
==120515== 
==120515== 

```
Since the above errors are not related to this PR, I will dig more and send a new PR to fix.

7. Tested with manually set the token urls:
```
VALID_OIDC_URL=http://localhost:8080/retrieve INVALID_OIDC_URL=http://localhost:8080/retrieve/badformat EXPIRED_TOKEN_OIDC_URL=http://localhost:8080/retrieve/expire TESTS=0126 make

[0126_oauthbearer_oidc       / 29.667s] ================= Test 0126_oauthbearer_oidc PASSED =================
[<MAIN>                      / 30.100s] ALL-TESTS: duration 30099.377ms
TEST 20211121131641 (bare, scenario default) SUMMARY
| <MAIN>                                   |     PASSED |  30.100s |
| 0126_oauthbearer_oidc                    |     PASSED |  29.667s |
[<MAIN>                      / 30.101s] 0 thread(s) in use by librdkafka
[<MAIN>                      / 30.101s] 
###
###  ./test-runner in bare mode PASSED! ###
###

```

### DIFF
--- a/src/rdhttp.c
+++ b/src/rdhttp.c
@@ -224,14 +224,14 @@ rd_http_error_t *rd_http_get(const char *url, rd_buf_t **rbufp) {
 
 
 /**
- * @brief Extract the json string from \p hreq and returned it from \p *jsonp.
+ * @brief Extract the JSON object from \p hreq and return it in \p *jsonp.
  *
  * @returns Returns NULL on success, or an JSON parsing error - this
  *          error object must be destroyed by calling rd_http_error_destroy().
  */
-rd_http_error_t *rd_http_parse_json (rd_http_req_t *hreq, cJSON **jsonp) {
-        size_t len = 0;
-        char *raw_json = NULL;
+rd_http_error_t *rd_http_parse_json(rd_http_req_t *hreq, cJSON **jsonp) {
+        size_t len;
+        char *raw_json;
         const char *end = NULL;
         rd_slice_t slice;
         rd_http_error_t *herr = NULL;
@@ -250,7 +250,7 @@ rd_http_error_t *rd_http_parse_json (rd_http_req_t *hreq, cJSON **jsonp) {
         if (!*jsonp)
                 herr = rd_http_error_new(hreq->hreq_code,
                                          "Failed to parse JSON response "
-                                         "at %"PRIusz"/%"PRIusz,
+                                         "at %" PRIusz "/%" PRIusz,
                                          (size_t)(end - raw_json), len);
         rd_free(raw_json);
         return herr;
@@ -258,33 +258,34 @@ rd_http_error_t *rd_http_parse_json (rd_http_req_t *hreq, cJSON **jsonp) {
 
 
 /**
- * @brief Check if the error returned from HTTP(S) is tempoorary or not.
+ * @brief Check if the error returned from HTTP(S) is temporary or not.
  *
  * @returns If the \p error_code is temporary, return rd_true,
  *          otherwise return rd_false.
  *
  * @locality Any thread.
  */
-static rd_bool_t rd_http_is_failure_temporary (int error_code) {
-        switch (error_code)
-        {
-                case REQUEST_TIMEOUT:
-                case TOO_EARLY:
-                case INTERNAL_SERVER_ERROR:
-                case BAD_GATEWAY:
-                case SERVICE_UNAVAILABLE:
-                case GATEWAY_TIMEOUT:
-                     return rd_true;
+static rd_bool_t rd_http_is_failure_temporary(int error_code) {
+        switch (error_code) {
+        case 408: /**< Request timeout */
+        case 425: /**< Too early */
+        case 500: /**< Internal server error */
+        case 502: /**< Bad gateway */
+        case 503: /**< Service unavailable */
+        case 504: /**< Gateway timeout */
+                return rd_true;
+
+        default:
+                return rd_false;
         }
-        return rd_false;
 }
 
 
 /**
- * @brief Perform a blocking HTTP(S) request to token url from \p rk with
- *        HTTP(S) headers and data with \p timeout.
+ * @brief Perform a blocking HTTP(S) request to \p url with
+ *        HTTP(S) headers and data with \p timeout_s.
  *        If the HTTP(S) request fails, will retry another \p retries times
- *        with multiplying backoff \p interval.
+ *        with multiplying backoff \p retry_ms.
  *
  * @returns The result will be returned in \p *jsonp.
  *          Returns NULL on success (HTTP response code < 400), or an error
@@ -293,95 +294,70 @@ static rd_bool_t rd_http_is_failure_temporary (int error_code) {
  *
  * @locality Any thread.
  */
-rd_http_error_t
-*rd_http_post_expect_json (rd_kafka_t *rk,
-                           const struct curl_slist *headers,
-                           const char *post_fields,
-                           size_t post_fields_size,
-                           long timeout,
-                           int retries,
-                           int interval,
-                           cJSON **jsonp) {
+rd_http_error_t *rd_http_post_expect_json(rd_kafka_t *rk,
+                                          const char *url,
+                                          const struct curl_slist *headers,
+                                          const char *post_fields,
+                                          size_t post_fields_size,
+                                          rd_ts_t timeout_s,
+                                          int retries,
+                                          int retry_ms,
+                                          cJSON **jsonp) {
         rd_http_error_t *herr;
         rd_http_req_t hreq;
         int i;
         size_t len;
         const char *content_type;
 
-        herr = rd_http_req_init(
-                &hreq,
-                rk->rk_conf.sasl.oauthbearer.token_endpoint_url);
+        herr = rd_http_req_init(&hreq, url);
         if (unlikely(herr != NULL))
                 return herr;
 
         curl_easy_setopt(hreq.hreq_curl, CURLOPT_HTTPHEADER, headers);
-        curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT, timeout);
+        curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT, timeout_s);
 
-        curl_easy_setopt(hreq.hreq_curl,
-                         CURLOPT_POSTFIELDSIZE,
+        curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDSIZE,
                          post_fields_size);
         curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDS, post_fields);
 
         for (i = 0; i <= retries && !rd_kafka_terminating(rk); i++) {
                 herr = rd_http_req_perform_sync(&hreq);
-                len = rd_buf_len(hreq.hreq_buf);
+                len  = rd_buf_len(hreq.hreq_buf);
 
-                /* Break the for loop directly if HTTP(S) request is succeed,
-                   return with permanent error code or last retry. */
-                if ((!herr && len > 0) ||
-                    (herr && !rd_http_is_failure_temporary(herr->code))
-                    || i == retries)
-                        break;
-                /* Empty response: should be retry */
-                else if (!herr && len == 0)
-                        rd_kafka_log(rk, LOG_WARNING, "HTTP",
-                                     "HTTP connection returns an empty "
-                                     "JSON string");
                 /* Retry if HTTP(S) request returns temporary error. */
-                else if (herr && rd_http_is_failure_temporary(herr->code)) {
-                        rd_kafka_log(rk, LOG_ERR, "HTTP",
-                                     "Failed to connect to HTTP server with "
-                                     "error code: %d, error string: %s",
-                                    herr->code, herr->errstr);
+                if (i < retries && herr &&
+                    rd_http_is_failure_temporary(herr->code)) {
                         rd_http_error_destroy(herr);
-                }
-
-                rd_usleep(interval * (i + 1), &rk->rk_terminate);
+                        rd_usleep(retry_ms * 1000 * (i + 1), &rk->rk_terminate);
+                } else
+                        break;
         }
 
-        if (herr) {
-                rd_kafka_log(rk, LOG_ERR, "HTTP",
-                             "Failed to connect to HTTP server with "
-                             "error code: %d, error string: %s",
-                             herr->code, herr->errstr);
+        if (herr || len == 0) {
                 rd_http_req_destroy(&hreq);
                 return herr;
         }
 
-        if (len == 0) {
-                /* Empty response: create empty JSON object */
-                rd_kafka_log(rk, LOG_WARNING, "HTTP",
-                             "HTTP connection returns an empty JSON string");
-                *jsonp = cJSON_CreateObject();
-                rd_http_req_destroy(&hreq);
-                return NULL;
-        }
-
         content_type = rd_http_req_get_content_type(&hreq);
 
-        if (!content_type ||
-            rd_strncasecmp(content_type,
-                           "application/json", strlen("application/json"))) {
+        if (!content_type || rd_strncasecmp(content_type, "application/json",
+                                            strlen("application/json"))) {
                 if (!herr)
                         herr = rd_http_error_new(
-                                hreq.hreq_code,
-                                "Response is not JSON encoded: %s",
-                                content_type ? content_type : "(n/a)");
+                            hreq.hreq_code, "Response is not JSON encoded: %s",
+                            content_type ? content_type : "(n/a)");
                 rd_http_req_destroy(&hreq);
                 return herr;
         }
 
         herr = rd_http_parse_json(&hreq, jsonp);
+
+        if (!cJSON_HasObjectItem(*jsonp, "access_token")) {
+                rd_kafka_op_err(rk, RD_KAFKA_RESP_ERR__AUTHENTICATION,
+                                "Expected JSON response with "
+                                "\"access_token\" field");
+        }
+
         rd_http_req_destroy(&hreq);
 
         return herr;
--- a/src/rdhttp.c
+++ b/src/rdhttp.c
@@ -222,6 +222,155 @@ rd_http_error_t *rd_http_get(const char *url, rd_buf_t **rbufp) {
         return NULL;
 }
 
+/**
+ * @brief Extract the jwt from \p *cjson.
+ *
+ * @returns The token will be responsed in \p *res.
+ *          Returns NULL on success or an error
+ *          object on error - this error object must be destroyed
+ *          by calling rd_http_error_destroy().
+ */
+rd_http_error_t *rd_http_extract_jwt_from_json (cJSON **cjson,
+                                               char **res,
+                                               const char *key) {
+        char *token;
+        cJSON *jval;
+        cJSON *parsed_token = NULL;
+        rd_http_error_t *herr;
+
+        cJSON_ArrayForEach(jval, *cjson) {
+                parsed_token = cJSON_GetObjectItem(*cjson, key);
+                break;
+        }
+        if (parsed_token == NULL) {
+                herr = rd_http_error_new(-1,
+                        "Expected non-empty JSON response");
+                return herr;
+        }
+
+        token = cJSON_Print(parsed_token);
+        *res = rd_malloc(strlen(token) + 1);
+        strcpy(*res, token);
+        rd_free(token);
+        return NULL;
+}
+
+
+/**
+ * @brief Extract the json string that includes jwt from \p hreq.
+ *
+ * @returns Returns the response (even if there's a HTTP error code returned)
+ *          in \p *json.
+ */
+rd_http_error_t *rd_http_parse_token_to_json (rd_http_req_t *hreq,
+                                              cJSON **jsonp) {
+        size_t len = 0;
+        char *raw_json = NULL;
+        const char *end = NULL;
+        rd_slice_t slice;
+        rd_http_error_t *herr = NULL;
+        *jsonp = NULL;
+
+        /* cJSON requires the entire input to parse in contiguous memory. */
+        rd_slice_init_full(&slice, hreq->hreq_buf);
+        len = rd_buf_len(hreq->hreq_buf);
+
+        raw_json = rd_malloc(len + 1);
+        rd_slice_read(&slice, raw_json, len);
+        raw_json[len] = '\0';
+
+        /* Parse JSON */
+        end = NULL;
+        *jsonp = cJSON_ParseWithOpts(raw_json, &end, 0);
+
+        if (!*jsonp && !herr)
+                herr = rd_http_error_new(hreq->hreq_code,
+                                         "Failed to parse JSON response "
+                                         "at %"PRIusz"/%"PRIusz,
+                                         (size_t)(end - raw_json), len);
+        rd_free(raw_json);
+        rd_http_req_destroy(hreq);
+        return herr;
+}
+
+
+/**
+ * @brief Perform a blocking HTTP(S) request to \p url with HTTP(S)
+ *        headers and data with 20s timeout.
+ *        If the HTTP(S) request fails, will retry another \p retry times.
+ *
+ * @returns The result will be responsed in \p *jsonp.
+ *          Returns NULL on success (HTTP response code < 400), or an error
+ *          object on transport or HTTP error - this error object must be
+ *          destroyed by calling rd_http_error_destroy().
+ *
+ * @locality Any thread.
+ */
+rd_http_error_t
+*rd_http_post_json (const char *url,
+                    struct curl_slist **headers,
+                    const char *data_to_token,
+                    const size_t data_to_token_size,
+                    const long timeout,
+                    const size_t retry,
+                    cJSON **jsonp) {
+        rd_http_error_t *herr;
+        rd_http_req_t hreq;
+        size_t i;
+        size_t len;
+        const char *content_type;
+
+        herr = rd_http_req_init(&hreq, url);
+        if (unlikely(herr != NULL))
+                return herr;
+
+        curl_easy_setopt(hreq.hreq_curl, CURLOPT_HTTPHEADER, *headers);
+        curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT, timeout);
+
+        curl_easy_setopt(hreq.hreq_curl,
+                         CURLOPT_POSTFIELDSIZE,
+                         data_to_token_size);
+        curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDS, data_to_token);
+
+        for (i = 0; i <= retry; i++) {
+                herr = rd_http_req_perform_sync(&hreq);
+                len = rd_buf_len(hreq.hreq_buf);
+                if (!herr && len > 0) {
+                        break;
+                }
+        }
+
+        if (herr && len == 0) {
+                rd_http_req_destroy(&hreq);
+                return herr;
+        }
+
+        if (len == 0) {
+                /* Empty response: create empty JSON object */
+                *jsonp = cJSON_CreateObject();
+                rd_http_req_destroy(&hreq);
+                return NULL;
+        }
+
+        content_type = rd_http_req_get_content_type(&hreq);
+
+        if (!content_type ||
+            rd_strncasecmp(content_type,
+                           "application/json", strlen("application/json"))) {
+                if (!herr)
+                        herr = rd_http_error_new(
+                                hreq.hreq_code,
+                                "Response is not JSON encoded: %s",
+                                content_type ? content_type : "(n/a)");
+                rd_http_req_destroy(&hreq);
+                return herr;
+        }
+
+        herr = rd_http_parse_token_to_json(&hreq, &*jsonp);
+
+        return herr;
+}
+
 
 /**
  * @brief Same as rd_http_get() but requires a JSON response.
--- a/src/rdhttp.c
+++ b/src/rdhttp.c
@@ -299,7 +299,7 @@ rd_http_error_t *rd_http_post_expect_json(rd_kafka_t *rk,
                                           const struct curl_slist *headers,
                                           const char *post_fields,
                                           size_t post_fields_size,
-                                          rd_ts_t timeout_s,
+                                          int timeout_s,
                                           int retries,
                                           int retry_ms,
                                           cJSON **jsonp) {
@@ -351,12 +351,6 @@ rd_http_error_t *rd_http_post_expect_json(rd_kafka_t *rk,
         }
 
         herr = rd_http_parse_json(&hreq, jsonp);
-
-        if (!cJSON_HasObjectItem(*jsonp, "access_token")) {
-                rd_kafka_op_err(rk, RD_KAFKA_RESP_ERR__AUTHENTICATION,
-                                "Expected JSON response with "
-                                "\"access_token\" field");
-        }
 
         rd_http_req_destroy(&hreq);
 
--- a/src/rdhttp.c
+++ b/src/rdhttp.c
@@ -320,22 +320,32 @@ rd_http_error_t *rd_http_post_expect_json(rd_kafka_t *rk,
                          post_fields_size);
         curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDS, post_fields);
 
-        for (i = 0; i <= retries && !rd_kafka_terminating(rk); i++) {
+        for (i = 0; i <= retries; i++) {
+                if (rd_kafka_terminating(rk)) {
+                        rd_http_req_destroy(&hreq);
+                        return rd_http_error_new(-1, "Terminating");
+                }
+
                 herr = rd_http_req_perform_sync(&hreq);
                 len  = rd_buf_len(hreq.hreq_buf);
 
-                /* Retry if HTTP(S) request returns temporary error. */
-                if (i < retries && herr &&
-                    rd_http_is_failure_temporary(herr->code)) {
-                        rd_http_error_destroy(herr);
-                        rd_usleep(retry_ms * 1000 * (i + 1), &rk->rk_terminate);
-                } else
-                        break;
-        }
+                if (!herr) {
+                        if (len > 0)
+                                break; /* Success */
+                        /* Empty response */
+                        rd_http_req_destroy(&hreq);
+                        return NULL;
+                }
+                /* Retry if HTTP(S) request returns temporary error and there
+                 * are remaining retries, else fail. */
+                if (i == retries || !rd_http_is_failure_temporary(herr->code)) {
+                        rd_http_req_destroy(&hreq);
+                        return herr;
+                }
 
-        if (herr || len == 0) {
-                rd_http_req_destroy(&hreq);
-                return herr;
+                /* Retry */
+                rd_http_error_destroy(herr);
+                rd_usleep(retry_ms * 1000 * (i + 1), &rk->rk_terminate);
         }
 
         content_type = rd_http_req_get_content_type(&hreq);
--- a/src/rdhttp.c
+++ b/src/rdhttp.c
@@ -222,54 +222,19 @@ rd_http_error_t *rd_http_get(const char *url, rd_buf_t **rbufp) {
         return NULL;
 }
 
-/**
- * @brief Extract the jwt from \p *cjson.
- *
- * @returns The token will be responsed in \p *res.
- *          Returns NULL on success or an error
- *          object on error - this error object must be destroyed
- *          by calling rd_http_error_destroy().
- */
-rd_http_error_t *rd_http_extract_jwt_from_json (cJSON **cjson,
-                                               char **res,
-                                               const char *key) {
-        char *token;
-        cJSON *jval;
-        cJSON *parsed_token = NULL;
-        rd_http_error_t *herr;
-
-        cJSON_ArrayForEach(jval, *cjson) {
-                parsed_token = cJSON_GetObjectItem(*cjson, key);
-                break;
-        }
-        if (parsed_token == NULL) {
-                herr = rd_http_error_new(-1,
-                        "Expected non-empty JSON response");
-                return herr;
-        }
-
-        token = cJSON_Print(parsed_token);
-        *res = rd_malloc(strlen(token) + 1);
-        strcpy(*res, token);
-        rd_free(token);
-        return NULL;
-}
-
 
 /**
- * @brief Extract the json string that includes jwt from \p hreq.
+ * @brief Extract the json string from \p hreq and returned it from \p *jsonp.
  *
- * @returns Returns the response (even if there's a HTTP error code returned)
- *          in \p *json.
+ * @returns Returns NULL on success, or an JSON parsing error - this
+ *          error object must be destroyed by calling rd_http_error_destroy().
  */
-rd_http_error_t *rd_http_parse_token_to_json (rd_http_req_t *hreq,
-                                              cJSON **jsonp) {
+rd_http_error_t *rd_http_parse_json (rd_http_req_t *hreq, cJSON **jsonp) {
         size_t len = 0;
         char *raw_json = NULL;
         const char *end = NULL;
         rd_slice_t slice;
         rd_http_error_t *herr = NULL;
-        *jsonp = NULL;
 
         /* cJSON requires the entire input to parse in contiguous memory. */
         rd_slice_init_full(&slice, hreq->hreq_buf);
@@ -280,73 +245,123 @@ rd_http_error_t *rd_http_parse_token_to_json (rd_http_req_t *hreq,
         raw_json[len] = '\0';
 
         /* Parse JSON */
-        end = NULL;
         *jsonp = cJSON_ParseWithOpts(raw_json, &end, 0);
 
-        if (!*jsonp && !herr)
+        if (!*jsonp)
                 herr = rd_http_error_new(hreq->hreq_code,
                                          "Failed to parse JSON response "
                                          "at %"PRIusz"/%"PRIusz,
                                          (size_t)(end - raw_json), len);
         rd_free(raw_json);
-        rd_http_req_destroy(hreq);
         return herr;
 }
 
 
 /**
- * @brief Perform a blocking HTTP(S) request to \p url with HTTP(S)
- *        headers and data with 20s timeout.
- *        If the HTTP(S) request fails, will retry another \p retry times.
+ * @brief Check if the error returned from HTTP(S) is tempoorary or not.
  *
- * @returns The result will be responsed in \p *jsonp.
+ * @returns If the \p error_code is temporary, return rd_true,
+ *          otherwise return rd_false.
+ *
+ * @locality Any thread.
+ */
+static rd_bool_t rd_http_is_failure_temporary (int error_code) {
+        switch (error_code)
+        {
+                case REQUEST_TIMEOUT:
+                case TOO_EARLY:
+                case INTERNAL_SERVER_ERROR:
+                case BAD_GATEWAY:
+                case SERVICE_UNAVAILABLE:
+                case GATEWAY_TIMEOUT:
+                     return rd_true;
+        }
+        return rd_false;
+}
+
+
+/**
+ * @brief Perform a blocking HTTP(S) request to token url from \p rk with
+ *        HTTP(S) headers and data with \p timeout.
+ *        If the HTTP(S) request fails, will retry another \p retries times
+ *        with multiplying backoff \p interval.
+ *
+ * @returns The result will be returned in \p *jsonp.
  *          Returns NULL on success (HTTP response code < 400), or an error
- *          object on transport or HTTP error - this error object must be
- *          destroyed by calling rd_http_error_destroy().
+ *          object on transport, HTTP error or a JSON parsing error - this
+ *          error object must be destroyed by calling rd_http_error_destroy().
  *
  * @locality Any thread.
  */
 rd_http_error_t
-*rd_http_post_json (const char *url,
-                    struct curl_slist **headers,
-                    const char *data_to_token,
-                    const size_t data_to_token_size,
-                    const long timeout,
-                    const size_t retry,
-                    cJSON **jsonp) {
+*rd_http_post_expect_json (rd_kafka_t *rk,
+                           const struct curl_slist *headers,
+                           const char *post_fields,
+                           size_t post_fields_size,
+                           long timeout,
+                           int retries,
+                           int interval,
+                           cJSON **jsonp) {
         rd_http_error_t *herr;
         rd_http_req_t hreq;
-        size_t i;
+        int i;
         size_t len;
         const char *content_type;
 
-        herr = rd_http_req_init(&hreq, url);
+        herr = rd_http_req_init(
+                &hreq,
+                rk->rk_conf.sasl.oauthbearer.token_endpoint_url);
         if (unlikely(herr != NULL))
                 return herr;
 
-        curl_easy_setopt(hreq.hreq_curl, CURLOPT_HTTPHEADER, *headers);
+        curl_easy_setopt(hreq.hreq_curl, CURLOPT_HTTPHEADER, headers);
         curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT, timeout);
 
         curl_easy_setopt(hreq.hreq_curl,
                          CURLOPT_POSTFIELDSIZE,
-                         data_to_token_size);
-        curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDS, data_to_token);
+                         post_fields_size);
+        curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDS, post_fields);
 
-        for (i = 0; i <= retry; i++) {
+        for (i = 0; i <= retries && !rd_kafka_terminating(rk); i++) {
                 herr = rd_http_req_perform_sync(&hreq);
                 len = rd_buf_len(hreq.hreq_buf);
-                if (!herr && len > 0) {
+
+                /* Break the for loop directly if HTTP(S) request is succeed,
+                   return with permanent error code or last retry. */
+                if ((!herr && len > 0) ||
+                    (herr && !rd_http_is_failure_temporary(herr->code))
+                    || i == retries)
                         break;
+                /* Empty response: should be retry */
+                else if (!herr && len == 0)
+                        rd_kafka_log(rk, LOG_WARNING, "HTTP",
+                                     "HTTP connection returns an empty "
+                                     "JSON string");
+                /* Retry if HTTP(S) request returns temporary error. */
+                else if (herr && rd_http_is_failure_temporary(herr->code)) {
+                        rd_kafka_log(rk, LOG_ERR, "HTTP",
+                                     "Failed to connect to HTTP server with "
+                                     "error code: %d, error string: %s",
+                                    herr->code, herr->errstr);
+                        rd_http_error_destroy(herr);
                 }
+
+                rd_usleep(interval * (i + 1), &rk->rk_terminate);
         }
 
-        if (herr && len == 0) {
+        if (herr) {
+                rd_kafka_log(rk, LOG_ERR, "HTTP",
+                             "Failed to connect to HTTP server with "
+                             "error code: %d, error string: %s",
+                             herr->code, herr->errstr);
                 rd_http_req_destroy(&hreq);
                 return herr;
         }
 
         if (len == 0) {
                 /* Empty response: create empty JSON object */
+                rd_kafka_log(rk, LOG_WARNING, "HTTP",
+                             "HTTP connection returns an empty JSON string");
                 *jsonp = cJSON_CreateObject();
                 rd_http_req_destroy(&hreq);
                 return NULL;
@@ -366,7 +381,8 @@ rd_http_error_t
                 return herr;
         }
 
-        herr = rd_http_parse_token_to_json(&hreq, &*jsonp);
+        herr = rd_http_parse_json(&hreq, jsonp);
+        rd_http_req_destroy(&hreq);
 
         return herr;
 }
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2342,25 +2342,6 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
                 rk->rk_conf.security_protocol = RD_KAFKA_PROTO_PLAINTEXT;
         }
 
-        /* Create background thread and queue if background_event_cb()
-         * RD_KAFKA_EVENT_BACKGROUND has been enabled.
-         * Do this before creating the main thread since after
-         * the main thread is created it is no longer trivial to error
-         * out from rd_kafka_new(). */
-
-        if (rk->rk_conf.background_event_cb ||
-            (rk->rk_conf.enabled_events & RD_KAFKA_EVENT_BACKGROUND)) {
-                rd_kafka_resp_err_t err;
-                rd_kafka_wrlock(rk);
-                err =
-                    rd_kafka_background_thread_create(rk, errstr, errstr_size);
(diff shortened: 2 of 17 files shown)


src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+}
+
+
+/**
+ * @brief Implementation of Oauth/OIDC token refresh call back function,
+ *        will receive the json response after HTTP call from token provider,
+ *        then extract the jwt from the json response, and forward it to
+ *        the broker.
+ */
+void rd_kafka_conf_set_oauthbearer_oidc_token_refresh_cb (rd_kafka_t *rk,
```
edenhill, line 195: This function name implies that it sets a token_refresh_cb on a conf object.

Rename to `rd_kafka_oidc_token_refresh_cb()`

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        const char *oauthbearer_config, void *opaque) {
+        static const char *data = "grant_type=client_credentials,scope=%s";
```
edenhill, line 197: Predefining these makes the code harder to read below, like the variable 'data' doesnt tell me much.
I recommend using the strings where they are needed instead, and employ the rule of 3 (duplicate something twice is okay, abstract and unify when used 3 times).

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        char *jwt_token = NULL;
+        struct curl_slist *headers = NULL;
+        char *data_to_token;
+        size_t data_to_token_size;
+        const char *scope;
+        const char *token_url;
+        rd_bool_t build_headers_status;
+        char *error;
+
+        token_url = rk->rk_conf.sasl.oauthbearer.token_endpoint_url;
```
edenhill, line 213: We shouldnt even get here if this is not set, should be checked in conf_finalize

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+                rd_kafka_log(rk, LOG_ERR, "OAUTH/OIDC",
+                             "token_url is a required field");
+                return;
+        }
+
+        build_headers_status = rd_false;
+        build_headers_status = build_headers(&rk->rk_conf, &headers, &error);
+        if (!build_headers_status) {
+                rd_kafka_log(rk, LOG_ERR, "OAUTH/OIDC",
+                             "Failed to build headers: %s",
```
edenhill, line 224: "Failed to build OAUTHBEARER OIDC headers: %s"

But this should be an error op, see rd_kafka_op_err() et.al.

Also, for logging facilities, dont use anything but A-Z. (no /)

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        char *error;
+
+        token_url = rk->rk_conf.sasl.oauthbearer.token_endpoint_url;
+        if (token_url == NULL) {
+                rd_kafka_log(rk, LOG_ERR, "OAUTH/OIDC",
+                             "token_url is a required field");
+                return;
+        }
+
+        build_headers_status = rd_false;
```
edenhill, line 220: This variable is not needed

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+                             "Failed to build headers: %s",
+                             error);
+                rd_free(error);
+                return;
+        }
+
+        /* Build data field */
+        scope = rk->rk_conf.sasl.oauthbearer.scope;
+        if (scope == NULL) {
+                rd_kafka_log(rk, LOG_ERR, "OAUTH/OIDC",
```
edenhill, line 233: should be checked by conf_finalize

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        rd_snprintf(data_to_token, data_to_token_size, data, scope);
+        herr = rd_http_post_json(token_url,
+                                 &headers,
+                                 data_to_token,
+                                 data_to_token_size,
+                                 timeout,
+                                 retry,
+                                 &json);
+        if (unlikely(herr != NULL)) {
+                rd_kafka_log(rk, LOG_ERR, "OAUTH/OIDC",
```
edenhill, line 248: Make this an op_err() error code _AUTH

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+                             "Failed to receive json result from HTTP(S), "
+                             "returned error code: %d, returned error string: "
+                             "%s", herr->code, herr->errstr);
+                rd_http_error_destroy(herr);
+                goto exit;
+        }
+        herr = rd_http_extract_jwt_from_json(&json, &jwt_token, token_key);
+        if (unlikely(herr != NULL)) {
+                rd_kafka_log(rk, LOG_ERR, "OAUTH/OIDC",
```
edenhill, line 257: this too

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+                                 timeout,
+                                 retry,
+                                 &json);
+        if (unlikely(herr != NULL)) {
+                rd_kafka_log(rk, LOG_ERR, "OAUTH/OIDC",
+                             "Failed to receive json result from HTTP(S), "
+                             "returned error code: %d, returned error string: "
+                             "%s", herr->code, herr->errstr);
+                rd_http_error_destroy(herr);
+                goto exit;
```
edenhill, line 253: `exit` may give the reader the idea that `exit()` will be called eventually. We typically use `done` for these types of gotos.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        RD_IF_FREE(headers, curl_slist_free_all);
+        RD_IF_FREE(jwt_token, rd_free);
+        return;
+}
+
+
+/**
+ * @brief make sure the jwt is able to be extracted from HTTP(S) requests.
+ */
+int unittest_sasl_oauthbearer_oidc_should_succeed (void) {
```
edenhill, line 279: use ut_ for the static sub-functions in the file itself. (also make this static)
use unittest_ for the single exported unittest function.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        static const char *token_key = "access_token";
+        static const char *expected_jwt_token =
+                "\"eyJhbGciOiJIUzI1NiIsInR5"
+                "cCI6IkpXVCIsImtpZCI6ImFiY2VkZmcifQ"
+                "."
+                "eyJpYXQiOjE2MzIzNzUzMjAsInN1YiI6InN"
+                "1YiIsImV4cCI6MTYzMjM3NTYyMH0"
+                "."
+                "bT5oY8K-rS2gQ7Awc40844bK3zhzBhZb7sputErqQHY\"";
+        const char *token_format = "{\"%s\":%s}";
```
edenhill, line 290: Dont use formats in variables. Error prone, will give warnings.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        rd_http_req_t hreq;
+        rd_http_error_t *herr;
+        cJSON *json = NULL;
+        char *token = NULL;
+
+        RD_UT_BEGIN();
+
+        herr = rd_http_req_init(&hreq, "");
+        RD_UT_ASSERT(!herr,
+                     "Expected initializa succeed, "
```
edenhill, line 302: typo

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+        cJSON_Delete(json);
+        RD_UT_PASS();
+}
+
+
+/**
+ * @brief make sure the return expected error if received empty json
+ *        from HTTP(S)
+ */
+int unittest_sasl_oauthbearer_oidc_with_empty_key (void) {
```
edenhill, line 346: Same comments for this function as the previous one

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,397 @@
...
+                     "but failed with error code: %d, error string: %s",
+                     herr->code, herr->errstr);
+
+        token_len = strlen(empty_token_format);
+
+        hreq.hreq_buf = rd_buf_new(1, token_len + 1);
+        rd_buf_write(hreq.hreq_buf , empty_token_format, token_len);
+
+        herr = rd_http_parse_token_to_json(&hreq, &json);
+
```
edenhill, line 372: These are good tests 👍

src/rdunittest.c
```diff
@@ -47,6 +47,7 @@
 
 #include "rdsysqueue.h"
 #include "rdkafka_sasl_oauthbearer.h"
+#include "rdkafka_sasl_oauthbearer_oidc.h"
```
edenhill, line 50: should be WITH_CURL protected

edenhill (changes requested):

src/rdhttp.c
```diff
@@ -223,6 +223,155 @@ rd_http_error_t *rd_http_get (const char *url, rd_buf_t **rbufp) {
...
+ *          Returns NULL on success (HTTP response code < 400), or an error
+ *          object on transport or HTTP error - this error object must be
+ *          destroyed by calling rd_http_error_destroy().
+ *
+ * @locality Any thread.
+ */
+rd_http_error_t
+*rd_http_post_json (const char *url,
+                    struct curl_slist **headers,
+                    const char *data_to_token,
```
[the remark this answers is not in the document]
jliunyu: Hi @edenhill, thanks for reviewing.

Yeah, the function is actually post headers and data and expect json response, sorry for the naming.

For the content-type, it's one of fields in the header, the default one from the python HTTP server is "application/x-www-form-urlencoded", same as our expectation. So I didn't add it to the header, but I will add it in case no surprise from other provider.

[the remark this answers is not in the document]
edenhill: Sounds good, we want to be explicit and not rely on one server's default value :+1:

tests/0126-oauthbearer_oidc.c
```diff
@@ -33,64 +33,140 @@
...
+        rd_kafka_t *c1;
+        int msgcounter;
+        rd_kafka_topic_t *rkt;
 
-        SUB_TEST("Test producer with oidc configuration");
+        SUB_TEST("Test producer and consumer with oidc configuration");
 
-        test_conf_init(&conf, NULL, 60);
+        test_conf_set(conf, "sasl.oauthbearer.token.endpoint.url",
+                      "https://dev-531534.okta.com/oauth2/default/v1/token");
```
edenhill, line 56: Where's this URL coming from? Who owns it?

tests/0126-oauthbearer_oidc.c
```diff
@@ -33,64 +33,140 @@
...
         const char *topic;
         uint64_t testid;
-        rd_kafka_t *rk;
-        rd_kafka_conf_t *conf;
-        rd_kafka_conf_res_t res;
-        char errstr[512];
+        rd_kafka_t *p1;
+        rd_kafka_t *c1;
+        int msgcounter;
+        rd_kafka_topic_t *rkt;
```
edenhill, line 51: Avoid using rkt in tests, it is a complex (and somewhat legacy) interface. There are test methods that take a topic string.

tests/0126-oauthbearer_oidc.c
```diff
@@ -33,64 +33,140 @@
...
 /**
- * @brief After config OIDC fields, make sure the producer gets created
- *        successfully.
+ * @brief After config OIDC, make sure the producer and consumer
+ *        can work successfully.
  *
  */
-static void do_test_create_producer() {
+static void do_test_produce_consumer_with_OIDC(rd_kafka_conf_t *conf) {
+        if (test_needs_auth()) {
```
edenhill, line 41: Doesn't this test actually require auth to be configured?

See comment in main function below.

tests/0126-oauthbearer_oidc.c
```diff
@@ -33,64 +33,140 @@
...
-        rd_kafka_destroy(rk);
+        rd_kafka_consumer_poll(c1, 10 * 1000);
 
+        test_consumer_close(c1);
+        rd_kafka_destroy(c1);
         SUB_TEST_PASS();
 }
 
 
 int main_0126_oauthbearer_oidc(int argc, char **argv) {
```
edenhill, line 138: Since these tests supposedly requires a cluster configured with SASL OAUTHBEARER OIDC I think you need to:
 * skip all tests here if OIDC is not configured
 * rely on the test conf (from test_conf_init()) to have correct OIDC configurations, since htat is set up by trivup, et.al.

tests/0126-oauthbearer_oidc.c
```diff
@@ -33,64 +33,140 @@
...
+        TEST_SAY("Topic: %s is created\n", topic);
+
+        rkt = test_create_producer_topic(p1, topic);
+        TEST_SAY("Auto-creating topic %s\n", topic);
+        test_auto_create_topic_rkt(p1, rkt, tmout_multip(5000));
+
+        test_produce_msgs_nowait(p1, rkt, testid, 0, 0, 1, NULL, 0, 0,
+                                 &msgcounter);
+        TEST_SAY("Flushing..\n");
+        rd_kafka_flush(p1, 1000);
```
edenhill, line 75: 1s is pretty short, and you are not checking that the flush succeeded.
Use test_flush() with a timeout of 10 or so.

edenhill: But since you are flushing after nowait you can just as well just use `test_produce_msgs2()` which will perform the wait for you

tests/0126-oauthbearer_oidc.c
```diff
@@ -33,64 +33,140 @@
...
-                      "supportFeatureX=true");
-        test_conf_set(conf, "sasl.oauthbearer.token.endpoint.url",
-                      "https://localhost:1/token");
+        SUB_TEST("Test OAUTHBEARER/OIDC failing with invalid JWT");
 
-        testid = test_id_generate();
-        rd_kafka_conf_set_dr_msg_cb(conf, test_dr_msg_cb);
-        rk = test_create_handle(RD_KAFKA_PRODUCER, conf);
+        test_conf_set(conf, "sasl.oauthbearer.token.endpoint.url",
+                      "http://localhost:8080/retrieve/badformat");
```
edenhill, line 121: Is this supposed to be a URL that doesnt work?
If so, please use a port less likely to have something running. E.g., 2 or 47637 or something.

tests/0126-oauthbearer_oidc.c
```diff
@@ -33,64 +33,140 @@
...
-        test_produce_msgs2(rk, topic, testid, 1, 0, 0, NULL, 0);
+        testid = test_id_generate();
 
-        /* Verify messages were actually produced by consuming them back. */
-        test_consume_msgs_easy(topic, topic, 0, 1, 1, NULL);
+        c1 = test_create_consumer("OIDC.fail.C1", NULL, rd_kafka_conf_dup(conf),
+                                  NULL);
 
-        rd_kafka_destroy(rk);
+        rd_kafka_consumer_poll(c1, 10 * 1000);
```
edenhill, line 130: Nothing checks that the auth error callback was actually fired with the expected error code.

edenhill (changes requested):

src/rdkafka.c
```diff
@@ -2330,6 +2342,24 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
                 rk->rk_conf.security_protocol = RD_KAFKA_PROTO_PLAINTEXT;
         }
 
+        /* Create background thread and queue if background_event_cb()
```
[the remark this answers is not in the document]
jliunyu: Since rd_kafka_sasl_oauthbearer_init(rk) called rd_kafka_sasl_background_callbacks_enable() already, it will create a new background thread.  if this part is after calling rd_kafka_sasl_oauthbearer_init(), it will fail to create background thread.

src/rdkafka.c
```diff
@@ -2238,11 +2241,20 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
...
+            rk->rk_conf.sasl.oauthbearer.method !=
+                RD_KAFKA_SASL_OAUTHBEARER_METHOD_OIDC)
                 rk->rk_conf.enabled_events |=
                     RD_KAFKA_EVENT_OAUTHBEARER_TOKEN_REFRESH;
 #endif
 
+#if WITH_CURL
+        if (rk->rk_conf.sasl.oauthbearer.method ==
+                RD_KAFKA_SASL_OAUTHBEARER_METHOD_OIDC &&
+            !rk->rk_conf.sasl.oauthbearer.token_refresh_cb)
```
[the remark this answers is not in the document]
jliunyu: I saw the similar operation for the existing oauthbearer is here, should we separate these two?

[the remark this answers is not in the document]
jliunyu: > I think this should be done in conf_finalize() instead.

I saw the similar operation for the existing oauthbearer is here, should we separate these two?

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,356 @@
...
+        rd_free(client_authorization_in.ptr);
+        rd_free(client_authorization_out.ptr);
+}
+
+
+/**
+ * @brief Build headers for HTTP(S) requests based on \p client_id
+ *        and \p client_secret. The result will be returned in \p *headers.
+ *
+ * @returns If succeed, return NULL, else return the error message.
```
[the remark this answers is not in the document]
jliunyu: I'm sorry, I changed it void, but this part is back caused by rebase.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,356 @@
...
+        jwt_token     = parsed_token->valuestring;
+        extension_str = rk->rk_conf.sasl.oauthbearer.extensions_str;
+        extension =
+            rd_string_split(extension_str, ',', rd_true, &extension_cnt);
+
+        extension_key_value = rd_kafka_conf_kv_split(
+            (const char **)extension, extension_cnt, &extension_key_value_cnt);
+
+        if (rd_kafka_oauthbearer_set_token(
+                rk, jwt_token, rd_uclock(), "",
```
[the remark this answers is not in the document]
jliunyu: The token expires is actually in the JWT itself. The broker doesn't use the expires here actually.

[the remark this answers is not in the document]
edenhill: If you look at the set_token() implementation there's two things that makes the current time problematic:
 1. it checks that the lifetime is in the future, which will cause your set_token() calls to fail (was this observed? if not, make sure the lifetime is sent as milliseconds since epoch, not microseconds)
 2. the refresh interval timer is scheduled for 80% of the lifetime.

src/rdkafka.c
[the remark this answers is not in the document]
edenhill: Huhm okay, maybe it is better to leave the original order of the code and instead check to see that the background has not been created (check rk->rk_background.q while holding the wrlock)

src/rdkafka.c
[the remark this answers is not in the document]
edenhill: Okay, makes sense.

src/rdkafka_sasl_oauthbearer_oidc.c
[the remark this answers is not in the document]
jliunyu: -> "was this observed?" 
I didn't see any error caused by this before. 

Thanks, I will update this part.

tests/0126-oauthbearer_oidc.c
jliunyu: This is from okta development account, we can use the provider from there, "https://dev-531534.okta.com/oauth2/default/v1/token" can provide us a JWT.

tests/0126-oauthbearer_oidc.c
jliunyu: This URL is valid in our trivup, but will generated a badformat JWT so we can use it for negative test.

src/rdkafka_conf.c
```diff
@@ -3686,12 +3685,46 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                        if (!conf->sasl.oauthbearer.scope) {
+                                return "`sasl.oauthbearer.scope` "
+                                       "is mandatory when "
+                                       "`sasl.oauthbearer.method` is set";
+                        }
+
+                        if (!conf->sasl.oauthbearer.extensions_str) {
+                                return "`sasl.oauthbearer.extensions` "
+                                       "is mandatory when "
+                                       "`sasl.oauthbearer.method` is set";
```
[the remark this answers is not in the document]
jliunyu: Synced up offline, this is mandatory.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,356 @@
...
+                             "token from \"%s\": %s (%d)",
+                             token_url, herr->errstr, herr->code);
+                rd_kafka_oauthbearer_set_token_failure(rk, herr->errstr);
+                rd_http_error_destroy(herr);
+                goto done;
+        }
+
+        parsed_token = cJSON_GetObjectItem(json, "access_token");
+
+        jwt_token     = parsed_token->valuestring;
```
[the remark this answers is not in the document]
jliunyu: cJSON_GetStringValue() will do a type check first, then get the value. I will use cJSON_GetStringValue() to get the string value.
https://github.com/edenhill/librdkafka/blob/master/src/cJSON.c#L95

tests/0126-oauthbearer_oidc.c
```diff
@@ -31,52 +31,41 @@
...
-        TEST_SAY("Flushing..\n");
-        rd_kafka_flush(p1, 1000);
+        test_produce_msgs2(p1, topic, testid, 0, 0, 5, NULL, 0);
 
         test_conf_set(conf, "auto.offset.reset", "earliest");
         c1 = test_create_consumer(topic, NULL, rd_kafka_conf_dup(conf), NULL);
         test_consumer_subscribe(c1, topic);
+
+        /* If token valid time is less than 3s, it will fresh the toke. */
+        rd_usleep(3 * 1000, NULL);
```
edenhill, line 68: usleep is microseconds, so this is 3 ms, not 3s.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+
+        if (!cJSON_HasObjectItem(json, "access_token")) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON response with "
+                    "\"access_token\" field");
+                goto done;
+        }
+
+        parsed_token = cJSON_GetObjectItem(json, "access_token");
```
edenhill, line 325: Can you call GetObjectItem() directly and check it for NULL? It voids the need for the separate HasObjectItem call above.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+        jwt_token = cJSON_GetStringValue(parsed_token);
+        if (jwt_token == NULL) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON "
+                    "response as a value string");
+                goto done;
+        }
+
+        decoded_payloads = rd_kafka_jwt_b64_decode(jwt_token);
```
edenhill, line 336: Needs error handling if the jwt_token is malformed.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+        payloads         = cJSON_Parse(decoded_payloads);
```
edenhill, line 337: Needs error handling if not valid JSON.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+
+        if (!cJSON_HasObjectItem(payloads, "exp")) {
```
edenhill, line 339: Same here, can we just call GetObjectItem()?  From looking at the source it seems HasObjectItem performs a GetObjectItem, so no need to do it twice for each field.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON response with "
```
edenhill, line 342: Since this is not in the outer JSON payload it needs to be more specific and say something like: "Expected JSON JWT response with .."

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+#include "cJSON.h"
+#include <curl/curl.h>
+#include <math.h>
+#include "rdhttp.h"
+#include "rdkafka_sasl_oauthbearer_oidc.h"
+
+/* define isnan for ANSI C, if in C99 or above, isnan has
+ * been defined in math.h */
+#ifndef isnan
+#define isnan(d) (d != d)
```
edenhill, line 45: Huhm, do you have a link or some reference that this implementation is correct and portable?

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+        if (!cJSON_HasObjectItem(payloads, "exp")) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON response with "
+                    "\"exp\" field");
+                goto done;
+        }
+        jwt_exp = cJSON_GetObjectItem(payloads, "exp");
+        exp     = cJSON_GetNumberValue(jwt_exp);
+        if (isnan(exp)) {
```
edenhill, line 348: Maybe we can make this even simpler and not rely on isnan by checking that exp is positive (e.g., greater than 0.1 or something)? We'll want exp to be in the future anyway.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON response with "
```
edenhill, line 351: JSON JWT

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+                    "valid \"exp\" field");
+                goto done;
+        }
+
+        if (!cJSON_HasObjectItem(payloads, "sub")) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON response with "
```
edenhill, line 359: JSON JWT ..

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+        sub     = cJSON_GetStringValue(jwt_sub);
+        if (sub == NULL) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON response with "
+                    "valid \"sub\" field");
+                goto done;
+        }
+
+        extension_str = rk->rk_conf.sasl.oauthbearer.extensions_str;
```
edenhill, line 373: nit: since this variable is only used once I recommend not having it and simply passing rk->rk..extension_str directly to string_split. Less noise.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+        extensions =
+            rd_string_split(extension_str, ',', rd_true, &extension_cnt);
+
+        extension_key_value = rd_kafka_conf_kv_split(
```
edenhill, line 377: this memory is never freed

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+        if (sub == NULL) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk,
+                    "Expected JSON response with "
+                    "valid \"sub\" field");
+                goto done;
+        }
+
+        extension_str = rk->rk_conf.sasl.oauthbearer.extensions_str;
+        extensions =
```
edenhill, line 374: this memory is never freed

src/rdkafka_sasl_oauthbearer_oidc.c
edenhill: Run the tests with valgrind to find memory leaks. E.g., `TESTS=0126 ./run-tests.sh valgrind` (on linux)

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+        cJSON *json = NULL;
+        char *token;
+        cJSON *parsed_token;
+
+        RD_UT_BEGIN();
+
+        herr = rd_http_req_init(&hreq, "");
+
+        RD_UT_ASSERT(!herr,
+                     "Expected initialize succeed, "
```
edenhill, line 436: to succeed

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+                     "but failed with error code: %d, error string: %s",
+                     herr->code, herr->errstr);
+
+        token_len = strlen("access_token") + strlen(expected_jwt_token) + 8;
```
edenhill, line 440: What's the +8?

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,529 @@
...
+                     "Expected JSON token parsing to succeed, "
+                     "but it failed with error code: %d, error string: %s",
+                     herr->code, herr->errstr);
+
+        RD_UT_ASSERT(json, "Expected non-empty json.");
+
+        parsed_token = cJSON_GetObjectItem(json, "access_token");
+
+        RD_UT_ASSERT(!parsed_token,
+                     "Did not expected access_token in JSON response");
```
edenhill, line 510: Did not expect

edenhill (changes requested):

tests/0126-oauthbearer_oidc.c
```diff
@@ -31,66 +31,148 @@
...
+        test_create_topic(p1, topic, 1, 3);
+        TEST_SAY("Topic: %s is created\n", topic);
+
+        test_produce_msgs2(p1, topic, testid, 0, 0, 5, NULL, 0);
+
+        test_conf_set(conf, "auto.offset.reset", "earliest");
+        c1 = test_create_consumer(topic, NULL, rd_kafka_conf_dup(conf), NULL);
+        test_consumer_subscribe(c1, topic);
+
+        /* If token valid time is less than 3s, it will fresh the toke. */
```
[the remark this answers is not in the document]
jliunyu: Add this sleep for to make the test a little bit longer, to cover the refresh token part.

src/rdkafka_sasl_oauthbearer_oidc.c
jliunyu: I used the cJSON.c file as reference: https://github.com/edenhill/librdkafka/blob/master/src/cJSON.c#L72

src/rdkafka_sasl_oauthbearer_oidc.c
jliunyu: I used the isnan() because the cJSON_GetNumberValue will return NAN if the "exp" is not a number.

edenhill: Problem is there is no guarantee that isnan() in this file and in cJSON.c behaves the same way. One definition might be from the fallback define, while the other one is using a gcc builtin, all depending on include file order and stuff.  This makes it hard to guarantee that the outcome is identical on all platforms, compilers, architectures, etc, and why I'm urging a different approach, even though it might seem uglier.

src/rdkafka_sasl_oauthbearer_oidc.c
jliunyu: Because when we build the token, the format is `"{\"%s\":\"%s\"}"`, we need to add 4 `"`, `{}`, `:` and the null-termination.

tests/interactive_broker_version.py
```diff
@@ -100,12 +100,36 @@ def test_version(version, cmd=None, deploy=True, conf={}, debug=False,
...
+                         'ascii')))
+                os.write(
+                    fd, ('sasl.oauthbearer.extensions=\
+                         ExtensionworkloadIdentity=develC348S,\
+                         Extensioncluster=lkc123\n'.encode(
+                         'ascii')))
+                os.write(
+                    fd, ('sasl.oauthbearer.scope=test\n'.encode(
+                         'ascii')))
+                cmd_env['VALID_OIDC_URL'] = "http://localhost:8080/retrieve"
```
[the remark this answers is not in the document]
jliunyu: I tried this way, but since trivup part is not merged yet, I met the build issue from CI, in my local it works fine. 
```
./interactive_broker_version.py -c "make quick" 2.7.0) || travis_terminate 1 ; fi
Traceback (most recent call last):
  File "./interactive_broker_version.py", line 14, in <module>
    from trivup.apps.OauthbearerOIDCApp import OauthbearerOIDCApp
ImportError: No module named 'trivup.apps.OauthbearerOIDCApp'
```
Can we keep this way for current? cmd_env['VALID_OIDC_URL'] = "http://localhost:8080/retrieve"
Or waiting for CI to merge.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,507 @@
...
+                        else {
+                                if (payloads_end > 0) {
+                                        errstr =
+                                            "The token is invalid with more "
+                                            "than 2 delimiters";
+                                        goto done;
+                                }
+                                payloads_end = i;
+                        }
+
```
edenhill, line 171: Is there a missing break here?
If not, then add a comment like so:
```c
  /*FALLTHRU*/
```

to indicate that the implicit fallthru is intentional.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,507 @@
...
+ *        Extract and decode payloads from JWT \p src.
+ *        The decoded payloads will be returned in \p *bufplainp.
+ *
+ * @returns Return error message while decoding the payload.
+ */
+static char *rd_kafka_jwt_b64_decode_payload(const char *src,
+                                             char **bufplainp) {
+        char *converted_src;
+        char *payload = NULL;
+        char *errstr  = NULL;
```
edenhill, line 136: make const

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,507 @@
...
+}
+
+/**
+ * @brief The format of JWT is Header.Payload.Signature.
+ *        Extract and decode payloads from JWT \p src.
+ *        The decoded payloads will be returned in \p *bufplainp.
+ *
+ * @returns Return error message while decoding the payload.
+ */
+static char *rd_kafka_jwt_b64_decode_payload(const char *src,
```
edenhill, line 132: return value should be const

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,507 @@
...
+        payload_len = payloads_end - payloads_start;
+        payload     = rd_malloc(payload_len + 4);
+        strncpy(payload, (converted_src + payloads_start), payload_len);
+
+        padding = 4 - (payload_len % 4);
+        if (padding < 4) {
+                while (padding--)
+                        payload[payload_len++] = '=';
+        }
+        payload[payload_len] = '\0';
```
edenhill, line 191: Don't think the nul-term is needed since the payload is passed without the nul (due to payload_len) to EVP_DecodeBlock()

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,507 @@
...
+                    rk,
+                    "Expected JSON "
+                    "response as a value string");
+                goto done;
+        }
+
+        errstr = rd_kafka_jwt_b64_decode_payload(jwt_token, &decoded_payloads);
+        if (errstr != NULL) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk, "Failed to decode payload");
```
edenhill, line 301: The error string is lost here. Suggest doing something like:
```c
  char errstr2[512];
   rd_snprintf(errstr2, sizeof(errstr2), "Failed to decode JWT payload: %s", errstr);
  ... set_token_failure(..., err2str);
```

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -0,0 +1,507 @@
...
+                goto done;
+        }
+
+        payloads = cJSON_Parse(decoded_payloads);
+        if (payloads == NULL) {
+                rd_kafka_oauthbearer_set_token_failure(
+                    rk, "Expected JSON JWT is valid JSON");
```
edenhill, line 308: "Failed to parse JSON JWT payload"

tests/0126-oauthbearer_oidc.c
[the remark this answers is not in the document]
edenhill: Okay, can you clarify the comment to say that we need to give it some to trigger the refresh token?
Also increase the sleep to 5 seconds, when running tests in parallell on slow systems (e.g., CI) it is not unusual for stalls that last a couple of seconds.

tests/0126-oauthbearer_oidc.c
```diff
@@ -31,66 +31,148 @@
...
+        rd_usleep(3 * 1000 * 1000, NULL);
+        test_consumer_poll("OIDC.C1", c1, testid, 1, -1, 1, NULL);
```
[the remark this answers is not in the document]
edenhill: ^

tests/0126-oauthbearer_oidc.c
```diff
@@ -31,66 +31,148 @@
...
+        SUB_TEST("Test OAUTHBEARER/OIDC failing with invalid JWT");
 
-        /* Produce messages */
-        test_produce_msgs2(rk, topic, testid, 1, 0, 0, NULL, 0);
+        expected_failure = rd_false;
 
-        /* Verify messages were actually produced by consuming them back. */
-        test_consume_msgs_easy(topic, topic, 0, 1, 1, NULL);
+        test_conf_set(conf, "sasl.oauthbearer.token.endpoint.url",
+                      rd_getenv("INVALID_OIDC_URL", NULL));
```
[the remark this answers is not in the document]
edenhill: No need to get the env again, it is already in `url` (and perhaps call it invalid_url?)

tests/0126-oauthbearer_oidc.c
```diff
@@ -31,66 +31,181 @@
...
+        rd_kafka_t *c1;
+        uint64_t testid;
+        rd_kafka_conf_t *conf;
+
+        const char *url = test_getenv("INVALID_OIDC_URL", NULL);
+
+        SUB_TEST("Test OAUTHBEARER/OIDC failing with invalid JWT");
+
+        if (!url) {
+                TEST_SKIP("INVALID_OIDC_URL environment variable is not set\n");
```
edenhill, line 158: Sub test must be skipped with SUB_TEST_SKIP

tests/0126-oauthbearer_oidc.c
```diff
@@ -31,66 +31,181 @@
...
+
+        sec = test_conf_get(conf, "security.protocol");
+        if (strcmp(sec, "sasl_plaintext")) {
+                TEST_SKIP("Apache Kafka cluster does not config SSL/SASL\n");
+                return 0;
+        }
+
+        oidc = test_conf_get(conf, "sasl.oauthbearer.method");
+        if (strcmp(oidc, "oidc")) {
+                TEST_SKIP("`sasl.oauthbearer.method=OIDC` is required\n");
```
edenhill, line 200: Is capital OIDC allowed?
If so the strcmp must be rd_strcasecmp

tests/0126-oauthbearer_oidc.c
```diff
@@ -31,66 +31,181 @@
...
+}
+
+
+/**
+ * @brief After config OIDC, if the token is not valid, make sure the
+ *        authentication fail as expected.
+ *
+ */
+static void
+do_test_produce_consumer_with_OIDC_should_fail(rd_kafka_conf_t *base_conf) {
```
edenhill, line 148: Make base_conf const in all these functions

tests/interactive_broker_version.py
```diff
@@ -283,6 +302,13 @@ def test_version(version, cmd=None, deploy=True, conf={}, debug=False,
         type=str,
         default=None,
         help='SASL mechanism (PLAIN, SCRAM-SHA-nnn, GSSAPI, OAUTHBEARER)')
+    parser.add_argument(
+        '--oidc-method',
```
edenhill, line 306: Should be named oauthbearer-method since oidc is one of the methods.
E.g., "oidc_method == oidc" and "oidc_method == default" are kind of ambiguous.

edenhill (changes requested):

src/rdhttp.c
```diff
@@ -223,6 +223,151 @@ rd_http_error_t *rd_http_get(const char *url, rd_buf_t **rbufp) {
...
+                }
+
+                herr = rd_http_req_perform_sync(&hreq);
+                len  = rd_buf_len(hreq.hreq_buf);
+
+                if (!herr) {
+                        if (len > 0)
+                                break; /* Success */
+                        /* Empty response */
+                        rd_http_req_destroy(&hreq);
```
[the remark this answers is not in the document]
jliunyu: Hi @edenhill, hanks for the review comments. The jsonp is initialized as NULL from the caller already: https://github.com/jliunyu/librdkafka/blob/kip768-token/src/rdkafka_sasl_oauthbearer_oidc.c#L223.
From the current function, we didn't use it yet before this line. So I think we don't need to set jsonp to NULL here?

src/rdkafka_sasl_oauthbearer_oidc.c
jliunyu: Yeah, this fallthru is intentional. I still want to copy the `.` to the tmp token holder converted_src.

tests/0126-oauthbearer_oidc.c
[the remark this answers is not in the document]
jliunyu: I changed this from the producer part to produce only 1 message.

edenhill (approved):
LGTM!
GREAT STUFF, JING!

PR #3572, opened 2021-10-05
Support for triggering SASL refresh callbacks on background thread

edenhill:
This PR solves two problems:
 1. Jing will need it for the internal OIDC refresh callback so it is triggered automatically on the background thread.
 2. Bindings, and advanced applications that provide their own refresh callbacks, no longer need to call poll() (et.al) prior to performing any other broker-needed APIs. Instead they need to set `enable.sasl.callback.queue=true` and call `rd_kafka_sasl_background_callbacks_enable()`. This may seem a bit complex and it would be much nicer if we could sort this out from rd_kafka_new(), but we need to guarantee that the background callback is triggered after rd_kafka_new() returns and the high-level client bindings have set up whatever high-level client instance objects they need. Failure to do so would trigger a refresh callback possibly referencing an object, or Null, that has not yet been returned to the user/app. So.. this extra call is needed to make sure that doesn't happen. And yes, that is really needed.

Any high-level binding that exposes a refresh callback API to the user needs to set `enable.sasl.callback.queue=true` and call `rd_kafka_sasl_background_callbacks_enable()` after `rd_kafka_new()` has returned.


CHANGELOG.md
```diff
@@ -3,6 +3,18 @@
 ## Enhancements
 
  * Added `ssl.ca.pem` to add CA certificate by PEM string. (#2380)
+ * SASL OAUTHBEARER refresh callbacks can now scheduled for execution
```
mhowlett, line 6: can now *be* scheduled for execution on librdkafka's background *thread*.

PR #3582, opened 2021-10-16
Fix memory leak in admin requests

benesch:
Fix a memory leak introduces in ca1b30e0 in which the arguments to an
admin request were not being freed. Discovered by the test suite for
rust-rdkafka [0].

[0]: https://github.com/fede1024/rust-rdkafka/pull/397/checks?check_run_id=3914902373


edenhill:
Thanks @benesch !

PR #3597, opened 2021-10-28
Fix out-of-bounds memory accesses in regex parsing

alpire:
This PR fixes out-of-bounds memory accesses in regex parsing found by OSS-Fuzz.

#  Out-of-bounds write

The implicit casting from integer to unsigned char of min & max in `parserep` allowed a maliciously crafted input to bypass the `max < min` check in parserep. This would later lead to a mismatch between the result of `count` and the number of `emit` calls, resulting in an heap out-of-bounds write in `emit`.

Example input: `3{3,65}3{55555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555553,65}35{61,}`

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27631
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27735

# Out-of-bounds read in `parseatom`

Example input: `((((((((((((((((\17+`

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28561

### DIFF
--- a/src/regexp.c
+++ b/src/regexp.c
@@ -512,7 +512,8 @@ static int empty(Renode *node) {
         }
 }
 
-static Renode *newrep(Restate *g, Renode *atom, int ng, int min, int max) {
+static Renode *
+newrep(Restate *g, Renode *atom, int ng, unsigned char min, unsigned char max) {
         Renode *rep = newnode(g, P_REP);
         if (max == REPINF && empty(atom))
                 die(g, "infinite loop matching the empty string");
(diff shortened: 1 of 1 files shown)


src/regexp.c
```diff
@@ -617,11 +619,14 @@ static Renode *parserep(Restate *g) {
...
-                int min = g->yymin, max = g->yymax;
+                unsigned char min = g->yymin, max = g->yymax;
                 next(g);
+                if ((g->yymin & ~(0xff)) || (g->yymax & ~(0xff)))
+                        die(g,
+                            "invalid quantifier: min & max should be in "
+                            "[0-255]");
                 if (max < min)
                         die(g, "invalid quantifier");
-                return newrep(g, atom, re_accept(g, '?'), min, max);
```
edenhill, line 624: Should this line really be removed?

alpire: Apologies, it's been a while, and I've lose context. I'll close this PR.

PR #3607, opened 2021-11-04
Fix MinGW Travis build issues

neptoess:
Relative discussion in #3130

### DIFF
--- a/packaging/mingw-w64/configure-build-msys2-mingw-static.sh
+++ b/packaging/mingw-w64/configure-build-msys2-mingw-static.sh
@@ -50,6 +50,3 @@ cp ./librdkafkacpp-static.a ../dest/lib/librdkafka++-static.a
 popd
 rm -rf ./mergescratch
 
-export PATH="$PWD/dest/bin:/mingw64/bin/:${PATH}"
-cd tests
-./test-runner.exe -l -Q -p1 0000
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,7 @@ matrix:
     - source ./packaging/mingw-w64/travis-before-install.sh
    before_script:
     - ./packaging/mingw-w64/configure-build-msys2-mingw-static.sh
+    - ./packaging/mingw-w64/run-tests.sh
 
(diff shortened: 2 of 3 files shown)


edenhill:
Perfect, thank you!

packaging/mingw-w64/configure-build-msys2-mingw-static.sh
```diff
@@ -50,6 +50,3 @@ cp ./librdkafkacpp-static.a ../dest/lib/librdkafka++-static.a
 popd
 rm -rf ./mergescratch
 
-export PATH="$PWD/dest/bin:/mingw64/bin/:${PATH}"
```
edenhill, line 53: late comment, but these should be moved for the dynamic build script too.

edenhill:
Hey @neptoess , pinging you here in lack of better communication alternatives:
The dynamic mingw build is failing (silently) because msys2 now uses gcc 11 which introduces a new `_ZSt28__throw_bad_array_new_lengthv` symbol,  but the libstdc++ on the system is compiled with an older gcc that  does not have this symbol, so trying to execute the test runner fails silently.
It is basically this issue:
https://github.com/stan-dev/pystan/issues/294

But I'm not sure that solution (symlink/replace) is viable for our msys2 builder.

Do you have any ideas?

https://app.travis-ci.com/github/edenhill/librdkafka/jobs/564513967

neptoess:
[v1.9.0-PRE11.build.log](https://github.com/edenhill/librdkafka/files/8363343/v1.9.0-PRE11.build.log)
Upgraded my msys2 packages and did the build. As the old joke goes, "works on my machine".

I also just realized that the build you linked was the dynamic build (pinging @ed-alertedh @edbordin since he was involved there), but the static build worked (https://app.travis-ci.com/github/edenhill/librdkafka/jobs/564513968). The build log you linked seems to cut off without even attempting to run the tests. Bizarre.

Adding a `$msys2 pacman -Q` to the end of `travis-before-install.sh` might be useful, so we can easily see exactly which packages, and which versions, are installed during the build.

PR #3625, opened 2021-11-21
Fix: fix callback missing when metadata request failed

aiquestion:
We experience this issue when main thread cgrp start to do a rebalance
and failed when parsing metadata, the cgrp callback will be lost 
and result in consumer unassigned.

If rd_kafka_handle_Metadata failed and will not retry, need to enque 
the callback with err.


edenhill:
Good find!

edenhill:
Thank you!

PR #3665, opened 2022-01-03
Update INTRODUCTION.md

amitkumar50:


edenhill:
Thank you but I think the short intro blurb should be before the ToC so people understand what the ToC relates to.

PR #3682, opened 2022-01-20
sasl.oauthbearer.extensions should be optional

lpsinger:
Fixes confluentinc/confluent-kafka-python#1269.


src/rdkafka_conf.c
```diff
@@ -3709,12 +3709,6 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                                        "is mandatory when "
                                        "`sasl.oauthbearer.method=oidc` is set";
                         }
-
-                        if (!conf->sasl.oauthbearer.extensions_str) {
```
edenhill, line 3713: @kirktrue @jliunyu Are extensions indeed optional for OIDC?

lpsinger: My reading of https://datatracker.ietf.org/doc/html/rfc7628#section-3.1 is that they are optional (emphasis is my own):

> Client responses are a GS2 [RFC5801] header followed by **zero or more** key/value pairs, or it **may be empty**.

src/rdkafka_conf.c
```diff
@@ -952,7 +952,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "is used. If set it to \"oidc\", OAuth/OIDC login method will "
      "be used. "
      "sasl.oauthbearer.client.id, sasl.oauthbearer.client.secret, "
-     "sasl.oauthbearer.scope, sasl.oauthbearer.extensions, "
+     "sasl.oauthbearer.scope, "
```
jliunyu, line 955: Hi @edenhill , do we need to separate this description for standard AK and CC?

edenhill: @jliunyu No I don't think so. The ccloud config requirements should not be expressed here but in ccloud docs&examples.

edenhill (approved):
LGTM

edenhill:
Thank you @lpsinger !

lpsinger:
Thank you!

Issue #3699, opened 2022-02-01
Unit tests fails with Kafka v3.0.0


ladislavmacoun:
Description
===========
The direct access to ZK is deprecated in Kafka v3.0.0 and the flag `--zookeeper` is removed. This causes some unit tests to fail

```bash
TEST 20220201134018 (bare, scenario default) SUMMARY
#==================================================================#
| <MAIN>                                   |     PASSED | 928.988s |
| 0000_unittests                           |     PASSED |   3.387s |
...
| 0052_msg_timestamps                      |     FAILED | 841.665s | test_kafka_topics():4539: system("/Users/lada/Workspace/librdkafka/tests/tmp/LibrdkafkaTestCluster/KafkaBrokerApp/kafka/3.0.0/bin/kafka-topics.sh --zookeeper localhost:57223 --create --topic "rdkafkatest_rnd6c63475269df9015_0052_msg_timestamps_CreateTime_0.10.1.0_none" --replication-factor 1 --partitions 1 --config message.timestamp.type=CreateTime") failed with exit status 1
...
| 0077_compaction                          |     FAILED | 792.952s | test_kafka_topics():4539: system("/Users/lada/Workspace/librdkafka/tests/tmp/LibrdkafkaTestCluster/KafkaBrokerApp/kafka/3.0.0/bin/kafka-topics.sh --zookeeper localhost:57223 --create --topic "rdkafkatest_rnd892f12067147045_0077-compaction.c" --partitions 1 --replication-factor 1 --config cleanup.policy=compact --config segment.ms=10000 --config segment.bytes=10000 --config min.cleanable.dirty.ratio=0.01 --config delete.retention.ms=86400 --config file.delete.delay.ms=10000") failed with exit status 1
```

More can be found at those KIPS
* [KIP-555](https://cwiki.apache.org/confluence/display/KAFKA/KIP-555%3A+Deprecate+Direct+Zookeeper+access+in+Kafka+Administrative+Tools)
* [KIP-604](https://cwiki.apache.org/confluence/display/KAFKA/KIP-604%3A+Remove+ZooKeeper+Flags+from+the+Administrative+Tools)

This [PR](https://github.com/apache/kafka/pull/10457) removed the `--zookeeper` option from topic command

How to reproduce
================
run unit tests with Kafka >3.0.0

```bash
./interactive_broker_version.py 3.0.0
make
```

Checklist
=========

Please provide the following information:

 - [x] librdkafka version: `current`
[...]

PR #3700, opened 2022-02-01
Replace deprecated zookeeper flag with bootstrap

ladislavmacoun:
Fixes: #3699

Signed-off-by: Ladislav Macoun <ladislavmacoun@gmail.com>

### DIFF
--- a/tests/test.c
+++ b/tests/test.c
@@ -4507,17 +4507,18 @@ void test_kafka_topics(const char *fmt, ...) {
         int r;
         va_list ap;
         test_timing_t t_cmd;
-        const char *kpath, *zk;
+        const char *kpath, *brokers;
 
         kpath = test_getenv("KAFKA_PATH", NULL);
-        zk    = test_getenv("ZK_ADDRESS", NULL);
+        brokers = test_getenv("BROKERS", NULL);
 
-        if (!kpath || !zk)
-                TEST_FAIL("%s: KAFKA_PATH and ZK_ADDRESS must be set",
+        if (!kpath || !brokers)
+                TEST_FAIL("%s: KAFKA_PATH and BROKERS must be set",
                           __FUNCTION__);
 
         r = rd_snprintf(cmd, sizeof(cmd),
-                        "%s/bin/kafka-topics.sh --zookeeper %s ", kpath, zk);
+                        "%s/bin/kafka-topics.sh --bootstrap-server %s ",
+                        kpath, brokers);
         TEST_ASSERT(r < (int)sizeof(cmd));
 
         va_start(ap, fmt);
@@ -5164,7 +5165,7 @@ void test_report_add(struct test *test, const char *fmt, ...) {
 }
 
 /**
- * Returns 1 if KAFKA_PATH and ZK_ADDRESS is set to se we can use the
+ * Returns 1 if KAFKA_PATH and BROKERS is set to se we can use the
  * kafka-topics.sh script to manually create topics.
  *
  * If \p skip is set TEST_SKIP() will be called with a helpful message.
@@ -5181,11 +5182,11 @@ int test_can_create_topics(int skip) {
 #else
 
         if (!test_getenv("KAFKA_PATH", NULL) ||
-            !test_getenv("ZK_ADDRESS", NULL)) {
+            !test_getenv("BROKERS", NULL)) {
                 if (skip)
                         TEST_SKIP(
                             "Cannot create topics "
-                            "(set KAFKA_PATH and ZK_ADDRESS)\n");
+                            "(set KAFKA_PATH and BROKERS)\n");
                 return 0;
         }
 
--- a/tests/test.c
+++ b/tests/test.c
@@ -4503,7 +4503,7 @@ void test_kafka_topics(const char *fmt, ...) {
 #ifdef _WIN32
         TEST_FAIL("%s not supported on Windows, yet", __FUNCTION__);
 #else
-        char cmd[512];
+        char cmd[1024];
         int r, bytes_left;
         va_list ap;
         test_timing_t t_cmd;
--- a/tests/test.c
+++ b/tests/test.c
@@ -4507,18 +4507,26 @@ void test_kafka_topics(const char *fmt, ...) {
         int r;
         va_list ap;
         test_timing_t t_cmd;
-        const char *kpath, *brokers;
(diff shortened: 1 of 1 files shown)


tests/test.c
```diff
@@ -4507,17 +4507,18 @@ void test_kafka_topics(const char *fmt, ...) {
         int r;
         va_list ap;
         test_timing_t t_cmd;
-        const char *kpath, *zk;
```
edenhill, line 4510: zk is still required for older Kafka versions, which we continue to test with, so this needs some more logic.

ladislavmacoun: I've tested it locally with Kafka v2.7.0 and it runs fine (tried even v2.5.0 and still works, see bellow[1]), I think the problem is with the static size buffer for [cmd](https://github.com/edenhill/librdkafka/blob/master/tests/test.c#L4506).

Here https://app.travis-ci.com/github/edenhill/librdkafka/jobs/557875940#L8186 it looks like the command is not complete, there is a missing setting for `--config delete.retention.ms`, the `test_kafka_topics()` should make boundary check for cmd since the `kpath` and input `fmt` are of variable length

>zk is still required for older Kafka versions

what is the oldest version you want to support if there would be a still problem with backwards capability, we can create `cmd` dynamically based on the `TEST_KAFKA_VERSION` env variable?

[1]:
```bash
bash-3.2$ TESTS=0077 make
Running tests in parallel
./run-test.sh
############## ./test-runner ################
### Running test ./test-runner in bare mode ###
[<MAIN>                      /  0.000s] Setting test timeout to 10s * 1.0
[<MAIN>                      /  0.000s] Git version: a98d26e5@master
[<MAIN>                      /  0.000s] Broker version: 2.5.0 (2.5.0.0)
[<MAIN>                      /  0.000s] Tests to run : 0077
[<MAIN>                      /  0.000s] Test mode    : bare
[<MAIN>                      /  0.000s] Test scenario: default
[<MAIN>                      /  0.000s] Test filter  : no filter
[<MAIN>                      /  0.000s] Test timeout multiplier: 2.7
[<MAIN>                      /  0.000s] Action on test failure: continue other tests
[<MAIN>                      /  0.000s] Current directory: /Users/lada/Workspace/librdkafka/tests
[<MAIN>                      /  0.000s] Setting test timeout to 30s * 2.7
[<MAIN>                      /  0.000s] 1 test(s) running: 0077_compaction
[0077_compaction             /  0.000s] ================= Running test 0077_compaction =================
[0077_compaction             /  0.000s] ==== Stats written to file stats_0077_compaction_4265141123287969475.json ====
[0077_compaction             /  0.000s] Using topic "rdkafkatest_rnd78f2c57b0a834c46_0077-compaction.c"
[0077_compaction             /  0.000s] Test compaction on topic rdkafkatest_rnd78f2c57b0a834c46_0077-compaction.c with no compression (40 messages)
[0077_compaction             /  0.000s] Executing: /Users/lada/Workspace/librdkafka/tests/tmp/LibrdkafkaTestCluster/KafkaBrokerApp/kafka/2.5.0/bin/kafka-topics.sh --bootstrap-server PLAINTEXT://0.0.0.0:50808,PLAINTEXT://0.0.0.0:50811,PLAINTEXT://0.0.0.0:50814 --create --topic "rdkafkatest_rnd78f2c57b0a834c46_0077-compaction.c" --partitions 1 --replication-factor 1 --config cleanup.policy=compact --config segment.ms=10000 --config segment.bytes=10000 --config min.cleanable.dirty.ratio=0.01 --config delete.retention.ms=86400 --config file.delete.delay.ms=100
WARNING: Due to limitations in metric names, topics with a period ('.') or underscore ('_') could collide. To avoid issues it is best to use either, but not both.
[<MAIN>                      /  1.005s] 1 test(s) running: 0077_compaction
Created topic rdkafkatest_rnd78f2c57b0a834c46_0077-compaction.c.
[0077_compaction             /  1.483s] exec: duration 1483.113ms
[0077_compaction             /  1.486s] Setting test timeout to 120s * 2.7
[0077_compaction             /  1.502s] Created    kafka instance 0077_compaction#producer-1
[0077_compaction             /  1.502s] Producing 20 messages (total of 20000 bytes) of compactable messages
[0077_compaction             /  1.503s] Created    kafka instance 0077_compaction#producer-2
[0077_compaction             /  1.503s] 0077_compaction#producer-2: Flushing 20 messages
[0077_compaction             /  1.593s] FLUSH: duration 89.913ms
[0077_compaction             /  1.593s] Producing 40 messages for 4 keys
[0077_compaction             /  1.639s] PRODUCE.DELIVERY.WAIT: duration 45.441ms
[0077_compaction             /  1.639s] Producing 20 messages (total of 20480 bytes) of compactable messages
[0077_compaction             /  1.639s] Created    kafka instance 0077_compaction#producer-3
```
[1]: https://github.com/edenhill/librdkafka/blob/master/tests/test.c#L4506

edenhill: We support Kafka 0.8 and later.

tests/test.c
```diff
@@ -5180,12 +5194,15 @@ int test_can_create_topics(int skip) {
...
+                        ? "BROKERS" : "ZK_ADDRESS";
+
         if (!test_getenv("KAFKA_PATH", NULL) ||
-            !test_getenv("ZK_ADDRESS", NULL)) {
+            !test_getenv(bootstrap, NULL)) {
                 if (skip)
                         TEST_SKIP(
                             "Cannot create topics "
-                            "(set KAFKA_PATH and ZK_ADDRESS)\n");
+                            "(set KAFKA_PATH and BROKERS)\n");
```
edenhill, line 5205: use %s  bootstrap here, as you did above.

tests/test.c
```diff
@@ -5164,12 +5176,14 @@ void test_report_add(struct test *test, const char *fmt, ...) {
 }
 
 /**
- * Returns 1 if KAFKA_PATH and ZK_ADDRESS is set to se we can use the
+ * Returns 1 if KAFKA_PATH and BROKERS is set to se we can use the
```
edenhill, line 5179: BROKERS (or ZK_ADDRESS)

edenhill (changes requested):
Good solution :+1:  Just some minor nits

tests/test.c
ladislavmacoun: Thank you for the review, this has slipped. I have changed that and recommitted.

edenhill:
Thank you!

PR #3704, opened 2022-02-02
sasl.oauthbearer.scope should be optional + fix buffer overflow when generating request body for auth token request

lpsinger:
According to the section 4.4.2 of RFC 6749, the scope is optional in the access token request in client credentials flow.

And indeed, for OIDC providers that I find in the wild such as Amazon Cognito, the scope _is_ optional. If the scope is omitted from the request, then the returned access token will contain any and all scope(s) that are configured for the client.

See https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2


lpsinger:
Note that in this PR, I also fixed a subtle buffer overflow / off-by-1 error: `post_fields_size` should equal `strlen(post_fields)`, _not_ `strlen(post_fields) + 1`. See example code at https://curl.se/libcurl/c/CURLOPT_POSTFIELDSIZE.html.

edenhill:
Thank you :+1:

PR #3713, opened 2022-02-07
Additional OIDC stuff

edenhill:


examples/misc.c
```diff
@@ -92,23 +88,23 @@ static void usage (const char *reason, ...) {
...
  */
-static void conf_set (rd_kafka_conf_t *conf,
-                      const char *name, const char *val) {
+static void conf_set(rd_kafka_conf_t *conf, const char *name, const char *val) {
         char errstr[512];
 
-        if (rd_kafka_conf_set(conf, name, val,
-                              errstr, sizeof(errstr)) != RD_KAFKA_CONF_OK)
+        if (rd_kafka_conf_set(conf, name, val, errstr, sizeof(errstr)) !=
+            RD_KAFKA_CONF_OK)
```
mhowlett, line 107: i like the other formatting better because it doesn't break the inequality over two lines, better still i'd put the errstr on the first line and the full inequality on the second line.

but i know you don't agree, and this is hardly important.

mhowlett (approved):

mhowlett (commented):
not very OIDCy but lgtm

examples/misc.c
edenhill: I agree, but formatting is fully automated now using clang-format :)

Issue #3720, opened 2022-02-10
Idempotent producer does not check for epoch exhaustion


ladislavmacoun:
Description
===========

The idempotent producer is not checking for epoch exhaustion, which can lead to an integer overflow during the `rd_kafka_idemp_drain_epoch_bump().

Looking at the reference AK producer [1], there is a check for integer overflow, in which case a new producer ID is requested to stay in sync with the broker.

Here is an example from the debug log when this happens, and in figure 1 you can see what happened in our production setup during a network outage. 

```
%7|1644510814.395|DRAIN|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: All partitions drained, bumped epoch to PID{Id:9,Epoch:32767}
%7|1644510814.395|IDEMPSTATE|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: Idempotent producer state change DrainBump -> Assigned
%7|1644510814.395|WAKEUP|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: localhost:53144/4: Wake-up
%7|1644510814.395|WAKEUP|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: localhost:53141/3: Wake-up
%7|1644510814.395|WAKEUP|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: localhost:53138/2: Wake-up
%7|1644510814.395|WAKEUP|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: 0.0.0.0:53144/bootstrap: Wake-up
%7|1644510814.395|WAKEUP|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: 0.0.0.0:53141/bootstrap: Wake-up
%7|1644510814.395|WAKEUP|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: 0.0.0.0:53138/bootstrap: Wake-up
%7|1644510814.395|TIMEOUT|0094_idempotence_msg_timeout#producer-1| [thrd:localhost:53144/4]: localhost:53144/4: rdkafkatest_rnd28f894cf57f1b70c_0094_idempotence_msg_timeout [0]: timed out 0+19 message(s) (MsgId 25..43): message.timeout.ms exceeded
[...]

PR #3721, opened 2022-02-10
Fix exhausted producer epochs

ladislavmacoun:
Fixes: #3720


edenhill (changes requested):
This is great stuff, but I think we need to keep the changes to the idempotence code down to a minimum. See suggestion inline.

src/rdkafka_idempotence.c
```diff
@@ -496,59 +496,77 @@ void rd_kafka_idemp_pid_update(rd_kafka_broker_t *rkb,
 }
 
 
+/**
+ * @brief Requests a new PID and triggers a re-request.
+ *
+ * @locality any
+ * @locks none
+ */
+static void rd_kafka_drain_reset_pid(rd_kafka_t *rk) {
```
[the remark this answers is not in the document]
ladislavmacoun: Thank you for the review!

I thought it would be better to narrow the critical sections used in the `..drain_done()` which includes debug logging, also the minimal nesting without auxiliary variables makes it more easier to understand, however, if it is a fragile code section I can change it in the `drain_done()` txn path directly.

tests/0094-idempotence_msg_timeout.c
```diff
@@ -213,6 +215,84 @@ static void do_test_produce_timeout(const char *topic, const int msgrate) {
...
+
+        rk  = test_create_handle(RD_KAFKA_PRODUCER, conf);
+        rkt = test_create_producer_topic(rk, topic, "message.timeout.ms",
+                                         "100", NULL);
+
+        /* Create the topic to make sure connections are up and ready. */
+        err = test_auto_create_topic_rkt(rk, rkt, tmout_multip(5000));
+        TEST_ASSERT(!err, "topic creation failed: %s", rd_kafka_err2str(err));
+
+        initial_pid = rk->rk_eos.pid.id;
```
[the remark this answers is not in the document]
ladislavmacoun: Make sense, thank you. This is the guidance that I needed for the proper test!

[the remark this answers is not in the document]
ladislavmacoun: I am not sure if I understand the use of `rd_kafka_test_idemp_set_pid(..., int64_t id, int16_t epoch);` correctly. Am I able to set a valid PID from a client without mocking the broker?

Would it be okay to use `rd_kafka_test_idemp_set_pid_epoch(rd_kafka_t *rk, int16_t epoch);` that would se the epoch for the currently assigned PID with proper locking, and then retrieving the pid with `rd_kafka_test_get_pid(rk)`

>It would also be good if we could rely on the broker to verify that the epoch is correct and not wrapped, rather than just reading it back from rd_kafka_t which doesn't really prove correct system behaviour.

I will add a produce request that should not fail after this.

src/rdkafka_idempotence.c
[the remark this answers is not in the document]
ladislavmacoun: Resolved with e4cb5c41e03d9212c0e355e59642bc9e2518d8d6

src/rdkafka.c
```diff
@@ -889,6 +889,39 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
 }
 
 
+int rd_kafka_test_idemp_set_pid(rd_kafka_t *rk, int64_t id, int16_t epoch) {
```
edenhill, line 892: Let's return `rd_kafka_resp_err_t` instead

src/rdkafka.c
```diff
@@ -889,6 +889,39 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
...
+        rd_kafka_pid_t pid = {id, epoch};
+
+        if (!rd_kafka_is_idempotent(rk) || !rd_kafka_pid_valid(pid)) {
+                return 1;
```
edenhill, line 896: Split this into two if-statements:
 * for !is_idempotent return `RD_KAFKA_RESP_ERR__NOT_CONFIGURED`
 * for !pid_valid return `RD_KAFKA_RESP_ERR__INVALID_ARG`

src/rdkafka.c
```diff
@@ -889,6 +889,39 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
...
+        }
+
+        rd_kafka_wrlock(rk);
```
edenhill, line 899:
```suggestion
        rd_kafka_log(rk, LOG_WARNING, "TESTPID", "Forcibly overwriting idempotency PID");
 
        rd_kafka_wrlock(rk);
```

src/rdkafka.c
```diff
@@ -889,6 +889,39 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
...
+
+        return 0;
+}
+
+
+int rd_kafka_test_idemp_get_pid(rd_kafka_t *rk, int64_t *id, int16_t *epoch) {
+        rd_kafka_pid_t pid;
+
+        if (!rd_kafka_is_idempotent(rk)) {
+                return 1;
```
edenhill, line 911: `NOT_CONFIGURED`

src/rdkafka.c
```diff
@@ -889,6 +889,39 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
...
+        }
+
+        pid = rd_kafka_idemp_get_pid(rk);
+        if (!rd_kafka_pid_valid(pid)) {
+                return 1;
```
edenhill, line 916: `ERR__STATE`

src/rdkafka.c
```diff
@@ -889,6 +889,39 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
...
+        }
+
+        *id    = pid.id;
+        *epoch = pid.epoch;
+
+        return 0;
```
edenhill, line 922: `ERR_NO_ERROR`

src/rdkafka_proto.h
```diff
@@ -641,6 +641,16 @@ rd_kafka_pid_bump(const rd_kafka_pid_t old) {
         return new_pid;
 }
 
+
+/**
+ * @brief Checks whether the PID epoch has been exhausted
+ */
+static RD_UNUSED RD_INLINE rd_bool_t
+rd_kafka_pid_epoch_exhausted(const rd_kafka_pid_t pid) {
+        return pid.epoch >= INT16_MAX;
```
edenhill, line 650: This can't really be >, right? Change it to ==

tests/0094-idempotence_msg_timeout.c
```diff
@@ -213,6 +215,114 @@ static void do_test_produce_timeout(const char *topic, const int msgrate) {
...
+
+        rk  = test_create_handle(RD_KAFKA_PRODUCER, conf);
+        rkt = test_create_producer_topic(rk, topic, "message.timeout.ms", "100",
+                                         NULL);
+
+        /* Create the topic to make sure connections are up and ready. */
+        err = test_auto_create_topic_rkt(rk, rkt, tmout_multip(5000));
+        TEST_ASSERT(!err, "topic creation failed: %s", rd_kafka_err2str(err));
+
+        if (rd_kafka_test_idemp_get_pid(rk, &pid.id, &pid.epoch) != 0) {
```
edenhill, line 257: With the new error signature this can be simplified by:
`TEST_CALL_ERR__(rd_kafka_test_idemp..(...));` which automatically checks for error and asserts accordingly

tests/0094-idempotence_msg_timeout.c
```diff
@@ -213,6 +215,114 @@ static void do_test_produce_timeout(const char *topic, const int msgrate) {
...
+                TEST_FAIL(
+                    "rd_kafka_test_idemp_get_pid() failed for "
+                    "current PID");
+        }
+
+        /* Mock epoch near exhaustion point to make overflow faster */
+        if (rd_kafka_test_idemp_set_pid(rk, pid.id, INT16_MAX) != 0) {
```
edenhill, line 264: Dito TEST_CALL_ERR__

tests/0094-idempotence_msg_timeout.c
```diff
@@ -213,6 +215,114 @@ static void do_test_produce_timeout(const char *topic, const int msgrate) {
...
+                TEST_FAIL("rd_kafka_test_idemp_set_pid() failed");
+        }
+
+        /* Add delay to socket to create message timeouts */
+        sockem_ctrl_set_delay(&ctrl, 500, 3 * 100);
+
+        for (int i = 0; i < msgcnt; i++) {
```
edenhill, line 271: This variable def is not compatible with older compilers. Move `int i;` to beginning of this scope (the function scope)

tests/0094-idempotence_msg_timeout.c
```diff
@@ -213,6 +215,114 @@ static void do_test_produce_timeout(const char *topic, const int msgrate) {
...
+        test_flush(rk, 1000);
+
+        /* Check if we got some timeouted messages */
+        if (rd_kafka_test_idemp_get_pid(rk, &new_pid.id, &new_pid.epoch) != 0) {
+                TEST_FAIL(
+                    "rd_kafka_test_idemp_get_pid() failed for "
+                    "a new PID");
+        }
+        if (new_pid.epoch >= pid.epoch && new_pid.id == pid.id) {
+                TEST_FAIL("Expected Producer epoch exhaustion");
```
edenhill, line 296: nit: I find it easier to read TEST_ASSERT()s, e.g. what we are expecting:
`TEST_ASSERT(new_pid != pid, "Expected PID&Epoch bump");`

edenhill (changes requested):
Great stuff, just some minor comments!

ladislavmacoun:
@edenhill Thank you for the review!

I've addressed the issues you have pointed out.

ladislavmacoun:
Hi @edenhill, don't want to push on this, but had you and chance to check on this?

ladislavmacoun:
Hi @emasab, any update on this?

PR #3733, opened 2022-02-16
Save one superfluous message timeout toppar scan

larry-cdn77:
C short-circuit evaluation seems to cobble the message timeout scanning condition in `rd_kafka_broker_producer_serve` such that instead of first iteration only (as intended according to adjacent code comment) you get a timeout in the first and second, so typically two scans instead of one

The side effect appears to have been introduced in 8af1b39 and getting rid of it is a minor optimisation, but consideration worthy for high-partition count topics

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -4008,6 +4008,7 @@ static void rd_kafka_broker_producer_serve(rd_kafka_broker_t *rkb,
                (abs_timeout > (now = rd_clock()))) {
                 rd_bool_t do_timeout_scan;
                 rd_ts_t next_wakeup = abs_timeout;
+                int overshoot;
 
                 rd_kafka_broker_unlock(rkb);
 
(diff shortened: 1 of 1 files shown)


src/rdkafka_broker.c
```diff
@@ -3995,6 +3995,7 @@ static void rd_kafka_broker_producer_serve(rd_kafka_broker_t *rkb,
         rd_interval_t timeout_scan;
         unsigned int initial_state = rkb->rkb_state;
         rd_ts_t now;
+        rd_ts_t now, overshoot;
```
edenhill, line 3998: `now` is duplicated.

But we'll want `overshoot` to be defined in the scope where it is used, namely in the while loop below.

edenhill:
Good catch!
How did you find this?

larry-cdn77:
Thanks for the quick response. How did I spot this? Lots of digging 😄 I have corrected the change as per your comments.

larry-cdn77:
I accidentally created this PR from my fork's master – move to #3739

PR #3739, opened 2022-02-21
Save one superfluous message timeout toppar scan

larry-cdn77:
C short-circuit evaluation seems to cobble the message timeout scanning condition in rd_kafka_broker_producer_serve such that instead of first iteration only (as intended according to adjacent code comment) you get a timeout in the first and second, so typically two scans instead of one

The side effect appears to have been introduced in https://github.com/edenhill/librdkafka/commit/8af1b39842de8baf61ef9612b9b730f8d455c713 and getting rid of it is a minor optimisation, but consideration worthy for high-partition count topics


larry-cdn77:
Supersede #3733 which was accidentally created from my fork's master

edenhill:
Thank you!

PR #3741, opened 2022-02-22
fix: acl binding enum checks

emasab:
If some kinds of enums like ANY or MATCH are passed when creating an AclBinding, an error should be given before calling the CreateAcls operation on the Kafka broker, in order to return a better explaination, because in these cases it returns an "unknown error". Also when a different value is returned from a future version of Kafka that value is mapped to UNKNOWN.

The ACL binding destroy array function is added too

### DIFF
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -4229,6 +4229,14 @@ static void rd_kafka_AclBinding_free(void *ptr) {
         rd_kafka_AclBinding_destroy(ptr);
 }
 
+
+void rd_kafka_AclBinding_destroy_array(rd_kafka_AclBinding_t **acl_bindings,
+                                       size_t acl_bindings_cnt) {
+        size_t i;
+        for (i = 0; i < acl_bindings_cnt; i++)
+                rd_kafka_AclBinding_destroy(acl_bindings[i]);
+}
+
 /**
  * @brief Parse CreateAclsResponse and create ADMIN_RESULT op.
  */
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -4366,7 +4366,7 @@ rd_kafka_DescribeAclsResponse_parse(rd_kafka_op_t *rko_req,
                                     char *errstr,
                                     size_t errstr_size) {
         const int log_decode_errors = LOG_ERR;
-        rd_kafka_broker_t *rkb      = rko_req->rko_rk->rk_internal_rkb;
+        rd_kafka_broker_t *rkb      = reply->rkbuf_rkb;
         rd_kafka_resp_err_t err     = RD_KAFKA_RESP_ERR_NO_ERROR;
         rd_kafka_op_t *rko_result   = NULL;
         int32_t res_cnt;
@@ -4414,7 +4414,7 @@ rd_kafka_DescribeAclsResponse_parse(rd_kafka_op_t *rko_req,
 
                 if (res_type <= RD_KAFKA_RESOURCE_UNKNOWN ||
                     res_type >= RD_KAFKA_RESOURCE__CNT) {
-                        rd_rkb_log(rkb, LOG_WARNING, "DESCRIBEACLRESPONSE",
+                        rd_rkb_log(rkb, LOG_WARNING, "DESCRIBEACLSRESPONSE",
                                    "DescribeAclsResponse returned unknown "
                                    "resource type %d",
                                    res_type);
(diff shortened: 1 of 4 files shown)


src/rdkafka_admin.c
```diff
@@ -4346,6 +4411,16 @@ rd_kafka_DescribeAclsResponse_parse(rd_kafka_op_t *rko_req,
                         rd_kafka_buf_read_i8(reply, &resource_pattern_type);
                 }
 
+                if (res_type <= RD_KAFKA_RESOURCE_UNKNOWN ||
```
edenhill, line 4414: Maybe we should emit a warning log here if the broker returns an (for us) unknown resource or pattern type, something like:
`rd_rkb_log(rkb, LOG_WARNING, "ACLRESPONSE", "DescribeAclsResponse returned unknown resource type %d");`

tests/0080-admin_ut.c
```diff
@@ -694,6 +694,610 @@ static void do_test_DeleteConsumerGroupOffsets(const char *what,
...
+        rd_bool_t valid_acl_operation[] = {
+            rd_false, rd_false, rd_true, rd_true, rd_true, rd_true, rd_true,
+            rd_true,  rd_true,  rd_true, rd_true, rd_true, rd_true, rd_false};
+        rd_bool_t valid_acl_permission_type[] = {rd_false, rd_false, rd_true,
+                                                 rd_true, rd_false};
+        const char *topic     = test_mk_topic_name(__FUNCTION__, 1);
+        const char *principal = "User:test";
+        const char *host      = "*";
+
+        SUB_TEST_QUICK("AclBinding");
```
edenhill, line 721: nit: no need for the "AclBinding" string, just leave it empty `SUB_TEST_QUICK()`  since the function name is emitted and already contains the name.

tests/0080-admin_ut.c
```diff
@@ -694,6 +694,610 @@ static void do_test_DeleteConsumerGroupOffsets(const char *what,
...
+
+        // Valid acl binding
+        new_acl = rd_kafka_AclBinding_new(
+            RD_KAFKA_RESOURCE_TOPIC, topic, RD_KAFKA_RESOURCE_PATTERN_LITERAL,
+            principal, host, RD_KAFKA_ACL_OPERATION_ALL,
+            RD_KAFKA_ACL_PERMISSION_TYPE_ALLOW, errstr, 512);
+        TEST_ASSERT(new_acl, "expected AclBinding");
+        rd_kafka_AclBinding_destroy(new_acl);
+
```
edenhill, line 730: Need to zero errstr prior to each of these calls (`*errstr = '\0';`) or errstr may have undefined contents if ..new() doesn't fail.

edenhill (changes requested):
Great stuff, great code, great tests!

emasab:
Thanks Magnus! I've addressed those issues

src/rdkafka_admin.c
```diff
@@ -4331,7 +4397,7 @@ rd_kafka_DescribeAclsResponse_parse(rd_kafka_op_t *rko_req,
                      rd_kafka_AclBinding_free);
 
         for (i = 0; i < (int)res_cnt; i++) {
-                int8_t res_type;
+                rd_kafka_ResourceType_t res_type = RD_KAFKA_RESOURCE_UNKNOWN;
```
edenhill, line 4400: This needs to be an int8_t since ..buf_read_i8() just performs a 1-byte memory copy, and we can't really know the size or layout of the enum.

src/rdkafka_admin.c
```diff
@@ -4346,6 +4412,26 @@ rd_kafka_DescribeAclsResponse_parse(rd_kafka_op_t *rko_req,
                         rd_kafka_buf_read_i8(reply, &resource_pattern_type);
                 }
 
+                if (res_type <= RD_KAFKA_RESOURCE_UNKNOWN ||
+                    res_type >= RD_KAFKA_RESOURCE__CNT) {
+                        rd_rkb_log(rkb, LOG_WARNING, "DESCRIBEACLRESPONSE",
+                                   "DescribeAclsResponse returned unknown "
```
edenhill, line 4418: :+1:

src/rdkafka_admin.c
```diff
@@ -4509,6 +4617,7 @@ rd_kafka_DeleteAclsResponse_parse(rd_kafka_op_t *rko_req,
                                   char *errstr,
                                   size_t errstr_size) {
         const int log_decode_errors = LOG_ERR;
+        rd_kafka_broker_t *rkb      = rko_req->rko_rk->rk_internal_rkb;
```
edenhill, line 4620: What's this for? we'll want to reference the reply rkb, if any.

src/rdkafka_admin.c
```diff
@@ -4549,19 +4658,20 @@ rd_kafka_DeleteAclsResponse_parse(rd_kafka_op_t *rko_req,
                 /* #maching_acls */
                 rd_kafka_buf_read_arraycnt(reply, &matching_acls_cnt, 100000);
                 for (j = 0; j < (int)matching_acls_cnt; j++) {
-                        int16_t acl_error_code;
-                        int8_t res_type;
+                        rd_kafka_resp_err_t acl_error_code;
```
edenhill, line 4661: Dito, these need to be int16 and int8

edenhill (changes requested):
Great! Just some minor stuff now

src/rdkafka_admin.c
emasab: Hi, it's for the `rd_rkb_log` . I've tried `rko_req->rko_replyq.q->rkq_rk->rk_internal_rkb` also now, but gives segmentation fault. Is it the correct way to access it?

emasab: ok, I've found it's like in the other response parse functions

PR #3750, opened 2022-02-28
Combine message timeout drain bumps

larry-cdn77:
Amend #2163

Issue an idempotent producer drain-bump only once per entire scanning cycle rather than for each toppar

Passes test number 94 (idempotence msg timeout) extended to spot absence of detected timeouts

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3630,9 +3630,12 @@ rd_kafka_broker_outbufs_space(rd_kafka_broker_t *rkb) {
 /**
  * @brief Serve a toppar for producing.
  *
+ * @param now is current time used for detecting timed-out messages
  * @param next_wakeup will be updated to when the next wake-up/attempt is
  *                    desired, only lower (sooner) values will be set.
  * @param do_timeout_scan perform msg timeout scan
+ * @param timeout_cnt number of timed-out messages found, variable incremented
+ *                    not set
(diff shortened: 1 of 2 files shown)


src/rdkafka_broker.c
```diff
@@ -3649,6 +3652,7 @@ static int rd_kafka_toppar_producer_serve(rd_kafka_broker_t *rkb,
                                           rd_ts_t now,
                                           rd_ts_t *next_wakeup,
                                           rd_bool_t do_timeout_scan,
+                                          int *timeout_cnt,
```
edenhill, line 3655: name this `timeout_cntp` (p indicating that it is a pointer, otherwise it is too easy to do `timeout_cnt++`)

edenhill (commented):
There's a problem with this solution in that on timeout producer_serve() will continue to be called for the remaining partitions but without bumping the epoch (and thus causing other partitions not to produce anything), so the other partitions may continue to send ProduceRequests for an epoch that will be bumped, when we actually want to drain and wait for all outstanding to finish.

larry-cdn77:
I think I understand the problem in principle, although at a glance I am unsure how any sending could occur during the scanning process that takes place in a toppar lock. If at all possible, making broker wake-ups lighter weight sounds like a better way of reducing the overhead of repeated drain-bump.

I am happy to close this PR

PR #3754, opened 2022-03-03
KIP 368: SASL Reauthentication

vctoriawu:
Added SASL reauthentication for the following feature: [KIP-368](https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate)

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -1706,7 +1706,8 @@ rd_kafka_broker_controller(rd_kafka_t *rk, int state, rd_ts_t abs_timeout) {
  * Find a waitresp (rkbuf awaiting response) by the correlation id.
  */
 static rd_kafka_buf_t *rd_kafka_waitresp_find(rd_kafka_broker_t *rkb,
-                                              int32_t corrid) {
+                                              int32_t corrid,
+                                              rd_bool_t skipStats) {
         rd_kafka_buf_t *rkbuf;
         rd_ts_t now = rd_clock();
 
@@ -1715,8 +1716,10 @@ static rd_kafka_buf_t *rd_kafka_waitresp_find(rd_kafka_broker_t *rkb,
         TAILQ_FOREACH(rkbuf, &rkb->rkb_waitresps.rkbq_bufs, rkbuf_link)
         if (rkbuf->rkbuf_corrid == corrid) {
                 /* Convert ts_sent to RTT */
-                rkbuf->rkbuf_ts_sent = now - rkbuf->rkbuf_ts_sent;
-                rd_avg_add(&rkb->rkb_avg_rtt, rkbuf->rkbuf_ts_sent);
+                if (unlikely(!skipStats)) {
+                    rkbuf->rkbuf_ts_sent = now - rkbuf->rkbuf_ts_sent;
+                    rd_avg_add(&rkb->rkb_avg_rtt, rkbuf->rkbuf_ts_sent);
+                }
 
                 if (rkbuf->rkbuf_flags & RD_KAFKA_OP_F_BLOCKING &&
                     rd_atomic32_sub(&rkb->rkb_blocking_request_cnt, 1) == 1)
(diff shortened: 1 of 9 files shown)


showuon:
@edenhill , could you please take a look? Thank you.

showuon:
@julesbovet , please also take a look. Thank you.

src/rdkafka_broker.c
```diff
@@ -2274,6 +2373,13 @@ static void rd_kafka_broker_connect_auth(rd_kafka_broker_t *rkb) {
                                 : RD_KAFKA_BROKER_STATE_AUTH_LEGACY);
                         rd_kafka_broker_unlock(rkb);
 
+                        // Check if reauth is in progress, if so then client close
+                        if (rkb->rkb_rk->rk_conf.sasl.enable_refresh) {
+                                if (rkb->rkb_sasl.session_lifetime_ms > 0) {
```
showuon, line 2378: nit: `if (rkb->rkb_rk->rk_conf.sasl.enable_refresh && rkb->rkb_sasl.session_lifetime_ms > 0)`

src/rdkafka_conf.c
```diff
@@ -926,6 +926,12 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "This builtin handler should only be used for development "
      "or testing, and not in production.",
      0, 1, 0, _UNSUPPORTED_OAUTHBEARER},
+    {_RK_GLOBAL, "enable.sasl.token.refresh", _RK_C_BOOL,
+     _RK(sasl.enable_refresh),
+     "Enable token refresh for SASL connections. "
+     "This verification can be extended by the application by "
+     "implementing a oauthbearer_token_refresh_cb.",
```
showuon, line 933: nit: by implementing oauthbearer_token_refresh_cb callback. (remove `a`, add `callback`)

showuon (commented):
@vctoriawu , thanks for the PR. This is a very important feature to get supported! Sorry, I'm not quite familiar with the code base. Left some minor comments. Thanks.

vctoriawu:
@edenhill any updates on the additional requirements?

src/rdkafka_conf.c
```diff
@@ -926,6 +926,12 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "This builtin handler should only be used for development "
      "or testing, and not in production.",
      0, 1, 0, _UNSUPPORTED_OAUTHBEARER},
+    {_RK_GLOBAL, "enable.sasl.token.refresh", _RK_C_BOOL,
```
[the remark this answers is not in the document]
SamBarker: I've no insight into the need for the flag but I do think re-authentication is generally valid.

The intention of re-auth is to ensure that the client permissions are still valid so it is useful across all SASL auth types.

Every re-auth period the PLAIN and SCRAM tokens would be re-presented (unchanged on the client side) to allow the server to re-validate the session. If the account represented by the token had been suspended the re-auth call will fail and thus the session will be terminated.

showuon:
@edenhill , any update? Thanks.

showuon:
@edenhill , any update for the requirements?

k-wall:
Any way we can help with this feature?

edenhill:
I think the proposed solution is overly complex, and I don't think the special handling of the reauth requests is really needed.

Can't we get away with just sending the SaslHandshakeRequest upon reauth timer expiry and handling the results on response; if reauth succeeds nothing needs to be done (just restart the timer), if it fails we close the broker connection and perhaps raise a fatal error.

There's also the addition of sasl_set_credentials() in #4033 that is also needed for reauth to be meaningful.

vctoriawu:
I agree that that could be a potential solution, but reading the description of the KIP leads me to believe that this is a requirement to hold back other requests:

"the broker will communicate the session expiration time as part of the final SASL_AUTHENTICATE response.  If this value is positive, then the client will automatically re-authenticate before anything else unrelated to re-authentication is sent beyond that expiration point."

also:

"Note also that the client cannot queue up additional send requests beyond the one that triggers re-authentication to occur until re-authentication succeeds and the triggering one is sent."

Not sure if I am misinterpreting

milindl:
Thanks a lot @vctoriawu for starting this.

We discussed this quite a bit internally, and we decided to do it with the new broker state as Magnus suggested in the comment [above](https://github.com/confluentinc/librdkafka/pull/3754#pullrequestreview-933416062). The approach does take care of the case you mention [here.](https://github.com/confluentinc/librdkafka/pull/3754#issuecomment-1411572826). 

The PR #4301 is a draft of that approach. I've taken the commits from this PR and built upon those. I've also added a few tests. (The credential refresh callback is yet to be implemented, though, it's only the reauthentication bit -- any credentials, if the user wants to reset them, have to be done on their own for SASL PLAIN and SCRAM.).

If you have any comment or question about this approach, please send them to that PR.

milindl:
Closing this with the 2.2.0 release which has this feature. Thanks

PR #3767, opened 2022-03-18
Grooming (compilation warning and potential issue)

sarroutbi:
Signed-off-by: Sergio Arroutbi <sarroutb@redhat.com>

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -778,7 +778,7 @@ restart:
                 rd_kafka_bufq_deq(rkbq, rkbuf);
 
                 if (now && cnt < log_first_n) {
-                        char holbstr[128];
+                        char holbstr[256];
                         /* Head of line blocking:
                          * If this is not the first request in queue, but the
                          * initial first request did not time out,
--- a/src/rdkafka_idempotence.c
+++ b/src/rdkafka_idempotence.c
@@ -313,14 +313,16 @@ redo:
                             rd_kafka_handle_InitProducerId, NULL);
                 }
 
-                rd_kafka_broker_destroy(rkb);
-
                 if (err) {
                         rd_rkb_dbg(rkb, EOS, "GETPID",
                                    "Can't acquire ProducerId from "
                                    "this broker: %s",
                                    errstr);
+                }
 
+                rd_kafka_broker_destroy(rkb);
+
+                if (err) {
                         if (rd_kafka_idemp_check_error(rk, err, errstr,
                                                        is_fatal))
                                 return; /* Fatal error */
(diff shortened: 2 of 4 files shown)


src/rdkafka_broker.c
```diff
@@ -76,6 +76,7 @@
 
 
 static const int rd_kafka_max_block_ms = 1000;
+static const unsigned short HOLBSTR_MAX_SIZE = 256;
```
edenhill, line 79: Why is this change needed?
That buffer size is local to the timeout_scan() function and only used once, so no need for a predefined value.

src/rdkafka_idempotence.c
```diff
@@ -313,14 +313,16 @@ void rd_kafka_idemp_pid_fsm(rd_kafka_t *rk) {
...
-                rd_kafka_broker_destroy(rkb);
-
                 if (err) {
                         rd_rkb_dbg(rkb, EOS, "GETPID",
                                    "Can't acquire ProducerId from "
                                    "this broker: %s",
                                    errstr);
+                }
 
+                rd_kafka_broker_destroy(rkb);
```
edenhill, line 323: Good find :+1:

src/rdkafka_mock.c
```diff
@@ -1765,6 +1765,7 @@ void rd_kafka_mock_push_request_errors(rd_kafka_mock_cluster_t *mcluster,
                 errors[i] = va_arg(ap, rd_kafka_resp_err_t);
 
         rd_kafka_mock_push_request_errors_array(mcluster, ApiKey, cnt, errors);
+        va_end(ap);
```
edenhill, line 1768: nit: Move this to the line after the for-loop

edenhill (changes requested):

src/rdkafka_broker.c
sarroutbi: The important thing here is the change in the value to avoid compilation warning. Regarding the usage of a constant, I just prefer using constants that define what is a buffer for (to avoid magic numbers). However, if not necessary, I can return it to where it was.

edenhill: Ah, I see. 
Yes, please remove the constant and just the use the value directly.

sarroutbi: Sure. Done.

src/rdkafka_mock.c
sarroutbi: Done

edenhill:
Thank you!

I notice that your branch is called covscan.. were these changes made as a result of a Coverity scan? If so; are there more to come?

I did a coverity scan a couple of years ago, which perhaps found one or two issues, but had hordes of false positives. Maybe the situation has improved?

sarroutbi:
Hello @edenhill. You are welcome.

Yes, this changes were result of a covscan execution in RHEL version of librdkafka. As you say, there were many issues, and the majority of them were false positives, but some of them were also "true positives". I analyzed all them and ended up with this PR, so from my side I expect no more changes related to covscan in the short trem.

PR #3770, opened 2022-03-20
cmake: Use CMAKE_INSTALL_LIBDIR

kraj:
this ensures that it is portable across platforms e.g. ppc64/linux
uses lib64 not lib

Signed-off-by: Khem Raj <raj.khem@gmail.com>


edenhill (approved):
LGTM

edenhill:
Thank you!

PR #3774, opened 2022-03-22
Reset stored offset on assign() and prevent offsets_store() for unassigned partitions

edenhill:


CHANGELOG.md
```diff
@@ -6,6 +6,26 @@ librdkafka v1.9.0 is a feature release:
  * Added KIP-140 Admin API ACL support (by @emasab, #2676)
 
 
+## Upgrade considerations
+
+ * Consumer:
+   `rd_kafka_offsets_store()` (et.al) will now return an error for any
+   partition that is not currently assigned (through `rd_kafka_*assign()`).
```
mhowlett, line 13: does the following issue comment indicate there may additionally be a different issue with similar effect?
https://github.com/confluentinc/confluent-kafka-dotnet/issues/1672#issuecomment-1000346590

src/rdkafka_op.c
```diff
@@ -897,6 +897,8 @@ void rd_kafka_op_offset_store(rd_kafka_t *rk, rd_kafka_op_t *rko) {
         rd_kafka_toppar_lock(rktp);
         rktp->rktp_app_offset = offset;
         if (rk->rk_conf.enable_auto_offset_store)
-                rd_kafka_offset_store0(rktp, offset, 0 /*no lock*/);
+                rd_kafka_offset_store0(rktp, offset,
+                                       /* force: ignore assignment state */
+                                       rd_true, RD_DONT_LOCK);
```
mhowlett, line 902: i'm unsure why this needs to be true, just checking you've thought it through.

mhowlett (approved):

src/rdkafka_op.c
edenhill: That's explained in the offset_store0 doc:
```
.....
 * Enforcing assignment for store in this state is tricky with a bunch of
 * corner cases, so instead we let those places forcibly store the offset, but
 * then in assign() we reset the stored offset to .._INVALID, just like we do
 * on revoke.
...
```

CHANGELOG.md
edenhill: Yep, that does look like the same issue.
A background rebalance is taking place between poll() and offsets_store(), effectively overwriting the reset offset with a proper one, which is then committed.

varunkamra:
@edenhill I had the exact consumer failure with 
```
offset reset (at offset 178229) to BEGINNING: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
```
When will the new version of librdkafka be released which fixes it?

plachor:
Hi @edenhill, we were recently affected by this issue I believe. In logs we notice such messages (for random partitions-CG):

> [thrd:main]: TOPIC_NAME [19]: offset reset (at offset 2060961) to BEGINNING: fetch failed due to requested offset not available on the broker: Broker: Offset out of range

It was logged with `Warning` severity level. For our application this was a critical issue we were flooded with millions of messages that introduce lag on our microservices causing inaccessibility to fresh data for our customers. 

We believe it correlates with network partition between brokers which led to reelection of new controller.  

In your latest release `1.9.0+` will it be thrown as an exception with some reasonable code or will it still be just logged with `Warning` severity level in such cases ? We cannot afford random reset to begging of partition without a reason.

varunkamra:
> @edenhill I had the exact consumer failure with following error:
> 
> ```
> offset reset (at offset 178229) to BEGINNING: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
> ```
> 
> When will the new version of librdkafka be released which fixes it?

My issue was resolved when I started committing offsets of consumed messages.

nik-hil:
@varunkamra 
I m facing the same issue. Will you share doc/code of your soln?

Madhurdev007:
Any update on this issue please... how we can fix this.. do we need to upgrade to 1.9.2 or some code changes with existing version 1.8.2

srolija:
@Madhurdev007 You have to upgrade to 1.9.0 or newer, you can see it in [release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.9.0), do note the additional error this change can yield in your code.

For the comment of @varunkamra I assume he was probably running with configuration `auto.commit.enable=false` so he had to call the one of the commit methods (depends on language how exactly they are called; ie in golang version of this library it is [CommitOffsets](https://docs.confluent.io/platform/current/clients/confluent-kafka-go/index.html#Consumer.CommitOffsets) or [rd_kafka_commit](https://github.com/edenhill/librdkafka/blob/cd955b13eabb29f8c4dbbb01a3943b17da909f25/src/rdkafka.h#L4254) for C version) to explicitly store the offsets that the consumers have processed. Because without it the consumer group would never had them stored. So upon start it would fallback onto the default reset policy.

Madhurdev007:
Thanks @srolija  ... I'll check this

Issue #3745, opened 2022-02-25
unexcepted reset offsets occasionally


intfish123:
Description
===========
unexcepted reset offsets occasionally

How to reproduce
================
it's just occasionally.

**IMPORTANT**: Always try to reproduce the issue on the latest released version (see https://github.com/edenhill/librdkafka/releases), if it can't be reproduced on the latest version the issue has been fixed.

Checklist
=========
 - [x] librdkafka version (release number or git tag): `v1.8.2`
 - [x] Apache Kafka version: `v2.2.0`
 - [x] librdkafka client configuration: 
 ```
c, err := kafka.NewConsumer(&kafka.ConfigMap{
		"bootstrap.servers":        strings.Join(address, ","),
		"group.id":                 groupId,
		"auto.offset.reset":        "latest",
		"enable.auto.offset.store": "false",
		"enable.auto.commit":       true,
		"auto.commit.interval.ms":  1000,
		"fetch.wait.max.ms":        1000,
	}) 
```
 - [x] Operating system: `alpine:3.14`
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 ```
 %4|1645709989.739|OFFSET|rdkafka#consumer-1| [thrd:main]: raw_box_group_01 [37]: offset reset (at offset 2067600870) to END: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
%4|1645709989.744|OFFSET|rdkafka#consumer-1| [thrd:main]: raw_box_group_01 [32]: offset reset (at offset 2067607114) to END: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
%4|1645709989.745|OFFSET|rdkafka#consumer-1| [thrd:main]: raw_box_group_01 [29]: offset reset (at offset 2067828573) to END: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
%4|1645709989.746|OFFSET|rdkafka#consumer-1| [thrd:main]: raw_box_group_01 [31]: offset reset (at offset 2067937001) to END: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
%4|1645709989.747|OFFSET|rdkafka#consumer-1| [thrd:main]: raw_box_group_01 [36]: offset reset (at offset 2067584343) to END: fetch failed due to requested offset not available on the broker: Broker: Offset out of range
[...]

Issue #3710, opened 2022-02-04
Invalid Offsets Committed during a rebalance on shutdown


mattatcha:
Description
===========
Invalid Offsets committed for a partition when the partition has been `postponed` while a rebalance is in progress while using the `EAGER` load load balancing strategy.

This is causing a single partitions offsets to be either committed out of range or committed back in time to a previous offset. this is causing our consumers to re-process millions of messages that were already processed.

We are using [kafka-client-go](https://github.com/confluentinc/confluent-kafka-go/tree/v1.8.2) version **1.8.2**. for reference.

See details below about `EAGER` vs `COOPERATIVE` rebalance strategies but we switched to using `COOPERATIVE` and it happens far less than with `EAGER` however it still does happen daily.

How to reproduce
================
When a consumer in a consumer group shuts down it causes a chain of events, at times, that commits the incorrect offsets for a partition.

Checklist
=========

Please provide the following information:

 - [x] librdkafka version: 
 **v1.8.2**

 - [x] Apache Kafka version: 
 **2.6.1**

 - [x] librdkafka client configuration: 
 ```shell
 {
		"client.id":                     "clearbit/kafka-go",
		"bootstrap.servers":             "localhost:9092",
		"enable.partition.eof":          "false",
		"session.timeout.ms":            "45000",
		"max.poll.interval.ms":          "86400000",
		"heartbeat.interval.ms":         "3000",
		"partition.assignment.strategy": "roundrobin",
		"request.required.acks":         "-1",
		"enable.auto.commit":            "true",
		"auto.commit.interval.ms":       "5000",
		"enable.auto.offset.store":      "false",
		"auto.offset.reset":             "earliest",
		"compression.codec":             "lz4",
		"compression.type":              "lz4",
		"partitioner":                   "random",
		"batch.num.messages":            "10000",
		"queue.buffering.max.ms":        "5",
		"statistics.interval.ms":        "15000",
		"security.protocol":             "ssl",
		"ssl.ca.location":               "...",
	}
```
 - [x] Operating system: 
 **Debian Linux (x64)>**

 - [x] Provide logs (with `debug=..` as necessary) from librdkafka
 - [x] Provide broker log excerpts
 - [x] Critical issue

The first line shows that the correct offsets were fetched but the old versions offset (-1) is what was stored.
The stored offset `411018846` is what was committed instead of `411193513` which was the previously committed offset.
```
[...]

PR #3782, opened 2022-03-25
fix: Incorrect stored offset will be use after rebalanced

intfish123:
Reset the stored offset in rebalanced, so that it will not commit previous stored offset to broker, (issue #3745, #3710).
And the codes in `rdkafka_assignment.c` -> `rd_kafka_assignment_serve_removals()` -> `rd_kafka_offset_store0(rktp, RD_KAFKA_OFFSET_INVALID,RD_DONT_LOCK);` can not fix the incorrect offset completely.

### DIFF
--- a/src/rdkafka_request.c
+++ b/src/rdkafka_request.c
@@ -824,7 +824,12 @@ rd_kafka_handle_OffsetFetch (rd_kafka_t *rk,
 				/* Update toppar's committed offset */
 				rd_kafka_toppar_lock(rktp);
 				rktp->rktp_committed_offset = rktpar->offset;
-                                /** Update toppar's stored offset to INVALID, issue #3745 */
+
+                                /* Update toppar's stored offset to INVALID.
+                                 * Reset the stored offset in rebalanced,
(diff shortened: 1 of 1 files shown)


intfish123:
The base tag is v1.8.2, and I hope master can fix this problem quickly, because it happend daily on our production environment.

edenhill:
Thanks for your contribution! 

We already have a fix for this issue in this PR: https://github.com/edenhill/librdkafka/pull/3774

edenhill:
Can you try out the fix in PR #3774 and verify that it fixes the problem for your use-case?
Thanks

intfish123:
> Can you try out the fix in PR #3774 and verify that it fixes the problem for your use-case? Thanks

ok, I will try it and verify on my `test` environment. 
And branch `resetrestored` can be used with confidence on `production` environment?

edenhill:
The branch has been reviewed and tested, but it hasn't been released so there are no guarantees.
It would however be very valuable for everyone if it could be verified before merging,

PR #3787, opened 2022-03-29
Make dynamic MinGW build copy DLLs instead of trying to manipulate PATH

neptoess:
My stab at fixing the issues highlighted in the #3607 chat recently. When running locally, the two librdkafka DLLs were the only issue. Copying the DLLs to the same folder as the EXE is a much more common practice on Windows than adding the folder the DLLs live in to PATH, so I'm giving this a shot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,6 @@ matrix:
     - ./configure --install-deps --source-deps-only --disable-lz4-ext --prefix="$PWD/dest" --enable-static --enable-strip
 
  - name: "Windows MinGW-w64 Dynamic"
-   if: tag IS PRESENT
    os: windows
    env:
     - LINKAGE=std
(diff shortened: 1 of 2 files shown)


neptoess:
Hmm, Travis didn't fire off the dynamic MinGW build (https://app.travis-ci.com/github/edenhill/librdkafka/builds/248597005). I don't think I'm able to trigger it either. Can you help me out there @edenhill ? Thanks

neptoess:
> @neptoess The dynamic build requires a tag, you can just remove that "if: .." statement in .travis.yml

Thank you. Removed the `if`. Looks like my fix worked. My guess is that something changed on the Travis end where the old approach of updating `PATH` instead of copying the DLLs stopped working. Relying on environment variable updates is a major dice roll on Windows, so the "copy the DLLs" approach is, in my experience, a much more common practice.

edenhill (approved):
LGTM!
Can you rebase on latest master and reinstate the Dynamic travis builder that I commented out and reinstate the "if: tag IS PRESENT" ? Thanks

edenhill (approved):
LGTM
Thanks!

edenhill:
Thank you for so quickly helping out with this, much appreciated!

Issue #3471, opened 2021-07-26
Kafka consumer `seek` (before `pause`) doesn't work


kenneth-jia:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
If Kafka consumer `seeks to some offsets` and then `pause the partition`, after `resuming the partition`, we could find the previous `seeks to some offsets` doesn't work at all!

How to reproduce
================
1. Start a Kafka consumer, which keeps reading messages from a topic
2. After polling some messages (to offset `A`), Kafka consumer seek to a earlier offset (`B`), and then pause the partition.
3. Resume the partition, and poll messages again, -- we can see it begins from offset `A` (instead of `B`)! I.e, the `seek` operation (before the `pause`) doesn't work!
Note, if we `pause` the partition before the `seek`, everything is fine, just as expected.

We also created some testcase in `modern-cpp-kafka` project, which verifies/reproduces the problem.
1. The case which would fail, https://github.com/morganstanley/modern-cpp-kafka/blob/main/tests/integration/TestKafkaConsumer.cc#L1503
2. The case which would succeed, https://github.com/morganstanley/modern-cpp-kafka/blob/main/tests/integration/TestKafkaConsumer.cc#L1452
3. The PR which could fail, https://github.com/morganstanley/modern-cpp-kafka/pull/63
```
Mon, 26 Jul 2021 13:21:20 GMT
2: /home/runner/work/modern-cpp-kafka/modern-cpp-kafka/tests/integration/TestKafkaConsumer.cc:1549: Failure
Mon, 26 Jul 2021 13:21:20 GMT
2: Expected equality of these values:
Mon, 26 Jul 2021 13:21:20 GMT
2:   messages.size()
Mon, 26 Jul 2021 13:21:20 GMT
2:     Which is: 3
Mon, 26 Jul 2021 13:21:20 GMT
2:   records.size()
Mon, 26 Jul 2021 13:21:20 GMT
2:     Which is: 2
```

Checklist
=========

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `1.7.0`
 - [x ] Apache Kafka version: `2.7.1`
 - [x] librdkafka client configuration: `auto.reset.offset=earliest`
 - [x] Operating system: `ubuntu 20.04`
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka

Log
=========
```
[...]

PR #3811, opened 2022-04-11
seek() now resets the app_offset so that resume() uses the seeked offset (#3471)

edenhill:


mhowlett (approved):
lgtm, though consider comment.

PR #3813, opened 2022-04-12
fix compilation error (minor)

niamster:
`rdkafka.c:4381:63: error: adding 'int' to a string does not append to`


edenhill:
Now that's a silly error, pointer arithmetics is a thing, what compiler and with what -W.. flags reports this?

niamster:
> Now that's a silly error, pointer arithmetics is a thing, what compiler and with what -W.. flags reports this?

That's generated by [-Wstring-plus-int](https://clang.llvm.org/docs/DiagnosticsReference.html#wstring-plus-int) which is enabled by default in `clang v13.0.0`

edenhill (approved):

edenhill:
Thank you!

niamster:
Thank you Magnus 🙇

PR #3814, opened 2022-04-12
fix undefined behaviour in config parsing

niamster:
Undefined behaviour error:
```
rdkafka_conf.c:2866:21: runtime error: index 20 out of bounds for type 'struct (anonymous struct at rdkafka_conf.c:87:9) const[20]'
```

### DIFF
--- a/src/rdkafka_conf.c
+++ b/src/rdkafka_conf.c
@@ -2863,7 +2863,7 @@ static size_t rd_kafka_conf_flags2str(char *dest,
 
         /* Phase 1: scan for set flags, accumulate needed size.
          * Phase 2: write to dest */
-        for (j = 0; prop->s2i[j].str; j++) {
+        for (j = 0; j < (int)RD_ARRAYSIZE(prop->s2i) && prop->s2i[j].str; j++) {
                 if (prop->type == _RK_C_S2F && ival != -1 &&
                     (ival & prop->s2i[j].val) != prop->s2i[j].val)
                         continue;


src/rdkafka_conf.c
```diff
@@ -2863,7 +2863,8 @@ static size_t rd_kafka_conf_flags2str(char *dest,
 
         /* Phase 1: scan for set flags, accumulate needed size.
          * Phase 2: write to dest */
-        for (j = 0; prop->s2i[j].str; j++) {
+        const int s2i_cnt = sizeof(prop->s2i) / sizeof(prop->s2i[0]);
+        for (j = 0; j < s2i_cnt && prop->s2i[j].str; j++) {
```
edenhill, line 2867: use RD_ARRAYSIZE() instead and skip the temp var, e.g.:
https://github.com/edenhill/librdkafka/blob/1100743cb213fc2a6ead47532a33abcbf2c80a3c/src/rdkafka_conf.c#L1787

niamster: Ah, thanks, I didn't notice it exists. Thanks!

niamster: 🤔  for some reason I don't have an option to apply your suggestion. Let me do it manually. I will squash commit per [contribution guidelines](https://github.com/edenhill/librdkafka/blob/4ad5481e13f68ad2033ca167abae4f2578dbf356/CONTRIBUTING.md#how-to-get-your-changes-into-the-main-sources).

edenhill:
Very good, thank you!

niamster:
> Very good, thank you!

Thanks!

Issue #3537, opened 2021-09-10
maximum poll interval (90000ms) exceeded by 846018558ms


WhiteWind:
===========
High-level consumer sometimes fails with error:
Application maximum poll interval (300000ms) exceeded by 846018558ms: Local: Maximum application poll interval (max.poll.interval.ms) exceeded

846018s = 235h = 9days, but application was started a few hours ago.

I don't know what debug=... setting I should set. App may run without issue for a few days, so debug=all will produce too much output

Checklist
=========

Please provide the following information:

 - [x] librdkafka version v1.7.0
 - [x] Apache Kafka version: 2.3.1
 - [x] librdkafka client configuration: api.version.request= true,auto.offset.reset=earliest,enable.auto.commit=false,fetch.message.max.bytes= 50000,heartbeat.interval.ms=10000,message.max.bytes= 17000000,queued.max.messages.kbytes= 17000,session.timeout.ms=30000,reconnect.backoff.max.ms=5000,reconnect.backoff.ms=200,retry.backoff.ms=200
 - [x] Operating system: Windows Server 2016 standard
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
I'm guessing your system clock was changed.

WhiteWind:
Maybe. But it's just native Windows time sync service on virtual machine.
It does change time by jumps, not by adjustments, so it may introduce negative time offsets.

Isn't it better to calculate timeouts using GetTickCount64/CLOCK_MONOTONIC_COARSE instead of wall clock time ?

WhiteWind:
I see, that you use CLOCK_MONOTONIC on Linux and QPC on Windows. My problem is that my VM does not have an invariant TSC. 
Is such precision really necessary? Wouldn't GetTickCount64 be enough?

edenhill:
We need sub-millisecond granularity for some timers, and it'd be a headache to split up into two different implementations.
GetTickCount64 seems to have a granularity of 10-16ms, which is not good enough:
https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-gettickcount64

WhiteWind:
Ok, I see. 
May there be a bug with integer overflow:

```exceeded = (int)((rd_clock() - last_poll) / 1000ll) -
                rk->rk_conf.max_poll_interval_ms;

[...]

PR #3815, opened 2022-04-13
Fix MAXPOLL error on Win32 (#3537)

WhiteWind:
Using InterlockedCompareExchange64() in rd_atomic64_get() on Windows, because reading 64bit value on 32bit platform is not atomic. This avoids errors when calculating max poll timeout


edenhill:
Thank you for this!

PR #3816, opened 2022-04-14
KIP-601: Added `socket.connection.setup.timeout.ms`

edenhill:
The Java client has a more aggressive 10s default for this property followed
by an exponential backoff up to a `socket.connection.setup.timeout.max.ms`.
They need this approach since they try brokers one by one, but that's not
really the case in librdkafka. And since we want to avoid emitting an
`ERR__ALL_BROKERS_DOWN` too early/aggressively we can afford a higher
default value for the connect timeout and not use the exponential backoff
at all.


mhowlett (approved):
awesome, 1.9.0 is shaping up to be a good vintage.

edenhill:
Nice turnaround!

PR #3820, opened 2022-04-19
Fix fast leader query back-off

### LINKED ISSUE #3690: Metadata leader query timer backoff never applied
Further to #3237, `timers_run` [unschedules](https://github.com/ajbarb/librdkafka/blob/efcdcee986d8998116e024e7de3541e41abfd530/src/rdkafka_timer.c#L281) the leader query timer before invoking its callback, `rd_kafka_metadata_leader_query_tmr_cb`. The callback's attempt to back-off then always fails on the [scheduled condition](https://github.com/ajbarb/librdkafka/blob/efcdcee986d8998116e024e7de3541e41abfd530/src/rdkafka_timer.c#L176).

```
        if (rd_kafka_timer_scheduled(rtmr)) {
                rtmr->rtmr_interval *= 2;
                rd_kafka_timer_unschedule(rkts, rtmr);
        }
```

To be sure of the above analysis I printed backoff attempts using version 1.8.2 and the backoff function would indeed always come in with an unscheduled timer, and actually not do the backoff.

larry-cdn77:
Fix #3690: Metadata leader query timer backoff never applied

### DIFF
--- a/src/rdkafka_timer.c
+++ b/src/rdkafka_timer.c
@@ -30,6 +30,7 @@
 #include "rd.h"
 #include "rdtime.h"
 #include "rdsysqueue.h"
+#include "rdunittest.h"
 
 #include "rdkafka_queue.h"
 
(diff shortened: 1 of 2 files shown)


larry-cdn77:
Test checks to see that after 6 seconds 6 triggers occur of an exponentially backed off 100ms timer (you expect triggers at 100 ms, 300, 700, 1500, 3100 and 5500)

I thought testing against the somewhat private rdkafka_timer.h interface is worth it given the benefit of having the test

larry-cdn77:
I don't immediately see why the rdkafka_timer symbols failed to link in the CI build

image

edenhill:
Use in code (src/rdtimer.c) unittests to test the private implementation, see for example [unittest_broker](https://github.com/edenhill/librdkafka/blob/master/src/rdunittest.c#L450).

Also make sure the test runs quickly, preferably in under a handful of seconds.

You can run a single unittest by:
`TESTS=0000 RD_UT_TEST=backoff make` in the `tests/` directory.

larry-cdn77:
Thank you. I had neglected the unit test framework. It is a unit test now reduced in duration to one second.

larry-cdn77:
In order for my test to run reliably I had to round up the microsecond-millisecond conversion of sleep time in `rd_kafka_timers_run`. Otherwise the value is rounded down and I get a 'dribbling' effect of waiting a bit more, a bit more and a tiny bit more still...

```
                        sleeptime = rd_kafka_timers_next(rkts, timeout_us,
                                                         0 /*no-lock*/);

                        if (sleeptime > 0) {
                                cnd_timedwait_ms(&rkts->rkts_cond,
                                                 &rkts->rkts_lock,
                                                 (int)(sleeptime / 1000) + 1);
                        }
```

PR #3822, opened 2022-04-21
fix config property introduction for custom partitioner callback

uncrownedyd:
I've noticed some confusion about introduction for the configuration property of 'custom partitioner callback', `partitioner` is a bit of different from `partitioner_cb`, according to the description below, it might be `partitioner_cb` here ?


edenhill:
Good find, thank you!

PR #3835, opened 2022-04-28
NuGet packaging: librdkafka.redist is now built with VS 2019 and msvcr v142

edenhill:


mhowlett (approved):
LGTM provided you know what you are doing regarding platform compatibility, dependency clashes etc (i don't). at this point i would expect 142 is better than 140, but have no idea.

PR #3837, opened 2022-05-02
Allow any form of unassign*() during consumer close

edenhill:
https://github.com/confluentinc/confluent-kafka-go/issues/767#issuecomment-1107465737

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -6,6 +6,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: |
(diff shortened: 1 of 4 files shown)


mhowlett (approved):
i assume this resolves an issue? can you link?

It seems a bit dirty to be hijacking a request to do one thing and making it do another. It seems like there probably should be a separate mechanism doing the unassign, and the command that shouldn't happen return an error.

but i don't have any issue merging this.

kevinconaway:
Thank you!

PR #3849, opened 2022-05-17
Default of linger.ms mentioned in intro was out of sync with latest d…

Schm1tz1:
…efault value - fixed that.


edenhill (approved):
Thank you!

PR #3850, opened 2022-05-18
fix: test 103 fails on Kafkas < 2.5.0

emasab:
in versions < 2.5.0 a sleep is necessary before checking the committed offsets


tests/0103-transactions.c
```diff
@@ -1003,6 +1003,10 @@ static void do_test_empty_txn(rd_bool_t send_offsets, rd_bool_t do_commit) {
         else
                 TEST_CALL_ERROR__(rd_kafka_abort_transaction(p, -1));
 
+        /* Wait before checking the committed offsets (Kafka < 2.5.0) */
+        if (test_broker_version < TEST_BRKVER(2, 5, 0, 0))
+                rd_usleep(tmout_multip(2000 * 1000), NULL);
```
edenhill, line 1008: I think 2s is too short in the standard case too, which will make the test flaky. Let's make it 5s by default.

edenhill (changes requested):
Looks good, but increase the sleep to avoid flaky tests.

Also squash/fixup your commits into a single commit and with a more specific commit message, e.g.:
`Test 0103: AK < 2.5 needs sleep to synchronize committed txn offsets`

edenhill (approved):
Great stuff!

PR #3863, opened 2022-05-31
Added async:able consumer_close_queue() and consumer_closed()

edenhill:
This is mainly for the Go client, but can be used by applications as well.

This is blocking the v1.9.0 Go release.

### DIFF
--- a/src-cpp/rdkafkacpp.h
+++ b/src-cpp/rdkafkacpp.h
@@ -2790,13 +2790,13 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
 
 
   /**
-   * @brief Close and shut down the proper.
+   * @brief Close and shut down the consumer.
    *
    * This call will block until the following operations are finished:
-   *  - Trigger a local rebalance to void the current assignment
-   *  - Stop consumption for current assignment
-   *  - Commit offsets
-   *  - Leave group
+   *  - Trigger a local rebalance to void the current assignment (if any).
+   *  - Stop consumption for current assignment (if any).
+   *  - Commit offsets (if any).
+   *  - Leave group (if applicable).
    *
    * The maximum blocking time is roughly limited to session.timeout.ms.
    *
@@ -2931,6 +2931,31 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
    */
   virtual Error *incremental_unassign(
       const std::vector<TopicPartition *> &partitions) = 0;
+
+  /**
+   * @brief Close and shut down the consumer.
+   *
+   * Same behaviour as RdKafka::KafkaConsumer::close() but with the following
+   * changes:
+   *
+   * Rebalance events/callbacks (etc) will be forwarded to the
+   * application-provided \p queue. The application must poll this queue (in
+   * another thread) until this close function returns or
+   * RdKafka::KafkaConsumer::closed() returns true.
+   *
+   * @remark Depending on consumer group join state there may or may not be
+   *         rebalance events emitted on \p rkqu.
+   *
+   * @returns an error object if the consumer close failed, else NULL.
+   *
+   * @sa RdKafka::KafkaConsumer::closed()
+   */
+  virtual Error *close(Queue *queue) = 0;
+
+
+  /** @returns true if the consumer has been closed with
+   *           RdKafka::KafkaConsumer::close(), else false. */
+  virtual bool closed() = 0;
 };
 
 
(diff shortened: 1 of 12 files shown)


emasab (approved):

src-cpp/rdkafkacpp.h
```diff
@@ -2931,6 +2931,31 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
...
+
+  /**
+   * @brief Close and shut down the consumer.
+   *
+   * Same behaviour as RdKafka::KafkaConsumer::close() but with the following
+   * changes:
+   *
+   * Rebalance events/callbacks (etc) will be forwarded to the
+   * application-provided \p queue. The application must poll this queue (in
+   * another thread) until this close function returns or
```
mhowlett, line 2943: > in another thread

two thoughts:

it would be better UX if close(queue) was non-blocking, then you wouldn't need another thread.

they way you describe it, it seems like you need to check two conditions in order to decide when to exit the poll loop.

mhowlett, line 2943: > in another thread

two thoughts:

it would be better UX if close(queue) was non-blocking, then you wouldn't need another thread.

they way you describe it, it seems like you need to check two conditions in order to decide when to exit the poll loop.

src-cpp/rdkafkacpp.h
```diff
@@ -2931,6 +2931,31 @@ class RD_EXPORT KafkaConsumer : public virtual Handle {
...
+   * @returns an error object if the consumer close failed, else NULL.
+   *
+   * @sa RdKafka::KafkaConsumer::closed()
+   */
+  virtual Error *close(Queue *queue) = 0;
+
+
+  /** @returns true if the consumer has been closed with
+   *           RdKafka::KafkaConsumer::close(), else false. */
+  virtual bool closed() = 0;
```
mhowlett, line 2958: i was thinking it would be more natural to have another event type for this (is this a queue based API or not!?), but this works, and in practice is probably more ergonomic i think.

mhowlett, line 2958: i was thinking it would be more natural to have another event type for this (is this a queue based API or not!?), but this works, and in practice is probably more ergonomic i think.

src/rdkafka.c
```diff
@@ -3226,10 +3275,15 @@ rd_kafka_resp_err_t rd_kafka_consumer_close(rd_kafka_t *rk) {
                         rd_kafka_op_res_t res;
```
mhowlett, line 3275: seems a bit messy to be working with `rkq` in the `rebalance_rkq` case. logic is fine, just probably not as clear as it could be.

mhowlett, line 3275: seems a bit messy to be working with `rkq` in the `rebalance_rkq` case. logic is fine, just probably not as clear as it could be.

src/rdkafka.c
```diff
@@ -3238,16 +3292,62 @@ rd_kafka_resp_err_t rd_kafka_consumer_close(rd_kafka_t *rk) {
...
+                 * application can exit its poll loop. */
+                rd_kafka_q_yield(rebalance_rkq);
+
+                /* Drop our reference */
+                rd_kafka_q_destroy(rebalance_rkq);
+        }
+
+        /* Reset cgrp queue forwarding as no further events will/must be
+         * emitted. */
         rd_kafka_q_fwd_set(rkcg->rkcg_q, NULL);
```
mhowlett, line 3306: i don't understand how the above code blocks until here until the consumer has closed in the rebalance_rkq case. i suspect you know what you are doing and i'm missing something obvious, but you might want to a comment about this.

src/rdkafka.c
```diff
@@ -3181,30 +3202,58 @@ rd_kafka_message_t *rd_kafka_consumer_poll(rd_kafka_t *rk, int timeout_ms) {
...
+ *                      to this queue which must be served by the application
+ *                      until this call to close_queue() returns.
+ *                      If the consumer is not in a joined state, no rebalance
+ *                      events will be emitted.
+ *                      If not specified; this function will serve the rebalance
+ *                      callbacks.
+ *
+ */
+static rd_kafka_error_t *
+rd_kafka_consumer_close_q(rd_kafka_t *rk, rd_kafka_q_t *rebalance_rkq) {
```
mhowlett, line 3218: i'd consider two separate methods, even though there would be some duplication, though you could maybe factor that out.

mhowlett, line 3218: i'd consider two separate methods, even though there would be some duplication, though you could maybe factor that out.

mhowlett (approved):
consider my comments carefully, then do what you want with them :-).

Issue #3867, opened 2022-06-03
Different location of Openssl on M1 macs


alicerum:
Since on m1 homebrew seems to be installed to a different location (`/opt/homebrew/`), it is probably necessary to check for location `/opt/homebrew/opt/openssl/`, or for `openssl-1.1`, `/opt/homebrew/opt/openssl@1.1/` (which is a symlink to `/opt/homebrew/Cellar/openssl@1.1/`).
Otherwise, `librdkafka-static.a` does not link `libssl` and `libcrypto` statically.

https://github.com/edenhill/librdkafka/blob/d50099f9d8f0c8a62edca91bebc107cf9a3ec8a8/mklove/modules/configure.libssl#L32

alicerum:
Basically, I have changed it to this on my local build, and error is fixed:

```
mkl_env_append PKG_CONFIG_PATH "/usr/local/opt/openssl/lib/pkgconfig" ":"
mkl_env_append PKG_CONFIG_PATH "/opt/homebrew/opt/openssl@1.1/lib/pkgconfig" ":"
```

PR #3868, opened 2022-06-04
OpenSSL libraries path for M1 MacOS

alicerum:
As mentioned in the issue #3867 
On m1 MacOS homebrew is located in a different place, so configure script is not able to locate libssl and libcrypto.
This change makes it possible to use those libraries during build on M1.


edenhill:
Thanks for the explanation 👍

reidgraviton:
> @edenhill afaiu, m1 has a new location, because you can still install and use `amd64` brew on m1 via rosetta. this is why different arch brews have different prefixes. `/usr/local/` is location for `amd64` `/opt/homebrew/` is location for `arm64` [here](https://docs.brew.sh/Manpage) you can find default value for prefixes.

I think you meant `/usr/local/` is location for `x86` no?

silviucpp:
@reidgraviton yes amd64 = x86 64 bits.

PR #3870, opened 2022-06-07
Mark II: Added asynchronous rd_kafka_consumer_close_queue() and .._consumer_closed()

edenhill:
This is mainly for the Go client, but can be used by applications as well.

This is **Mark II**, which compared to #3863 makes the close truly asynchronous.

Most things are identical to #3863, main change is that consumer_close_queue() will now return as soon as the close is triggered, and it is up to the application to wait for consumer_closed() to return true in its poll loop.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -50,6 +50,7 @@
 #include "rdkafka_assignor.h"
 #include "rdkafka_request.h"
 #include "rdkafka_event.h"
+#include "rdkafka_error.h"
 #include "rdkafka_sasl.h"
 #include "rdkafka_interceptor.h"
 #include "rdkafka_idempotence.h"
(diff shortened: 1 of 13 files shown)


emasab (approved):

emasab (approved):
LGTM

src/rdkafka.c
```diff
@@ -3181,33 +3202,79 @@ rd_kafka_message_t *rd_kafka_consumer_poll(rd_kafka_t *rk, int timeout_ms) {
...
+                                                   rd_kafka_q_t *rkq) {
         rd_kafka_cgrp_t *rkcg;
-        rd_kafka_op_t *rko;
-        rd_kafka_resp_err_t err = RD_KAFKA_RESP_ERR__TIMED_OUT;
-        rd_kafka_q_t *rkq;
+        rd_kafka_error_t *error = NULL;
 
         if (!(rkcg = rd_kafka_cgrp_get(rk)))
-                return RD_KAFKA_RESP_ERR__UNKNOWN_GROUP;
+                return rd_kafka_error_new(RD_KAFKA_RESP_ERR__UNKNOWN_GROUP,
```
mhowlett, line 3220: i think i would err on the side of making this a noop (mark consumer as closed and return) vs an error.

you could argue that the more constrained behavior -> informative, but I feel like if there is a method as generic as close, on a class, it generally shouldn't be an error to call it.

PR #3937, opened 2022-08-03
Add timeout_ms option for rd_kafka_commit, rd_kafka_commit_message

kenneth-jia:
Add an additional parameter (`timeout_ms`) for `rd_kafka_commit`/`rd_kafka_commit_message`.
To avoid stuck there forever while a consumer is trying to sync-commit but no response from brokers (e.g. brokers unexpected down)


kenneth-jia:
Hi, @edenhill ,
I found it's really hard to fix the `check/style (pull_request)` failure. I've tried `clang-format -style=.clang-format`, -- it didn't work; also tried `make style-fix`, but didn't work either.
I guess the `check/style` is newly added but most source files would fail the checking at present (even with no change from this PR)? 
Thanks,
Kenneth

edenhill:
Regarding the style:  clang-format seems to be broken on OSX.  Running it on Linux yields good results.

src/rdkafka.h
```diff
@@ -4197,7 +4200,8 @@ RD_EXPORT int rd_kafka_assignment_lost(rd_kafka_t *rk);
 RD_EXPORT rd_kafka_resp_err_t
 rd_kafka_commit(rd_kafka_t *rk,
                 const rd_kafka_topic_partition_list_t *offsets,
-                int async);
+                int async,
+                int timeout_ms);
```
edenhill, line 4204: These C APIs can't be changed since it would break both the API and ABI compatibility guarantees.

What you would need to do is add new APIs, but this is something we typically try to avoid - to keep the API surface area down and since designing new APIs in a future-proof manner is tricky.

Additionally, I'm not convinced a timeout is the best option here. A timed out commit call could still have sent a commit request to the broker.
I think a better option is to use the existing rd_kafka_commit_queue() API with an callback.

kenneth-jia: Thank you, @edenhill 
I agree on keeping the C APIs unchanged. Apart from that, I'm just wondering if `use the existing rd_kafka_commit_queue() API with an callback` would help, -- e.g. if the brokers are down, and a `sync commit` has just been sent, is there a way to make `rd_kafka_commit` (or, `rd_kafka_commit_queue`) not wait there for ever?

PR #3950, opened 2022-08-23
Do not allocate more memory than configured

daib:
When decompressed batch sizes are big, e.g., `200MB`, the memory allocated for the buffer to store the decompressed batch sizes can become `600MB` based on this buffer growth formular: `out_bufsize += RD_MAX(out_bufsize * 2, 4000);`.  This can happen easily especially with `zstd` when the compression ratio is very good.

If we multiply 600MB by the number of partitions one consumer need to handle, e.g., 30 partitions, then the total memory usage can become 18GB. This is a huge amount of memory used to store decompressed batches. This can lead to a big overshoot for memory usage in `librdkafka` and potentially leads to out of memory issues.

We should not allocate more memory than configured.

### DIFF
--- a/src/rdkafka_zstd.c
+++ b/src/rdkafka_zstd.c
@@ -90,9 +90,23 @@ rd_kafka_resp_err_t rd_kafka_zstd_decompress(rd_kafka_broker_t *rkb,
 
                 /* Check if the destination size is too small */
                 if (ZSTD_getErrorCode(ret) == ZSTD_error_dstSize_tooSmall) {
+                        /* If we already reach the maximum buffer size, don't grow the buffer anymore */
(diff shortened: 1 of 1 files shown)


edenhill:
This is in stark contrast to #3935 :)

daib:
> This is in stark contrast to #3935 :)

I think the problem here is that, `librdkafka` allocates more memory than specified by the config and creates a problem for us. At DataDog, we have had lots of problem out of memory pods because the `zstd` compression is very code. We actually had to set the limit to 1Gb in order for us to process our messages and it is very risky.

Removing the limit for a small system is fine, but for a data intensive system, it is very dangerous.

PR #3952, opened 2022-08-24
Remove warning for allow.auto.create.topics with producer

milindl:
Also improves the documentation around this property.

This solves the following problem:

The property "allow.auto.create.topics" is supposed to be a consumer
property, but we are setting it (and it is affecting the behaviour of)
both the consumer and producer. It gives a warning if we change it in
the producer, but works nevertheless.

(the default value for the producer is true)

One of the effects is that if a user is using their producer as an
adminclient, a call to get metadata for a topic might create that
topic, and if the user specifies allow.auto.create.topics, then they
get a warning.

Unfortunately, we even recommend using a producer with the above
setting (see INTRODUCTION.md).

A knock on effect is that both the go and python clients use a producer
internally for their adminclients so the user has to either live with
a call to GetMetadata creating topics, or with the warning.
(ref: [confluent-kafka-go/#780](https://github.com/confluentinc/confluent-kafka-go/issues/780))

The java client only allows this property to be set on the consumer,
which makes it more confusing.


milindl:
Hi @edenhill :

I'm stuck as to what's the best solution to the problem described above. I discussed a few different approaches with @emasab / @mhowlett  and we thought this was the best way. The other approaches both won't change the property itself. All the approaches are:

* change allow.auto.create.topics scope in librdkafka to include producer, this removes warning. we will need to change the doc also (this PR)
* change all clients individually to use consumer rather than producer for their adminclients
* some other approach where clients use producer by default but use consumer if this config is specified

edenhill:
allow.auto.create.topics is used by the producer too, so I think the proper fix is to remove _RK_CONSUMER from the configuration property and update the docstring to also explain the producer behaviour (i.e., referenced topics will be auto-created).

milindl:
Thanks, it sounds good and that's what I already have done in this PR (removed _RK_CONSUMER, and added doc for the different behaviors of this for producer and consumer)

edenhill:
Please rebase this on latest master.

Also make sure that you do not edit CONFIGURATION.md directly since it is auto-generated by `make`.

Move the CHANGELOG entry to the 1.9.3 Enhancements section.

edenhill:
Make sure to run the style-fixer (on linux, not osx): `make style-fix` (or `make style-fix-changed` prior to committing your stuff, since that's quicker)

src/rdkafka_conf.c
```diff
@@ -58,21 +58,22 @@
 struct rd_kafka_property {
         rd_kafka_conf_scope_t scope;
         const char *name;
-        enum { _RK_C_STR,
```
edenhill, line 61: avoid style/whitespace changes

src/rdkafka_conf.c
```diff
@@ -732,16 +733,30 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
     /* Security related global properties */
     {_RK_GLOBAL | _RK_HIGH, "security.protocol", _RK_C_S2I,
      _RK(security_protocol), "Protocol used to communicate with brokers.",
      .vdef = RD_KAFKA_PROTO_PLAINTEXT,
      .s2i  = {{RD_KAFKA_PROTO_PLAINTEXT, "plaintext"},
-             {RD_KAFKA_PROTO_SSL, "ssl", _UNSUPPORTED_SSL},
-             {RD_KAFKA_PROTO_SASL_PLAINTEXT, "sasl_plaintext"},
-             {RD_KAFKA_PROTO_SASL_SSL, "sasl_ssl", _UNSUPPORTED_SSL},
-             {0, NULL}}},
+              {RD_KAFKA_PROTO_SSL, "ssl", _UNSUPPORTED_SSL},
```
edenhill, line 756: avoid whitespace changes

edenhill (changes requested):
Lots of whitespace/style changes.  Inspect your diffs before committing.

milindl:
Fixed whitespace issues. Sorry about that, clang-format v14.0.0 did not work as expected, had to use v10.0.0.

edenhill (approved):

edenhill:
LGTM

PR #3955, opened 2022-08-26
Adds callback to obtain SASL PLAIN credentials

josvisser66:
This PR adds a callback for obtaining the SASL PLAIN username and password. The use case here is to support changing the credentials to be used for a Kafka cluster without bouncing the Kafka client (e.g. in the case of credential rotating).

The user can set a credentials callback using the new API function rd_kafka_conf_set_plain_creds_cb. That callback will be called by the SASL implementation whenever it needs a SASL username, password, or both. The SASL provider does not cache the provided credentials as this would defeat the purpose.

Trivial and not terribly correct example:

```
int sasl_creds_cb(rd_kafka_t *rk,
                  char *username,
                  int username_size,
                  char *password,
                  int password_size) {
    if (username) {
        strcpy(username, kUsername);
    }
    if (password) {
        strcpy(password, kPassword);
    }
    return 0;
}

...
    conf = rd_kafka_conf_new();
    /* Do not set sasl.username and sasl.password */
    rd_kafka_conf_set_plain_creds_cb(conf, sasl_creds_cb);
```


src/rdkafka_sasl_cyrus.c
```diff
@@ -379,17 +394,31 @@ static int rd_kafka_sasl_cyrus_cb_getsecret(sasl_conn_t *conn,
...
+                int size = 256;
 
-        password = rktrans->rktrans_rkb->rkb_rk->rk_conf.sasl.password;
+                password_ptr = rd_alloca(size);
+
+                if (rk->rk_conf.sasl.plain_creds_cb(rk, NULL, NULL,
+                                                    password_ptr, &size) < 0) {
+                        rd_rkb_log(rkb, LOG_ERR, "SASLPLAIN",
+                                   "SASL password does not fit in "
+                                   "128-byte receiving buffer");
```
ttimonen, line 410: bug/misleading: The implementation has 256 bytes. 

Also, maybe say something like "SASL password does not fit in 256-byte receiving buffer; needs %i", size)
(This is, unless of course we didn't fail for some other reason, in which case the whole buf length story is misleading)
These later stories apply to the username above as well.

src/rdkafka_sasl_plain.c
```diff
@@ -70,30 +70,56 @@ int rd_kafka_sasl_plain_client_new(rd_kafka_transport_t *rktrans,
                                    size_t errstr_size) {
         rd_kafka_broker_t *rkb = rktrans->rktrans_rkb;
         rd_kafka_t *rk         = rkb->rkb_rk;
+
+        char *username_ptr = rk->rk_conf.sasl.username;
+        char *password_ptr = rk->rk_conf.sasl.password;
+
+        if (rk->rk_conf.sasl.plain_creds_cb) {
+                size_t username_size = 128;
```
ttimonen, line 78: consistency nit: Here you use 128 bytes, earlier 256 bytes. Any particular reason for that?

src/rdkafka_sasl_plain.c
```diff
@@ -70,30 +70,56 @@ int rd_kafka_sasl_plain_client_new(rd_kafka_transport_t *rktrans,
...
+                size_t password_size = 128;
+                int attempt;
+                int rc;
+
+                for (attempt = 0; attempt < 2; ++attempt) {
```
ttimonen, line 83: nit: int attempt = 0 here since you don't use it outside the scope of for loop, no?

src/rdkafka_sasl_plain.c
```diff
@@ -70,30 +70,56 @@ int rd_kafka_sasl_plain_client_new(rd_kafka_transport_t *rktrans,
...
+                        username_ptr = rd_alloca(username_size);
+                        password_ptr = rd_alloca(password_size);
```
ttimonen, line 85: Thought: Per RFC4616 the username and pw are limited to 255 bytes anyway, is there an use-case that justifies the additional complexity imposed by memory management and retrying? If not, we can trim this down all over.

josvisser66:
We decided on another approach.

PR #3961, opened 2022-09-04
feat: Add max.poll.records

dark0dave:
Attempting to make the max.poll.records a usable field, I know its available through the method call, however it would be useful to be able to call this method and have it use that value. This would be useful to downstream libraries as well, as implementing a batch call would become very simple.


dark0dave:
Let me know what I have to do to get this landed. Just trying to get the field to work, whist maintaining backwards compatibility.

dark0dave:
@edenhill is this sufficient? Would you like me to write this a different way? Please let me know. Thank you for your time.

edenhill (commented):
I believe the Java consumer max.poll.records is protocol implementation leaking into the public API, it conflates API usage with instance configuration, and not something we want to replicate in librdkafka.

PR #3966, opened 2022-09-06
rdkafkacpp: Eliminate "return by const value" pessimizations.

Quuxplusone:
Returning a `const string` means "My caller can have a copy of
this string, but they aren't allowed to modify it" -- for example,
they aren't allowed to use move semantics to move-out-of the
returned string. Thus `x = y.name();` will use copy assignment
rather than move assignment. This is a needless pessimization
in C++11 and later.

See also the errata marked "nxd" in
https://www.aristeia.com/BookErrata/ec++3e-errata.html


edenhill (commented):
Is this an API and ABI safe change to make?  Probably doesn't affect the API, unless people downgrade to the const-version. ABI wise?

Quuxplusone:
> Is this an [...] ABI safe change to make?

Yes, as far as I know (and I'm fairly confident). Adding or removing `const` from the return type doesn't change the mangling on Itanium ABI (because the return type is not mangled unless it's dependent), and also doesn't change the mangling on MSVC as far as I can tell. https://godbolt.org/z/4c1ear6M9

Obviously this will change observable behavior for someone who's doing something like `decltype(Handle().name())`, but that's true of pretty much any change, because C++. :)

It certainly won't change the vtable layout or anything like that.

Quuxplusone:
Gentle ping!

edenhill:
Can you add a line to the 1.9.3 release's Enhancements list in CHANGELOG.md and mention this allows move optimization?

edenhill:
I believe this will break the ABI if const is part of the symbol name mangling, so what about we just add non-const alternatives?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ librdkafka v1.9.3 is a maintenance release:
 ## Enhancements
 
  * Bundled zlib upgraded to version 1.2.13.
+ * The C++ API no longer returns strings "by const value," which enables better move optimization in callers.
```
edenhill, line 18:
```suggestion
 * The C++ API no longer returns strings "by const value" which enables better move optimization in callers.
```

edenhill (changes requested):

CHANGELOG.md
Quuxplusone: Well, we want to make it (grammatically) clear that it's the _removal_ of "return by const value" that enables the better optimization. We're not removing something _which enables optimization._ We're removing something, _which enables optimization._

Could change to `* To enable better move optimization in callers, the C++ API no longer returns strings "by const value."` if you like that better.

edenhill: Or simply remove the quotes, I think that's probably the best option.

edenhill:
Thank you!

PR #3967, opened 2022-09-06
rdkafkacpp: Fix all -Wsuggest-override warnings

Quuxplusone:
Then, also in this branch:
- rdkafkacpp: Remove an unused non-virtual function.

The non-virtual function has the same name as a bunch of virtual functions, and so I'm guessing that it was an oversight — it meant to override something from the base class, but in fact there was nothing to override with that particular signature, so it's simply an unused (dead) function.


Quuxplusone:
Gentle ping! and question: what's the best way to deal with the current test failure, which indicates that you can't use `override` in C++98 mode? I'd argue that C++98 is 24 years and 5 revisions-of-the-language ago, and it'd be fine to drop support for it now. But one could also add
```
#if __cplusplus >= 201103L
 #define RDKAFKA_OVERRIDE override
#else
 #define RDKAFKA_OVERRIDE
#endif
```
somewhere near the top of `rdkafkacpp_int.h` (and optionally `#undef` it at the bottom).

edenhill:
Older MSVC toolchains typically have very old C++/C standard version support, so we need to be conservative with requiring modern things. So I'd prefer the ifdef approach.

Quuxplusone:
Updated! I named the new macro `RD_OVERRIDE` for consistency with `RD_EXPORT`. clang-format seems mostly happy with it; I tried also adding `AttributeMacros: ['RD_OVERRIDE']` to the clang-format file, but that (A) seemed maybe unnecessary and (B) made the CI unhappy because its version of clang-format is pretty old.

At some future point, when C++11 support can be assumed, someone can make a PR that's just `s/RD_OVERRIDE/override/g`.

Quuxplusone:
Rebased — post-Thanksgiving ping!

I think it would be nice to add `-Wsuggest-override` to the project's default CXXFLAGS as well (when Clang is being used), just to avoid regressions in this department. But a brief glance at https://github.com/edenhill/librdkafka/blob/master/mklove/modules/configure.good_cflags did not enlighten me as to how to _conditionally_ add something to CXXFLAGS (since not all compilers support `-Wsuggest-override`), so I'll leave that idea out-of-scope for now.

Quuxplusone:
Ping!

Quuxplusone:
@edenhill ping! This is still just waiting to be merged, AFAIK.

PR #3968, opened 2022-09-07
CLIENTS-2514: Fixed broken links

pranavrth:
Fixing 4 broken links in the documentation.

1. Statistics Schema JSON
2. transaction.c example file
3. rdkafka.h src file
4. rdkafka_ssl.c src file

### DIFF
--- a/INTRODUCTION.md
+++ b/INTRODUCTION.md
@@ -241,7 +241,7 @@ on the producer:
 
 See [STATISTICS.md](STATISTICS.md) for the full definition of metrics.
 A JSON schema for the statistics is available in
-[statistics-schema.json](src/statistics-schema.json).
+[statistics_schema.json](https://github.com/edenhill/librdkafka/blob/master/src/statistics_schema.json).
 
 
 ### Compression
@@ -776,7 +776,7 @@ automatically.
 librdkafka supports Exactly One Semantics (EOS) as defined in [KIP-98](https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging).
 For more on the use of transactions, see [Transactions in Apache Kafka](https://www.confluent.io/blog/transactions-apache-kafka/).
 
-See [examples/transactions.c](examples/transactions.c) for an example
+See [examples/transactions.c](https://github.com/edenhill/librdkafka/blob/master/examples/transactions.c) for an example
 transactional EOS application.
 
 **Warning**
@@ -791,7 +791,7 @@ See KIP-447 for more information.
 
 ### Documentation
 
-The librdkafka API is documented in the [`rdkafka.h`](src/rdkafka.h)
+The librdkafka API is documented in the rdkafka.h
 header file, the configuration properties are documented in
 [`CONFIGURATION.md`](CONFIGURATION.md)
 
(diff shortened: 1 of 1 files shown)


INTRODUCTION.md
```diff
@@ -241,7 +241,7 @@ on the producer:
 
 See [STATISTICS.md](STATISTICS.md) for the full definition of metrics.
 A JSON schema for the statistics is available in
-[statistics-schema.json](src/statistics-schema.json).
+[statistics_schema.json](https://github.com/edenhill/librdkafka/blob/master/src/statistics_schema.json).
```
jonchiu, line 244: I think we can still use relative links here. The broken link was the result of using `-` instead of `_`.

I would update this to 
```
[statistics-schema.json](src/statistics_schema.json).
```

INTRODUCTION.md
```diff
@@ -776,7 +776,7 @@ automatically.
 librdkafka supports Exactly One Semantics (EOS) as defined in [KIP-98](https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging).
 For more on the use of transactions, see [Transactions in Apache Kafka](https://www.confluent.io/blog/transactions-apache-kafka/).
 
-See [examples/transactions.c](examples/transactions.c) for an example
+See [examples/transactions.c](https://github.com/edenhill/librdkafka/blob/master/examples/transactions.c) for an example
```
jonchiu, line 779: this link wasn't broken for me

INTRODUCTION.md
```diff
@@ -791,7 +791,7 @@ See KIP-447 for more information.
 
 ### Documentation
 
-The librdkafka API is documented in the [`rdkafka.h`](src/rdkafka.h)
+The librdkafka API is documented in the rdkafka.h
```
jonchiu, line 794: what was the reason for removing this link?

INTRODUCTION.md
```diff
@@ -1090,7 +1090,7 @@ by setting `enable.ssl.certificate.verification=false`, but this is not
 recommended since it allows for rogue brokers and man-in-the-middle attacks,
 and should only be used for testing and troubleshooting purposes.
 
-CA location probe paths (see [rdkafka_ssl.c](src/rdkafka_ssl.c) for full list)
+CA location probe paths (see [rdkafka_ssl.c](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_ssl.c) for full list)
```
jonchiu, line 1093: this link wasn't broken for me

jonchiu (changes requested):

pranavrth:
The links are broken in this doc - https://docs.confluent.io/platform/current/clients/librdkafka/html/md_INTRODUCTION.html

Please verify here. In github, all the links except statistics_schema.json works fine.

pranavrth:
If we see Doxyfile, we can see that we are not generating examples or src file documentations except for rdkafka.h . Hence, I kept absolute links.

https://github.com/edenhill/librdkafka/blob/master/Doxyfile#L767

jonchiu:
Ah, thanks for clarifying -- so it seems like in the example for `rdkafka.h`, it should point to https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html. I am not sure how we fix this for doxygen while not hardcoding the path.

pranavrth:
> Ah, thanks for clarifying -- so it seems like in the example for `rdkafka.h`, it should point to https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html. I am not sure how we fix this for doxygen while not providing the absolute path.

If we use `rdkafka.h`, it directly points to https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html in the docs. (search for rdkafka.h in https://docs.confluent.io/platform/current/clients/librdkafka/html/md_INTRODUCTION.html)

In github, we just specify `rdkafka.h`. (example - https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#simple-consumer-api-legacy)

PR #3971, opened 2022-09-08
Fix for fenced producer after disconnection

emasab:
This is a collection of fixes developed while running different long running chaos tests with single and multiple brokers and several configurations. 

* When requesting an abort with drain and bump after message timeouts, the bump was executed first, causing a series of retries on InitProducerIdRequest until a transaction timeout happen and the producer was fenced. This commit changes the order of the operations, starting the drain bump after the complete transaction ack.
 * While a commit operation was in queue, a timeout happens that can cause an abort. The state changes from COMMITTING_TRANSACTION to ABORTABLE_ERROR to ABORTING_TRANSACTION. When the broker comes up the error is permanent or fatal because the state has changed from the initial one.
 * When doing a drain and bump txn_curr_coord is not null in state WAIT_TRANSPORT, but if txn_coord has to be requested or broker is down it's retried. During this retry the txn_curr_coord can be set to null after a COORDINATOR_NOT_AVAILABLE error so when it comes back to the WAIT_TRANSPORT case a fatal error happens in the assert. It has to query for a transaction coordinator and retry.
 * When aborting a transaction, if a local TIMED_OUT, TIMED_OUT_QUEUE or OUTDATED error happens, the error was not retriable (nor fenced or abortable) so rd_kafka_txn_op_abort_transaction_ack is never called, caller code doesn't know what to do so starts a new transaction but begin transaction fails because transactional state is not READY. These errors are now retriable.
 * When calling TxnOffsetCommit, if retries were needed they were made at full speed, causing too much logging and network calls. A one second timeout has been added between retries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,19 @@
 
 librdkafka v1.9.3 is a maintenance release:
 
- *  Self-contained static libraries can now be built on Linux arm64 too
+ * Self-contained static libraries can now be built on Linux arm64 too
+ * Fixes to the transactional and idempotent producer (#3971).
+
+
+## Fixes
+
+### Transactional producer fixes
+
+ * When requesting an abort with drain and bump after message timeouts, the bump was executed first, causing a series of retries on InitProducerIdRequest until a transaction timeout happen and the producer was fenced. This commit changes the order of the operations, starting the drain bump after the complete transaction ack (#3971).
+ * While a commit operation was in queue, a timeout happens that can cause an abort. The state changes from COMMITTING_TRANSACTION to ABORTABLE_ERROR to ABORTING_TRANSACTION. When the broker comes up the error is permanent or fatal because the state has changed from the initial one (#3971).
+ * When doing a drain and bump txn_curr_coord is not null in state WAIT_TRANSPORT, but if txn_coord has to be requested or broker is down it's retried. During this retry the txn_curr_coord can be set to null after a COORDINATOR_NOT_AVAILABLE error so when it comes back to the WAIT_TRANSPORT case a fatal error happens in the assert. It has to query for a transaction coordinator and retry (#3971).
+ * When aborting a transaction, if a local TIMED_OUT, TIMED_OUT_QUEUE or OUTDATED error happens, the error was not retriable (nor fenced or abortable) so rd_kafka_txn_op_abort_transaction_ack is never called, caller code doesn't know what to do so starts a new transaction but begin transaction fails because transactional state is not READY. These errors are now retriable (#3971).
+ * When calling TxnOffsetCommit, if retries were needed they were made at full speed, causing too much logging and network calls. A one second timeout has been added between retries (#3971).
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 librdkafka v1.9.3 is a maintenance release:
 
- * Self-contained static libraries can now be built on Linux arm64 too
+ * Self-contained static libraries can now be built on Linux arm64.
  * Fixes to the transactional and idempotent producer (#3971).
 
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,19 @@
 
 librdkafka v1.9.3 is a maintenance release:
 
- *  Self-contained static libraries can now be built on Linux arm64 too
+ * Self-contained static libraries can now be built on Linux arm64 too
```
mhowlett, line 5: get rid of the "too". Just: "Self-contained static libraries can now be built on Linux arm64"

mhowlett: and remember the full stop.

CHANGELOG.md
```diff
@@ -2,7 +2,19 @@
...
+ * Fixes to the transactional and idempotent producer (#3971).
+
+
+## Fixes
+
+### Transactional producer fixes
+
+ * When requesting an abort with drain and bump after message timeouts, the bump was executed first, causing a series of retries on InitProducerIdRequest until a transaction timeout happen and the producer was fenced. This commit changes the order of the operations, starting the drain bump after the complete transaction ack (#3971).
```
mhowlett, line 13: maybe: "In certain circumstances, an `InitProducerId` request would be issued to bump the producer epoch before the open transaction was aborted, causing the producer to enter a failed state. The producer will now wait until the transaction is aborted before requesting an epoch bump."?

CHANGELOG.md
```diff
@@ -2,7 +2,19 @@
...
+ * While a commit operation was in queue, a timeout happens that can cause an abort. The state changes from COMMITTING_TRANSACTION to ABORTABLE_ERROR to ABORTING_TRANSACTION. When the broker comes up the error is permanent or fatal because the state has changed from the initial one (#3971).
```
mhowlett, line 14: it's not clear from this wording what was existing behavior and what you changed.

src/rdkafka_txnmgr.c
```diff
@@ -2027,21 +2099,37 @@ rd_kafka_error_t *rd_kafka_send_offsets_to_transaction(
...
+            rk->rk_eos.idemp_state != RD_KAFKA_IDEMP_STATE_ASSIGNED;
+
         rk->rk_eos.txn_requires_epoch_bump = rd_false;
         rk->rk_eos.txn_req_cnt             = 0;
 
-        rd_kafka_txn_set_state(rk, RD_KAFKA_TXN_STATE_READY);
+        if (drain_bump) {
+                rd_kafka_wrunlock(rk);
+                rd_kafka_idemp_drain_epoch_bump_start(
+                    rk, "txn_requires_epoch_bump");
```
mhowlett, line 2126: underscores not needed

src/rdkafka_txnmgr.c
```diff
@@ -2086,14 +2174,26 @@ static void rd_kafka_txn_handle_EndTxn(rd_kafka_t *rk,
 err:
         rd_kafka_wrlock(rk);
 
-        if (rk->rk_eos.txn_state == RD_KAFKA_TXN_STATE_COMMITTING_TRANSACTION) {
+        int state_committing =
```
mhowlett, line 2177: these booleans do make it more readable IMO (good idea!), however:
1. you need to declare the variables at the top of the function, not here.
2. i don't think it's consistent with the librdkafka coding style to use them - just use the == expressions (even though it'll be harder to read)

src/rdkafka_txnmgr.c
```diff
@@ -2086,14 +2174,26 @@ static void rd_kafka_txn_handle_EndTxn(rd_kafka_t *rk,
...
+            rk->rk_eos.txn_state == RD_KAFKA_TXN_STATE_COMMITTING_TRANSACTION;
+        int state_aborting =
+            rk->rk_eos.txn_state == RD_KAFKA_TXN_STATE_ABORTING_TRANSACTION;
+        int state_abortable_error =
+            rk->rk_eos.txn_state == RD_KAFKA_TXN_STATE_ABORTABLE_ERROR;
+        // While request was in queue, state can have changed from:
```
mhowlett, line 2183: librdkafka coding style has /* */ only.

src/rdkafka_txnmgr.c
mhowlett: actually I take back point #2 - the logic for `state_different_from_request` really does benefit enough to keep them...

src/rdkafka_txnmgr.c
```diff
@@ -1593,20 +1653,32 @@ static void rd_kafka_txn_handle_TxnOffsetCommit(rd_kafka_t *rk,
...
-                            rd_timeout_remains_limit0(
-                                remains_ms, rk->rk_conf.socket_timeout_ms),
-                            RD_KAFKA_REPLYQ(rk->rk_ops, 0),
-                            rd_kafka_txn_handle_TxnOffsetCommit, rko);
+                        rd_ts_t diff =
+                            rd_interval(&last_retry, 1000 * 1000 /* 1s */, 0);
+                        if (diff < 0) {
+                                rd_kafka_dbg(
+                                    rk, EOS, "TXNOFFSETCOMMIT",
+                                    "Waiting %ld µs before next retry. rkb:%s "
```
mhowlett, line 1661: i think there's a macro you should use rather than %ld, look at other examples.
and do you really want to report in micro-seconds? i think better to use milliseconds. or if that's too coarse, no log message at all.

src/rdkafka_idempotence.c
```diff
@@ -599,6 +613,32 @@ void rd_kafka_idemp_drain_reset(rd_kafka_t *rk, const char *reason) {
...
+ * @param fmt is a human-readable reason for the bump.
+ *
+ * @locality any
+ * @locks none
+ */
+void rd_kafka_idemp_drain_epoch_bump_start(rd_kafka_t *rk, const char *reason) {
+        rd_kafka_wrlock(rk);
+        rd_kafka_dbg(rk, EOS, "DRAIN",
+                     "Beginning partition drain for %s epoch bump "
+                     "for %d partition(s) with in-flight requests: %s",
```
mhowlett, line 625: there's a macro for %d - look to other examples.

mhowlett (approved):
some minor comments.
LGTM, for an RC release, given the amount of testing you've done and earlier discussions.

Issue #3929, opened 2022-07-29
Multi-Threaded rd_jitter Broken on Windows


brimonk:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
This has been discussed in the past (#2795), but the output of `rand()` is predictable. This predictability is usually alleviated with the use of `srand()`, or by calling `rand_r()`.

While POSIX went and defined `rand_r` to allow for reentrant PRNG, the Microsoft solution to this problem was to make `rand` and similar C functions thread-safe by [storing the CRT global state in thread-safe storage](https://docs.microsoft.com/en-us/cpp/c-runtime-library/global-state?view=msvc-170#global-state-affected-by-appos-separation). Given two threads, A and B, if thread A is the one to call `srand()` followed by thread B calling `rand()`, the PRGN seed will be the default for thread B, resulting in the same fixed sequence.

Long story short, this means that librdkakfa cannot ensure the calling application may call any of the API functions from any of its own threads at any time on Windows.

How to reproduce
================

Reproduction steps below assume Windows:

1. Enable debug output
2. Call `rd_kafka_new` and obtain an `rd_kafka_t` handle on thread A
3. In a loop (more than once to validate), spawn a new thread, passing this handle to the new thread
4. Produce a single message to the configured topic on thread B
5. Analyze the logging output to see messages like the following:

```
Debug : [thrd:main]: TOPIC [5] is the new sticky partition
```

In a production scenario where this was spotted, the particular topic had 12 partitions, and partition 5 was the one that was selected every single time. To demonstrate this without using librdkafka, you can look at the following example:

```c
// Compile with:
//     cl randtest.c
//
// Given PARTITION_CNT == 12, this will output:
//
//   0 0
//   1 0
//   2 0
//   3 0
//   4 0
//   5 1000
//   6 0
//   7 0
//   8 0
//   9 0
//   10 0
//   11 0

#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>

#define WIN32_LEAN_AND_MEAN
#include <windows.h>

#define PARTITION_CNT 12

long RANDVALS[PARTITION_CNT];

// rd_jitter: the rd_jitter definition from v1.2.0 (effectively the same from v1.8.0 without rand_r)
static inline int rd_jitter(int low, int high)
{
	return (low + (rand() % ((high - low) + 1)));
}

// randtest: does the rand test
DWORD __stdcall randtest(void *p)
{
[...]

Issue #2795, opened 2020-04-01
Potential broken rand() in rdrand.h


by700git:
Read the FAQ first: https://github.com/edenhill/librdkafka/wiki/FAQ

Description
===========
Not sure if anyone noticed this before, but seems this `rd_jitter` is used across multiple functions such as assigning partitions and select broker to send request:

in `rdrand.h`
```c
static RD_INLINE int rd_jitter (int low, int high) {
	return (low + (rand() % ((high-low)+1)));
	
}
```
I tested the behavior of `rand()` on multiple machines with same OS version (Windows Server 2016), it seems they are generating the same random sequence.

In our prod scenario, this potentially cause all newly started consumers send metadata request to the same broker, and the broker's request queue get filled up.

Potential fix might be:
1. Use something like `srand(time(NULL))`?
2. Or use `random(3)`? https://linux.die.net/man/3/rand

How to reproduce
================
In description

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `master` `commit 71a9d8568edf3044af28756d22b44b7bcb1ac7ca` `v1.4.0-RC7`
 - [x] Apache Kafka version: `2.3.1`
 - [x] librdkafka client configuration: `all confs`
 - [x] Operating system: `Windows Server 2016`
 - [x] Provide logs `N/A`
 - [x] Provide broker log excerpts
 - [x] Critical issue: In our case, the broken rand() potentially cause consumers on different machines to send request to one broker, which will fill up the request queue if these are some heavy requests.

edenhill:
We rely on the application initializing the random generator with a seed by calling srand(), but this isn't documented, which is clearly a flaw.

We could call srand()/srandom() ourselves, but I'm not completely certain that's a good idea for a library to do since it will affect the application's use of rand(), i.e., if an application relies on predictable random numbers by supplying its own seed with srand() we will silently break that by calling srand() internally.
One way around that would be to have a `enable.random.seed` which defaults to `true`, having librdkafka call srand() once on startup, but that the application can disable that by setting `enable.random.seed=false`.

Comments?

rolandyoung:
Better to default to false, so that correctly written apps don't have to make any change? Just documenting the setting should provide a sufficient hint to those to whom this matters.

edenhill:
[...]

PR #3975, opened 2022-09-09
rand: Ensure rd_jitter functions properly for new threads (#3929)

brimonk:
In cases where a Windows producer spins up a new thread to produce a single message, the value emitted by rd_jitter is exactly the same for every new thread.

On Windows, calls to srand() modify thread-local state, with subsequent calls to rand() reading that same thread-local state.

The previous fix to call srand() once during rd_kafka_new (#2795) does not fully fix this situation, as a brand new thread can take the object without ever calling rd_kafka_new() or rd_kafka_global_srand().

This patch rectifies this problem by using a version of rand_r() from the musl c library for Win32 platforms. On POSIX platforms, rd_rand_r() is a #define that points to the system's existing rand_r().

### DIFF
--- a/src/rdrand.c
+++ b/src/rdrand.c
@@ -33,21 +33,15 @@
 
 int rd_jitter(int low, int high) {
         int rand_num;
-#if HAVE_RAND_R
+        struct timeval tv;
         static RD_TLS unsigned int seed = 0;
 
-        /* Initial seed with time+thread id */
-        if (unlikely(seed == 0)) {
-                struct timeval tv;
-                rd_gettimeofday(&tv, NULL);
-                seed = (unsigned int)(tv.tv_usec / 1000);
-                seed ^= (unsigned int)(intptr_t)thrd_current();
-        }
+        rd_gettimeofday(&tv, NULL);
+        seed = (unsigned int)(tv.tv_usec / 1000);
+        seed ^= (unsigned int)(intptr_t)thrd_current();
 
-        rand_num = rand_r(&seed);
-#else
-        rand_num = rand();
-#endif
+        rand_num = rd_rand_r(&seed);
+
         return (low + (rand_num % ((high - low) + 1)));
 }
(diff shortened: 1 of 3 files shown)


src/rdrand.c
```diff
@@ -32,23 +32,16 @@
 #include "tinycthread.h"
 
 int rd_jitter(int low, int high) {
-        int rand_num;
-#if HAVE_RAND_R
```
edenhill, line 36: This makes it so that rand_r() is required on non-windows platforms, which we can't really assert.

I think a better option is to use HAVE_RAND_R and only provide the musl-inspired rd_rand_r() if HAVE_RAND_R is false.

src/rdwin32.h
```diff
@@ -377,6 +377,32 @@ static RD_UNUSED int rd_pipe_nonblocking(rd_socket_t *fds) {
...
+ */
+
+/**
+ * @brief rand_r() for Windows systems to ensure thread agnostic rand() results
+ * @returns a pseudo-random integer in the range 0 to RAND_MAX inclusive [0,
+ * RAND_MAX].
+ */
+static RD_UNUSED RD_INLINE int
+rd_rand_r(unsigned int *seed) {
+        /* Source: https://git.musl-libc.org/cgit/musl/tree/src/prng/rand_r.c */
```
edenhill, line 391: This has license implications.

Since your implementation here is not a verbatim copy of musl's implementation, I believe it is safe to say that this implementation is inspired by the above link, rather than calling it source.

Alternatively we add a rdrandr.c that is only built if HAVE_RAND_R is not set and then add a LICENSE.rdrandr to the top-level directoy containing which file is covered (src/rdrandr.c) and the musl license (MIT).

brimonk:
@edenhill the commit history on this PR is a little messy. If you'd like me to clean it up as per the contributing guidelines, I can.

PR #3987, opened 2022-09-26
List alter group offset helper functions

pranavrth:
Added helper functions for ListConsumerGroupOffsets and AlterConsumerGroupOffsets being used in Python client.


pranavrth:
Not Required.

PR #3995, opened 2022-09-29
Implemented KIP-88, KIP-222, KIP-518 and partially KIP-396

emasab:
Started from the initial KIP-396/KIP-222 implementation of @lesterfan, this PR contains implementation for the following KIPs

 * [KIP-88](https://cwiki.apache.org/confluence/display/KAFKA/KIP-88%3A+OffsetFetch+Protocol+Update) OffsetFetch Protocol Update.
 * [KIP-222](https://cwiki.apache.org/confluence/display/KAFKA/KIP-222+-+Add+Consumer+Group+operations+to+Admin+API) Add Consumer Group operations to Admin API (started by @lesterfan).
 * [KIP-518](https://cwiki.apache.org/confluence/display/KAFKA/KIP-518%3A+Allow+listing+consumer+groups+per+state) Allow listing consumer groups per state.
 * [KIP-396](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=97551484) Partially implemented: support for AlterConsumerGroupOffsets (started by @lesterfan).


src/rdkafka.h
```diff
@@ -4763,6 +4765,11 @@ typedef struct rd_kafka_metadata {
 } rd_kafka_metadata_t;
 
 
+/**
+ * @brief Node information.
```
edenhill, line 4769: Might be good to indicate that Node represents a broker, so:
`* @brief Node (broker) information.`

src/rdkafka.h
```diff
@@ -4798,6 +4805,33 @@ RD_EXPORT
...
+ * @return The node id, or -1 is \p node is NULL.
+ */
+RD_EXPORT
+int rd_kafka_Node_id(rd_kafka_Node_t *node);
+
+/**
+ * @brief Get the host of \p node.
+ *
+ * @param node The Node instance.
+ * @return The node host, or NULL is \p node is NULL.
```
edenhill, line 4821: Add:
`@remark The lifetime of the returned memory is the same as the lifetime of \p node object.`

src/rdkafka.h
```diff
@@ -4798,6 +4805,33 @@ RD_EXPORT
...
 
 
+/**
+ * @brief Get the id of \p node.
+ *
+ * @param node The Node instance.
+ * @return The node id, or -1 is \p node is NULL.
+ */
+RD_EXPORT
+int rd_kafka_Node_id(rd_kafka_Node_t *node);
```
edenhill, line 4815: If we don't foresee this object to be mutable then we should consider consting the node param.

src/rdkafka.h
```diff
@@ -4827,8 +4861,24 @@ struct rd_kafka_group_member_info {
         void *member_assignment;    /**< Member assignment (binary),
                                      *    format depends on \p protocol_type. */
         int member_assignment_size; /**< Member assignment size in bytes */
+        rd_kafka_topic_partition_list_t
```
edenhill, line 4864: Public types may not be altered since it would break the ABI compatibility guarantees, so you will probably need to add a new Admin type here that is private.

src/rdkafka.h
```diff
@@ -4827,8 +4861,24 @@ struct rd_kafka_group_member_info {
...
+            *member_assignment_toppars; /**< Member assignment topic partitions
+                                         */
 };
 
+/**
+ * @brief Consumer group states
+ */
+typedef enum {
+        RD_KAFKA_CGRP_STATE_UNKNOWN              = 0,
```
edenhill, line 4873: CONSUMER_GROUP

"cgrp" is an internal librdkafka thing.

src/rdkafka.h
```diff
@@ -4827,8 +4861,24 @@ struct rd_kafka_group_member_info {
         void *member_assignment;    /**< Member assignment (binary),
                                      *    format depends on \p protocol_type. */
         int member_assignment_size; /**< Member assignment size in bytes */
+        rd_kafka_topic_partition_list_t
+            *member_assignment_toppars; /**< Member assignment topic partitions
+                                         */
 };
 
+/**
+ * @brief Consumer group states
```
edenhill, line 4870: `@enum`

src/rdkafka.h
```diff
@@ -4837,10 +4887,12 @@ struct rd_kafka_group_info {
         char *group;                            /**< Group name */
         rd_kafka_resp_err_t err;                /**< Broker-originated error */
         char *state;                            /**< Group state */
-        char *protocol_type;                    /**< Group protocol type */
-        char *protocol;                         /**< Group protocol */
+        rd_kafka_consumer_group_state_t state_code; /**< Group state code */
```
edenhill, line 4890: dito

src/rdkafka.h
```diff
@@ -4853,11 +4905,22 @@ struct rd_kafka_group_list {
         int group_cnt;                      /**< Group count */
 };
 
+/**
+ * @brief Options for rd_kafka_list_consumer_groups
```
edenhill, line 4909: These are named rd_kafka_ListConsumerGroups() and .. _DescribeConsumerGroups(), right?

src/rdkafka.h
```diff
@@ -4853,11 +4905,22 @@ struct rd_kafka_group_list {
...
+ */
+typedef struct rd_kafka_list_consumer_groups_options_s
+    rd_kafka_list_consumer_groups_options_t;
```
edenhill, line 4912: If the type is specific to a certain Admin API then use the CamelCase format to match the API name for the type, e.g.; rd_kafka_ListConsumerGroups_options_t;

If the type is more generic, such as topic_result_t, then use snake_case.

src/rdkafka.h
```diff
@@ -4880,14 +4943,153 @@ struct rd_kafka_group_list {
...
+ * @returns A pointer to the newly created
+ * rd_kafka_list_consumer_groups_options_t,
+ * or NULL if options parameters are invalid.
+ *
+ * @sa Use rd_kafka_list_consumer_groups_options_destroy() to release it.
+ */
+RD_EXPORT
+rd_kafka_list_consumer_groups_options_t *
+rd_kafka_list_consumer_groups_options_new(
+    int request_timeout,
```
edenhill, line 4973: This is already on the standard AdminOptions, isn't it?

src/rdkafka.h
```diff
@@ -4880,14 +4943,153 @@ struct rd_kafka_group_list {
...
 rd_kafka_resp_err_t
 rd_kafka_list_groups(rd_kafka_t *rk,
                      const char *group,
                      const struct rd_kafka_group_list **grplistp,
                      int timeout_ms);
 
+/**
+ * @brief Creates an options struct for rd_kafka_list_consumer_groups.
+ *
+ * \p request_timeout the request timeout.
```
edenhill, line 4960: Use `@param ..` when list parameters like this. `\p ..` is for use in sentences.

src/rdkafka.h
```diff
@@ -4880,14 +4943,153 @@ struct rd_kafka_group_list {
...
+ * \p states An array with the states to query, NULL to query for all the
+ * states. \p states_cnt Size of the state array or 0 if NULL.
+ *
+ * @returns A pointer to the newly created
```
edenhill, line 4964: If there are multiple error cases then maybe it is better to have this return an error_t* object so the caller will know what went wrong.

src/rdkafka.h
edenhill: (and timeouts should always include the unit, e.g., request_timeout_ms)

src/rdkafka.h
```diff
@@ -4880,14 +4943,153 @@ struct rd_kafka_group_list {
...
+ * \p request_timeout the request timeout.
+ *
+ * @returns A pointer to the newly created
+ * rd_kafka_describe_consumer_groups_options_t.
+ *
+ * @sa Use rd_kafka_describe_consumer_groups_options_destroy() to release it.
+ */
+RD_EXPORT
+rd_kafka_describe_consumer_groups_options_t *
+rd_kafka_describe_consumer_groups_options_new(int request_timeout);
```
edenhill, line 4999: rd_kafka_AdminOptions_set_request_timeout

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+
+
+/**
+ * @brief Gets the partition assignor for the \p desc group.
+ *
+ * @param desc The group description.
+ * @return The partition assignor, or NULL if \p desc is NULL.
+ */
+RD_EXPORT
+char *rd_kafka_ConsumerGroupDescription_partition_assignor(
```
edenhill, line 7732: const char *

Also add:
`@remark The lifetime of the returned string is the same as the \p grpdesc object.`

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+    const rd_kafka_ConsumerGroupDescription_t *desc);
+
+/**
+ * @brief Gets the members count of \p desc group.
+ *
+ * @param desc The group description.
+ * @return The member count, or 0 if \p desc is NULL.
+ */
+RD_EXPORT
+int rd_kafka_ConsumerGroupDescription_member_cnt(
```
edenhill, line 7763: you can spell this out: _count(), since these APIs are generally quite expressive.

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+    int idx);
+
+/**
+ * @brief Gets client id of a \p member.
+ *
+ * @param member The group member.
+ * @return The client id, or NULL if \p member is NULL.
+ */
+RD_EXPORT
+char *rd_kafka_MemberDescription_client_id(
```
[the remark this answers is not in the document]
edenhill: add remark on lifetime for any API that returns a pointer.

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+    const rd_kafka_MemberDescription_t *member);
+
+/**
+ * @brief Gets host of a \p member.
+ *
+ * @param member The group member.
+ * @return The host, or NULL if \p member is NULL.
+ */
+RD_EXPORT
+char *
```
edenhill, line 7806: const

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+/**
+ * @brief Get an array of group results from a DescribeGroups result.
+ *
+ * The returned groups life-time is the same as the \p result object.
+ *
+ * @param result Result to get group results from.
+ * @param cntp is updated to the number of elements in the array.
+ */
+RD_EXPORT
+const rd_kafka_ConsumerGroupDescription_t **
```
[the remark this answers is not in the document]
edenhill: Is the Java client type named ConsumerGroupDescription or GroupDescription?

Reason I'm asking is that it is a bit misaligned with the API named DescribeGroups() but the returned value is ConsumerGroupDescription, and the protocol requests seem to be group-type agnostic.

src/rdkafka.h
```diff
@@ -7410,9 +7905,196 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ * @brief Helper function to destroy all ListConsumerGroupOffsets objects in
+ *        the \p list_grpoffsets array (of \p list_grpoffsets_cnt elements).
+ *        The array itself is not freed.
+ */
+RD_EXPORT void rd_kafka_ListConsumerGroupOffsets_destroy_array(
+    rd_kafka_ListConsumerGroupOffsets_t **list_grpoffsets,
+    size_t list_grpoffset_cnt);
+
+/**
+ * @brief List committed offsets for a set of partitions in a conusmer
```
edenhill, line 7955: consumer

src/rdkafka.h
```diff
@@ -7410,9 +7905,196 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ *        group.
+ *
+ * @param rk Client instance.
+ * @param list_grpoffsets Array of group committed offsets to list.
```
edenhill, line 7959: Why limited to 1?

src/rdkafka.h
```diff
@@ -7410,9 +7905,196 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ * @param partitions Partitions to alter committed offsets for.
+ *                   Only the topic and partition fields are used.
+ *
+ * @returns a new allocated AlterConsumerGroupOffsets object.
+ *          Use rd_kafka_AlterConsumerGroupOffsets_destroy() to free
+ *          object when done.
+ */
+RD_EXPORT rd_kafka_AlterConsumerGroupOffsets_t *
+rd_kafka_AlterConsumerGroupOffsets_new(
+    const char *group,
```
edenhill, line 8027: Make sure we're consistent with group vs group_id (the latter is better I think)

src/rdkafka.h
```diff
@@ -7410,9 +7905,196 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ * @brief Helper function to destroy all AlterConsumerGroupOffsets objects in
+ *        the \p alter_grpoffsets array (of \p alter_grpoffsets_cnt elements).
+ *        The array itself is not freed.
+ */
+RD_EXPORT void rd_kafka_AlterConsumerGroupOffsets_destroy_array(
+    rd_kafka_AlterConsumerGroupOffsets_t **alter_grpoffsets,
+    size_t alter_grpoffset_cnt);
+
+/**
+ * @brief Alter committed offsets for a set of partitions in a conusmer
```
edenhill, line 8047: consumer

edenhill (commented):
I started out by commenting on the first incarnation of the API before I realized I was supposed to the Admin part, but I think there's some useful comments there as well.

src/rdkafka.h
emasab: These are from the old implementation. I intended to remove them as last thing, going to do it today.

src/rdkafka.h
emasab: dito

src/rdkafka.h
emasab: dito

src/rdkafka.h
emasab: dito

src/rdkafka.h
```diff
@@ -4880,14 +4943,153 @@ struct rd_kafka_group_list {
...
+ * @brief Returns a code for a state name.
+ *
+ * \p state The state name.
+ */
+RD_EXPORT
+rd_kafka_consumer_group_state_t
+rd_kafka_consumer_group_state_code(const char *name);
+
+/**
+ * @brief List client groups in cluster.
```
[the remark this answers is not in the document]
emasab: dito

src/rdkafka.h
```diff
@@ -4880,14 +4943,153 @@ struct rd_kafka_group_list {
...
+ *
+ *           The \p grplistp remains untouched if any error code is returned,
+ *           with the exception of RD_KAFKA_RESP_ERR__PARTIAL which behaves
+ *           as RD_KAFKA_RESP_ERR__NO_ERROR (success) but with an incomplete
+ *           group list.
+ *
+ * @sa Use rd_kafka_group_list_destroy() to release list memory.
+ */
+RD_EXPORT
+rd_kafka_resp_err_t rd_kafka_list_consumer_groups(
```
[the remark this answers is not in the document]
emasab: dito

src/rdkafka.h
emasab: dito

src/rdkafka.h
emasab: dito

src/rdkafka.h
```diff
@@ -4880,14 +4943,153 @@ struct rd_kafka_group_list {
...
+rd_kafka_describe_consumer_groups_options_new(int request_timeout);
+
+/**
+ * @brief Destroys the options struct.
+ *
+ * \p options The options struct to destroy.
+ */
+RD_EXPORT
+void rd_kafka_describe_consumer_groups_options_destroy(
+    const rd_kafka_describe_consumer_groups_options_t *options);
```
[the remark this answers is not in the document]
emasab: dito

src/rdkafka.h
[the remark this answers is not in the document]
emasab: the Java type is named ConsumerGroupDescription. but the functions are called rd_kafka_ListGroups and rd_kafka_DescribeGroups as there was rd_kafka_DeleteGroups already. Also in Java they are called listConsumerGroups, describeConsumerGroups and deleteConsumerGroups. Not sure how to proceed here.

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+ *
+ * @param member The group member.
+ * @return The client id, or NULL if \p member is NULL.
+ */
+RD_EXPORT
+char *rd_kafka_MemberDescription_client_id(
+    const rd_kafka_MemberDescription_t *member);
+
+/**
+ * @brief Gets consumer id of a \p member.
```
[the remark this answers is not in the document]
emasab: It's the member id, I've used the same naming as Java

src/rdkafka.h
```diff
@@ -6533,6 +6793,28 @@ rd_kafka_AdminOptions_set_broker(rd_kafka_AdminOptions_t *options,
...
+ * @param errstr_size Writable size in \p errstr.
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success or an
+ *          error code on failure in which case an error string will
+ *          be written \p errstr.
+ *
+ * @remark This option is valid for ListConsumerGroupOffsets.
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_AdminOptions_set_require_stable(rd_kafka_AdminOptions_t *options,
```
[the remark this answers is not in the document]
emasab: The Java option is called [requireStable](https://kafka.apache.org/33/javadoc/org/apache/kafka/clients/admin/ListConsumerGroupOffsetsOptions.html#requireStable-boolean-)  but given here options are flat it can be better to disambiguate

src/rdkafka.h
emasab: It's because [KIP-709](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=173084258) is not implemented yet, but this way the API is forward compatible

src/rdkafka.h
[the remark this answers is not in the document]
edenhill: Let's use the Java naming with Consumer in the method and type names.
For DeleteGroups we can add an alias function and deprecate the old one.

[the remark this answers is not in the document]
edenhill: Or just document on DeleteGroups that the Java counterpart is called `DeleteConsumerGroups` so that people will find it when searching, probably enough.

src/rdkafka.h
```diff
@@ -6714,7 +6713,7 @@ rd_kafka_AdminOptions_set_require_stable(rd_kafka_AdminOptions_t *options,
 RD_EXPORT rd_kafka_resp_err_t rd_kafka_AdminOptions_set_consumer_group_states(
     rd_kafka_AdminOptions_t *options,
     rd_kafka_consumer_group_state_t *consumer_group_states,
-    size_t consumer_group_states_cnt,
+    int consumer_group_states_cnt,
```
edenhill, line 6716: size_t is better

src/rdkafka_admin.c
```diff
@@ -5939,11 +5939,11 @@ rd_kafka_MemberDescription_new(const char *client_id,
                 return NULL;
         rd_kafka_MemberDescription_t *member;
```
edenhill, line 5940: All declarations must come before statements.

src/rdkafka_admin.c
```diff
@@ -5939,11 +5939,11 @@ rd_kafka_MemberDescription_new(const char *client_id,
                 return NULL;
```
edenhill, line 5939: Defensive checks like this is typically not done internally in C apps/libs, and it makes things more complex since now MemberDescription_new() can return NULL, which needs to be checked by the caller.

src/rdkafka_admin.c
```diff
@@ -6280,14 +6280,14 @@ rd_kafka_admin_DescribeGroupsRequest(rd_kafka_broker_t *rkb,
                                      void *opaque) {
         int i;
         char *group;
-        size_t groups_cnt       = rd_list_cnt(groups);
+        int groups_cnt          = rd_list_cnt(groups);
         const char **groups_arr = calloc(groups_cnt, sizeof(*groups_arr));
         RD_LIST_FOREACH(group, groups, i) {
```
edenhill, line 6285: Please put some air between logical sections of code;
```c
  variable declarations;

  some code();
  that does things();

  some other code that does other
  things();

  return;
```

tests/0081-admin.c
```diff
@@ -2536,7 +2536,7 @@ static void do_test_ListGroups(const char *what,
         TEST_ASSERT(found == TEST_LIST_GROUPS_CNT,
                     "expected to find %" PRId32
                     " started groups,"
-                    " got %" PRId32,
+                    " got %" PRIu64,
```
edenhill, line 2539: Use PRIusz for size_t

edenhill (changes requested):

src/rdkafka.h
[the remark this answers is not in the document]
emasab: it's the state of a single consumer group

src/rdkafka.h
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+
+/**
+ * @brief Gets the group id for the \p desc group.
+ *
+ * @param desc The group description.
+ * @return The group id, or NULL if \p desc is NULL.
+ */
+RD_EXPORT
+const char *rd_kafka_ConsumerGroupDescription_group_id(
+    const rd_kafka_ConsumerGroupDescription_t *desc);
```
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+
+/**
+ * @brief Gets the coordinator for the \p desc group.
+ *
+ * @param desc The group description.
+ * @return The group coordinator, or NULL if \p desc is NULL.
+ */
+RD_EXPORT
+rd_kafka_Node_t *rd_kafka_ConsumerGroupDescription_coordinator(
```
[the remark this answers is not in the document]
emasab: Was created when parsing the result from the broker. Set to const.

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+/**
+ * @brief Gets a member of \p desc group.
+ *
+ * @param desc The group description.
+ * @param idx The member idx.
+ * @return A member at index \p idx, or NULL if \p desc is NULL
+ *         or \p idx is out of range.
+ */
+RD_EXPORT
+rd_kafka_MemberDescription_t *rd_kafka_ConsumerGroupDescription_member(
```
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+    const rd_kafka_MemberDescription_t *member);
+
+/**
+ * @brief Gets consumer id of a \p member.
+ *
+ * @param member The group member.
+ * @return The consumer id, or NULL if \p member is NULL.
+ */
+RD_EXPORT
+char *rd_kafka_MemberDescription_consumer_id(
```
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+ */
+RD_EXPORT
+const rd_kafka_MemberAssignment_t *rd_kafka_MemberDescription_assignment(
+    const rd_kafka_MemberDescription_t *member);
+
+/**
+ * @brief Gets assigned partitions of a member \p assignment.
+ *
+ * @param assignment The group member assignment.
+ * @return The assigned partitions, or NULL if \p assignment is NULL.
```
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
emasab: reverted rd_kafka_group_member_info to previous state

src/rdkafka.h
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
```diff
@@ -7337,8 +7629,209 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+rd_kafka_DescribeGroups_result_groups(
+    const rd_kafka_DescribeGroups_result_t *result,
+    size_t *cntp);
+
+
+/**
+ * @brief Gets the group id for the \p desc group.
+ *
+ * @param desc The group description.
+ * @return The group id, or NULL if \p desc is NULL.
```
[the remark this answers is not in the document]
emasab: done

src/rdkafka.h
```diff
@@ -6533,6 +6793,28 @@ rd_kafka_AdminOptions_set_broker(rd_kafka_AdminOptions_t *options,
...
+ *               The \p errstr is only written in case of error.
+ * @param errstr_size Writable size in \p errstr.
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success or an
+ *          error code on failure in which case an error string will
+ *          be written \p errstr.
+ *
+ * @remark This option is valid for ListConsumerGroupOffsets.
+ */
+RD_EXPORT rd_kafka_resp_err_t
```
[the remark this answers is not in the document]
emasab: done

src/rdkafka_admin.c
emasab: done

tests/0081-admin.c
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka.h
emasab: done

src/rdkafka_admin.c
emasab: done

src/rdkafka_admin.c
emasab: done

src/rdkafka.h
```diff
@@ -6631,6 +6794,38 @@ rd_kafka_AdminOptions_set_broker(rd_kafka_AdminOptions_t *options,
...
+ * @param options Admin options.
+ * @param true_or_false Defaults to false.
+ *
+ * @return NULL on success, a new error instance that must be
+ *         released with rd_kafka_error_destroy() in case of error.
+ *
+ * @remark This option is valid for ListConsumerGroupOffsets.
+ */
+RD_EXPORT
+rd_kafka_error_t *rd_kafka_AdminOptions_set_require_stable_offsets(
```
milindl, line 6810: More than a review comment, a question - Why are we changing the return type of admin option from rd_kafka_resp_err_t coupled with two out params, to a single rd_kafka_error_t?

Is this the new way going forward?

milindl: Looking at it a bit:
Is this comment the norm for all API including AdminOptions_set* going forward? 
https://github.com/edenhill/librdkafka/pull/3995#discussion_r1029752282

src/rdkafka.h
```diff
@@ -6457,20 +6614,26 @@ rd_kafka_group_result_partitions(const rd_kafka_group_result_t *groupres);
...
-        RD_KAFKA_ADMIN_OP_DELETERECORDS,    /**< DeleteRecords */
-        RD_KAFKA_ADMIN_OP_DELETEGROUPS,     /**< DeleteGroups */
+        RD_KAFKA_ADMIN_OP_ANY = 0,                /**< Default value */
+        RD_KAFKA_ADMIN_OP_CREATETOPICS,           /**< CreateTopics */
+        RD_KAFKA_ADMIN_OP_DELETETOPICS,           /**< DeleteTopics */
+        RD_KAFKA_ADMIN_OP_CREATEPARTITIONS,       /**< CreatePartitions */
+        RD_KAFKA_ADMIN_OP_ALTERCONFIGS,           /**< AlterConfigs */
+        RD_KAFKA_ADMIN_OP_DESCRIBECONFIGS,        /**< DescribeConfigs */
+        RD_KAFKA_ADMIN_OP_DELETERECORDS,          /**< DeleteRecords */
+        RD_KAFKA_ADMIN_OP_LISTCONSUMERGROUPS,     /**< ListConsumerGroups */
```
milindl, line 6624: Another question: We are adding things to the middle of a publicly exposed enum, will this break ABI compatibility? 

Say for example someone is dynamically linking librdkafka, and using 
`rd_kafka_AdminOptions_new(rk, RD_KAFKA_ADMIN_OP_DELETEGROUPS)`

He/she now replaces the .so file of librdkafka without recompiling code, and since now enum value points to RD_KAFKA_ADMIN_OP_LISTCONSUMERGROUPS, won't it cause failure?

src/rdkafka.h
emasab: Yes, we want to return only error objects, so the user doesn't check the error code but relies on flags related to actions like retriable, abortable, fatal and so on.

src/rdkafka.h
emasab: Yes, thanks. I'm moving it at the end of the enum.

emasab: done

src/rdkafka.h
```diff
@@ -7435,8 +7640,377 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+RD_EXPORT
+const char *rd_kafka_ConsumerGroupDescription_partition_assignor(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc);
+
+
+/**
+ * @brief Gets state for the \p grpdesc group.
+ *
+ * @param grpdesc The group description.
+ * @return A group state, or RD_KAFKA_CONSUMER_GROUP_STATE_UNKNOWN if
```
milindl, line 7873: Comment is not consistent with actual behaviour, which is rd_assert(grplist != NULL)

examples/alter_group_offsets.c
```diff
@@ -0,0 +1,336 @@
...
+                                           * AlterConsumerGroupOffsets() */
+        rd_kafka_event_t *event; /* AlterConsumerGroupOffsets result event */
+        const int min_argc = 2;
+        int num_partitions = 0;
+
+        /*
+         * Argument validation
+         */
+        int print_usage = argc < min_argc;
+        print_usage |= (argc - min_argc) % 2 != 0;
```
edenhill, line 180: since print_usage is only used once, why not make this an easier to read if statement with an or?
```c
if (argc < min_argc || (argc - min_argc) % 2 != 0)
    print_usage("...");
```

examples/alter_group_offsets.c
```diff
@@ -0,0 +1,336 @@
...
+print_partition_list(FILE *fp,
+                     const rd_kafka_topic_partition_list_t *partitions,
+                     int print_offset,
+                     const char *prefix) {
+        if (partitions->cnt == 0) {
+                fprintf(fp, "%sNo partition found", prefix);
+        }
+        int i;
+        char offset_string[521] = {};
+        for (i = 0; i < partitions->cnt; i++) {
```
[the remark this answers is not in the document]
emasab: Done

examples/alter_group_offsets.c
```diff
@@ -0,0 +1,336 @@
...
+static void
+print_partition_list(FILE *fp,
+                     const rd_kafka_topic_partition_list_t *partitions,
+                     int print_offset,
+                     const char *prefix) {
+        if (partitions->cnt == 0) {
+                fprintf(fp, "%sNo partition found", prefix);
+        }
+        int i;
+        char offset_string[521] = {};
```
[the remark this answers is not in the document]
emasab: Done

examples/alter_group_offsets.c
emasab: Done

examples/alter_group_offsets.c
```diff
@@ -0,0 +1,336 @@
...
+         */
+        int print_usage = argc < min_argc;
+        print_usage |= (argc - min_argc) % 2 != 0;
+        if (print_usage) {
+                usage("Wrong number of arguments");
+        }
+
+        num_partitions       = (argc - min_argc) / 2;
+        const char *group_id = argv[0];
+        const char *topic    = argv[1];
```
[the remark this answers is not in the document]
emasab: Done

src/rdkafka.h
```diff
@@ -7435,8 +7640,376 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+ * @param grplist The group listing.
+ *
+ * @return A group state.
+ */
+RD_EXPORT
+rd_kafka_consumer_group_state_t rd_kafka_ConsumerGroupListing_state(
+    const rd_kafka_ConsumerGroupListing_t *grplist);
+
+/**
+ * @brief Get an array of valid list groups from a ListConsumerGroups result.
```
edenhill, line 7715: What's a valid group? and what's a non valid group?
I don't see this nomenclature in Java.

If this is rather the non-error-groups, which is the successful path, then just go with `rd_kafka_ListconsumerGorups_result_groups()`.
See existing admin APIs, e.g., `rd_kafka_CreateTopics_result_topics()`.

Also drop the valid word from the description, I don't think it adds anything but rather creates this notion of non-valid groups, which is unclear.

src/rdkafka.h
```diff
@@ -7435,8 +7640,376 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+RD_EXPORT
+const char *rd_kafka_ConsumerGroupDescription_partition_assignor(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc);
+
+
+/**
+ * @brief Gets state for the \p grpdesc group.
+ *
+ * @param grpdesc The group description.
+ * @return A group state, or RD_KAFKA_CONSUMER_GROUP_STATE_UNKNOWN if
```
edenhill, line 7872: add empty line between params and return.

src/rdkafka.h
edenhill: Yes, remove the NULL check and update the return documentation.

src/rdkafka.h
```diff
@@ -7435,8 +7640,376 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+
+/**
+ * @brief Gets the members count of \p grpdesc group.
+ *
+ * @param grpdesc The group description.
+ *
+ * @return The member count.
+ */
+RD_EXPORT
+int rd_kafka_ConsumerGroupDescription_member_count(
```
edenhill, line 7902: size_t

src/rdkafka.h
```diff
@@ -7435,8 +7640,376 @@ RD_EXPORT const rd_kafka_topic_partition_list_t *
...
+ * @return A member at index \p idx, or NULL if
+ *         \p idx is out of range.
+ *
+ * @remark The lifetime of the returned memory is the same
+ *         as the lifetime of the \p grpdesc object.
+ */
+RD_EXPORT
+const rd_kafka_MemberDescription_t *rd_kafka_ConsumerGroupDescription_member(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc,
+    int idx);
```
edenhill, line 7920: size_t , to match member_count()

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ *
+ * @param group_id Consumer group id.
+ * @param partitions Partitions to list committed offsets for.
+ *                   Only the topic and partition fields are used.
+ *
+ * @returns a new allocated ListConsumerGroupOffsets object.
+ *          Use rd_kafka_ListConsumerGroupOffsets_destroy() to free
+ *          object when done.
+ */
+RD_EXPORT const rd_kafka_ListConsumerGroupOffsets_t *
```
edenhill, line 8113: can't return const here, this memory belongs to the application as it is passed to destroy later.

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+rd_kafka_ListConsumerGroupOffsets_new(
+    const char *group_id,
+    const rd_kafka_topic_partition_list_t *partitions);
+
+/**
+ * @brief Destroy and free a ListConsumerGroupOffsets object previously
+ *        created with rd_kafka_ListConsumerGroupOffsets_new()
+ */
+RD_EXPORT void rd_kafka_ListConsumerGroupOffsets_destroy(
+    const rd_kafka_ListConsumerGroupOffsets_t *list_grpoffsets);
```
edenhill, line 8123: can't pass const to a destructor

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+
+/**
+ * @brief Helper function to destroy all ListConsumerGroupOffsets objects in
+ *        the \p list_grpoffsets array (of \p list_grpoffsets_cnt elements).
+ *        The array itself is not freed.
+ */
+RD_EXPORT void rd_kafka_ListConsumerGroupOffsets_destroy_array(
+    const rd_kafka_ListConsumerGroupOffsets_t **list_grpoffsets,
```
edenhill, line 8131: no const

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ *
+ * @remark The result event type emitted on the supplied queue is of type
+ *         \c RD_KAFKA_EVENT_LISTCONSUMERGROUPOFFSETS_RESULT
+ *
+ * @remark The current implementation only supports one group per invocation.
+ */
+RD_EXPORT
+void rd_kafka_ListConsumerGroupOffsets(
+    rd_kafka_t *rk,
+    const rd_kafka_ListConsumerGroupOffsets_t **list_grpoffsets,
```
edenhill, line 8154: It's tricky to call `const foo_t **arr` functions , so I recommend dropping the const here - thjat's what we've done with the other admin APIs.
Not 100% correct, but helps usability alot.

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ *
+ * @param group_id Consumer group id.
+ * @param partitions Partitions to alter committed offsets for.
+ *                   Only the topic and partition fields are used.
+ *
+ * @returns a new allocated AlterConsumerGroupOffsets object.
+ *          Use rd_kafka_AlterConsumerGroupOffsets_destroy() to free
+ *          object when done.
+ */
+RD_EXPORT const rd_kafka_AlterConsumerGroupOffsets_t *
```
edenhill, line 8208: can't have const

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+rd_kafka_AlterConsumerGroupOffsets_new(
+    const char *group_id,
+    const rd_kafka_topic_partition_list_t *partitions);
+
+/**
+ * @brief Destroy and free a AlterConsumerGroupOffsets object previously
+ *        created with rd_kafka_AlterConsumerGroupOffsets_new()
+ */
+RD_EXPORT void rd_kafka_AlterConsumerGroupOffsets_destroy(
+    const rd_kafka_AlterConsumerGroupOffsets_t *alter_grpoffsets);
```
edenhill, line 8218: drop const

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+
+/**
+ * @brief Helper function to destroy all AlterConsumerGroupOffsets objects in
+ *        the \p alter_grpoffsets array (of \p alter_grpoffsets_cnt elements).
+ *        The array itself is not freed.
+ */
+RD_EXPORT void rd_kafka_AlterConsumerGroupOffsets_destroy_array(
+    const rd_kafka_AlterConsumerGroupOffsets_t **alter_grpoffsets,
```
edenhill, line 8226: drop const

src/rdkafka.h
```diff
@@ -7508,9 +8085,202 @@ RD_EXPORT const rd_kafka_group_result_t **rd_kafka_DeleteGroups_result_groups(
...
+ *
+ * @remark The result event type emitted on the supplied queue is of type
+ *         \c RD_KAFKA_EVENT_ALTERCONSUMERGROUPOFFSETS_RESULT
+ *
+ * @remark The current implementation only supports one group per invocation.
+ */
+RD_EXPORT
+void rd_kafka_AlterConsumerGroupOffsets(
+    rd_kafka_t *rk,
+    const rd_kafka_AlterConsumerGroupOffsets_t **alter_grpoffsets,
```
edenhill, line 8250: drop const

src/rdkafka_admin.h
```diff
@@ -343,4 +356,124 @@ struct rd_kafka_DeleteAcls_result_response_s {
...
+ * @{
+ */
+
+/**
+ * @struct ListConsumerGroups result for a single group
+ */
+typedef struct rd_kafka_ConsumerGroupListing_s {
+        char *group_id; /**< Group id */
+        rd_bool_t is_simple_consumer_group;
+        /**< Is it a simple consumer group? That means empty protocol_type. */
```
edenhill, line 414: If the comment can't fit on the same line, put it above the field instead and use `/**` instead of `/**<`

src/rdkafka_admin.h
```diff
@@ -343,4 +356,124 @@ struct rd_kafka_DeleteAcls_result_response_s {
...
+        rd_kafka_consumer_group_state_t state; /**< Consumer group state. */
+} rd_kafka_ConsumerGroupListing_t;
+
+
+/**
+ * @struct ListConsumerGroups results and errors
+ */
+typedef struct rd_kafka_ListConsumerGroupsResult_s {
+        rd_list_t valid;  /**< List of valid ConsumerGroupListing
```
edenhill, line 423: rename to groups

edenhill: drop "valid"

src/rdkafka_admin.h
```diff
@@ -343,4 +356,124 @@ struct rd_kafka_DeleteAcls_result_response_s {
...
+ * @name DescribeConsumerGroups
+ * @{
+ */
+
+/**
+ * @struct Assignment of a consumer group member.
+ *
+ */
+typedef struct rd_kafka_MemberAssignment_s {
+        rd_kafka_topic_partition_list_t *topic_partitions;
```
edenhill, line 440: we typically just call it "partitions" in librdkafka, the topic is implicit.

src/rdkafka_admin.h
```diff
@@ -343,4 +356,124 @@ struct rd_kafka_DeleteAcls_result_response_s {
...
+        /**< Partitions assigned to current member. */
```
edenhill, line 441: but comment on the line above and drop <

src/rdkafka_admin.h
```diff
@@ -343,4 +356,124 @@ struct rd_kafka_DeleteAcls_result_response_s {
...
+
+/**
+ * @name ListConsumerGroups
+ * @{
+ */
+
+/**
+ * @struct ListConsumerGroups result for a single group
+ */
+typedef struct rd_kafka_ConsumerGroupListing_s {
```
edenhill, line 411: There is already a typedef for this in rdkafka.h, just do the struct here.

same for other structs below

src/rdkafka_admin.h
```diff
@@ -343,4 +356,124 @@ struct rd_kafka_DeleteAcls_result_response_s {
...
+        rd_kafka_MemberAssignment_t assignment; /**< Member assignment */
+} rd_kafka_MemberDescription_t;
+
+/**
+ * @struct DescribeConsumerGroups result
+ */
+typedef struct rd_kafka_ConsumerGroupDescription_s {
+        char *group_id; /**< Group id */
+        rd_bool_t is_simple_consumer_group;
+        /**< Is it a simple consumer group? That means empty protocol_type. */
```
edenhill, line 462: put comment above the field, not below

src/rdkafka.h
emasab: valid is [here](https://kafka.apache.org/33/javadoc/org/apache/kafka/clients/admin/ListConsumerGroupsResult.html#valid--)

tests/test.c
```diff
@@ -6090,8 +6164,6 @@ rd_kafka_resp_err_t test_DeleteGroups_simple(rd_kafka_t *rk,
 
         rd_kafka_queue_destroy(q);
 
-        rd_kafka_DeleteGroup_destroy_array(del_groups, group_cnt);
```
edenhill, line 6093: :+1:

edenhill (commented):
Make a grep spin to make sure you are only use rd_malloc, rd_calloc, rd_strdup, rd_realloc, rd_free, and so on, otherwise there will be weird crashes on windows.

src/rdkafka.h
emasab: just rd_list_cnt returns int

src/rdkafka.h
emasab: here too rd_list_elem takes a signed int, isn't there risk of overflow?

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+ * @param opaque Not used.
+ */
+static void *rd_kafka_MemberDescription_list_copy(const void *elem,
+                                                  void *opaque) {
+        return rd_kafka_MemberDescription_copy(elem);
+}
+
+static void
+rd_kafka_MemberDescription_destroy(rd_kafka_MemberDescription_t *member) {
+        if (!member)
```
edenhill, line 6023: no null checks

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        rd_free(member);
+}
+
+static void rd_kafka_MemberDescription_free(void *member) {
+        rd_kafka_MemberDescription_destroy(member);
+}
+
+const char *rd_kafka_MemberDescription_client_id(
+    const rd_kafka_MemberDescription_t *member) {
+        rd_assert(member != NULL);
```
edenhill, line 6042: no null checks

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        return member->client_id;
+}
+
+const char *rd_kafka_MemberDescription_group_instance_id(
+    const rd_kafka_MemberDescription_t *member) {
+        rd_assert(member != NULL);
```
edenhill, line 6048: no null checks

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        return member->group_instance_id;
+}
+
+const char *rd_kafka_MemberDescription_consumer_id(
+    const rd_kafka_MemberDescription_t *member) {
+        rd_assert(member != NULL);
```
edenhill, line 6054: no null checks

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        return member->consumer_id;
+}
+
+const char *
+rd_kafka_MemberDescription_host(const rd_kafka_MemberDescription_t *member) {
+        rd_assert(member != NULL);
```
edenhill, line 6060: no null checks

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        return member->host;
+}
+
+const rd_kafka_MemberAssignment_t *rd_kafka_MemberDescription_assignment(
+    const rd_kafka_MemberDescription_t *member) {
+        rd_assert(member != NULL);
```
edenhill, line 6066: no null checks

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        return &member->assignment;
+}
+
+const rd_kafka_topic_partition_list_t *
+rd_kafka_MemberAssignment_topic_partitions(
+    const rd_kafka_MemberAssignment_t *assignment) {
+        rd_assert(assignment != NULL);
```
edenhill, line 6073: and so on

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        grpdesc->is_simple_consumer_group = is_simple_consumer_group;
+        if (members == NULL) {
+                rd_list_init(&grpdesc->members, 0,
+                             rd_kafka_MemberDescription_free);
+        } else {
+                rd_list_init_copy(&grpdesc->members, members);
+                rd_list_copy_to(&grpdesc->members, members,
+                                rd_kafka_MemberDescription_list_copy, NULL);
+        }
+        grpdesc->partition_assignor = !partition_assignor
```
edenhill, line 6112: This reads a bit weird, be more explicit with your intention in this case, so:
`partition_assignor = partition_assignor ? rd_strdup(partition_assignor) : NULL;`

edenhill: But you don't really need the NULL assign since the struct is calloced, so do an if instead.

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+                                          ? (char *)partition_assignor
+                                          : rd_strdup(partition_assignor);
+        grpdesc->state = state;
+        if (coordinator != NULL)
+                grpdesc->coordinator = rd_kafka_Node_copy(coordinator);
+        grpdesc->error =
```
edenhill, line 6118: if (error)
..

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+ *        rd_list_copy(). The \p opaque is ignored.
+ */
+static void *rd_kafka_ConsumerGroupDescription_copy_opaque(const void *grpdesc,
+                                                           void *opaque) {
+        return rd_kafka_ConsumerGroupDescription_copy(grpdesc);
+}
+
+static void rd_kafka_ConsumerGroupDescription_destroy(
+    rd_kafka_ConsumerGroupDescription_t *grpdesc) {
+        if (!grpdesc)
```
edenhill, line 6166: nulll...

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        rd_free(grpdesc);
+}
+
+static void rd_kafka_ConsumerGroupDescription_free(void *ptr) {
+        rd_kafka_ConsumerGroupDescription_destroy(ptr);
+}
+
+const char *rd_kafka_ConsumerGroupDescription_group_id(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc) {
+        rd_assert(grpdesc != NULL);
```
edenhill, line 6186: null

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        return grpdesc->group_id;
+}
+
+const rd_kafka_error_t *rd_kafka_ConsumerGroupDescription_error(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc) {
+        rd_assert(grpdesc != NULL);
```
edenhill, line 6192: and so on

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+    size_t errstr_size,
+    rd_kafka_replyq_t replyq,
+    rd_kafka_resp_cb_t *resp_cb,
+    void *opaque) {
+        int i;
+        char *group;
+        rd_kafka_resp_err_t err;
+        int groups_cnt          = rd_list_cnt(groups);
+        rd_kafka_error_t *error = NULL;
+        const char **groups_arr = calloc(groups_cnt, sizeof(*groups_arr));
```
edenhill, line 6316: must use rd_*() allocators or it will crash horrendouesly on windows

edenhill: drop const

edenhill (changes requested):
Really think we should drop the group state enum, serves no point and just makes it harder to keep in synch with what happens on the broker side, protocol wise.

edenhill:
Massive PR this, but finally fully reviewed.

This is overall excellent work once again, @emasab !

There are some habits that needs to go, like all the NULL checking :). It is part of the API contract that it will be used correctly, and if the application developer doesn't honour their part there is no point in us trying to do anything about it. If they aren't passing null they can pass garbage, you know, so it is not something we need to protect against, or can.

src/rdkafka_admin.h
emasab: see previous comment with J. naming

src/rdkafka.h
edenhill: huhm, ok, but I really don't like the name, and I don't think it makes any sense in our implementation because in theirs they have all() for a mix of the two.
I think we should go with groups()

but I'll let you decide.

src/rdkafka.h
edenhill: but that's just an internal implementation detail, cast it to size_t

tests/0081-admin.c
```diff
@@ -2233,7 +2233,7 @@ typedef struct expected_group_result {
 static void do_test_DeleteGroups(const char *what,
                                  rd_kafka_t *rk,
                                  rd_kafka_queue_t *useq,
-                                 int op_timeout) {
+                                 int request_timeout) {
```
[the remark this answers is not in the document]
emasab: It's request timeout for DeleteGroups, not operation timeout, that is the broker timeout that is sent as a protocol field and supported only for some operations

src/rdkafka_request.c
```diff
@@ -1730,43 +1838,135 @@ void rd_kafka_HeartbeatRequest(rd_kafka_broker_t *rkb,
...
+ *        for handling by \p resp_cb (with \p opaque passed).
+ *
+ * @return NULL on success, a new error instance that must be
+ *         released with rd_kafka_error_destroy() in case of error.
  */
-void rd_kafka_ListGroupsRequest(rd_kafka_broker_t *rkb,
-                                rd_kafka_replyq_t replyq,
-                                rd_kafka_resp_cb_t *resp_cb,
-                                void *opaque) {
+rd_kafka_error_t *rd_kafka_ListGroupsRequest(rd_kafka_broker_t *rkb,
```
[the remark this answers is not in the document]
emasab: I think it's handled in rd_kafka_list_groups.

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+
+        rd_assert(rd_list_cnt(alter_grpoffsets) == 1);
+
+        rd_kafka_topic_partition_list_t *offsets = grpoffsets->partitions;
+        rd_kafka_consumer_group_metadata_t *cgmetadata =
+            rd_kafka_consumer_group_metadata_new(grpoffsets->group_id);
+
+        int ret = rd_kafka_OffsetCommitRequest_group(
+            rkb, cgmetadata, offsets, replyq, resp_cb, opaque,
+            "rd_kafka_AlterConsumerGroupOffsetsRequest");
```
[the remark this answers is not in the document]
emasab: It's the reason paramer there

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+/**
+ * @name List consumer groups
+ * @{
+ *
+ *
+ *
+ *
+ */
+
+#define CONSUMER_PROTOCOL_TYPE "consumer"
```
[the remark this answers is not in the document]
emasab: No, the SIMPLE_CONSUMER_PROTOCOL_TYPE has an empty string, the normal one is "consumer".

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+                rd_kafka_buf_read_str(reply, &ProtocolType);
+                if (api_version >= 4) {
+                        rd_kafka_buf_read_str(reply, &GroupState);
+                }
+                rd_kafka_buf_skip_tags(reply);
+
+                group_id   = RD_KAFKAP_STR_DUP(&GroupId);
+                proto_type = RD_KAFKAP_STR_DUP(&ProtocolType);
+                if (api_version >= 4) {
+                        group_state = RD_KAFKAP_STR_DUP(&GroupState);
```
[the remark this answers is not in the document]
emasab: J. has an [enum](https://kafka.apache.org/33/javadoc/org/apache/kafka/common/ConsumerGroupState.html), I think it's better to do the same. The reason I think is they make state selection unambiguous for the user when filtering.

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+                        state = rd_kafka_consumer_group_state_code(group_state);
+                }
+
+                is_simple_consumer_group = *proto_type == '\0';
+                is_consumer_protocol_type =
+                    !strcmp(proto_type, CONSUMER_PROTOCOL_TYPE);
+                if (is_simple_consumer_group || is_consumer_protocol_type) {
+                        group_listing = rd_kafka_ConsumerGroupListing_new(
+                            group_id, is_simple_consumer_group, state);
+                        rd_list_add(&valid, group_listing);
```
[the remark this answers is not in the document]
emasab: I think it's better to follow J. naming for least astonishment, see comment above.

src/rdkafka_admin.c
```diff
@@ -4801,3 +5055,1608 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+                error = rd_kafka_error_new(
+                    error_code,
+                    "Broker [%" PRId16
+                    "] "
+                    "ListConsumerGroups response protocol parse failure: %s",
+                    nodeid, rd_kafka_err2str(error_code));
+                rd_list_add(&errors, error);
+        }
+
+        list_result = rd_kafka_ListConsumerGroupsResult_new(&valid, &errors);
```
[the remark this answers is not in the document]
emasab: rd_kafka_ListConsumerGroupsResult_new is used also by copy, I think it would create some problems.

src/rdkafka_broker.c
```diff
@@ -5774,7 +5833,61 @@ void rd_kafka_broker_monitor_del(rd_kafka_broker_monitor_t *rkbmon) {
...
+ * @param id The node id.
+ * @param host The node host.
+ * @param port The node port.
+ * @param rack_id (optional) The node rack id.
+ * @return A new allocated Node object.
+ *         Use rd_kafka_Node_destroy() to free when done.
+ */
+rd_kafka_Node_t *
+rd_kafka_Node_new(int id, const char *host, int port, const char *rack_id) {
+        rd_kafka_Node_t *ret = calloc(1, sizeof(*ret));
```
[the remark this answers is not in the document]
emasab: You mean with rd_malloc? I prefer using rd_calloc for the automatic zero initialization.

src/rdkafka.h
```diff
@@ -6631,6 +6794,38 @@ rd_kafka_AdminOptions_set_broker(rd_kafka_AdminOptions_t *options,
...
+ *
+ * @param options Admin options.
+ * @param consumer_group_states Array of consumer group states.
+ * @param consumer_group_states_cnt Size of the \p consumer_group_states array.
+ *
+ * @return NULL on success, a new error instance that must be
+ *         released with rd_kafka_error_destroy() in case of error.
+ */
+RD_EXPORT
+rd_kafka_error_t *rd_kafka_AdminOptions_set_consumer_group_states(
```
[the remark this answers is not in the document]
emasab: I change it as it's common to all the operations and there could be conflicts.

src/rdkafka_admin.c
```diff
@@ -4801,3 +5058,1552 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        rd_kafka_MemberDescription_t *member;
+        member              = rd_calloc(1, sizeof(*member));
+        member->client_id   = rd_strdup(client_id);
+        member->consumer_id = rd_strdup(consumer_id);
+        if (group_instance_id)
+                member->group_instance_id = rd_strdup(group_instance_id);
+        member->host       = rd_strdup(host);
+        member->assignment = assignment;
+        member->assignment.partitions =
+            rd_kafka_topic_partition_list_copy(member->assignment.partitions);
```
milindl, line 5955: Two things regarding this line:

1. We're currently copying member->assignment.partitions into itself. Rather, it should be:
```
member->assignment.partitions =
            rd_kafka_topic_partition_list_copy(assignment.partitions);
```

2. We should have a NULL check somewhere on assignment.partitions, since `rd_kafka_topic_partition_list_copy` assumes the argument passed to it will be not NULL. But assignment.partitions can indeed be NULL. 
How we will handle the NULL value is something I can't confirm, we can either do member->assignment.partitions = NULL or else member->assignment.partitions = empty partition list

(2) sometimes leads to a segfault, for which I'm attaching a partial bt.

```
#0  rd_kafka_topic_partition_list_copy (src=src@entry=0x0) at rdkafka_partition.c:2755
#1  0x00007ffff7e77b3f in rd_kafka_MemberDescription_new (
    client_id=0x7fffa4005db0 "test.client.1", 
    consumer_id=0x7fffa4004b80 "test.client.1-641dfe50-6d1f-4b84-8ff2-ad84094b86dd", 
    group_instance_id=0x0, host=0x7fffa40051c0 "/172.19.0.1", assignment=...)
    at rdkafka_admin.c:5956
#2  0x00007ffff7e549a5 in rd_list_copy_to (dst=dst@entry=0x7fffa40051f0, 
    src=src@entry=0x7fffc15d66a0, 
    copy_cb=copy_cb@entry=0x7ffff7e77b90 <rd_kafka_MemberDescription_list_copy>, 
    opaque=opaque@entry=0x0) at rdlist.c:457
#3  0x00007ffff7e77d9b in rd_kafka_ConsumerGroupDescription_new (
    group_id=group_id@entry=0x7fffa8001e30 "testgroup-108073486942453323", 
    is_simple_consumer_group=is_simple_consumer_group@entry=0 '\000', 
    members=members@entry=0x7fffc15d66a0, 
    partition_assignor=partition_assignor@entry=0x7fffa40035f0 "range", 
    state=RD_KAFKA_CONSUMER_GROUP_STATE_STABLE, coordinator=coordinator@entry=0x7fffa4005190, 
    error=0x0) at rdkafka_admin.c:6072
#4  0x00007ffff7e83a77 in rd_kafka_DescribeConsumerGroupsResponse_parse (rko_req=<optimized out>, 
    rko_resultp=0x7fffc15d6960, reply=0x7fffac000f00, errstr=0x7fffc15d6970 "h\177]\301\377\177", 
    errstr_size=512) at rdkafka_admin.c:6446
```

edenhill: Good find, Milind :+1:
I have a fix on the way

edenhill: https://github.com/confluentinc/librdkafka/pull/3995/commits/ec016519d4623cfee566cb951e2fa1f8062929d8

src/rdkafka_request.c
```diff
@@ -945,60 +945,69 @@ void rd_kafka_op_handle_OffsetFetch(rd_kafka_t *rk,
...
 
+        if (parts) {
+                parts_size = parts->cnt * 32;
+        }
+
         rkbuf = rd_kafka_buf_new_flexver_request(
             rkb, RD_KAFKAP_OffsetFetch, 1,
-            RD_KAFKAP_STR_SIZE(rkb->rkb_rk->rk_group_id) + 4 +
-                (parts->cnt * 32) + 1,
+            32 /* GroupId */ + 4 + parts_size /* Topics */ + 1,
```
milindl, line 985: Question: why aren't we using `RD_KAFKAP_STR_SIZE` anymore for GroupId?

tests/0080-admin_ut.c
```diff
@@ -1557,6 +2134,7 @@ static void do_test_options(rd_kafka_t *rk) {
                     RD_KAFKA_ADMIN_OP_CREATEPARTITIONS,                        \
                     RD_KAFKA_ADMIN_OP_ALTERCONFIGS,                            \
                     RD_KAFKA_ADMIN_OP_DESCRIBECONFIGS,                         \
+                    RD_KAFKA_ADMIN_OP_DESCRIBECONSUMERGROUPS,                  \
```
milindl, line 2137: Question: here, why do we avoid adding listconsumergroups, and alter/listconsumergroupoffsets? They seem to allow request timeout, opaque, and broker options too?

src/rdkafka_admin.c
```diff
@@ -4801,3 +5058,1551 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        rd_kafka_topic_partition_list_t *partitions = NULL;
+        rd_kafka_resp_err_t err = RD_KAFKA_RESP_ERR_NO_ERROR;
+        const rd_kafka_AlterConsumerGroupOffsets_t *alter_grpoffsets =
+            rd_list_elem(&rko_req->rko_u.admin_request.args, 0);
+        partitions =
+            rd_kafka_topic_partition_list_copy(alter_grpoffsets->partitions);
+
+        rk  = rko_req->rko_rk;
+        rkb = reply->rkbuf_rkb;
+        err = rd_kafka_handle_OffsetCommit(rk, rkb, err, reply, NULL,
```
milindl, line 5176: We are keeping the `request` parameter of this function NULL.
But it's possible that, in case we encounter `RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART`, we will set the action to take as `RD_KAFKA_ERR_ACTION_RETRY`, which will eventually call `rd_kafka_buf_retry(rkb, request)` which will segfault.


Writing this segfault here, I'm not sure how to fix it yet, I'll update the PR with the fix commit when I find one (maybe using ignore_cgrp in some way is the best fix)

milindl: It's tricky because we can't ignore things like `RD_KAFKA_RESP_ERR_COORDINATOR_LOAD_IN_PROGRESS` in which case we _actually_ want to retry.

edenhill: I'm not sure why it is reusing handle_OffsetCommit() in the first place? That's for another request type,  this should have its own response handler.

src/rdkafka_admin.c
```diff
@@ -4801,3 +5058,1551 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+        rko_result = rd_kafka_admin_result_new(rko_req);
+        rd_list_init(&rko_result->rko_u.admin_result.results, 1,
+                     rd_kafka_group_result_free);
+        rd_list_add(&rko_result->rko_u.admin_result.results,
+                    rd_kafka_group_result_new(alter_grpoffsets->group_id, -1,
+                                              partitions, NULL));
+        rd_kafka_topic_partition_list_destroy(partitions);
+        *rko_resultp = rko_result;
+
+        rd_snprintf(
```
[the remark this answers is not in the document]
edenhill: And the value from `err` is lost here.

src/rdkafka_admin.c
milindl: Ref the KIP 396:
> Regarding the implementation, instead of relying on KafkaConsumer to perform the alter, I propose to instead directly send OffsetCommit requests. This is principally for 2 reasons [...]

We're doing the same thing and inside `rd_kafka_AlterConsumerGroupOffsetsResponse_parse` we call handle_OffsetCommit

src/rdkafka_request.c
```diff
@@ -1153,6 +1153,7 @@ rd_kafka_handle_OffsetCommit(rd_kafka_t *rk,
                 rd_kafka_buf_read_throttle_time(rkbuf);
 
         rd_kafka_buf_read_i32(rkbuf, &TopicArrayCnt);
+
```
edenhill, line 1156: Avoid whitespace diffs

src/rdkafka_admin.c
```diff
@@ -5186,10 +5186,10 @@ rd_kafka_AlterConsumerGroupOffsetsResponse_parse(rd_kafka_op_t *rko_req,
         rd_kafka_topic_partition_list_destroy(partitions);
         *rko_resultp = rko_result;
 
-        rd_snprintf(
-            errstr, errstr_size,
-            "AlterConsumerGroupOffset response protocol parse failure: %s",
-            rd_kafka_err2str(reply->rkbuf_err));
+        rd_snprintf(errstr, errstr_size,
```
edenhill, line 5189: This seems to be a reindent, not sure clang-format will be happy about.

Also, change it so it is conditional and uses the correct error.
```c
if (!err && rkbuf->rkbuf_err)
   err = rkbuf->rkbuf_err;

if (err)
   rd_snprintf(errstr, .....);

return err;
```

edenhill (changes requested):

src/rdkafka_admin.c
milindl: Did this partly - for now , as :
```
if (rkbuf->rkbuf_err)
   rd_snprintf(errstr, .....);
```

Because `err` should actually be conditionally ignored (handle_OfffsetCommit raises errors for those cases where all partitions have partitions level errors which shouldn't be a global level error for alter offsets)

src/rdkafka_admin.c
```diff
@@ -4801,3 +5058,1551 @@ void rd_kafka_DeleteAcls(rd_kafka_t *rk,
...
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc,
+    size_t idx) {
+        return (rd_kafka_MemberDescription_t *)rd_list_elem(&grpdesc->members,
+                                                            idx);
+}
+
+/**
+ * @brief Group arguments comparator for DescribeConsumerGroups args
+ */
+static int rd_kafka_DescribeConsumerGroups_cmp(const void *a, const void *b) {
```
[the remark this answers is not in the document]
milindl: I don't think we need this, the other de-dups are for the input.

tests/0138-admin_mock.c
```diff
@@ -0,0 +1,191 @@
...
+
+        q = rd_kafka_queue_get_main(rk);
+
+        if (req_timeout_ms > 0) {
+                /* Admin options */
+                options = rd_kafka_AdminOptions_new(
+                    rk, RD_KAFKA_ADMIN_OP_ALTERCONSUMERGROUPOFFSETS);
+                err = rd_kafka_AdminOptions_set_request_timeout(
+                    options, req_timeout_ms, errstr, sizeof(errstr));
+                TEST_ASSERT(!err, "%s", rd_kafka_err2str(err));
```
milindl, line 92: nit: this is better done with TEST_CALL__ or one of its related functions

milindl (approved):
LGTM. Thanks a lot @emasab for this PR.

tests/0137-barrier_batch_consume.c
```diff
@@ -178,106 +178,108 @@ static void do_test_consume_batch_with_seek(void) {
...
-        rd_kafka_topic_partition_list_t *pause_partition_list;
-        rd_kafka_message_t **rkmessages;
-        size_t msg_cnt;
-        const int timeout_ms            = 10000;
-        const int consume_msg_cnt       = 10;
-        const int produce_msg_cnt       = 8;
-        const int produce_partition_cnt = 2;
-        const int expected_msg_cnt      = 4;
-        int32_t pause_partition         = 0;
+// static void do_test_consume_batch_with_pause_and_resume(void) {
```
emasab, line 181: This avoids a warning that can be treated as error

emasab:
Thanks again to @lesterfan for starting this implementation

PR #4003, opened 2022-10-05
Fix cURL linking

stertingen:
Similar to https://github.com/edenhill/librdkafka/pull/3909, but somewhat more consistent with the surrounding CMake code linking to zlib or zstd.


edenhill:
Thank you!

PR #4004, opened 2022-10-05
Public API headers at <librdkafka/*.h>

stertingen:
Addressing https://github.com/edenhill/librdkafka/issues/3994. Currently building onto https://github.com/edenhill/librdkafka/pull/4003 to fix build issues.

As discussed in above issue, new public API headers are added to `include/librdkafka/` which include the headers in the source directories. The public API headers stay in the source directories to maintain compatibility with projects which bundle their own version ob librdkafka.

Examples and tests have their include directives adjusted.

I'm planning on adding a deprecation note if `<rdkafka.h>` is included directly.


include/librdkafka/rdkafka.h
```diff
@@ -0,0 +1,34 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef RDKAFKA_RDKAFKA_H
+#define RDKAFKA_RDKAFKA_H
+
+#include <rdkafka.h>
```
Quuxplusone, line 32: IIUC, the intent of this line is `#include "../../src/rdkafka.h"`; shouldn't it be spelled that way? Spelling it with angle brackets makes it seem like it would prefer `/usr/local/include/rdkafka.h`, which I think is not the intended connotation.

From first principles, I would have expected the _compatibility_ header in `./src/rdkafka.h` to be the one-liner `#include <librdkafka/rdkafka.h>`, and then `./include/librdkafka/rdkafka.h` would contain the actual code moving forward. But the commit messages/discussion sound like you considered that option and already rejected it? (I don't quite understand why, though.)

FWIW, for the non-compatibility headers that are actually going to be installed in `/usr/local/include` (which I would _hope_ means everything from `./include/` and nothing from `./src/`, even though that's not what this PR currently does), I strongly agree with the direction of this PR to use `#include <librdkafka/...>` consistently throughout. I just think the compatibility headers (which I would _hope_ means only things in `./src/`) should continue to use `#include "..."`.

stertingen: Actually, `#include "../../src/rdkafka.h"` did not come across my mind.

The motivation of having the compatibility header at `include/librdkafka/` was to maintain compatibility with projects which add `src/` and `src-cpp/` to their include dirs, but not `include/` with the actual header files. Adding `include/` to the include dirs would have been necessary, as the compatibility headers would have had `#include <librdkafka/rdkafka.h>` as their content.

I'm trying to rebuild this PR with `#include "../include/librdkafka/rdkafka.h"` later today.

Quuxplusone: I see you've flipped the headers around per my suggestion (compatibility headers in `src/`, "real" headers in `include/librdkafka/`). Just for the record, yes, I now think you're doing it right. I don't _think_ I'm going to come back in a week and say "no no, I meant to suggest something radically different from that." :)  (And I hope this direction will be acceptable to @edenhill and also to whatever's going on with the VS project include directories. ;))

stertingen:
Due to the file movements, this PR now adds and deletes 12k lines. I tried to convince git to recognize the file movements without any changes by splitting out the creation of compatibility headers into a new commit. (This makes rebasing the new commits onto anything a lot easier.)

After rebasing, every commit after '[Fix references to public API header locations in docs and scripts](https://github.com/edenhill/librdkafka/pull/4004/commits/c7e1cd05770a617b6d097e7bacdef3b125c4c13d)' should produce a working build, except for Win32 and warnings.

### On the deprecation warning

I'm not 100% satisfied with the deprecation warning.

`#warning` is a GNU extension supported by many compilers excluding MSVC. It will be standard with C23 and C++23. (https://wg21.link/p2437), however, there is no official way for checking C23 or C++23 yet (like `__cplusplus`).

Similarly, `#pragma message` also is a nonstandard compiler extension supported by many compilers, including MSVC, but still not all. This will not be standardized AFAIK. However, most compilers just ignore unknown pragmas.

Some compilers support both.

Currently, the warning is only displayed for GCC, clang and MSVC. We could agree on a set of compilers and get the deprecation warning working for these. But how do we proceed on unofficially supported compilers?
1. Do not display any warning for these
2. Use `#warning` because it will be standard
3. Use `#pragma message` because most compilers ignore it if unsupported
4. Generate the deprecation warning by a completely different way instead of the preprocessor (Runtime warning? Yikes.).

edenhill:
I think it is important to keep the complexity and size of this PR down to a minimum since it only solves a very small and niche problem.

That's why I recommend keeping all source files as-is, except for the examples, and then simply symlinking include/rdkafka/rdkafka.h (et.al) to their in-source-tree locations, not the other way around.

stertingen:
Closing for now in favor of https://github.com/edenhill/librdkafka/pull/4050.

PR #4006, opened 2022-10-07
use s1-prod-macos-arm64 semaphore build agent

emasab:
This updates the Semaphore pipeline to use the s1-prod-macos-arm64 build agent. This new agent was created in a more scalable and maintainable way than the old one.


edenhill (commented):
LGTM

PR #4011, opened 2022-10-11
Incorrect complete transaction and pid bump order

emasab:
Changed the order of the abort and PID bump operations,  starting the drain and bump after completing the transaction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,21 @@ librdkafka v1.9.3 is a maintenance release:
 
  * Self-contained static libraries can now be built on Linux arm64 (#4005).
  * Fix for using PKCS#12 keystores on Windows.
+ * Fixes to the transactional and idempotent producer.
 
 
 ## Fixes
 
 ### General fixes
 
- * Windows: couldn't read a PKCS#12 keystore correctly because binary mode wasn't explicitly set and Windows defaults to text mode.
+ * Windows: couldn't read a PKCS#12 keystore correctly because binary
+   mode wasn't explicitly set and Windows defaults to text mode (#3999).
+
+
+### Transactional producer fixes
+
+ * Changed the order of the abort and PID bump operations,
+   starting the drain and bump after completing the transaction (#4011).
 
 
 
--- a/src/rdkafka_mock.c
+++ b/src/rdkafka_mock.c
@@ -1876,6 +1876,33 @@ rd_kafka_mock_broker_push_request_error_rtts(rd_kafka_mock_cluster_t *mcluster,
         return RD_KAFKA_RESP_ERR_NO_ERROR;
 }
 
+rd_kafka_resp_err_t
+rd_kafka_mock_broker_error_stack_cnt(rd_kafka_mock_cluster_t *mcluster,
+                                     int32_t broker_id,
+                                     int16_t ApiKey,
+                                     size_t *cnt) {
+        rd_kafka_mock_broker_t *mrkb;
+        rd_kafka_mock_error_stack_t *errstack;
+
+        if (!mcluster || !cnt)
+                return RD_KAFKA_RESP_ERR__INVALID_ARG;
+
+        mtx_lock(&mcluster->lock);
+
+        if (!(mrkb = rd_kafka_mock_broker_find(mcluster, broker_id))) {
+                mtx_unlock(&mcluster->lock);
+                return RD_KAFKA_RESP_ERR__UNKNOWN_BROKER;
+        }
+
+        errstack = rd_kafka_mock_error_stack_get(&mrkb->errstacks, ApiKey);
+
+        *cnt = errstack->cnt;
+
+        mtx_unlock(&mcluster->lock);
+
+        return RD_KAFKA_RESP_ERR_NO_ERROR;
+}
+
 
 void rd_kafka_mock_topic_set_error(rd_kafka_mock_cluster_t *mcluster,
                                    const char *topic,
--- a/src/rdkafka_mock.h
+++ b/src/rdkafka_mock.h
@@ -191,6 +191,26 @@ rd_kafka_mock_broker_push_request_error_rtts(rd_kafka_mock_cluster_t *mcluster,
 
 
 /**
+ * @brief Get the count of errors in
+ *        the broker's error stack for the given \p ApiKey.
+
+ * \p mcluster the mock cluster.
+ * \p broker_id id of the broker in the cluster.
+ * \p ApiKey is the Kafka protocol request type, e.g., ProduceRequest (0).
+ * \p cnt pointer for receiving the count.
+ *
+ * @returns \c RD_KAFKA_RESP_ERR_NO_ERROR if the count was retrieved,
+ * \c RD_KAFKA_RESP_ERR__UNKNOWN_BROKER if there was not broker with this id,
+ * \c RD_KAFKA_RESP_ERR__INVALID_ARG if some of the parameters are not valid.
+ */
+rd_kafka_resp_err_t
+rd_kafka_mock_broker_error_stack_cnt(rd_kafka_mock_cluster_t *mcluster,
+                                     int32_t broker_id,
+                                     int16_t ApiKey,
+                                     size_t *cnt);
+
+
+/**
  * @brief Set the topic error to return in protocol requests.
  *
  * Currently only used for TopicMetadataRequest and AddPartitionsToTxnRequest.
--- a/src/rdkafka_txnmgr.c
+++ b/src/rdkafka_txnmgr.c
@@ -463,14 +463,15 @@ static void rd_kafka_txn_curr_api_reply(rd_kafka_q_t *rkq,
  */
 void rd_kafka_txn_idemp_state_change(rd_kafka_t *rk,
                                      rd_kafka_idemp_state_t idemp_state) {
-        rd_bool_t reply_assigned = rd_false;
+        rd_bool_t reply         = rd_false;
+        rd_kafka_error_t *error = NULL;
 
         if (idemp_state == RD_KAFKA_IDEMP_STATE_ASSIGNED &&
             rk->rk_eos.txn_state == RD_KAFKA_TXN_STATE_WAIT_PID) {
                 /* Application is calling (or has called) init_transactions() */
                 RD_UT_COVERAGE(1);
                 rd_kafka_txn_set_state(rk, RD_KAFKA_TXN_STATE_READY_NOT_ACKED);
-                reply_assigned = rd_true;
+                reply = rd_true;
 
         } else if (idemp_state == RD_KAFKA_IDEMP_STATE_ASSIGNED &&
                    rk->rk_eos.txn_state ==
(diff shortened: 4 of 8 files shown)


src/rdkafka_mock.c
```diff
@@ -1876,6 +1876,33 @@ rd_kafka_mock_broker_push_request_error_rtts(rd_kafka_mock_cluster_t *mcluster,
         return RD_KAFKA_RESP_ERR_NO_ERROR;
 }
 
+rd_kafka_resp_err_t
```
edenhill, line 1879: RD_EXPORT is needed for public symbols to be exported on Windows

src/rdkafka_mock.c
```diff
@@ -1876,6 +1876,33 @@ rd_kafka_mock_broker_push_request_error_rtts(rd_kafka_mock_cluster_t *mcluster,
...
+                return RD_KAFKA_RESP_ERR__INVALID_ARG;
+
+        mtx_lock(&mcluster->lock);
+
+        if (!(mrkb = rd_kafka_mock_broker_find(mcluster, broker_id))) {
+                mtx_unlock(&mcluster->lock);
+                return RD_KAFKA_RESP_ERR__UNKNOWN_BROKER;
+        }
+
+        errstack = rd_kafka_mock_error_stack_get(&mrkb->errstacks, ApiKey);
```
edenhill, line 1897: _get() creates the error stack if it doesn't exist, use _find() instead.

src/rdkafka_mock.h
```diff
@@ -190,6 +190,26 @@ rd_kafka_mock_broker_push_request_error_rtts(rd_kafka_mock_cluster_t *mcluster,
...
+ * @brief Get the count of errors in
+ *        the broker's error stack for the given \p ApiKey.
+
+ * \p mcluster the mock cluster.
+ * \p broker_id id of the broker in the cluster.
+ * \p ApiKey is the Kafka protocol request type, e.g., ProduceRequest (0).
+ * \p cnt pointer for receiving the count.
+ *
+ * @returns \c RD_KAFKA_RESP_ERR_NO_ERROR if the count was retrieved,
+ * \c RD_KAFKA_RESP_ERR__UNKNOWN_BROKER if there was not broker with this id,
```
edenhill, line 203: no broker

edenhill (changes requested):

tests/0105-transactions_mock.c
```diff
@@ -564,7 +569,120 @@ static void do_test_txn_fenced_reinit(rd_kafka_resp_err_t error_code) {
...
+        rd_kafka_error_destroy(error);
+
+        fatal_err = rd_kafka_fatal_error(rk, errstr, sizeof(errstr));
+        TEST_ASSERT(fatal_err, "Expected a fatal error to have been raised");
+        TEST_SAY("Fatal error: %s: %s\n", rd_kafka_err2name(fatal_err), errstr);
+
+        if (rd_kafka_mock_broker_error_stack_cnt(
+                mcluster, txn_coord, RD_KAFKAP_EndTxn, &errors_cnt)) {
+                TEST_FAIL(
+                    "Broker error count should succeed for API %s"
```
edenhill, line 666: This sentence is a bit hard to decipher, also, it should print the error (for troubleshooting).
E.g., "Failed to retrieve broker %d error stack count for %s: %s"

Can also use the `TEST_CALL_ERR__(rd_kafka_mock_broker_error_stack_cnt(...))` helper, it will fail the test if the call returns an error.

tests/0105-transactions_mock.c
```diff
@@ -564,7 +569,120 @@ static void do_test_txn_fenced_reinit(rd_kafka_resp_err_t error_code) {
...
+ * @brief KIP-360: Test that fatal idempotence errors triggers abortable
+ *        transaction errors, but let the broker-side abort of the
+ *        transaction fail with a fencing error.
+ *        Should raise a fatal error.
+ *
+ * @param error_code Which error code EndTxn should fail with.
+ *                   Either RD_KAFKA_RESP_ERR_INVALID_PRODUCER_EPOCH (older)
+ *                   or RD_KAFKA_RESP_ERR_PRODUCER_FENCED (newer).
+ */
+static void do_test_txn_fenced_abort(rd_kafka_resp_err_t error_code) {
```
edenhill, line 594: Good test :+1:

Does the unfixed txnmgr implementation fail this test?

edenhill (changes requested):

CHANGELOG.md
```diff
@@ -4,13 +4,21 @@ librdkafka v1.9.3 is a maintenance release:
...
 ### General fixes
 
- * Windows: couldn't read a PKCS#12 keystore correctly because binary mode wasn't explicitly set and Windows defaults to text mode.
+ * Windows: couldn't read a PKCS#12 keystore correctly because binary
+   mode wasn't explicitly set and Windows defaults to text mode (#3999).
+
+
+### Transactional producer fixes
+
+ * Changed the order of the abort and PID bump operations,
```
edenhill, line 20: It is good to also briefly explain how the problem manifested itself so that people can tell if this is a problem they were hit by.
Typically something like (just an example):
 "The transactional producer was previously unable to recover from fatal errors of the underlying idempotent producer, such as `FENCED_PRODUCER`. This js now fixed by changing the order of the abort and PID bump operations...... (#4011)".

edenhill (changes requested):

edenhill (commented):
Try to structure the commit messages like so:
```
One-line explanation of what was done (#issuenumber) - preferably under 80 characters.

Addition details and
motiviation
if necessary
```

Otherwise frequently used tools like gitk and git log --oneline output is less usable

src/rdkafka_txnmgr.c
```diff
@@ -2027,21 +2034,37 @@ rd_kafka_error_t *rd_kafka_send_offsets_to_transaction(
...
+            rk->rk_eos.idemp_state != RD_KAFKA_IDEMP_STATE_ASSIGNED;
+
         rk->rk_eos.txn_requires_epoch_bump = rd_false;
         rk->rk_eos.txn_req_cnt             = 0;
 
-        rd_kafka_txn_set_state(rk, RD_KAFKA_TXN_STATE_READY);
+        if (drain_bump) {
+                rd_kafka_wrunlock(rk);
+                rd_kafka_idemp_drain_epoch_bump_start(
+                    rk, "txn_requires_epoch_bump");
```
edenhill, line 2061: This should be a human/user-readable reason, e.g:
`is_commit ? "Transaction successfully committed" : "Transaction successfully aborted"`

src/rdkafka_txnmgr.c
```diff
@@ -2027,21 +2034,37 @@ rd_kafka_error_t *rd_kafka_send_offsets_to_transaction(
...
 
+        rd_bool_t drain_bump =
+            rk->rk_eos.txn_requires_epoch_bump ||
+            rk->rk_eos.idemp_state != RD_KAFKA_IDEMP_STATE_ASSIGNED;
+
         rk->rk_eos.txn_requires_epoch_bump = rd_false;
         rk->rk_eos.txn_req_cnt             = 0;
 
-        rd_kafka_txn_set_state(rk, RD_KAFKA_TXN_STATE_READY);
+        if (drain_bump) {
```
edenhill, line 2058: This function, txn_complete(), should be called when the entire transaction is done, including bumps, so this drain_bump logic should not be here but in an earlier stage.
Recommend adding a `.._try_complete()` that contains this logic.

src/rdkafka_txnmgr.c
```diff
@@ -2399,17 +2424,35 @@ rd_kafka_txn_op_commit_transaction_ack(rd_kafka_t *rk,
         if ((error = rd_kafka_txn_require_state(
                  rk, RD_KAFKA_TXN_STATE_COMMIT_NOT_ACKED)))
                 goto done;
+        if (rk->rk_eos.txn_init_rkq != NULL) {
```
edenhill, line 2427: add empty line between logical blocks.

src/rdkafka_txnmgr.c
```diff
@@ -2390,6 +2413,8 @@ rd_kafka_txn_op_commit_transaction_ack(rd_kafka_t *rk,
                                        rd_kafka_q_t *rkq,
                                        rd_kafka_op_t *rko) {
         rd_kafka_error_t *error;
+        rd_bool_t complete    = rd_true;
```
edenhill, line 2416: ..transaction_ack() should be called when everything is done, the draining must be performed in an earlier stage before we transition not .._NOT_ACKED.

src/rdkafka_txnmgr.c
```diff
@@ -2588,49 +2631,6 @@ static rd_kafka_op_res_t rd_kafka_txn_op_abort_transaction(rd_kafka_t *rk,
                 goto done;
         }
 
-        if (rk->rk_eos.txn_requires_epoch_bump ||
-            rk->rk_eos.idemp_state != RD_KAFKA_IDEMP_STATE_ASSIGNED) {
-                /* If the underlying idempotent producer's state indicates it
```
edenhill, line 2593: Please do not remove comments as they're needed to explain the why of the code, and we're not really changing the behaviour so this comment is still mostly correct.

src/rdkafka_txnmgr.h
```diff
@@ -167,5 +167,4 @@ rd_bool_t rd_kafka_txn_coord_set(rd_kafka_t *rk,
 
 void rd_kafka_txns_term(rd_kafka_t *rk);
 void rd_kafka_txns_init(rd_kafka_t *rk);
-
```
edenhill, line 170: avoid whitespace diffs

edenhill (changes requested):
This changes the behaviour and semantics by reusing functions (such as txn_complete()) for new functionality it was not intended for.
commits and aborts are asynchronous and they perform a variety of different operations, but what is important is that all operations are finished before transitioning to .._NOT_ACKED, and that all that happens in the NOT_ACKED state is transition to READY when the application "acks".

tests/0105-transactions_mock.c
emasab: yes, it fails with
`### Test "0105_transactions_mock (do_test_txn_fenced_abort:600: With error INVALID_PRODUCER_EPOCH)" failed at 0105-transactions_mock.c:667:do_test_txn_fenced_abort() at Thu Oct 13 12:04:14 2022: ###
Test assertion failed: "errors_cnt == 0": Expected error count 0 for API EndTxn, found 1`

tests/0105-transactions_mock.c
```diff
@@ -564,7 +569,120 @@ static void do_test_txn_fenced_reinit(rd_kafka_resp_err_t error_code) {
...
+        if (rd_kafka_mock_broker_error_stack_cnt(
+                mcluster, txn_coord, RD_KAFKAP_EndTxn, &errors_cnt)) {
+                TEST_FAIL(
+                    "Broker error count should succeed for API %s"
+                    " on broker %" PRId32,
+                    rd_kafka_ApiKey2str(RD_KAFKAP_EndTxn), txn_coord);
+        }
+        /* Checks all the  RD_KAFKAP_EndTxn responses have been consumed */
+        TEST_ASSERT(errors_cnt == 0,
+                    "Expected error count 0 for API %s, found %zu",
```
edenhill, line 672: Should probably say ".. error stack count .." since it otherwise implies that it is counting (occurred) errors, which is the opposite of what it does :)

emasab:
Putting this on hold until a better motivation for the change is found. That is if we find some cases where an abort is sent after the re-init, causing a fenced producer. Currently either the abort without re-init or the re-init only is done. The latter causes the abort too.

emasab:
Putting this on hold until a better motivation for the change is found. That is if we find some cases where an abort is sent after the re-init, causing a fenced producer. Currently either the abort without re-init or the re-init only is done. The latter causes the abort too.

edenhill:
Superceeded by #4054

PR #4012, opened 2022-10-11
Copy public headers to include/librdkafka/*.h

stertingen:
As an alternative to https://github.com/edenhill/librdkafka/pull/4004.

This approach copies the public headers at build time when using CMake.
This behavior is not implemented for Make as I am not familiar with mklove.
When using `add_subdirectory` in CMake, the new include directory is added along the old one automatically, so linking to the `rdkafka` or `rdkafka++` target should work.

Currently, no deprecation warning is emitted.
It could be implemented by applying `configure_file` on each header and setting a flag when deploying it to `include/librdkafka/`.

### DIFF
--- a/src-cpp/CMakeLists.txt
+++ b/src-cpp/CMakeLists.txt
@@ -30,8 +30,18 @@ endif()
 
 target_link_libraries(rdkafka++ PUBLIC rdkafka)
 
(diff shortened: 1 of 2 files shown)


stertingen:
Closing for now in favor of https://github.com/edenhill/librdkafka/pull/4050.

PR #4016, opened 2022-10-13
Fix for invalid state transition when commit times out and abort is called

emasab:
While a commit operation was in queue, a timeout happens that can cause an abort. The state changes from COMMITTING_TRANSACTION to ABORTABLE_ERROR to ABORTING_TRANSACTION. When the broker comes up the error, that before was retriable, now is permanent or fatal because the state has changed from the initial one

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ librdkafka v1.9.3 is a maintenance release:
  * Fix for using PKCS#12 keystores on Windows.
  * OpenSSL 3.0.x support - the maximum bundled OpenSSL version is now 3.0.5 (previously 1.1.1q).
  * Updated to zlib 1.2.13 and zstd 1.5.2 in self-contained librdkafka bundles.
+ * Fixes to the transactional and idempotent producer.
 
 
 ## Upgrade considerations
@@ -47,6 +48,15 @@ configuration property.
 ## Enhancements
 
  * Bundled zlib upgraded to version 1.2.13.
+
+### Transactional producer fixes
+
+ * When a commit operation is in queue, a timeout can happen,
+ causing an abortable error. The subsequent abort can cause
+ the an assert to fail for an invalid state transition.
+ This fix allows to recover from that acknowledging the
+ previously succeeded commit operation. (#4016).
+
 
 
 # librdkafka v1.9.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@ configuration property.
 
  * When a commit operation is in queue, a timeout can happen,
  causing an abortable error. The subsequent abort can cause
- the an assert to fail for an invalid state transition.
+ an assert to fail for an invalid state transition.
  This fix allows to recover from that acknowledging the
  previously succeeded commit operation. (#4016).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,9 @@ configuration property.
  * When a commit operation is in queue, a timeout can happen,
  causing an abortable error. The subsequent abort can cause
  an assert to fail for an invalid state transition.
- This fix allows to recover from that acknowledging the
- previously succeeded commit operation. (#4016).
+ This fix allows to recover from that, retrying the
+ commit or abort operation until a response by the broker
+ is received (#4016).
 
 
 
--- a/src/rdkafka_txnmgr.c
+++ b/src/rdkafka_txnmgr.c
@@ -1192,16 +1192,15 @@ static rd_kafka_error_t *rd_kafka_txn_curr_api_req(rd_kafka_t *rk,
 
         rd_assert(for_reuse == reuse);
 
-        rd_snprintf(rk->rk_eos.txn_curr_api.name,
-                    sizeof(rk->rk_eos.txn_curr_api.name), "%s", name);
         /* When API changes, destroy previous pending queue. */
         if (rk->rk_eos.txn_init_rkq &&
-            strcmp(rk->rk_eos.txn_last_api.name, name)) {
+            strcmp(rk->rk_eos.txn_curr_api.name, name)) {
                 rd_kafka_q_destroy_owner(rk->rk_eos.txn_init_rkq);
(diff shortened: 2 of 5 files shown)


tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,67 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
 }
 
 
+/**
+ * @brief Test that transaction state different from the initial one,
+ * because of a local timed out, doesn't cause an illegal state transition.
+ */
+static void do_test_txn_state_different_from_the_initial_one() {
```
edenhill, line 2873:
```suggestion
static void do_test_txn_state_different_from_the_initial_one(void) {
```

Empty arg-list is an undefined arg-list in C, as opposed to C++ where it is indeed empty.

CHANGELOG.md
```diff
@@ -13,6 +14,15 @@ librdkafka v1.9.3 is a maintenance release:
  * Windows: couldn't read a PKCS#12 keystore correctly because binary mode wasn't explicitly set and Windows defaults to text mode.
 
 
+### Transactional producer fixes
+
+ * When a commit operation is in queue, a timeout can happen,
+ causing an abortable error. The subsequent abort can cause
+ the an assert to fail for an invalid state transition.
```
edenhill, line 21: -the

tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,67 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+
+        rd_kafka_mock_broker_push_request_error_rtts(
+            mcluster, coord_id, RD_KAFKAP_EndTxn, 1, RD_KAFKA_RESP_ERR_NO_ERROR,
+            70);
+
+        error_commit = rd_kafka_commit_transaction(rk, 50);
+        TEST_SAY_ERROR(error_commit, "commit_transaction() returned: ");
+        TEST_ASSERT(rd_kafka_error_txn_requires_abort(error_commit),
+                    "Expected an abortable error");
+        TEST_ASSERT(rd_kafka_error_code(error_commit) ==
```
edenhill, line 2910: If I understand correctly what will happen in this case is that the EndTxn(commit) will succeed on the broker, but commit_transaction() will fail locally due to timeout - at which point the application does an abort_transaction().

And we're assuming the abort_transaction() will succeed, which gives the application the idea that the transaction was indeed aborted, but in fact it was successfully committed on the broker - this yields an inconsistent non-repairable view of the last transaction between the cluster and application.

I'm not sure what the proper behaviour is here, but I think it should be that the abort_transaction() fails if there's an in-flight commit. Question is, should it be a retriable, permanent, or fatal error?

What does the Java producer do in this case?

edenhill (changes requested):

tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,67 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+
+        TEST_SAY("Starting transaction\n");
+        TEST_CALL_ERROR__(rd_kafka_init_transactions(rk, 5000));
+        TEST_CALL_ERROR__(rd_kafka_begin_transaction(rk));
+
+        test_produce_msgs2_nowait(rk, topic, 0, RD_KAFKA_PARTITION_UA, 0,
+                                  msgcnt / 2, NULL, 0, &remains);
+
+        rd_kafka_mock_broker_push_request_error_rtts(
+            mcluster, coord_id, RD_KAFKAP_EndTxn, 1, RD_KAFKA_RESP_ERR_NO_ERROR,
```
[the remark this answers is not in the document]
emasab: changed to 500ms, with 1000 it wasn't giving the same error but a timeout on rd_kafka_abort_transaction

tests/0105-transactions_mock.c
emasab: Yes it should give an error, not sure if fatal or permanent. I'm checking

emasab: In Java it's throwing TimeoutException [here](https://github.com/apache/kafka/blob/0507597597e1d2d9adaa33b87e5ea509e12fd2f0/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionalRequestResult.java#L58), that is retriable. So the commit should be retried in that case and not transition not an abortable error.

emasab: I've pushed a commit to make commits retriable too

src/rdkafka_txnmgr.c
```diff
@@ -1192,8 +1194,13 @@ static rd_kafka_error_t *rd_kafka_txn_curr_api_req(rd_kafka_t *rk,
 
         rd_snprintf(rk->rk_eos.txn_curr_api.name,
                     sizeof(rk->rk_eos.txn_curr_api.name), "%s", name);
-
-        tmpq = rd_kafka_q_new(rk);
+        /* When API changes, destroy previous pending queue. */
```
edenhill, line 1197: This is a nice solution, but do we really want to support API changes in the midst of an operation? It seems complex to recover cleanly from.

Say you have commit_transaction(timeout=300ms) and the API call times out while awaiting response from the broker, but the operation continues in the background.
Now the application calls abort_transaction(): what do we do here?
If we receive a succesful commit response then we should probably fail abort_transaction() - since the abort is no longer possible. But how do we make it meaningful for the application to recover from that?

I think it would be better to simply not allow other API calls if one has been started.

What do you think?

I have some work in progress for this that we can look at together.

emasab: The intention of this change was to keep having one transactional call at a time, just to make some retriable calls like commit and abort able to retry when timed out without sending a new EndTxn. Could add a test case for that situation, but first let's talk about your work.

emasab: Doing a quick check, if abort_transaction is called on the same thread after an initial retriable timeout on commit, it give a STATE error with message "Conflicting commit_transaction call already in progress". It's worth checking what happens from a different thread.

edenhill:
Superceeded by #4054

PR #4020, opened 2022-10-17
A NULL coordinator shouldn't be fatal

emasab:
When a PID epoch bump is requested and the producer is waiting
to reconnect to the transaction coordinator, a failure in a find coordinator
request could cause txn_curr_coord to be set to NULL and an assert to fail.
This was fixed by retrying when the coordinator is known.


tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,96 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+        rd_kafka_error_t *error;
+        rd_kafka_topic_t *rkt;
+        rd_kafka_resp_err_t err;
+        int32_t coord_id             = 1;
+        const char *topic            = "test";
+        const char *transactional_id = "txnid";
+        int msgcnt                   = 1;
+        int remains                  = 0;
+
+        SUB_TEST_QUICK("Test coordinator NULL shouldn't cause "
```
edenhill, line 2887: nit: this is fine, but the sub-test messages are optional and typically only used for showing variations/parameters to a test. There's enough information added by SUB_TEST*() itself to identify the test being run.

tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,96 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+        TEST_SAY("commit_transaction() failed (expectedly): %s\n",
+                 rd_kafka_error_string(error));
+
+        /* Needs to wait some time before closing to make sure it doesn't go
+         * into TERMINATING state before error is triggered */
+        rd_usleep(1400 * 1000, NULL);
+
+        rd_kafka_topic_destroy(rkt);
+        rd_kafka_error_destroy(error);
+        TEST_SAY("destroying rk");
```
edenhill, line 2949: missing \n

tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,96 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+
+        TEST_SAY("commit_transaction() failed (expectedly): %s\n",
+                 rd_kafka_error_string(error));
+
+        /* Needs to wait some time before closing to make sure it doesn't go
+         * into TERMINATING state before error is triggered */
+        rd_usleep(1400 * 1000, NULL);
+
+        rd_kafka_topic_destroy(rkt);
+        rd_kafka_error_destroy(error);
```
edenhill, line 2948: it is better to destroy temporary error objects after their last use, so on line 2942.

tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,96 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+
+        SUB_TEST_QUICK("Test coordinator NULL shouldn't cause "
+            "a fatal assertion "
+            "when transactional producer reconnects");
+
+        test_curr->is_fatal_cb = error_is_fatal_cb;
+        test_curr->exp_dr_err  = RD_KAFKA_RESP_ERR__MSG_TIMED_OUT;
+        allowed_error          = RD_KAFKA_RESP_ERR__TRANSPORT;
+
+        rk = create_txn_producer(&mcluster, transactional_id, 1, "debug", "all", NULL);
```
edenhill, line 2895: If you put the message.timeout.ms here and use test_produce_msgs2_nowait() below you can get rid of the rkt.

tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,96 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+        rd_kafka_mock_broker_push_request_error_rtts(
+            mcluster, coord_id, RD_KAFKAP_FindCoordinator, 1,
+                RD_KAFKA_RESP_ERR_COORDINATOR_NOT_AVAILABLE, 10);
+
+        /* This value is linked to message.timeout.ms, needs enough time
+         * so message time out and aren't purged */
+        rd_kafka_flush(rk, 500);
+        /* Needs a specific timeout so FindCoordinatorRequest is the first thing that
+         * is done when the coordinator comes up */
+        rd_usleep(605 * 1000, NULL);
```
edenhill, line 2930: Please explain in the comment where this exact time is coming from, e.g., something? + message.timeout.ms + some headroom.

tests/0105-transactions_mock.c
```diff
@@ -2866,6 +2866,96 @@ static void do_test_disconnected_group_coord(rd_bool_t switch_coord) {
...
+        error = rd_kafka_commit_transaction(rk, -1);
+
+        TEST_ASSERT(remains == 0, "%d message(s) were not produced\n", remains);
+        TEST_ASSERT(error != NULL, "Expected commit_transaction() to fail");
+
+        TEST_SAY("commit_transaction() failed (expectedly): %s\n",
+                 rd_kafka_error_string(error));
+
+        /* Needs to wait some time before closing to make sure it doesn't go
+         * into TERMINATING state before error is triggered */
```
edenhill, line 2944: same thing here, why 1400

edenhill (changes requested):
Great stuff, some minor changes needed

tests/0105-transactions_mock.c
emasab: yes, and I remove the debug all that I've left

tests/0105-transactions_mock.c
emasab: I remove this log as it was mainly for debugging

edenhill (approved):
LgTM!

PR #4025, opened 2022-10-18
Fix some comment typos. NFC.

Quuxplusone:
This is not remotely the result of an exhaustive proofread; just fixed a couple in `rd_kafka_poll` because I was reading it, and then a few more instances that I saw while reviewing _that_ diff.


edenhill (approved):
Superb! Thank you

PR #4030, opened 2022-10-24
Handle COORDINATOR_LOAD_IN_PROGRESS in err_action

mhowlett:
It has been reported that `rd_kafka_consumer_poll` is returning an `COORDINATOR_LOAD_IN_PROGRESS` error when a cluster is rolled, evidently originating from two different protocol requests, as two different error strings are reported: `Broker: Coordinator load in progress` and `JoinGroup failed: Broker: Coordinator load in progress`.

At leas in the JoinGroup case, the java client backs off and retries: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java#L639

This seems like an appropriate action for any protocol request to me, and propose adding this to `rd_kafka_err_action`.


edenhill:
Looks good but it would be good with a test-case (that fails prior to the fix), easiest approach is to use the mock API, you can use this test as a boiler plate (and you can add it to the same test file):
https://github.com/edenhill/librdkafka/blob/master/tests/0117-mock_errors.c#L181

your version will be simpler though since it only needs the single error push and a consumer poll that should not fail.

Also add an entry to the Consumer fixes section of the CHANGELOG.md

milindl:
I've added the test and the changelog changes, but I don't have permissions to push to the right branch. I've created a PR on matt's repository, and as soon as that's approved my changes should show here. https://github.com/mhowlett/librdkafka/pull/4/

src/rdkafka_coord.c
```diff
@@ -403,9 +403,6 @@ static void rd_kafka_coord_req_handle_FindCoordinator(rd_kafka_t *rk,
             RD_KAFKA_ERR_ACTION_RETRY,
             RD_KAFKA_RESP_ERR_COORDINATOR_NOT_AVAILABLE,
 
-            RD_KAFKA_ERR_ACTION_RETRY,
```
milindl, line 406: Should we remove the same action,error tuple from `rd_kafka_handle_OffsetCommit_error` inside src/rdkafka_request.c? Given that it is redundant now too.

edenhill: We can keep it, for clarity.

edenhill:
Recommend closing this one and creating a new PR based on your own cloned repo's branch so you can update the PR yourself.

milindl:
I created another PR based on my repo at #4048 . Unfortunately, I don't have the permissions to close this one.

PR #4031, opened 2022-10-24
Transaction offset commit shouldn't retry without waiting

emasab:
When calling transaction offset commit, if the coordinator was found but
 not available, retries were made without waits, causing too much logging
 and network calls. An interval has been added between retries.


edenhill:
I think this solution might be overly complicated and I believe a simpler solution is just to make creq's delayable, i.e., they can be instantiated with a delay before executing.

See this PR for an alternate take:
https://github.com/edenhill/librdkafka/pull/4078

edenhill:
Fixed in #4078

PR #4033, opened 2022-10-24
[KIP-86] Implements rd_kafka_sasl_set_credentials

josvisser66:
This PR adds a new API call rd_kafka_sasl_set_credentials which can be used to set or reset the credentials used for SASL PLAIN or SCRAM in a Kafka client. This is useful for environments that want to change their credentials (e.g. because the old credentials are going to expire) without destroying the Kafka client. The new credentials will be used the next time the Kafka client needs to make a new connection to a broker. Setting these credentials using the new API call will not break any connections that were set up using the old credentials.

Note: Although this is not the same as KIP-86, it allows for the same sort of behavior as KIP-86.


src/rdkafka.c
```diff
@@ -346,6 +346,29 @@ rd_kafka_resp_err_t rd_kafka_oauthbearer_set_token_failure(rd_kafka_t *rk,
 #endif
 }
 
+rd_kafka_error_t *rd_kafka_set_sasl_creds(rd_kafka_t *rk, const char *username, const char *password) {
+        char errstr[256];
+        rd_kafka_conf_t *conf = &rk->rk_conf; /* Not using rd_kafka_conf because of const qualifier */
+
+        rwlock_wrlock(&conf->sasl.lock);
```
edenhill, line 353: No need for an rwlock, this is not in the fast-path. Use a regular mtx_t (mtx_plain).

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```
edenhill, line 1: Please remove this file.

Also make sure to fixup the style by running `make style-fix` in the top-level directory prior to committing.

src/rdkafka.c
```diff
@@ -346,6 +346,29 @@ rd_kafka_resp_err_t rd_kafka_oauthbearer_set_token_failure(rd_kafka_t *rk,
 #endif
 }
 
+rd_kafka_error_t *rd_kafka_set_sasl_creds(rd_kafka_t *rk, const char *username, const char *password) {
+        char errstr[256];
+        rd_kafka_conf_t *conf = &rk->rk_conf; /* Not using rd_kafka_conf because of const qualifier */
```
edenhill, line 351: I'm not sure I understand this comment, rk_conf is non-const.

src/rdkafka.c
```diff
@@ -346,6 +346,29 @@ rd_kafka_resp_err_t rd_kafka_oauthbearer_set_token_failure(rd_kafka_t *rk,
...
+        if (RD_KAFKA_CONF_OK != rd_kafka_conf_set(conf, "sasl.password", password, errstr, sizeof(errstr))) {
+                goto error;
+        }
+
+        rwlock_wrunlock(&conf->sasl.lock);
+        return NULL;
+
+error:
+        rwlock_wrunlock(&conf->sasl.lock);
+        return rd_kafka_error_new(RD_KAFKA_RESP_ERR__INVALID_ARG, "rd_conf_set failed: %s", errstr);
```
edenhill, line 368:
```suggestion
        return rd_kafka_error_new(RD_KAFKA_RESP_ERR__INVALID_ARG, "Failed to update SASL credentials: %s", errstr);
```

src/rdkafka.c
```diff
@@ -346,6 +346,29 @@ rd_kafka_resp_err_t rd_kafka_oauthbearer_set_token_failure(rd_kafka_t *rk,
 #endif
 }
 
+rd_kafka_error_t *rd_kafka_set_sasl_creds(rd_kafka_t *rk, const char *username, const char *password) {
```
edenhill, line 349: I'd move this to rdkafka_sasl.c and rename it to `rd_kafka_sasl_set_credentials()`

src/rdkafka.c
```diff
@@ -2204,6 +2227,7 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
         rd_kafka_interceptors_on_new(rk, &rk->rk_conf);
 
         rwlock_init(&rk->rk_lock);
+        rwlock_init(&rk->rk_conf.sasl.lock);
```
edenhill, line 2230: use mtx, also add mtx_destroy() to destroy_final()

src/rdkafka_sasl_plain.c
```diff
@@ -70,10 +70,14 @@ int rd_kafka_sasl_plain_client_new(rd_kafka_transport_t *rktrans,
                                    size_t errstr_size) {
         rd_kafka_broker_t *rkb = rktrans->rktrans_rkb;
         rd_kafka_t *rk         = rkb->rkb_rk;
+        rd_kafka_conf_t *conf  = &rk->rk_conf; /* Not using rd_kafka_conf because of const modifier */
```
edenhill, line 73: Other places in this function already use rk->rk_conf directly, so no need for this alias.

edenhill (changes requested):
Great stuff!

src/rdkafka.c
josvisser66: const rd_kafka_conf_t *rd_kafka_conf(rd_kafka_t *rk) returns a const* to an rd_kafka_conf_t. The calls to lock/unlock break that const modifier.

src/rdkafka.c
josvisser66: Done

src/rdkafka.c
josvisser66: Done

src/rdkafka_sasl_plain.c
josvisser66: Done

src/rdkafka.c
edenhill: No need for rd_kafka_conf(), that's part of the public API. Inside librdkafka you should access rk->rk_conf directly.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# librdkafka v1.9.4
+
+librdkafka v1.9.4 is a maintenance release:
+
+* Added rd_kafka_sasl_set_credentials API function to set or reset the SASL PLAIN or SCRAM credentials used for new broker connections.
```
edenhill, line 5: This should go in the v1.9.3 section below - that's the next release.
Add it to the Enhancement chapter
Try to stick below 80 columns

josvisser66: Fixed

edenhill (approved):

edenhill:
Thanks Jos!

PR #4040, opened 2022-10-28
(Re)factor fetcher out to its own file

edenhill:
..and factored out per-partition fetch response handling into its own function.

No code changes other than what was required by the refactoring.


emasab (approved):
LGTM

PR #4048, opened 2022-11-04
Handle COORDINATOR_LOAD_IN_PROGRESS in err_action

milindl:
See #4030 -- 

> It has been reported that rd_kafka_consumer_poll is returning an COORDINATOR_LOAD_IN_PROGRESS error when a cluster is rolled, evidently originating from two different protocol requests, as two different error strings are reported: Broker: Coordinator load in progress and JoinGroup failed: Broker: Coordinator load in progress.
> 
> At leas in the JoinGroup case, the java client backs off and retries: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java#L639
> 
> This seems like an appropriate action for any protocol request to me, and propose adding this to rd_kafka_err_action

I don't have access to push to the relevant branch to update the same PR, so I've created this new PR which builds on top of it. This addresses the test addition comment, and the changelog comment, both.


tests/0117-mock_errors.c
```diff
@@ -255,6 +255,57 @@ static void do_test_offset_commit_request_timed_out(rd_bool_t auto_commit) {
         SUB_TEST_PASS();
 }
 
+/**
+ * @brief PR #4030. A COORDINATOR_LOAD_IN_PROGRESS error is returned while a
```
edenhill, line 259: It is fine referencing an issue for background information, but we typically don't reference PRs, so just remove the 'PR #4030' part.

edenhill: It is not clear what "returned" refers to here, so flesh it out to say "returned from consumer_poll()" or similar.

edenhill (changes requested):
Good stuff! Just comments on comments :)

tests/0117-mock_errors.c
milindl: Done

tests/0117-mock_errors.c
```diff
@@ -255,6 +255,58 @@ static void do_test_offset_commit_request_timed_out(rd_bool_t auto_commit) {
         SUB_TEST_PASS();
 }
 
+/**
+ * @brief A COORDINATOR_LOAD_IN_PROGRESS error is returned from
```
edenhill, line 259: Sorry but this might make sense now, but when reading this in the future it will seem as if we're expecting consumer_poll() to return COORDINATOR_LOAD_IN_PROGRESS. So maybe rephrase as:
"Verify that a cluster roll does not cause consumer_poll() to return the temporary and retriable COORDINATOR_LOAD_IN_PROGRESS error. We should backoff and retry in that case."

edenhill (changes requested):

tests/0117-mock_errors.c
milindl: Makes sense, especially since there's nothing to refer to besides this comment. I changed it to be like what you have recommended.

edenhill:
Fix this build error:
https://app.travis-ci.com/github/edenhill/librdkafka/jobs/587553400#L708

edenhill (approved):
LGTM!

PR #4049, opened 2022-11-04
Read all certificates from ssl.ca.pem, not just the first one

edenhill:


emasab (approved):
Great improvement! LGTM

PR #4050, opened 2022-11-04
Add symlinks to public API headers at include/librdkafka

stertingen:
Also addressing https://github.com/edenhill/librdkafka/issues/3994, as an alternative to https://github.com/edenhill/librdkafka/pull/4004 and https://github.com/edenhill/librdkafka/pull/4012, based on feedback in https://github.com/edenhill/librdkafka/pull/4004#issuecomment-1303431978.

This PR adds symlinks to the public API headers at `include/librdkafka/` to support `#include <librdkafka/rdkafka.h>` when using the library without installing it.


tests/0000-unittests.c
```diff
@@ -27,7 +27,7 @@
  */
 
 #include "test.h"
-#include "rdkafka.h"
+#include <librdkafka/rdkafka.h>
```
edenhill, line 30: Don't see a reason for having any of the tests to change since they'll never be built outside the source tree.

examples/Makefile
```diff
@@ -10,8 +10,8 @@ all: $(EXAMPLES)
 
 include ../mklove/Makefile.base
 
-CFLAGS += -I../src
-CXXFLAGS += -I../src-cpp
+CFLAGS += -I../include -I../src
```
edenhill, line 13: Why keep ../src?

tests/0000-unittests.c
stertingen: Yeah, your're right. I'm gonna remove this.

examples/Makefile
stertingen: Some examples include internal header files: https://github.com/edenhill/librdkafka/blob/master/examples/rdkafka_performance.c

stertingen:
Symlinks look broken on Windows, at least they do not work out-of-the-box. Trying with redirecting includes next.

stertingen:
@edenhill Would you take another look at this? Thanks!

I switched away from symlinks as they did not work very well on Windows. They require special treatment with Git, so they do not work without the user doing some things. (See https://stackoverflow.com/questions/5917249/git-symbolic-links-in-windows)
Introducing extra header files which include the actual files was the simpler and more robust approach.

PR #4051, opened 2022-11-06
Allow overriding DNS resolution

benesch:
Introduce a new `resolve_cb` configuration parameter for controlling DNS resolution. When provided, librdkafka invokes the callback when resolving the address of a Kafka broker instead of calling `getaddrinfo` directly.  The callback is expected to function exactly as `getaddrinfo`, except that it is additional provided with the `opaque` value.

The callback will enable tunneling to a Kafka broker (e.g., via SSH or AWS PrivateLink). The broker addresses returned by the bootstrap protocol will not be directly routable from the client; the address resolution callback allows the end user to rewrite the broker addresses as appropriate for the tunnel.


edenhill:
Rename test to 0136. 0135 is in #4033

edenhill (changes requested):
Remove the two checked in examples

src/rdkafka.h
```diff
@@ -2244,6 +2245,27 @@ void rd_kafka_conf_set_open_cb(
...
+ * @brief Set address resolution callback.
+ *
+ * The callback is responsible for resolving the hostname \p node and the
+ * service \p service into a list of socket addresses as \p getaddrinfo would.
+ * The \p hints and \p res parameters function as they do for \p getaddrinfo.
+ * The callback's \p opaque argument is the opaque set with
+ * rd_kafka_conf_set_opaque().
+ *
+ * The callback's return value is interpreted as the return value of
+ * \p getaddrinfo. It must be safe to invoke \c freeaddrinfo on \c res.
```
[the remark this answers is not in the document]
benesch: Sounds good to me. Updated as you suggest.

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,123 @@
...
+        const char *topic = test_mk_topic_name(__FUNCTION__, 1);
+
+        rd_kafka_conf_t *conf;
+        test_conf_init(&conf, NULL, 0);
+        rd_kafka_conf_set_dr_msg_cb(conf, test_dr_msg_cb);
+        rd_kafka_conf_set_resolve_cb(conf, resolve_cb);
+
+        const char *brokers =
+            rd_strdup(test_conf_get(conf, "metadata.broker.list"));
+        TEST_SAY("Discovered broker list: %s\n", brokers);
```
edenhill, line 86: Add `TEST_ASSERT(brokers != NULL, "No bootstraps configured");`

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,123 @@
...
+
+        {
+                char *n;
+                char *port = "9092";
+
+                if ((n = strchr(brokers, ',')))
+                        *n = '\0';
+
+                if ((n = strrchr(brokers, ':')) &&
+                    ((n == strchr(brokers, ':')) || *(n - 1) == ']')) {
```
edenhill, line 96: Need to check that n > brokers here or n-1 will be unsafe

edenhill (changes requested):
Add an entry to CHANGELOG.md 's Enhancements list of 1.9.3

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: I copied this out of the broker parsing inside of librdkakfa itself and I believe it's safe because we only compute `*(n - 1)` if `strrchr(brokers, ':')` and `strchr(brokers, ':')` are different, which implies that the string has at least two colons, which implies that `n - 1` is a valid index in the string.

benesch:
> Add an entry to CHANGELOG.md 's Enhancements list of 1.9.3

Done.

tests/0136-resolve_cb.c
edenhill: Ah, good! :)

src/rdkafka.h
```diff
@@ -47,6 +47,7 @@
 #include <stdio.h>
 #include <inttypes.h>
 #include <sys/types.h>
+#include <netdb.h>
```
edenhill, line 50: This isn't portable, and we don't actually need anything from that file.
Instead make a forward declaration of the `struct addrinfo;` just above the doc string.

benesch: Gotcha, done!

edenhill:
Tests fail to build due to netdb.h.

I'm thinking that instead of doing all the #ifdef hell to get netdb or winsock, what really want to know is that librdkafka attempts to use the resolve_cb and that it uses the results, and then calls the new free alternative, so how about this:?

 1. Create a LOCAL test (see other LOCAL tests in test.c), it means it won't need a cluster to operate.
 2. Let this test register a resolve_cb and a connect_cb.
 3. Configure a client with bootstrap.servers=TESTING_RESOLVE_CB:1234.
 4. In resolve_cb: verify the node matches TESTING_RESOLVE_CB, have it return one or more addrinfos with IP address 127.1.2.x (or whatever).
 5. In the connect_cb: verify that the sockaddr_in matches 127.1.2.x. Return -1 so we avoid the actual connection, it is not important.
 6. in resolve_cb, also verify that "free" is called (the null variant of resolve_cb()).
 7. Win!

benesch:
That sounds good to me, but it'll take me at least another day or two to make that change. Is there any time pressure for getting this into v1.9.3? (In the meantime I added the ifdef header block, since that was an easy fix, and maybe it'll be good enough for CI.)

benesch:
I've updated the test as you suggest.

benesch:
Mercifully, the tests are all passing. 🙌🏽

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,167 @@
...
+
+static int resolve_cb(const char *node,
+                      const char *service,
+                      const struct addrinfo *hints,
+                      struct addrinfo **res,
+                      void *opaque) {
+
+        int32_t cnt;
+
+        TEST_SAY("resolve_cb invoked: node=%s service=%s\n", node, service);
```
edenhill, line 69: For TEST_*() macros to work correctly from non-main threads you will need to set `test_curr = this_test` in the callback, recommend a global static variable, like:
```c
static struct test *this_test;

static int resolve_cb(..) {
     int32_t cnt;

    test_curr = this_test;
```

and set it at before you initialize the client below in main

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,167 @@
...
+static int connect_cb(int s,
+                      const struct sockaddr *addr,
+                      int addrlen,
+                      const char *id,
+                      void *opaque) {
+        /* Stage 3: assert address is expected bogus. */
+
+        int32_t cnt;
+        struct sockaddr_in *addr_in;
+
```
edenhill, line 121: set test_curr here too

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,167 @@
...
+                struct sockaddr_in *addr;
+
+                TEST_ASSERT(strcmp(node, "TESTING_RESOLVE_CB") == 0,
+                            "unexpected node: %s\n", node);
+                TEST_ASSERT(strcmp(service, "1234") == 0,
+                            "unexpected service: %s\n", service);
+
+                addr                  = calloc(1, sizeof(struct sockaddr_in));
+                addr->sin_family      = AF_INET;
+                addr->sin_port        = 4321;
```
edenhill, line 85:
```suggestion
                addr->sin_port        = htons(4321);
```

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,167 @@
...
+                addr->sin_addr.s_addr = 5432;
```
edenhill, line 86:
```suggestion
                addr->sin_addr.s_addr = htonl(0x7f010203) /* 127.1.2.3 */;
```

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,167 @@
...
+        cnt = rd_atomic32_get(&stage);
+        TEST_ASSERT(cnt == 2, "connect_cb invoked in unexpected stage: %d\n",
+                    cnt);
+
+        TEST_ASSERT(addr->sa_family == AF_INET,
+                    "address has unexpected type: %d\n", addr->sa_family);
+
+        addr_in = (struct sockaddr_in *)(void *)addr;
+
+        TEST_ASSERT(addr_in->sin_port == 4321,
```
edenhill, line 133:
```suggestion
        TEST_ASSERT(addr_in->sin_port == htons(4321),
```

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,167 @@
...
+                    "address has unexpected port: %d\n", addr_in->sin_port);
+        TEST_ASSERT(addr_in->sin_addr.s_addr == 5432,
```
edenhill, line 135:
```suggestion
        TEST_ASSERT(addr_in->sin_addr.s_addr == htonl(0x7f010203),
```

edenhill (changes requested):

src/rdkafka.h
```diff
@@ -2244,6 +2244,34 @@ void rd_kafka_conf_set_open_cb(
     int (*open_cb)(const char *pathname, int flags, mode_t mode, void *opaque));
 #endif
 
+struct addrinfo;
```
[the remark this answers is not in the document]
benesch: Done.

src/rdkafka.h
```diff
@@ -2244,6 +2244,34 @@ void rd_kafka_conf_set_open_cb(
...
+
+/**
+ * @brief Set address resolution callback.
+ *
+ * The callback is responsible for resolving the hostname \p node and the
+ * service \p service into a list of socket addresses as \p getaddrinfo would.
```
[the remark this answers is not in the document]
benesch: Done.

src/rdkafka.h
```diff
@@ -2244,6 +2244,34 @@ void rd_kafka_conf_set_open_cb(
...
+ * The \p hints and \p res parameters function as they do for \p getaddrinfo.
```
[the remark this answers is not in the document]
benesch: Done.

src/rdkafka.h
```diff
@@ -2244,6 +2244,34 @@ void rd_kafka_conf_set_open_cb(
...
+ * The callback's \p opaque argument is the opaque set with
+ * rd_kafka_conf_set_opaque().
+ *
+ * If the callback is invoked with a NULL \p node, \p service, and \p hints,
+ * the callback should instead free the addrinfo struct specified in \p res.
+ * In this case the callback must succeed; the return value will not be checked
+ * by the caller.
+ *
+ * The callback's return value is interpreted as the return value of
+ * \p getaddrinfo.
```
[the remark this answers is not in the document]
benesch: Done.

src/rdkafka_conf.c
```diff
@@ -683,6 +683,8 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
           rd_kafka_open_cb_generic
 #endif
     },
+    {_RK_GLOBAL, "resolve_cb", _RK_C_PTR, _RK(resolve_cb),
+     "Address resolution callback"},
```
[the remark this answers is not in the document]
benesch: Done.

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,167 @@
...
+ * to the connect callback. If the resolution callback is not invoked, or if the
+ * connect callback is not invoked with the output of the resolution callback,
+ * the test will fail.
+ */
+
+/**
+ * Stage of the test:
+ *   0: expecting resolve_cb to be invoked with TESTING_RESOLVE_CB:1234
+ *   1: expecting resolve_cb to be invoked with NULL, NULL
+ *   2: expecting connect_cb to invoked with socket address 56.21.0.0:57616
```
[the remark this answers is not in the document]
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

benesch:
@edenhill I believe I've addressed all your feedback! 🤞🏽

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,176 @@
...
+
+        int32_t cnt;
+        struct sockaddr_in *addr_in;
+
+        test_curr = this_test;
+
+        TEST_SAY("connect_cb invoked\n");
+
+        cnt = rd_atomic32_get(&stage);
+        TEST_ASSERT(cnt == 2, "connect_cb invoked in unexpected stage: %d\n",
```
edenhill, line 132: nit: TEST_ASSERT and TEST_FAIL should not have newlines

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,176 @@
...
+                    cnt);
+
+        TEST_ASSERT(addr->sa_family == AF_INET,
+                    "address has unexpected type: %d\n", addr->sa_family);
```
edenhill, line 136: dito

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,176 @@
...
+
+        addr_in = (struct sockaddr_in *)(void *)addr;
+
+        TEST_ASSERT(addr_in->sin_port == htons(4321),
+                    "address has unexpected port: %d\n", addr_in->sin_port);
```
edenhill, line 141: this will print the big endian port, use:   ntohs(..)

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,176 @@
...
+        TEST_ASSERT(addr_in->sin_addr.s_addr == htonl(0x7f010203),
+                    "address has unexpected host: %d\n",
```
edenhill, line 143:
```suggestion
                    "address has unexpected host: 0x%x\n",
```

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,176 @@
...
+                    addr_in->sin_addr.s_addr);
```
edenhill, line 144: ntohl()

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,176 @@
...
+        TEST_SAY("resolve_cb invoked: node=%s service=%s\n", node, service);
+
+        cnt = rd_atomic32_get(&stage);
+
+        if (cnt == 0) {
+                /* Stage 0: return a bogus address. */
+
+                struct sockaddr_in *addr;
+
+                TEST_ASSERT(strcmp(node, "TESTING_RESOLVE_CB") == 0,
```
edenhill, line 83: Add TEST_ASSERT that node and service are not NULL.

also remove \n from asserts

tests/0136-resolve_cb.c
```diff
@@ -0,0 +1,176 @@
...
+                      const char *service,
+                      const struct addrinfo *hints,
+                      struct addrinfo **res,
+                      void *opaque) {
+
+        int32_t cnt;
+
+        test_curr = this_test;
+
+        TEST_SAY("resolve_cb invoked: node=%s service=%s\n", node, service);
```
edenhill, line 74: move this after atomic get so you can print the stage as part of it, makes for easier troubleshooting

edenhill (changes requested):
almost there!

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

tests/0136-resolve_cb.c
benesch: Done.

edenhill (approved):
LGTM!

benesch:
Thanks very much, Magnus! Excited to have this officially part of librdkafka.

edenhill:
Thank you!

PR #4052, opened 2022-11-08
Add more explicit .clang-format files, warn if clang-format version != 10

milindl:
.clang-format files have been made more explicit using the --dump-files option on clang-format v10, this guards against any changes to the defaults, or changes to the 'BasedOnStyle' styles. 

For .clang-format-cpp, because of the way we are using it in the script by converting it to the key/value format, a number of options from the --dump-files option have to be omitted, hence why we need to retain the BasedOnStyle. For .clang-format, we no longer need this.

Additionally, style-format.sh now warns in case clang-format version is not 10.x.y.


packaging/tools/style-format.sh
```diff
@@ -70,6 +72,14 @@ for f in $*; do
         stylename="C"
     fi
 
+    if [[ $lang == c && $clang_version_warning == 0 ]]; then
+        # Check if clang-format is at the correct version, and if not, warn.
+        if ! clang-format --version | grep -o "version ${clang_format_version}\.[[:digit:]]\+\.[[:digit:]]\+" 1> /dev/null; then
```
edenhill, line 77: you can use `grep -q` instead of redirecting the output

packaging/tools/style-format.sh
```diff
@@ -70,6 +72,14 @@ for f in $*; do
         stylename="C"
     fi
 
+    if [[ $lang == c && $clang_version_warning == 0 ]]; then
```
edenhill, line 75: No reason to check this for each file, add the check above the loop, preferably on line 23, and I'd change it to erroring out instead of just warning.
Error messages should be formatted as:
"$0: clang-format version 10 or 11 required"

edenhill (changes requested):

packaging/tools/style-format.sh
milindl: Using sed now

packaging/tools/style-format.sh
milindl: I've moved it to that line and erroring out.
The original logic for having it inside the loop was that, we should allow make style-check-(changed) to run on things like python files without actually needing clang-format. But I think that might be a rare case to have someone need to check just non c/cpp files.

packaging/tools/style-format.sh
```diff
@@ -21,6 +21,12 @@ else
     fix=0
 fi
 
+clang_format_version=$(clang-format --version | sed -Ee 's/.*version ([[:digit:]]+)\.[[:digit:]]+\.[[:digit:]]+.*/\1/')
```
edenhill, line 24: Very good :+1:

edenhill (approved):
Great stuff!

edenhill:
Bonus points for rebasing and squashing/fixing up into a single commit :+1:

PR #4054, opened 2022-11-08
Rewrote TXN API call handling to properly handle timeouts and resumable calls

edenhill:
Fixes problem in #4016


tests/0103-transactions.c
```diff
@@ -635,9 +635,9 @@ static void do_test_misuse_txn(void) {
                 error = rd_kafka_begin_transaction(p);
                 TEST_ASSERT(error, "Expected begin_transactions() to fail");
                 TEST_ASSERT(rd_kafka_error_code(error) ==
-                                RD_KAFKA_RESP_ERR__STATE,
+                                RD_KAFKA_RESP_ERR__CONFLICT,
```
emasab, line 638: Is it possible to keep to same _STATE error here? Maybe some application or binding is expecting the previous code and not the new one. It's also similar to the exception Java throws: IllegalStateException

src/rdkafka_txnmgr.c
```diff
@@ -374,73 +407,266 @@ void rd_kafka_txn_set_abortable_error0(rd_kafka_t *rk,
...
-                                        const char *errstr_fmt,
-                                        ...) RD_FORMAT(printf, 4, 5);
-
-static void rd_kafka_txn_curr_api_reply(rd_kafka_q_t *rkq,
-                                        int actions,
-                                        rd_kafka_resp_err_t err,
-                                        const char *errstr_fmt,
-                                        ...) {
-        rd_kafka_error_t *error = NULL;
+static void rd_kafka_txn_curr_api_set_result0(const char *func,
```
emasab, line 622: Good idea to introduce this function!

tests/0103-transactions.c
edenhill: Users are not supposed to treat specific error codes differently, but look at the error attributes (retriable, requires_txn_abort, etc), and the reason for using __CONFLICT is to make it clearer that it is primarily conflictuous. The __STATE error code is still returned for more statey errors, like calling commit_..() when not in a transaction and so forth.

tests/0105-transactions_mock.c
```diff
@@ -3044,5 +3170,9 @@ int main_0105_transactions_mock(int argc, char **argv) {
 
         do_test_txn_coordinator_null_not_fatal();
 
+        do_test_txn_resumable_calls_timeout(rd_true);
+
+        do_test_txn_resumable_calls_timeout(rd_false);
```
emasab, line 3175: Do I add also the tests from #4016 about misuse for commit and abort and the misuse with concurrent call from a different thread? (do_test_txn_unexpected_operations and do_test_txn_concurrent_operations)

src/rdkafka_int.h
```diff
@@ -397,55 +400,32 @@ struct rd_kafka_s {
...
-             *   .._F_REUSE, blocking                                          \
-             *   any non-F_REUSE                                               \
-             *   curr API calls. */
-#define RD_KAFKA_TXN_CURR_API_F_REUSE                                          \
-        0x8 /**< Reuse/continue with                                           \
-             *   current API state.                                            \
-             *   This is used for                                              \
-             *   multi-stage APIs,                                             \
-             *   such as txn commit. */
+                        char name[64];     /**< API name, e.g.,
```
philipnee, line 403: the comments are a bit misaligned with others. not sure if this is intended. (by 6 spaces)

edenhill: Yeah it's a bit random, but we rely on clang-format to do formatting and the pros outweighs the cons :)

emasab (approved):
LGTM, great refactoring! The timeout in-flight and in-queue are handled in #4064

PR #4056, opened 2022-11-08
Fix the batch.size condition on the produce path

travisdowns:
When the batch.size is exceeded, we will close and send the current batch. In case compression is enabled, we "inflate" the batch size by 8x for the purposes of this check.

At least, that's the intent. Currently the condition is reversed: the 8x inflation happens if compression is disabled, not the other way around. The effect is that messages linger for longer than expected in the queue, as if the batch size was 8x larger. They are still sent respecting the batch size however, since this occurs at a different location.

This change reverses the sense of the condition so that the 8x multiplier is applied iff compression is enabled.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3803,11 +3803,8 @@ static int rd_kafka_toppar_producer_serve(rd_kafka_broker_t *rkb,
                     now, flushing ? 1 : rkb->rkb_rk->rk_conf.buffering_max_us,
                     /* Batch message count threshold */
                     rkb->rkb_rk->rk_conf.batch_num_messages,
-                    /* Batch size threshold.
-                     * When compression is enabled the
-                     * threshold is increased by x8. */
-                    (rktp->rktp_rkt->rkt_conf.compression_codec ? 1 : 8) *
-                        (int64_t)rkb->rkb_rk->rk_conf.batch_size);
+                    /* Batch total size threshold */
+                    rkb->rkb_rk->rk_conf.batch_size);
         }
 
         rd_kafka_toppar_unlock(rktp);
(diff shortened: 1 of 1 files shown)


edenhill:
Whoa, good find!

edenhill:
I wonder if this has adverse effects on the compression case.. maybe we should not do the x8 at all, just to be safe.

travisdowns:
> I wonder if this has adverse effects on the compression case.. maybe we should not do the x8 at all, just to be safe.

I think it depends a bit on the write/send path: it separately respects `batch.size`, but is this the "before" or "after" compression figure?

If it is before compression, then I think this 8x doesn't make any sense at all. If it is after compression, then it seems that this 8x is a problem in the case that the compression ratio is lower: we won't even try to send the batch even though it is already starting to exceed the size: by the time we trigger `batch_ready` maybe there are 4 batches ready to send or whatever. 

In that case, if we used a multiplier of 1, I guess the impact is performance:  in "server" we will be constantly triggering the batch_ready condition but nothing gets sent.

I do think it makes sense to look at the Java client and as long as the behavior is reasonable there, I think there is a strong incentive to align the approach. In fact, this was discovered exactly when switching from the Java client and then noting the sudden change in behavior.

edenhill:
There's this WIP branch for adding streaming compression, which should alleviate this problem:
https://github.com/edenhill/librdkafka/tree/streamingcompr

It's a bit abandoned though, but it was working quite well in testing.
There was a lot of tuning of compression buffer sizes, and that's where it is currently at.

travisdowns:
I guess the safest approach would be to use no multiplier since that's how it is effectively now (and I guess no one has complained) and I think that's how it was before the changes earlier this year that introduced this 8x multiplier (from what I can tell, I didn't do deep research on whether this also existed somewhere else before) and I _think_ it is consistent with the Java client.

edenhill:
Yep, I agree. Rip it out

travisdowns:
@edenhill - new push with no multiplier regardless of whether compression is enabled or not.

edenhill (changes requested):
There is another call down at line 3970

travisdowns:
@edenhill oops, good catch! Fixed that one too.

edenhill:
Very good, thank you!

PR #4062, opened 2022-11-11
Fixed issue with barrier fetch operation with batch consume

pranavrth:


tests/0134-purge_same_partition_messages_after_seek_in_same_batch_consume.c
```diff
@@ -0,0 +1,281 @@
+/*
```
edenhill, line 1: Following the repo style, this filename is way too long and is more of a test description (which goes in the test itself).
Suggest renaming to 0136-toppar_barrier.c.
Use number 0136:  134 and 135 are taken in other PRs.

src/rdkafka_queue.c
```diff
@@ -548,8 +549,8 @@ static size_t rd_kafka_purge_outdated_messages(int32_t version,
         for (i = 0; i < cnt; i++) {
                 rd_kafka_op_t *rko;
                 rko = rkmessages[i]->_private;
-                if (rd_kafka_op_version_outdated(rko, version)) {
-                        /* This also destroys the corresponding rkmessage. */
+                if (rko->rko_rktp == rktp &&
+                    rd_kafka_op_version_outdated(rko, version)) {
```
edenhill, line 553: Looking at op_version_outdated(), I think we might want to call it version set to 0 and let it just fall down to its 3rd if-statement:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_partition.h#L806

tests/test.c
```diff
@@ -480,6 +481,7 @@ struct test tests[] = {
     _TEST(0131_connect_timeout, TEST_F_LOCAL),
     _TEST(0132_strategy_ordering, 0, TEST_BRKVER(2, 4, 0, 0)),
     _TEST(0133_ssl_keys, TEST_F_LOCAL),
+    _TEST(0134_purge_same_partition_messages_after_seek_in_same_batch_consume, 0, TEST_BRKVER(2, 4, 0, 0)),
```
edenhill, line 484: Why does it require broker version >= 2.4?

pranavrth:
Fix as part of https://github.com/confluentinc/librdkafka/pull/4117

PR #4064, opened 2022-11-15
Treat EndTxn timeouts as retriable

edenhill:


emasab (approved):
LGTM

emasab:
Checking the failed test

edenhill:
It's a race condition in init_transactions() result return from the refactor, I'm on it :)

PR #4065, opened 2022-11-15
Fix compilation with boringssl

filimonov:
Sincle 1.7.0 compilation with boringssl  was not working out of the box (see a93482b7ef36d06e9483770e040ff31dbeb9ff74)   

It seems like after 3709caa91391cb115848ef9946ad228aa56ea476 only some tiny switches are needed to make it work again.

See https://boringssl.googlesource.com/boringssl/+/HEAD/PORTING.md


edenhill:
Thank you!

PR #4067, opened 2022-11-21
Fix memory leaks (mostly in tests)

edenhill:
Seems ASAN stopped reporting leaks properly, these are from valgrind runs.


emasab (approved):
LGTM

PR #4076, opened 2022-11-28
Some post-refactor transaction fixes

edenhill:


src/rdkafka_txnmgr.c
```diff
@@ -1329,6 +1329,16 @@ rd_kafka_error_t *rd_kafka_init_transactions(rd_kafka_t *rk, int timeout_ms) {
...
 
+        /* Cap actual timeout to transaction.timeout.ms * 2 when an infinite
+         * timeout is provided, this is to make sure the call doesn't block
+         * indefinitely in case a coordinator is not available.
+         * This is only needed for init_transactions() since there is no
+         * coordinator to time us out yet. */
+        if (timeout_ms == RD_POLL_INFINITE &&
+            /* Avoid overflow */
+            rk->rk_conf.eos.transaction_timeout_ms < INT_MAX / 2)
+                timeout_ms = rk->rk_conf.eos.transaction_timeout_ms * 2;
```
milindl, line 1340: We should use `timeout_ms = INT_MAX` or similar in case `rk->rk_conf.eos.transaction_timeout_ms >= INT_MAX / 2`
In this case, it would remain infinite whenever transaction_timeout_ms >= INT_MAX/2, and user provided timeout_ms is RD_POLL_INFINITE.

edenhill: For all practical purposes, infinite and >INT_MAX/2 (>12 days) are effectively the same, so didn't bother.

edenhill: The broker default max transaction timeout is 15 minutes, for reference.

emasab (approved):
LGTM

PR #4080, opened 2022-11-29
Change ssl.endpoint.identification.algorithm to be enabled by default

edenhill:
Since this a potentially breaking change we need to get it in the upcoming 2.0.0 release.


emasab (approved):
LGTM

PR #4083, opened 2022-11-30
Try to avoid master references when possible

sarroutbi:
References to some words such as master/slave, blacklist, whitelist, etc. shoud be avoided. 

References to master branch or blacklist in librdkafka also exist, but removing them would imply higher impact, such as backward uncompatibility when changing topic.blacklist, for example

Signed-off-by: Sergio Arroutbi <sarroutb@redhat.com>


edenhill (approved):
Thank you

PR #4089, opened 2022-12-01
Prevent offsets from being committed manually when a rebalance is in progress and partitions are owned by consumer

roxelo:
See https://github.com/edenhill/librdkafka/issues/4059

When using cooperative sticky, attempting to commit offsets during a rebalance can trigger a follow up rebalance.
One way of reducing the risk of this happening, is to add the same [check that is included in the auto commit function](https://github.com/edenhill/librdkafka/blob/v1.8.2/src/rdkafka_cgrp.c#L3366-L3370) in the manual commit method

### DIFF
--- a/src/rdkafka_offset.c
+++ b/src/rdkafka_offset.c
@@ -400,6 +400,12 @@ rd_kafka_commit(rd_kafka_t *rk,
                 rq   = RD_KAFKA_REPLYQ(repq, 0);
         }
 
+        /* Don't attempt auto commit when rebalancing or initializing since
+         * the rkcg_generation_id is most likely in flux. */
+        if (rkcg->rkcg_subscription &&
+            rkcg->rkcg_join_state != RD_KAFKA_CGRP_JOIN_STATE_STEADY)
+                return RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS;
+
         err = rd_kafka_commit0(rk, offsets, NULL, rq, NULL, NULL, "manual");
 
         if (!err && !async)
--- a/tests/0137-cooperative_commit_rebalance.c
+++ b/tests/0137-cooperative_commit_rebalance.c
@@ -1,0 +1,140 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2020, Magnus Edenhill
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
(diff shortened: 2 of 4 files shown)


src/rdkafka_offset.c
```diff
@@ -400,6 +400,12 @@ rd_kafka_commit(rd_kafka_t *rk,
                 rq   = RD_KAFKA_REPLYQ(repq, 0);
         }
 
+        /* Don't attempt auto commit when rebalancing or initializing since
+         * the rkcg_generation_id is most likely in flux. */
+        if (rkcg->rkcg_subscription &&
```
edenhill, line 405: This is not thread safe (this is on the application thread, but rkcg is local to the rdkafka main thread).

Check should probably be performed in rd_kafka_cgrp_offsets_commit().

But we need to make sure that it is ok to commit during the phase of the rebalance when the rebalance callback/event is triggered.

tests/0137-cooperative_commit_rebalance.c
```diff
@@ -0,0 +1,140 @@
+/*
```
edenhill, line 1: Bravo on test!

tests/0137-cooperative_commit_rebalance.c
```diff
@@ -0,0 +1,140 @@
...
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2020, Magnus Edenhill
```
edenhill, line 4: 2022

edenhill (changes requested):
Also make sure to run `make style-fix` (on linux, since it si broken on osx)

src/rdkafka_offset.c
roxelo: Fixed, I also removed the same check from `rd_kafka_cgrp_offset_commit_tmr_cb` since the function will also call `rd_kafka_cgrp_offsets_commit`. I am not sure if this is the ideal behavior for auto commit though as we would no longer "silently" skip committing offsets...

wmorgan6796:
@roxelo whats the status of this PR? Any help needed?

roxelo:
@wmorgan6796 Some of the tests failed and I haven't had a chance to look at it further. If you have some time to investigate, another set of eyes would help

wmorgan6796:
> @wmorgan6796 Some of the tests failed and I haven't had a chance to look at it further. If you have some time to investigate, another set of eyes would help

I believe I found the reason for the test failure. We fundamentally change how offset committing occurs by short circuiting even sending the OffsetCommitRequest to the brokers. When we send the request to the brokers, we will get an  `RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID` from the (mock) brokers. But since the request never makes it to the brokers, we end up changing the error returned to `RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS`.

I think the correct thing to do is allow for the error returned to change, but ultimately would leave it up to the maintainers to decide is ok.

wmorgan6796:
https://github.com/confluentinc/librdkafka/pull/4089#issuecomment-1468597923 

@milindl @emasab regarding this, what are your thoughts?

Issue #4103, opened 2022-12-14
Draft: rd_kafka_flush() may loss data


shenxingwuying:
Description
===========
I hava found a problem about 'rd_kafka_flush()', which can cause some data loss.

The scenarios is:

We use librdkafka to write record to KafkaCluster:
`
resp_code = rd_kafka_produce*(...)
resp_code = rd_kafka_flush(...);
`

I the code

When I stop the KafkaCluster.

rd_kafka_flush() may return an incorrect code, it's a low probability event. And I found when the case happens, it always print a log about producer, like this:

`
#producer.*|.*partition count changed from .* to 0
`

How to reproduce
================
I have reproduced the issue. My code file using librdkafka's unittest structure can be provide later. 
But it's not an very directly unit test because I am not familiar with librdkafka project.

I print some logs and then using shell command to prove it.

I am glad to receive your advices about the test codes.

PR #4104, opened 2022-12-14
DRAFT: Bug reproduce: add an unit test case for rd_kafka_flush. #4103

shenxingwuying:


tests/test.conf
```diff
@@ -0,0 +1,27 @@
+# Copy this file to test.conf and set up according to your configuration.
+
+#
+# Test configuration
+#
+# For slow connections: multiply test timeouts by this much (float)
+test.timeout.multiplier=100
```
shenxingwuying, line 7: Is there any other method to set the test's timeout ?

tests/0800-fuzzy.c
```diff
@@ -0,0 +1,195 @@
...
+        system(start_cmd);
+        sleep(6);
+    }
+    TEST_SAY("Last start kafka, make sure kafka is running\n");
+    system(start_cmd);
+    TEST_SAY("Fuzzy thread finish");
+    return NULL;
+}
+
+int main_0800_fuzzy(int argc, char** argv) {
```
shenxingwuying, line 158: The test case is an integration test, which code path should integration tests be?

tests/test.conf
shenxingwuying: How to control the test.conf file?

PR #4107, opened 2022-12-15
Update curl to 7.86.0

edenhill:
Fixes CVE-2022-35252, CVE-2022-42915, CVE-2022-42916


edenhill:
The style check is fixed in the semaphoreci PR.

emasab (approved):

PR #4110, opened 2022-12-16
Incremental alter configs implementation [KIP-339] 

PrasanthV454:
- KIP-339 Incremental Alter Configs Implementation. 

- Supporting Set, Remove, Append and Subtract Operations.


src/rdkafka.h
```diff
@@ -7190,6 +7211,26 @@ rd_kafka_ConfigResource_set_config(rd_kafka_ConfigResource_t *config,
...
+ * This will overwrite the current value.
+ *
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR if config was added to resource,
+ *          or RD_KAFKA_RESP_ERR__INVALID_ARG on invalid input.
+ */
+RD_EXPORT rd_kafka_resp_err_t
+rd_kafka_ConfigResource_incremental_set_config(rd_kafka_ConfigResource_t *config,
+                                   const char *name,
+                                   const char *value);
+
```
mahajanadhitya, line 7231: Add incremental append , remove and subtract as well to export

src/rdkafka_admin.c
```diff
@@ -2585,6 +2634,26 @@ rd_kafka_ConfigResource_delete_config(rd_kafka_ConfigResource_t *config,
...
+
+rd_kafka_resp_err_t
+rd_kafka_ConfigResource_incremental_subtract_config(rd_kafka_ConfigResource_t *config,
+                                   const char *name) {
+        if (!name || !*name)
+                return RD_KAFKA_RESP_ERR__INVALID_ARG;
+
+        return rd_kafka_admin_incremental_add_config0(&config->config, name, NULL,
+                                          RD_KAFKA_INCREMENTAL_ALTER_OP_SUBTRACT);
+}
```
mahajanadhitya, line 2655: incremental subtract does expect a value string as well just like set and append. So basically it omits only the mentioned values in the list. 
Eg: key1->[v1,v2,v3] 
Subtract(key1,"v1,v3")
Key1->[v2]

milindl (commented):
Did a first-pass review, will look into this again later

src/rdkafka.h
```diff
@@ -7495,6 +7592,65 @@ rd_kafka_AlterConfigs_result_resources(
...
+/**
+ * @brief Incrementally update the configuration for the specified resources.
+ *        Updates are not transactional so they may succeed for some resources
+ *        while fail for others. The configs for a particular resource are
+ * updated atomically, replacing values using the provided ConfigEntrys and
+ * reverting unspecified ConfigEntrys to their default values.
+ *
+ * @remark Requires broker version >=2.3.0
+ *
+ * @warning IncrementalAlterConfigs will replace all existing configuration for
```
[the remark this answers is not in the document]
PrasanthV454: Missed few things to change. It is one among these. Most of the changes are to be done in comments. btw yes this is the main disadvantage of AlterConfigs.

src/rdkafka.h
```diff
@@ -6621,9 +6638,11 @@ typedef enum rd_kafka_admin_op_t {
         RD_KAFKA_ADMIN_OP_DELETETOPICS,     /**< DeleteTopics */
         RD_KAFKA_ADMIN_OP_CREATEPARTITIONS, /**< CreatePartitions */
         RD_KAFKA_ADMIN_OP_ALTERCONFIGS,     /**< AlterConfigs */
-        RD_KAFKA_ADMIN_OP_DESCRIBECONFIGS,  /**< DescribeConfigs */
-        RD_KAFKA_ADMIN_OP_DELETERECORDS,    /**< DeleteRecords */
-        RD_KAFKA_ADMIN_OP_DELETEGROUPS,     /**< DeleteGroups */
+        RD_KAFKA_ADMIN_OP_INCREMENTALALTERCONFIGS,
```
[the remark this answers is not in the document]
PrasanthV454: got it.

src/rdkafka.h
pranavrth, line 1: `ConfigEntry` for us is not same as `ConfigEntry` in java. Our `ConfigEntry` is equivalent to `AlterConfigOp` in Java. Shall we do this change?

tests/test.c
```diff
@@ -6486,6 +6497,99 @@ rd_kafka_resp_err_t test_AlterConfigs_simple(rd_kafka_t *rk,
...
+/**
+ * @brief Delta Incremental Alter configuration for the given resource,
+ *        overwriting/setting the configs provided in \p configs.
+ *        Existing configuration remains intact.
+ *
+ * @param configs 'const char *name, const char *value' tuples
+ * @param config_cnt is the number of tuples in \p configs
+ */
+rd_kafka_resp_err_t
+test_IncrementalAlterConfigs_simple(rd_kafka_t *rk,
```
pranavrth, line 6509: I don't exactly why is this needed. Can you explain once?

pranavrth (changes requested):
First round of review related to example, test and structs. Checking implementation in details.

examples/incremental_alter_configs.c
```diff
@@ -0,0 +1,348 @@
...
+                        ? RD_KAFKA_RESOURCE_TOPIC
+                        : !strcmp(restype_s, "BROKER")
+                              ? RD_KAFKA_RESOURCE_BROKER
+                              : RD_KAFKA_RESOURCE_UNKNOWN;
+
+                if (restype == RD_KAFKA_RESOURCE_UNKNOWN) {
+                        usage("Invalid resource type: %s", restype_s);
+                }
+
+                if (restype != prev_restype || strcmp(resname, prev_resname)) {
```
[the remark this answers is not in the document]
emasab: Added a comment about that

src/rdkafka_admin.h
```diff
@@ -31,6 +31,7 @@
 
 
 #include "rdstring.h"
+#include "rdmap.h"
```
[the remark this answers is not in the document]
emasab: It's now used in `rd_kafka_IncrementalAlterConfigs` for finding duplicates

tests/test.c
emasab: Added tests for the test helper, there was something to fix

src/rdkafka.h
emasab: In J they have a Map from ConfigResource to ConfigEntry. In librdkafka the ConfigEntry is inside ConfigResource and the list of entries isn't public. One could add a new list of alter_config_ops to the ConfigResult and these AlterConfigOp would contain a ConfigEntry and an incremental op but that wait it would still needed to have this list even if it's not need, for example when describing configs.
I think this is a good compromise, similar to the previous operation that was added for AlterConfigs but it's always SET because later the functionality was added in a different RPC. I could also remove the previous operation field because it's only used internally and only leave the new one.

emasab: Latest commit b233fea removes the previous `rd_kafka_AlterOperation_t`

pranavrth (changes requested):

pranavrth (approved):
Good Work.

LGTM!

src/rdkafka_admin.c
```diff
@@ -2832,37 +2834,42 @@ rd_kafka_ConfigResource_add_ConfigEntry(rd_kafka_ConfigResource_t *config,
...
-rd_kafka_ConfigResource_set_config(rd_kafka_ConfigResource_t *config,
-                                   const char *name,
-                                   const char *value) {
-        if (!name || !*name || !value)
-                return RD_KAFKA_RESP_ERR__INVALID_ARG;
 
-        return rd_kafka_admin_add_config0(&config->config, name, value,
-                                          RD_KAFKA_ALTER_OP_SET);
-}
+rd_kafka_error_t *rd_kafka_ConfigResource_set_incremental_config(
```
pranavrth, line 2848: This should be `rd_kafka_ConfigResource_add_incremental_config`

emasab: Currently it'using the same naming as rd_kafka_ConfigResource_set_config, the key currently is (name), in future could be (name, op_type)

pranavrth: But it's adding incremental alter config to the config list of the resource. Set feels that it replaces the old one and set the current one.

PR #4111, opened 2022-12-16
free key in rd_kafka_msg_destroy

uananth:


src/rdkafka_msg.c
```diff
@@ -121,6 +121,9 @@ void rd_kafka_msg_destroy(rd_kafka_t *rk, rd_kafka_msg_t *rkm) {
 
         if (rkm->rkm_flags & RD_KAFKA_MSG_F_FREE && rkm->rkm_payload)
                 rd_free(rkm->rkm_payload);
+        
+        if (rkm->rkm_flags & RD_KAFKA_MSG_F_FREE && rkm->rkm_key)
```
uananth, line 125: or, have a new flag for freeing MSG and KEY if it breaks existing behavior that others are using?

edenhill:
The key is allocated along with the rkm itself, so it does not need separate freeing.

uananth:
@edenhill , i do see that. 
`        if (payload && msgflags & RD_KAFKA_MSG_F_COPY) {
                /* Copy payload to space following the ..msg_t */
                rkm->rkm_payload = p;
                memcpy(rkm->rkm_payload, payload, len);
                p += len;

        } else {
                /* Just point to the provided payload. */
                rkm->rkm_payload = payload;
        }

        if (key) {
                rkm->rkm_key     = p;
                rkm->rkm_key_len = keylen;
                memcpy(rkm->rkm_key, key, keylen);
        } else {
                rkm->rkm_key     = NULL;
                rkm->rkm_key_len = 0;
        }`
        
        is there a reason to copy the key when being provided one instead of using what is provided when MSG_FREE is being used? we were expecting the payload and key to be freed after calling rd_kafka_producer()

edenhill:
They key is typically so small that it is more efficient to copy it.

edenhill:
The payload is also copied, if it is small enough (below 64k), because it turned out to be more efficient that way.

uananth:
ok. thank your for the details. is it possible to update the documentation that the key provided by the user will have to be handled by the user even if using MSG_FREE? because we found a memory leak in our application when we were passing both a msg, and key assuming both would end up being freed from within librdkakfa.

PR #4112, opened 2022-12-16
Don't return NULL from rd_kafka_new()'s "This should never happen" codepath

### LINKED ISSUE #4100: Use-after-free when rd_kafka_new returns RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE
Description
===========

We're hitting [this codepath](https://github.com/edenhill/librdkafka/blob/3b50e1e/src/rdkafka.c#L2512-L2537) in `rd_kafka_new`. I don't know _why_ we're hitting it, yet, but it likely has something to do with all Kafka brokers being unreachable. This is in a large integration test where we're specifically testing network failures. Sadly I don't have any small reproducer yet (or maybe ever).

```
        if (rd_kafka_init_wait(rk, 60 * 1000) != 0) {
                /* This should never happen unless there is a bug
                 * or the OS is not scheduling the background threads.
                 * Either case there is no point in handling this gracefully
                 * in the current state since the thread joins are likely
                 * to hang as well. */
                mtx_lock(&rk->rk_init_lock);
                rd_kafka_log(rk, LOG_CRIT, "INIT",
                             "Failed to initialize %s: "
                             "%d background thread(s) did not initialize "
                             "within 60 seconds",
                             rk->rk_name, rk->rk_init_wait_cnt);
                if (errstr)
                                    "Timed out waiting for "
                                    "%d background thread(s) to initialize",
                                    rk->rk_init_wait_cnt);
                mtx_unlock(&rk->rk_init_lock);

                rd_kafka_set_last_error(RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE,
                                        EDEADLK);
                return NULL;
        }
```

Looking at this code, I see that it calls `rd_kafka_log(rk, ...)` and then returns null without destroying the `rk` at all. This seems like a memory leak, right?  That's bad, but not fatal.
What's _fatal_ is that this codepath leads to a use-after-free for us, because we assume that when `rd_kafka_new` fails, we should clean up the `rk_conf` that we passed to it. In other words, we do exactly what's documented in [INTRODUCTION.md](https://github.com/edenhill/librdkafka/blob/3b50e1e/INTRODUCTION.md?plain=1#L857-L863):

```
    rk = rd_kafka_new(RD_KAFKA_PRODUCER, conf, errstr, sizeof(errstr));
    if (!rk) {
        rd_kafka_conf_destroy(rk);  // [sic]; actually this should say `conf` and I'll submit a PR for that in a minute
        fail("Failed to create producer: %s\n", errstr);
    }
    /* Note: librdkafka takes ownership of the conf object on success */
```

But along this specific codepath, `rd_kafka_new` fails and returns NULL _and yet there are still background threads running,_ and those background threads are going to try to access the function-pointer callbacks registered in that `conf` object. If we `rd_kafka_conf_destroy(conf)` on failure, then we have a use-after-free situation on those function pointers, and the symptom is generally a segfault.

#2820 might be related; it mentions this same codepath.

My ideal outcome here would be a simple rule like "When `rd_kafka_new` fails by returning NULL, it _always_ relinquishes ownership of the `rk_conf`, so you can feel free to destroy it at your leisure," or "Whenever `rd_kafka_new` is called, it _always_ takes ownership of the `rk_conf`, so you should actually _never_ destroy the `rk_conf` after that point because now it belongs to the library."  (The former is what I always _thought_ the rule was. The latter would be awesome, but probably can't be achieved in practice because it would cause double-free bugs for all existing code.)

The second-best outcome would be if you could give us a simple distinguishing rule for when we should destroy the `rk_conf` and when we shouldn't. For example, "When `rd_kafka_new` fails by returning NULL, it relinquishes ownership of the `rk_conf` if and only if `rd_kafka_last_error() != RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE`. Client code should check for that situation and avoid destroying the `rk_conf` if `rd_kafka_last_error() == RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE`."  (However, notice that that rule is also wrong, because of [this codepath](https://github.com/edenhill/librdkafka/blob/3b50e1e/src/rdkafka.c#L2466-L2485).)

Checklist
=========

 - [x] librdkafka version (release number or git tag): `v1.8.2`, `v1.9.2`
 - [x] Apache Kafka version: `2.6.3`, I think
 - [ ] librdkafka client configuration: not sure yet
 - [x] Operating system: RHEL 7, also probably any OS

Quuxplusone:
As the new comment says: We shouldn't return NULL unless failure has actually occurred and our caller can proceed with cleaning up their resources. If librdkafka is still actively running background threads that need to touch those resources, we MUST NOT communicate otherwise to our caller.

Since this codepath "should never happen," it theoretically doesn't matter much what we do here. But `return NULL` in practice leads to use-after-free segfaults on overloaded VMs, so we shouldn't do that. Instead, just loop (and log) until the background threads have run, and then proceed. Slow success is still success.

Fixes #4100.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2148,6 +2148,7 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
 #endif
         char builtin_features[128];
         size_t bflen;
+        int i;
 
         rd_kafka_global_init();
 
@@ -2512,28 +2513,27 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
         /* Wait for background threads to fully initialize so that
          * the client instance is fully functional at the time it is
          * returned from the constructor. */
-        if (rd_kafka_init_wait(rk, 60 * 1000) != 0) {
-                /* This should never happen unless there is a bug
-                 * or the OS is not scheduling the background threads.
-                 * Either case there is no point in handling this gracefully
+        for (i = 1; rd_kafka_init_wait(rk, 60 * 1000) != 0; ++i) {
+                /* This should never happen unless there is a bug,
+                 * or the OS is not scheduling the background threads
+                 * (for example we're on a severely overloaded virtual machine).
+                 * Either way there is no point in handling this gracefully
                  * in the current state since the thread joins are likely
-                 * to hang as well. */
+                 * to hang as well. But it is also unsafe to return NULL
+                 * to our caller, since that would indicate that we have
+                 * "successfully failed" and invite the caller to destroy
+                 * app_conf and the callbacks it holds. But the background
+                 * threads will try to use those callbacks as soon as they
+                 * wake up. So we must not return NULL, either. Just keep
+                 * looping, since that is the only safe behavior in this
+                 * unlikely case. */
                 mtx_lock(&rk->rk_init_lock);
                 rd_kafka_log(rk, LOG_CRIT, "INIT",
                              "Failed to initialize %s: "
                              "%d background thread(s) did not initialize "
-                             "within 60 seconds",
-                             rk->rk_name, rk->rk_init_wait_cnt);
-                if (errstr)
-                        rd_snprintf(errstr, errstr_size,
-                                    "Timed out waiting for "
-                                    "%d background thread(s) to initialize",
-                                    rk->rk_init_wait_cnt);
+                             "after waiting at least %d seconds",
+                             rk->rk_name, rk->rk_init_wait_cnt, 60*i);
                 mtx_unlock(&rk->rk_init_lock);
-
-                rd_kafka_set_last_error(RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE,
-                                        EDEADLK);
-                return NULL;
         }
 
         rk->rk_initialized = 1;


src/rdkafka.c
```diff
@@ -2512,28 +2513,27 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
...
                              "%d background thread(s) did not initialize "
-                             "within 60 seconds",
-                             rk->rk_name, rk->rk_init_wait_cnt);
-                if (errstr)
-                        rd_snprintf(errstr, errstr_size,
-                                    "Timed out waiting for "
-                                    "%d background thread(s) to initialize",
-                                    rk->rk_init_wait_cnt);
+                             "within %d seconds",
+                             rk->rk_name, rk->rk_init_wait_cnt, 60*i);
```
edenhill, line 2535: 60*i is likely to not be correct on heavily loaded vms, so recommend just skipping the %d seconds timeout and instead add what it will do, e.g., "Still waiting for %s background thread(s) to initialize".

Replace the for loop with a while, since `i` is no longer needed, alternatively make it assert after 3 attempts (e.g., 3 minutes).

edenhill (changes requested):

src/rdkafka.c
Quuxplusone: These messages will show up in /var/log/messages with timestamps, so the durations _after the first one_ don't need to be the primary source of truth. I agree with your assessment that the first duration isn't likely to be exactly 60 seconds either. Does librdkafka have a way of telling what time it is? If so, we could update this codepath to print the _actual_ duration elapsed (`now() - start_time`). Alternatively, we can print "Waited _at least_ %d seconds for..."; pending more discussion, for now I'll just update the patch in that direction.

Assert-failing is clearly better than returning NULL and segfaulting; but it's not clearly better than just continuing to loop forever and relying on the operator to `kill -9` the process. At least for our own codebase, we'll prefer to loop forever. Notice that we (the client programmer) will be seeing this message sent to our logging callback every 60 seconds forever, and we can choose to `abort` after 1 or 3 or 300 printings of the message, if we feel like it. I don't think librdkafka needs to take on that responsibility itself.

edenhill: This is a really niche problem, so let's keep this as simple as possible.
If the system is loaded it couldn't spin up the background threads in over a minute, it won't be very useful for anything anyway, so why not just assert?

Quuxplusone: AFAICT right now: in our own codebase we won't assert, so if upstream asserts, we'll have to patch it. This is not a big deal (we certainly need to patch it anyway, from today until whenever we upgrade to 2.0+), but it would be _nice_ if the future-upstream just looked perfect from our POV and didn't still need patching.

I still think the simple thing is just "wait until all the threads have started, don't try to be clever by hard-coding any specific timeout." Which is what this proposed patch does, at the moment, right?

Alternatively, is there a way to actually fix the root cause here? Like, this code is just waiting 60 seconds and then going "oh no, we spawned off some worker threads and now we can't get them back, _and_ if we return NULL then those worker threads are going to access freed memory when they _do_ get around to running." Could you do something here to mark the leaked `rd_kafka_t` object as "dead" so that when the workers do get around to running, they know not to access any of its config but simply exit? Or ask pthreads to aggressively kill those worker threads so that we know they're dead before we return? or something else along those lines? I think that's probably impossible, but just in case it brings something to mind for you...?

edenhill: I have a hard time understanding how any of this is useful in practice; if your system is so slow it takes minutes to spin up a background thread, then the subsequent of the multi-threaded librdkafka client will most likely be so slow that nothing works anyway. I.e., connects and handshakes will time out, making it so that there is no usable broker connections, etc.

I do understand that the current behaviour of returning with the background thread still running is problematic for cleanup and late-reference reasons, and we could probably do better.

Killing the background threads might work, but will leak memory and possibly cause deadlocks due to held locks, so we're just transitioning into another unknown state.

So, my suggestion is just to rd_assert() if the background thread hasn't come alive in one minute - it is the only safe thing to do since we can't reason about the current state of things, and we can't clean up properly.

Quuxplusone: @edenhill: I continue to think that the simplest thing is "wait until all the threads have started, don't hard-code any specific timeout."  However, I can see that that idea is unlikely to get adopted by you upstream; meanwhile I'm able to patch the code any way I like downstream.

So I think the appropriate outcome on this PR would be if you'd just commandeer it to add the `rd_kafka_assert(rk, !*"Resource exhaustion: background threads didn't start in 60 seconds")` or `rd_assert(!*"BUG: this should never happen")` or whatever you think is the appropriate way to assert-fail here — and also presumably remove the `rd_kafka_set_last_error` and `rd_snprintf` but presumably _not_ the `rd_kafka_log`. I could try to guess at those changes, but there are so many different ways it _might_ be done that I think it's easier if you do it.
That'll address the use-after-free bug; and then everything else I can handle via downstream patches.

Quuxplusone:
We have handled this downstream by adding the loop exactly as in my original PR.

PR #4113, opened 2022-12-16
Update README.md

Kaustubh01:
fixed grammer


edenhill:
The existing copy is as is intended and is grammatically correct afaik.

PR #4114, opened 2022-12-16
Doc typo: rd_kafka_conf_destroy(rk) should be (conf)

Quuxplusone:
As mentioned in #4100.


INTRODUCTION.md
```diff
@@ -846,17 +846,21 @@ Configuration is applied prior to object creation using the
 
     res = rd_kafka_conf_set(conf, "compression.codec", "snappy",
                             errstr, sizeof(errstr));
-    if (res != RD_KAFKA_CONF_OK)
+    if (res != RD_KAFKA_CONF_OK) {
+        rd_kafka_conf_destroy(conf);
         fail("%s\n", errstr);
+    }
```
edenhill, line 852: since there is no return; in this if-statement it is implied that fail() exits the application, so no point in doing conf_destroy() here. This example is not aiming to be complete.

Alternatively, also add a return -1;  or something to each of these.

INTRODUCTION.md
```diff
@@ -846,17 +846,21 @@ Configuration is applied prior to object creation using the
...
 
     rk = rd_kafka_new(RD_KAFKA_PRODUCER, conf, errstr, sizeof(errstr));
     if (!rk) {
-        rd_kafka_conf_destroy(rk);
+        rd_kafka_conf_destroy(conf);
```
edenhill, line 863: :+1:

INTRODUCTION.md
Quuxplusone: Okay, but then we should also remove the `rd_kafka_conf_destroy` from old line 859, because it is _also_ followed by a `fail`.

Alternatively, the reader could assume from context (as I did) that `fail` must _not_ exit the application — e.g. probably it logs the error message and `longjmp`s back to safety — in which case the cleanup _is_ required in all three places.

The worst possible outcome would be to imply that a failed `rd_kafka_new` requires the caller to clean up `conf`, but a failed `rd_kafka_conf_set` does not.

Quuxplusone: Oh, unless `fail` is just like `fprintf(stderr, ...)` and _doesn't_ exit the scope. In that case I should remove the two `rd_kafka_conf_destroy`s I added, but also add a `return;` after old line 860 (and stylistically IMHO swap lines 859 and 860).

Quuxplusone:
I've rebased the patch.
We still haven't resolved [what is the intended semantics of `fail(...)` in this example](https://github.com/confluentinc/librdkafka/pull/4114#discussion_r1051214093). Being a C++ chauvinist myself, I prefer to think of `fail(...)` as doing a `throw` or `longjmp`, so that we still need to clean up before calling it. That's also the most useful semantics _for documentation purposes_, because it gives us a chance to show the reader what they actually need to do to clean up after each of these calls. If we assume `fail(...)` means `abort()`, then we don't need to clean up (in this example) but we also can't show the reader what _they_ need to do to clean up (in actual practice).

Quuxplusone:
New year's ping!

PR #4077, opened 2022-11-28
Added changelog and Barrier op priority for Consumer batch purging issue

pranavrth:
- Added Changelog for the Consumer batch APIs multithreading issue
- Updated Barrier op priority

PR #4117, opened 2022-12-21
Fixed multiple issues with interaction of BARRIER operation with consumer batch API

pranavrth:
* Added partition comparison while purging messages in consume batch API.
* Made sure that Barrier operation is never skipped due to barrier version in batch API.
* Increase priority of Barrier operation so that it is run as soon as possible after a barrier operation in raised.

Continuation of the PR - https://github.com/confluentinc/librdkafka/pull/4077 as the repo is moved.


edenhill:
the windows build failures can be ignored

edenhill (approved):
LGTM

PR #4132, opened 2023-01-03
Update Kafka and librdkafka max API versions

emasab:


INTRODUCTION.md
```diff
@@ -1950,33 +1950,33 @@ release of librdkafka.
...
-| 13      | LeaveGroup          | 3           | 1                       |
-| 14      | SyncGroup           | 3           | 3                       |
-| 15      | DescribeGroups      | 4           | 0                       |
-| 16      | ListGroups          | 2           | 0                       |
+| 0       | Produce             | 9           | 7                       |
+| 1       | Fetch               | 13          | 11                      |
+| 2       | ListOffsets         | 7           | 2                       |
+| 3       | Metadata            | 12          | 4                       |
+| 8       | OffsetCommit        | 8           | 7                       |
+| 9       | OffsetFetch         | 8           | 1                       |
```
ijuma, line 1958: Shouldn't this be `7` for librdkafka? https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_request.c#L971

ijuma:
Thanks for the PR. One thing that would be very useful is to also include the first librdkafka version where a given protocol version was first supported.

INTRODUCTION.md
emasab: Yes, as this change is included in #3995 I've overlooked this but actually the version 7 was already supported.

pranavrth (approved):

PR #4139, opened 2023-01-09
Debuging c99 mode

fachrozi-fahmi:
I have found an issue with the path and line 292:
https://github.com/confluentinc/librdkafka/blob/1f9f245ac409f50f724695c628c7a0d54a763b9a/examples/list_consumer_group_offsets.c

I found an issue with tags [v2.0.0-RC2](https://github.com/confluentinc/librdkafka/releases/tag/v2.0.0-RC2)

image


emasab:
@fachrozi-fahmi it seems you're using a gcc version before 5.1.0. It had still C89 as default standard and we don't support C89. Try configuring and compiling with `CFLAGS=-std=c99`

emasab:
Solved by #4140.

PR #4140, opened 2023-01-09
Remove declarations inside for loops

emasab:


PrasanthV454 (commented):
Have a small doubt here. Any particular reason behind making this change. Generally I think best practice is to use the smallest scope possible for variables.

pranavrth (approved):

pranavrth:
> Have a small doubt here. Any particular reason behind making this change. Generally I think best practice is to use the smallest scope possible for variables.

This format is not supported in older version of C (before C99).

emasab:
Let's check just declarations in for loops. As for declaration before statements, it's better to avoid them, but it's not a strict requirement as if you compile with `-Wdeclaration-after-statement` you already get a lot of warnings in many places.

PrasanthV454:
> > Have a small doubt here. Any particular reason behind making this change. Generally I think best practice is to use the smallest scope possible for variables.
> 
> This format is not supported in older version of C (before C99).

Got it. thanks

Quuxplusone:
I'd recommend _not_ merging this patch. It's true that C89 didn't support declarations in `for` loops, but it's also been 34 years since 1989, and you have to go _really really far_ out of your way in the year 2023 to convince a compiler to complain about things that have been legal in C for probably the whole lifetime of many of the humans reading this.

Also, as mentioned in https://github.com/confluentinc/librdkafka/pull/4140#issuecomment-1377179989, librdkafka already uses C99-isms in many places. If you took a time machine back to the '80s and tried to compile librdkafka with one of those compilers, you'd find that they complain about a lot more than just variable declarations — they'd complain about `//` comments, the `unsigned long long` data type, and so on. Go back another 10 years and you'd find them complaining about the use of the keyword `void`.

Basically, I think librdkafka should assume the use of a compiler about as old as Kafka itself (i.e. 2011), but going further back in time than that seems impractical and masochistic.

emasab:
@Quuxplusone this is not for being C89 compatible, we're following the project declaration:

```## C standard "C98"

This is a mix of C89 and C99, to be compatible with old MSVC versions.

Notable, it is C99 with the following limitations:

 * No variable declarations after statements.
 * No in-line variable declarations.
 ```
 
But, limitation 1 has to be dropped because of https://github.com/confluentinc/librdkafka/pull/4140#issuecomment-1377179989
 and MSVC only implemented variable declarations after statements in 2013. This PR is about limitation 2, but without the first it won't compile anyway on older MSVC. Also Visual Studio 2012 support ended [yesterday](https://devblogs.microsoft.com/visualstudio/support-ends-for-older-versions-of-visual-studio-feb2022/).
 
 This said, it seems we should move to C99

Quuxplusone:
@emasab: "this is not for being C89 compatible, we're following the project declaration" — Except that, as you say, you're _not_ following the description in the README. The description in the README describes a world that self-evidently is not our reality: https://github.com/confluentinc/librdkafka/pull/4140#issuecomment-1377179989 . So it seems like what you should do, really, is just make the two-line update to the README's description so that it matches the real world. This PR makes reality worse, but (as you say) _doesn't even succeed_ at implementing the dystopia described in the README.

emasab:
We need to merge this as we support [RHEL/CentOS 7](https://docs.confluent.io/platform/current/installation/versions-interoperability.html#operating-systems), whose EOL is in 2024. CentOS has a GCC version before 5.1.0 where the default was changed from gnu90.

With these fixes it compiles correctly configuring with gnu90, and gives error without, so that's the minimum C standard we need to support.

We're going to include this in the README and in the CI. Up to 2024 when CentOS will reach EOL.

PR #4141, opened 2023-01-10
Return error on duplicate partitions

emasab:
also on duplicate match states

### DIFF
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -5210,6 +5210,7 @@ void rd_kafka_AlterConsumerGroupOffsets(
             rd_kafka_AlterConsumerGroupOffsetsResponse_parse,
         };
         rd_kafka_op_t *rko;
+        rd_kafka_topic_partition_list_t *copied_offsets;
 
         rd_assert(rkqu);
 
@@ -5243,6 +5244,18 @@ void rd_kafka_AlterConsumerGroupOffsets(
                         goto fail;
                 }
         }
+
+        /* Copy offsets list for checking duplicated */
+        copied_offsets =
+            rd_kafka_topic_partition_list_copy(alter_grpoffsets[0]->partitions);
+        if (rd_kafka_topic_partition_list_has_duplicates(
+                copied_offsets, rd_false /* ignore partition */)) {
+                rd_kafka_topic_partition_list_destroy(copied_offsets);
+                rd_kafka_admin_result_fail(rko, RD_KAFKA_RESP_ERR__INVALID_ARG,
+                                           "Duplicate partitions not allowed");
+                goto fail;
+        }
+        rd_kafka_topic_partition_list_destroy(copied_offsets);
 
         rko->rko_u.admin_request.broker_id = RD_KAFKA_ADMIN_TARGET_COORDINATOR;
         rko->rko_u.admin_request.coordtype = RD_KAFKA_COORD_GROUP;
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -1573,9 +1573,9 @@ rd_kafka_error_t *rd_kafka_AdminOptions_set_match_consumer_group_states(
         }
 
         for (i = 0; i < consumer_group_states_cnt; i++) {
+                uint64_t state_bit;
                 rd_kafka_consumer_group_state_t state =
                     consumer_group_states[i];
-                uint64_t state_bit = 1 << state;
 
                 if (state < 0 || state >= RD_KAFKA_CONSUMER_GROUP_STATE__CNT) {
                         rd_list_destroy(states_list);
@@ -1584,6 +1584,7 @@ rd_kafka_error_t *rd_kafka_AdminOptions_set_match_consumer_group_states(
                             "Invalid group state value");
                 }
 
+                state_bit = 1 << state;
                 if (states_bitmask & state_bit) {
                         rd_list_destroy(states_list);
                         return rd_kafka_error_new(
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -5452,6 +5452,7 @@ void rd_kafka_ListConsumerGroupOffsets(
             rd_kafka_ListConsumerGroupOffsetsResponse_parse,
         };
         rd_kafka_op_t *rko;
+        rd_kafka_topic_partition_list_t *copied_offsets;
 
         rd_assert(rkqu);
 
@@ -5466,9 +5467,7 @@ void rd_kafka_ListConsumerGroupOffsets(
                                            "Exactly one "
                                            "ListConsumerGroupOffsets must "
                                            "be passed");
-                rd_kafka_admin_common_worker_destroy(rk, rko,
-                                                     rd_true /*destroy*/);
-                return;
+                goto fail;
         }
 
         if (list_grpoffsets[0]->partitions != NULL &&
@@ -5480,9 +5479,23 @@ void rd_kafka_ListConsumerGroupOffsets(
                     "NULL or "
                     "non-empty topic partition list must "
                     "be passed");
-                rd_kafka_admin_common_worker_destroy(rk, rko,
-                                                     rd_true /*destroy*/);
-                return;
+                goto fail;
+        }
+
+        /* TODO: add group id duplication check when implementing KIP-709 */
+        if (list_grpoffsets[0]->partitions != NULL) {
+                /* Copy offsets list for checking duplicated */
+                copied_offsets = rd_kafka_topic_partition_list_copy(
+                    list_grpoffsets[0]->partitions);
+                if (rd_kafka_topic_partition_list_has_duplicates(
+                        copied_offsets, rd_false /* ignore partition */)) {
+                        rd_kafka_topic_partition_list_destroy(copied_offsets);
+                        rd_kafka_admin_result_fail(
+                            rko, RD_KAFKA_RESP_ERR__INVALID_ARG,
+                            "Duplicate partitions not allowed");
+                        goto fail;
+                }
+                rd_kafka_topic_partition_list_destroy(copied_offsets);
         }
 
         rko->rko_u.admin_request.broker_id = RD_KAFKA_ADMIN_TARGET_COORDINATOR;
@@ -5498,6 +5511,9 @@ void rd_kafka_ListConsumerGroupOffsets(
                     rd_kafka_ListConsumerGroupOffsets_copy(list_grpoffsets[0]));
 
         rd_kafka_q_enq(rk->rk_ops, rko);
+        return;
+fail:
+        rd_kafka_admin_common_worker_destroy(rk, rko, rd_true /*destroy*/);
 }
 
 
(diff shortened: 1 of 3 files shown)


pranavrth:
We should check for duplicate in describeConsumerGroups and deleteConsumerGroups as well.

pranavrth:
Also, for listConsumerGroupOffsets

src/rdkafka_admin.c
```diff
@@ -5210,6 +5210,7 @@ void rd_kafka_AlterConsumerGroupOffsets(
             rd_kafka_AlterConsumerGroupOffsetsResponse_parse,
         };
         rd_kafka_op_t *rko;
+        rd_kafka_topic_partition_list_t *copied_offsets;
```
pranavrth, line 5213: Why are we creating a new copy? Do we really need to preserve the order?

src/rdkafka_admin.c
```diff
@@ -5244,6 +5245,18 @@ void rd_kafka_AlterConsumerGroupOffsets(
                 }
         }
 
+        /* Copy offsets list for checking duplicated */
+        copied_offsets =
+            rd_kafka_topic_partition_list_copy(alter_grpoffsets[0]->partitions);
+        if (rd_kafka_topic_partition_list_has_duplicates(
+                copied_offsets, rd_false /* ignore partition */)) {
+                rd_kafka_topic_partition_list_destroy(copied_offsets);
```
pranavrth, line 5253: We can move this to _fail_ instead of here. I feel its better if we have all free and destroy methods at one place.

tests/0080-admin_ut.c
```diff
@@ -1608,6 +1610,22 @@ static void do_test_AlterConsumerGroupOffsets(const char *what,
...
+                    rd_kafka_topic_partition_list_new(3);
+                rd_kafka_topic_partition_list_add(partitions_duplicate,
+                                                  "topic1", 9)
+                    ->offset = 9;
+                rd_kafka_topic_partition_list_add(partitions_duplicate,
+                                                  "topic3", 15)
+                    ->offset = 15;
+                rd_kafka_topic_partition_list_add(partitions_duplicate,
+                                                  "topic1", 9)
+                    ->offset           = 1;
```
pranavrth, line 1625: styling

tests/0080-admin_ut.c
```diff
@@ -1696,6 +1714,38 @@ static void do_test_AlterConsumerGroupOffsets(const char *what,
...
+        err                                = rd_kafka_event_error(rkev);
+        const char *event_errstr_duplicate = rd_kafka_event_error_string(rkev);
+        TEST_ASSERT(err, "expected AlterConsumerGroupOffsets to fail");
+        TEST_ASSERT(err == RD_KAFKA_RESP_ERR__INVALID_ARG,
+                    "expected RD_KAFKA_RESP_ERR__INVALID_ARG, not %s",
+                    rd_kafka_err2name(err));
+        TEST_ASSERT(strcmp(event_errstr_duplicate,
+                           "Duplicate partitions not allowed") == 0,
+                    "expected \"Duplicate partitions not allowed\", not \"%s\"",
+                    event_errstr_negative);
```
pranavrth, line 1747: event_errstr_duplicate

pranavrth (changes requested):

pranavrth:
We should check for duplicate group names as well. Not required right now as we take 1 group only but will be future proof.

src/rdkafka_admin.c
emasab: Actually it's better to destroy as soon as possible, because as happened sometimes, if the function is not copying the passed parameter and it can lead to a use after free that we don't detect if we free later (after reading the response for example)

src/rdkafka_admin.c
emasab: I thinks it's better to avoid having side effects for an operation that is read only, also done in other places

emasab:
We can add the duplicate check for groups names it when we add support for more groups. I add a TODO

src/rdkafka_admin.c
```diff
@@ -5244,6 +5245,18 @@ void rd_kafka_AlterConsumerGroupOffsets(
                 }
         }
 
+        /* Copy offsets list for checking duplicated */
+        copied_offsets =
+            rd_kafka_topic_partition_list_copy(alter_grpoffsets[0]->partitions);
+        if (rd_kafka_topic_partition_list_has_duplicates(
+                copied_offsets, rd_false /* ignore partition */)) {
```
pranavrth, line 5269: Shouldn't the comment be - "don't ignore partition"? Its confusing, better to remove the comment.

we are making `ignore_partition = rd_false` here.

src/rdkafka_admin.c
```diff
@@ -5244,6 +5245,18 @@ void rd_kafka_AlterConsumerGroupOffsets(
                 }
         }
 
+        /* Copy offsets list for checking duplicated */
```
pranavrth, line 5265: Add TODO like below here as well.

src/rdkafka_admin.c
```diff
@@ -5244,6 +5245,18 @@ void rd_kafka_AlterConsumerGroupOffsets(
                 }
```
pranavrth, line 5262: In the line 5231, we should also add a NULL check `alter_grpoffsets[0]->partitions == NULL || alter_grpoffsets[0]->partitions->cnt == 0`

src/rdkafka_admin.c
```diff
@@ -5467,9 +5479,23 @@ void rd_kafka_ListConsumerGroupOffsets(
...
+                goto fail;
+        }
+
+        /* TODO: add group id duplication check when implementing KIP-709 */
+        if (list_grpoffsets[0]->partitions != NULL) {
+                /* Copy offsets list for checking duplicated */
+                copied_offsets = rd_kafka_topic_partition_list_copy(
+                    list_grpoffsets[0]->partitions);
+                if (rd_kafka_topic_partition_list_has_duplicates(
+                        copied_offsets, rd_false /* ignore partition */)) {
```
pranavrth, line 5508: Same as above for ignore_partition comment.

tests/0080-admin_ut.c
```diff
@@ -1756,10 +1807,12 @@ static void do_test_ListConsumerGroupOffsets(const char *what,
         rd_kafka_queue_t *q;
 #define MY_LIST_CGRPOFFS_CNT 1
         rd_kafka_AdminOptions_t *options = NULL;
-        rd_kafka_topic_partition_list_t *empty_cgoffsets_list;
+        rd_kafka_topic_partition_list_t *cgoffsets_empty_list;
```
pranavrth, line 1822: I think its better to have more verbose name. Something like `listcgoffsets_empty_partitions`. The current name suggests that we are dealing with offsets here but we are just concerned about partitions.

tests/0080-admin_ut.c
```diff
@@ -1756,10 +1807,12 @@ static void do_test_ListConsumerGroupOffsets(const char *what,
...
         const rd_kafka_ListConsumerGroupOffsets_result_t *res;
         rd_kafka_ListConsumerGroupOffsets_t *cgoffsets[MY_LIST_CGRPOFFS_CNT];
-        rd_kafka_ListConsumerGroupOffsets_t *empty_cgoffsets;
+        rd_kafka_ListConsumerGroupOffsets_t *cgoffsets_empty;
```
pranavrth, line 1825: Shouldn't be an array?

tests/0080-admin_ut.c
```diff
@@ -1833,11 +1895,39 @@ static void do_test_ListConsumerGroupOffsets(const char *what,
...
+        rkev = rd_kafka_queue_poll(q, exp_timeout + 1000);
+        TEST_SAY("ListConsumerGroupOffsets: got %s\n",
+                 rd_kafka_event_name(rkev));
+
+        /* Expecting error */
+        err = rd_kafka_event_error(rkev);
+        TEST_ASSERT(err, "expected ListConsumerGroupOffsets to fail");
+
+        errstr_ptr = rd_kafka_event_error_string(rkev);
+        TEST_ASSERT(!strcmp(errstr_ptr, "Duplicate partitions not allowed"),
```
pranavrth, line 1928: Shall we use `rd_strcmp`?

pranavrth (changes requested):

src/rdkafka_admin.c
emasab: I put `/*check partition*/` like in the other place

src/rdkafka_admin.c
emasab: There should be no KIP for alterConsumerGroupOffsets still, I add a comment more generic

tests/0080-admin_ut.c
emasab: If you see in tests it's not used, it adds a check for null but in this case cannot be null because of the previous assert

tests/0080-admin_ut.c
emasab: It's because it's the list used by cgoffsets_empty, uniformed to the name like in do_test_AlterConsumerGroupOffsets.

src/rdkafka_admin.c
```diff
@@ -1566,16 +1566,33 @@ rd_kafka_error_t *rd_kafka_AdminOptions_set_match_consumer_group_states(
...
+        uint64_t states_bitmask = 0;
+
+        if (RD_KAFKA_CONSUMER_GROUP_STATE__CNT >= 64) {
+                rd_assert("BUG: cannot handle states with a bitmask anymore");
+        }
+
         for (i = 0; i < consumer_group_states_cnt; i++) {
                 rd_kafka_consumer_group_state_t state =
                     consumer_group_states[i];
+                uint64_t state_bit = 1 << state;
```
pranavrth, line 1578: This should be after the if statement

src/rdkafka_admin.c
pranavrth: Good to know.

pranavrth (changes requested):

tests/0080-admin_ut.c
pranavrth: Sure for here.

General comment - I think we should always think about using rd_* version of the function as we might have to implement things differently based on the architecture or version.

tests/0080-admin_ut.c
pranavrth: I feel it still gives an impression that we are dealing with offsets but here we want verify that we don't have duplicate partitions. In alter API, it made sense as we don't want duplicate offset.

emasab: Here too, we were checking that in J. is returns a Map with TopicPartition so it's better that the partitions in the request are unique too to avoid having a different count

pranavrth: I am just talking about the variable name. `cgoffsets_duplicate` implies that we are talking about offsets. In ListConsumerGroupOffsets, we don't want to have duplicate partitions, so I feel its better to change it to something like `*partitions_duplicate*`. Nothing critical though.

pranavrth (approved):
LGTM

pranavrth:
Build didn't run. It is enabled for dev_.* branches.

PR #4161, opened 2023-01-26
Fix admin integration test with old broker versions

emasab:


tests/0081-admin.c
```diff
@@ -2416,6 +2416,8 @@ static void do_test_ListConsumerGroups(const char *what,
         rd_kafka_resp_err_t exp_err = RD_KAFKA_RESP_ERR_NO_ERROR;
         const rd_kafka_ListConsumerGroups_result_t *res;
         const rd_kafka_ConsumerGroupListing_t **groups;
+        rd_bool_t has_match_states =
+            test_broker_version >= TEST_BRKVER(2, 7, 0, 0);
```
milindl, line 2420: Wasn't the AK ticket merged in v2.6? May be that I am mistaken somehow and the release notes don't reflect the true situation, please ignore this comment in that case

https://home.apache.org/~rhauch/kafka-2.6.0-rc0/RELEASE_NOTES.html -- KAFKA-9130

milindl (approved):
LGTM

tests/0081-admin.c
emasab: What I do is checking the commit that added the feature, to the ListGroupsRequest.json file in this case.
https://github.com/apache/kafka/commit/fe948d39e54cf2d57e9b9e0e2a203890dc8ce86d

There it's written in which tags it's present and 2.7.x is the first one.

PR #4177, opened 2023-02-03
msg.timeout <= transaction.timeout enforcement check

jainruchir:


tests/0004-conf.c
```diff
@@ -498,6 +498,37 @@ static void do_message_timeout_linger_checks(void) {
...
 }
 
+/**
+ * @brief Verify behaviour of checking that message.timeout.ms fits within
+ *        configured transaction.timeout.ms. By jainruchir.
+ */
+static void do_message_timeout_transaction_timeout_checks(void) {
+        rd_kafka_conf_t *conf;
+        rd_kafka_t *rk;
+        char numstr[8];
```
milindl, line 508: Not needed by this test

tests/0004-conf.c
```diff
@@ -498,6 +498,37 @@ static void do_message_timeout_linger_checks(void) {
...
+        char errstr[512];
+        const char *transactional_id = "txnid";
+
+        SUB_TEST_QUICK();
+        test_conf_init(&conf, NULL, 60);
+        test_conf_set(conf, "transactional.id", transactional_id);
+        test_conf_set(conf, "socket.connection.setup.timeout.ms", "5000");
```
milindl, line 515: socket.connection.setup.timeout.ms and reconnect.backoff.max.ms are not needed by this test

tests/0004-conf.c
```diff
@@ -498,6 +498,37 @@ static void do_message_timeout_linger_checks(void) {
...
+        test_conf_set(conf, "reconnect.backoff.max.ms", "2000");
+
+        rd_snprintf(numstr, sizeof(numstr), "%d", 2);
```
milindl, line 518: Not needed after removing numstr

tests/0004-conf.c
```diff
@@ -498,6 +498,37 @@ static void do_message_timeout_linger_checks(void) {
...
+        test_conf_set(conf, "message.timeout.ms", "10000");
+        test_conf_set(conf, "transaction.timeout.ms", "7000");
+        if (!strcmp(test_conf_get(conf, "client.id"), "rdkafka"))
```
milindl, line 521: I don't think this if clause is needed for this test

tests/0004-conf.c
```diff
@@ -498,6 +498,37 @@ static void do_message_timeout_linger_checks(void) {
...
+                test_conf_set(conf, "client.id", test_curr->name);
+        rk = rd_kafka_new(RD_KAFKA_PRODUCER, conf, errstr, sizeof(errstr));
+        if (rk)
+                TEST_FAIL(
+                    "rdkafka instance created. Expected test to fail: "
```
milindl, line 526: _Expected it to fail_, it referring to the creation of the instance.

tests/0004-conf.c
```diff
@@ -498,6 +498,37 @@ static void do_message_timeout_linger_checks(void) {
...
+                    "message.timeout.ms <= transaction.timeout.ms is not "
+                    "enforced");
+        TEST_SAY("message.timeout.ms <= transaction.timeout.ms is enforced\n");
```
milindl, line 529: We don't typically add a TEST_SAY for the success case, because SUB_TEST_PASS already does some logging to this effect

tests/0004-conf.c
```diff
@@ -498,6 +498,37 @@ static void do_message_timeout_linger_checks(void) {
...
+ */
+static void do_message_timeout_transaction_timeout_checks(void) {
+        rd_kafka_conf_t *conf;
+        rd_kafka_t *rk;
+        char numstr[8];
+        char errstr[512];
+        const char *transactional_id = "txnid";
+
+        SUB_TEST_QUICK();
+        test_conf_init(&conf, NULL, 60);
```
milindl, line 513: The memory allocated for `conf` is leaked, so you should add `rd_kafka_conf_destroy(conf);` in case that test doesn't fail.

milindl (changes requested):
Thanks @jainruchir  for your PR!
Mostly minor stuff, just one significant comment about leaking memory.

PR #4180, opened 2023-02-06
Fix CMake pkg-config issue with cURL require

emasab:


dvzrv:
Please also consider https://github.com/confluentinc/librdkafka/pull/4165 and https://github.com/confluentinc/librdkafka/pull/4193 since they fix related issues

milindl (approved):
Approving based on discussion yesterday

PR #4186, opened 2023-02-10
Set socket connection setup timeout in 0105:do_test_txn_concurrent_operations

milindl:
While using create_txn_producer for any test, we set the socket.connection.setup.timeout.ms to 5s, see https://github.com/confluentinc/librdkafka/pull/4076/commits/a6a5e53c7cfb6430de7dfafb4a49fc1f8d34bc16 . 

However, for this test, we are setting the mock broker's RTT to 5s, and due to the fact that we might make the ApiVersionRequest twice (once with version 3, and then with version 0 when the first one fails)

### DIFF
--- a/tests/0105-transactions_mock.c
+++ b/tests/0105-transactions_mock.c
@@ -3533,7 +3533,14 @@ static void do_test_txn_concurrent_operations(rd_bool_t do_commit) {
 
         test_timeout_set(90);
 
-        rk = create_txn_producer(&mcluster, transactional_id, 1, NULL);
+        /* We need to override the value of socket.connection.setup.timeout.ms
+         * to be twice the RTT of the mock broker. This is because the first
+         * ApiVersion request will fail, since we make the request with v3, and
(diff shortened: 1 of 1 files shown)


tests/0105-transactions_mock.c
```diff
@@ -3533,7 +3533,14 @@ static void do_test_txn_concurrent_operations(rd_bool_t do_commit) {
 
         test_timeout_set(90);
 
-        rk = create_txn_producer(&mcluster, transactional_id, 1, NULL);
+        /* We need to override the value of socket.connection.setup.timeout.ms
+         * to be twice the RTT of the mock broker. This is because the first
+         * ApiVersion request will fail, since we make the request with v3, and
+         * the mock broker's MaxVersion is 2, so the request is retried with v0.
+         */
+        rk = create_txn_producer(&mcluster, transactional_id, 1,
```
pranavrth, line 3541: I would suggest adding some buffer time as well.

pranavrth (approved):

PR #4207, opened 2023-03-03
Fix ListGroup protocol message for ApiVersion >= 3

milindl:
Earlier, we were adding extra tagged fields when the version was a flexver. (The tagged fields are added at a common place, so this was extra)

I've run tests 0081 and 0080


milindl:
The test failure, seems very unrelated
```
main_0001_multiobj():94: full create-produce-destroy cycle: expected duration 0 <= 1409 <= 999 ms 03
```

And the windows build is failing because of the coapp issue

emasab (approved):
LGTM!

PR #4208, opened 2023-03-06
Fixed issues related to store_offset and next_offset during Barrier op in consume batch API 

pranavrth:
1. Fixes app_offset and store_offset getting updated after reading each messages. This would have resulted in incorrect app_offset and store_offset in case of Barrier op purging some messages.
2. Fixes pause and resume skipping some offsets in case of Barrier op due to the above app_offset update.


emasab (approved):
LGTM! Let's remember to add a CHANGELOG entry, I've added a TODO.

Issue #4196, opened 2023-02-19
Polling consumer queue spends ~100% cpu time in rd_clock even though queue is full


Ten0:
Description
===========
Polling consumer queue spends ~100% cpu time in rd_clock even though queue is full:

![image](https://user-images.githubusercontent.com/9094255/219905393-f7561fd1-2d0e-4948-85f3-a8c6818d4969.png)

 Each of these calls represents 1us, or about 1 000 CPU instructions per message, which is very non neglictable (Once I've got the message, processing is super fast, so I'm expecting to be able to process ~50M messages per second, not ~150000).

How to reproduce
================
I'm using this through a Rust wrapper and it's in the middle of a very complex app so I can't setup a repo at the time.
However I can tell that this
- is based on librdkafka v1.9.2
- I'm capturing `perf` with
```bash
perf record -g --call-graph=dwarf -F999
```
- We are just creating a consumer using `rd_kafka_new` then polling with `rd_kafka_poll` and `rd_kafka_consumer_poll`. ([relevant rust source code](https://github.com/fede1024/rust-rdkafka/blob/35177be201eef182a66b2bdc3af695354bcc1e8b/src/consumer/base_consumer.rs#L125-L132))

Please let me know whether that rings a bell without a repro, or whether you really need me to spend a day converting that into a repro.

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `v1.9.2`
 - [X] Apache Kafka version: `3.2.3`
 - [X] librdkafka client configuration: `auto.offset.reset=beginning, queued.max.messages.kbytes=1000000, queued.min.messages=1000000, session.timeout.ms=6000, enable.auto.commit=true, auto.commit.interval.ms=1000, enable.auto.offset.store=false, enable.partition.eof=false, statistics.interval.ms=1000`
 - [X] Operating system: `NixOS 28319deb5ab05458d9cd5c7d99e1a24ec2e8fc4b`
 - [X] Provide logs (with `debug=..` as necessary) from librdkafka: Unless I made a mistake I got no debug log at the librdkafka level
 - [X] Provide broker log excerpts: consumer queue is full (~890K messages in there at all time), so that is probably not relevant
 - [X] Critical issue (this clock thingy costs the equivalent of ~1000 CPU instructions per message, which is ridiculously high)

Ten0:
[...]

PR #4217, opened 2023-03-12
Avoid clock calls on poll when enforcing max.poll.interval.ms - 60% cpu consumption reduction on poll

Ten0:
Helps #4196.

This reworks slightly max.poll.interval.ms enforcement so that it costs basically nothing as we poll for messages, while still achieving the same constraints in the worse-case (that is, it still takes 500ms at most to realize that app hasn't polled in those 500ms.)

This has a very significant performance impact, as described at https://github.com/confluentinc/librdkafka/issues/4196.

Notably, when making calls that don't involve timeouts:

Before:
image

After:
image

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2185,8 +2185,9 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
          */
         rk = rd_calloc(1, sizeof(*rk));
 
-        rk->rk_type       = type;
-        rk->rk_ts_created = rd_clock();
+        rk->rk_type         = type;
+        rk->rk_ts_created   = rd_clock();
+        rk->rk_ts_last_poll = rk->rk_ts_created;
 
         /* Struct-copy the config object. */
         rk->rk_conf = *conf;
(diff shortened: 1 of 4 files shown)


milindl:
Hi @Ten0, thanks for this PR. I'll take a further look into this.

emasab:
Hi @Ten0, thanks! This is interesting, especially when polling a single message at a time and not using the batch consumer.
Do you see an improvement in message rate with this, from the initial 150 K message/s?

Ten0:
Hello, yes I do see an improvement.
The original code that was reaching 150k/s was also polling with timeouts, and I've reworked it so that it does not use timeouts, getting rid of all leftover clock calls, and then with this addition I get up to 500k/s (now I get CPU-capped on my multi-threaded processing part whereas before the thread reading from the consumer (there is only one for algorithmic reasons) would be capping the whole system so I expect it could go even higher in theory)

Ten0:
Heyy! Would you be able to give an ETA by any chance? We're still hitting this as a significant bottleneck.

Ten0:
Still hitting this as a bottleneck...

PR #4218, opened 2023-03-13
Fix warning for duplicate OPENSSL_SUPPRESS_DEPRECATED definition

emasab:
Use a config.h definition instead of a CFLAG


pranavrth (approved):
LGTM!

Build failing though. I triggered a Rerun for now.

Issue #4059, opened 2022-11-09
Consumers should not be able to commit during a rebalance


roxelo:
Description
===========
Now that librdkafka supports cooperative sticky partitions assignment strategy, we should ensure that consumers that commit offsets manually can’t commit offsets during rebalance as it triggers a follow up rebalance. 

I don’t think there is any valid use cases for allowing this type of behavior:
1. This behavior could lead to an infinite amount of rebalances. Rebalances will only stop once none of the consumers attempt to commit offsets during a rebalance.
2. The [Java Kafka Consumer lib](https://github.com/apache/kafka/blob/fa10e213bfea306bb5d4590dbdc5e54854192782/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L1304) does not appear to allow clients to commit offsets during a rebalance
3. [Auto commit callback in librdkafka does not commit offsets when there is an ongoing rebalance](https://github.com/edenhill/librdkafka/blob/v1.8.2/src/rdkafka_cgrp.c#L3366-L3370)

I believe this issue with manual auto commit is isolated to `cooperative sticky` strategy because when a consumer uses the `eager` strategy, a rebalance starts with all the partitions being revoked and ends when new partitions have been assigned to the consumer. As a result, the consumer will never attempt to commit offsets because there are no offsets to be committed. Of course, if the consumer uses cooperative sticky, we can’t ensure that the consumer won’t attempt to commit offsets during a rebalance as the consumer might still own partitions during a rebalance. Furthermore, clients have no way of knowing that a rebalance is ongoing or not and so they can’t prevent consumers from committing offsets when necessary.

I see three potential solutions to this problem, but I think the first one makes the most sense to implement:
1. Add the same check that exists in the auto commit callback in [rd_kafka_commit](https://github.com/edenhill/librdkafka/blob/52949110d60e36c9eae59d31956a7fd760d783fa/src/rdkafka_offset.c#L387). We can throw an error to the client to let them know why we did not attempt to commit offsets
2. Provide a new endpoint to librdkafka clients that allows them to check whether a rebalance is ongoing or not, so they are able to prevent consumers from committing offsets on their end
3. Provided the option to client (by passing a flag of some sort) to decide whether they want `rd_kafka_commit` to not attempt to commit offsets during rebalances

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

[...]

PR #4220, opened 2023-03-14
Only commit when a rebalance is not in progress (Fix for Cooperative-sticky partition assignment strategy with manual commits)

wmorgan6796:
This is an attempt to bring #4089 by @roxelo over the finish line. I believe the main difference from there is fixing the test case and bringing the PR up to date with latest master branch.

This should make cooperative-sticky work with manual commits.

See #4059


roxelo:
@edenhill Are you still interested in this fix?

wmorgan6796:
@emasab would this PR be of interest? As the sticky-cooperative partition assignment is essentially broken with manual commits, I was hoping I could get this reviewed. (@edenhill reviewed the linked PR before so I don't foresee many issues)

emasab:
@wmorgan6796 @roxelo thanks for this contribution! We're doing changes to the cooperative-sticky assignor and we can check this too while doing that.

wmorgan6796:
@emasab is there a PR/branch that can be followed?

milindl:
I had added this earlier to the changes we're making to the assignors, #4252, but stripped the changes recently as we think there are some further complexities we need to address. 

The change seems to make it impossible to use this pattern inside a user-defined rebalance callback, which is a valid use case:
```
...
case RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS:
                  rd_kafka_commit(rk, partitions, 0); // sync commit
 ...
```

(it's a valid usecase in Java as well, to call `commitSync` from the `onPartitionsRevoked` method of the `ConsumerRebalanceListener`).

Additionally, we were also thinking it's best to include two more tests, first, a test added to 0118 which checks the fact that commit should work inside a rebalance cb (if it's called before calling assign()), and a second test which can replicate the situation defined in #4059 and then work towards fixing that.

Thanks for your patience, we're discussing it further internally, to see how we can best help with it.

milindl:
Hi @wmorgan6796 , no, it still isn't figured out. We can't simply disallow committing while rebalance is happening (it's a legitimate case to commit in a sync manner inside the rebalance cb), but at the same time, the issue with illegal generation persists. Please leave this PR open as we figure out the fix, it contains the entire context, and I'll add commits to the same branch too, thanks.

jzvelc:
Hi. Any progress on this? We have also hit this issue.

wmorgan6796:
Closing this as its been fixed with #4908

PR #4228, opened 2023-03-21
Fix test 0086 often failing

emasab:
increasing linger ms and reducing second batch size to avoid sending a second batch when the first batch is purged in flight


milindl (approved):
lgtm with rationale given

milindl:
The test still fails, however, at the same place?

PR #4229, opened 2023-03-21
PR fixes

emasab:
- purge TAILQ
- remove memory leaks
- fix type warnings

### DIFF
--- a/src/rdkafka_queue.c
+++ b/src/rdkafka_queue.c
@@ -544,7 +544,7 @@ rd_kafka_purge_outdated_messages(rd_kafka_toppar_t *rktp,
                                  int32_t version,
                                  rd_kafka_message_t **rkmessages,
                                  size_t cnt,
-                                 struct rd_kafka_op_tailq ctrl_msg_q) {
+                                 struct rd_kafka_op_tailq *ctrl_msg_q) {
         size_t valid_count = 0;
         size_t i;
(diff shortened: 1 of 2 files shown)


src/rdkafka_queue.c
```diff
@@ -563,13 +563,15 @@ rd_kafka_purge_outdated_messages(rd_kafka_toppar_t *rktp,
         }
 
         /* Discard outdated control msgs ops */
-        next = TAILQ_FIRST(&ctrl_msg_q);
-        while (next && next->rko_rktp == rktp &&
-               rd_kafka_op_version_outdated(next, version)) {
+        next = TAILQ_FIRST(ctrl_msg_q);
+        while (next) {
```
pranavrth, line 567: I feel that we don't need to iterate full queue as all the higher versions are going to be at the end. But its fine this way as well just that we will process more ops.

tests/0137-barrier_batch_consume.c
```diff
@@ -87,6 +87,8 @@ static int consumer_batch_queue(void *arg) {
                 rd_kafka_message_destroy(rkmessage[i]);
         }
 
+        rd_free(rkmessage);
```
pranavrth, line 90: This function is taken from TEST 0122 as it is. We should add this in that test as well.

PR #4252, opened 2023-04-13
[consumer] Add rack-awareness (KIP-881) to assignors

milindl:
This builds on the protocol change PR to add rack awareness to the Range Assignor.

I understand that this is quite a large PR, here are the main changes:
1. `rd_kafka_range_assignor_assign_cb` and its call to `rd_kafka_topic_assignment_state_new` form the entry points of the assignment algorithm, and should probably be seen first. The code is based on the AK [RangeAssignor.java](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/RangeAssignor.java) and it is recommended to keep that open.
2. Unit test related changes, some stuff from sticky_assignor was moved to common files, because it was reused in range assignor. Earlier, rdkafka_range_assignor.c did not have any unit tests (there were some, but in rdkafka_assignor.c).
3. Unit tests from the Java range assignor. The entry point for this is `rd_kafka_range_assignor_unittest`. Almost all the tests within [RangeAssignorTest.java](https://github.com/apache/kafka/blob/trunk/clients/src/test/java/org/apache/kafka/clients/consumer/RangeAssignorTest.java) have been converted, except a few.

For sticky assignor:

1. Start with rd_kafka_sticky_assignor_assign_cb. They match, almost completely, with the Java changes to AbstractStickyAssignor.java in [this PR.](https://github.com/apache/kafka/commit/1f0ae71fb328288152a938aa2a259f9646783f60#diff-2814643946ccefef81f480e1280e9b3e3397f4673166ca1ac47db7e3cb6fb760). Additionally, a struct matching RackInfo in the Java implementation, is added.

2. There are, what I think are two pre-existing bugs in the implementation that I found while porting tests. They are marked in the commit itself as a comment (I will remove the comment after first round of reviews).

3. Unit test addition: tests missing or different from AbstractStickyAssignorTest.java are added, existing tests are parametrized to include rack cases (rack consumer, rack brokers, rack both).

4. Some refactoring of common functions across assignors and unit tests into rdkafka_assignor.h/c to avoid duplication.


milindl:
Rebased onto latest protocol changes.

ZhukovAntonPls:
It would be a great change!

src/rdkafka_metadata.c
```diff
@@ -682,8 +766,14 @@ rd_kafka_parse_Metadata(rd_kafka_broker_t *rkb,
                                 rd_kafka_metadata_cache_topic_update(
                                     rk, &md->topics[i], &mdi->topics[i],
                                     rd_false /*propagate later*/,
-                                    has_client_rack, mdi->brokers,
-                                    md->broker_cnt);
+                                    has_client_rack, /* use has_client_rack
+                                                       rather than
```
emasab, line 770: Comment style, maybe change it to:

> use has_client_rack rather than compute_racks. We need cached rack ids only in case we need to rejoin the group if they change and client.rack is set. As specified in KIP-881.

milindl: I've made this change, but the GitHub UI makes it a bit difficult to understand the intent (I've changed the text and the position slightly)

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ librdkafka v2.2.0 is a feature release:
  * Store offset commit metadata in `rd_kafka_offsets_store` (@mathispesch, #4084).
  * Fix a bug that happens when skipping tags, causing buffer underflow in
    MetadataResponse (#4278).
+ * [KIP-881](https://cwiki.apache.org/confluence/display/KAFKA/KIP-881%3A+Rack-aware+Partition+Assignment+for+Kafka+Consumers):
+   Add support for rack-aware partition assignment for consumers
+   (#4184, #4291, #4252).
```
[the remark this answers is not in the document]
milindl: Done

emasab (approved):
LGTM! Amazing job @milindl !

PR #4256, opened 2023-04-14
Add field to rd_kafka_queue which denotes if it contains fetched msgs

milindl:
An issue in v2.1.0 was fixed in which max.poll.interval.ms was not honored,
    because it was reset on any queue poll, not just consumer poll.
    It was changed it so that only certain rdkafka.h functions which were polling
    would reset the timer.
    
   However, librdkafka exposes a method rd_kafka_queue_get_consumer, which returns
    the consumer queue, and the application can poll this queue for events rather
    than calling consume poll. There is no way to distinguish polls to this queue
    and an arbitrary queue, and it won't reset the timer.
    
   So, a new field is maintained inside the queue denoting if it might
    contain fetched messages, or not. It deals with forwarding of queues, so
    if a queue which receives fetched messages is forwarded multiple times,
    calling poll on the forwardee will also reset the timer.

Note that there are a few points where someone might forward/unforward into a queue _while_ polling it. (See cases with `does_q_contain_consumer_msgs = rd_kafka_q_consumer_cnt(rkq, 1);`) which might make the timer reset wrong on these occasions. 

I don't think it matters much since that's a somewhat convoluted use case and to fix it will require taking a lock over the duration of the entire poll. But let me know what you think.

Fixes [confluentinc/confluent-kafka-go#980](https://github.com/confluentinc/confluent-kafka-go/issues/980)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ librdkafka v2.1.1 is a maintenance release:
  * Fix CMake pkg-config cURL require and use
    pkg-config `Requires.private` field (@FantasqueX, @stertingen, #4180).
  * Fixes certain cases where polling would not rejoin the consumer group
-   when the cause of leaving the group was exceeding `max.poll.interval.ms`.
-   (#4256).
+ * Fixes certain cases where polling would not keep the consumer
+   in the group or make it rejoin it (#4256).
 
 ## Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ librdkafka v2.1.1 is a maintenance release:
  * Fix CMake pkg-config cURL require and use
    pkg-config `Requires.private` field (@FantasqueX, @stertingen, #4180).
  * Fixes certain cases where polling would not rejoin the consumer group
-   when the cause of leaving the group was exceeding `max.poll.interval.ms`
+   when the cause of leaving the group was exceeding `max.poll.interval.ms`.
    (#4256).
 
 ## Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@ librdkafka v2.1.1 is a maintenance release:
    expires while waiting for the result of a list offsets operation (#4254).
  * Fix CMake pkg-config cURL require and use
    pkg-config `Requires.private` field (@FantasqueX, @stertingen, #4180).
- * Fixes certain cases where polling would not rejoin the consumer group
  * Fixes certain cases where polling would not keep the consumer
    in the group or make it rejoin it (#4256).
 
@@ -31,7 +30,7 @@ librdkafka v2.1.1 is a maintenance release:
    `max.poll.interval.ms`. Only certain functions were made to reset the timer,
    but it is possible for the user to obtain the queue with messages from
    the broker, skipping these functions. This was fixed by encoding information
-   in a queue itself, that whether polling it should reset the timer.
+   in a queue itself, that, whether polling, resets the timer.
 
 
 # librdkafka v2.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ librdkafka v2.1.1 is a maintenance release:
    expires while waiting for the result of a list offsets operation (#4254).
  * Fix CMake pkg-config cURL require and use
    pkg-config `Requires.private` field (@FantasqueX, @stertingen, #4180).
+ * Fixes certain cases where polling would not rejoin the consumer group
+   when the cause of leaving the group was exceeding `max.poll.interval.ms`
+   (#4256).
 
 ## Fixes
 
@@ -24,8 +27,11 @@ librdkafka v2.1.1 is a maintenance release:
    allowing threads different from the main one to call
    the `rd_kafka_toppar_set_fetch_state` function, given they hold
    the lock on the `rktp`.
-
-
+ * In v2.1.0, a bug was fixed which caused polling any queue to reset the
+   `max.poll.interval.ms`. Only certain functions were made to reset the timer,
+   but it is possible for the user to obtain the queue with messages from
+   the broker, skipping these functions. This was fixed by encoding information
+   in a queue itself, that whether polling it should reset the timer.
 
 # librdkafka v2.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ librdkafka v2.1.1 is a maintenance release:
    the broker, skipping these functions. This was fixed by encoding information
    in a queue itself, that whether polling it should reset the timer.
 
+
 # librdkafka v2.1.0
 
 librdkafka v2.1.0 is a feature release:
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3427,7 +3427,7 @@ rd_kafka_error_t *rd_kafka_sasl_set_credentials(rd_kafka_t *rk,
  * @returns a reference to the librdkafka consumer queue.
  * This is the queue served by rd_kafka_consumer_poll().
  *
- * Use rd_kafka_queue_destroy() to lose the reference.
+ * Use rd_kafka_queue_destroy() to loose the reference.
  *
  * @remark rd_kafka_queue_destroy() MUST be called on this queue
  *         prior to calling rd_kafka_consumer_close().
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3427,10 +3427,16 @@ rd_kafka_error_t *rd_kafka_sasl_set_credentials(rd_kafka_t *rk,
  * @returns a reference to the librdkafka consumer queue.
  * This is the queue served by rd_kafka_consumer_poll().
  *
- * Use rd_kafka_queue_destroy() to loose the reference.
+ * Use rd_kafka_queue_destroy() to lose the reference.
  *
  * @remark rd_kafka_queue_destroy() MUST be called on this queue
  *         prior to calling rd_kafka_consumer_close().
+ * @remark Polling the returned queue counts as a consumer poll, and will reset
+ *         the timer for max.poll.interval.ms. If this queue is forwarded to a
+ *         "destq", polling destq also counts as a consumer poll (this works
+ *         for any number of forwards). However, even if this queue is
+ *         unforwarded or forwarded elsewhere, polling destq will continue
+ *         to count as a consumer poll.
  */
 RD_EXPORT
 rd_kafka_queue_t *rd_kafka_queue_get_consumer(rd_kafka_t *rk);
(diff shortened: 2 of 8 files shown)


src/rdkafka_queue.c
```diff
@@ -118,6 +124,33 @@ rd_kafka_q_t *rd_kafka_q_new0(rd_kafka_t *rk, const char *func, int line) {
         return rkq;
 }
 
+/*
+ * Handles the logic for increment/decrement of rkq_consumer_cnt.
+ * Only used internally when forwarding queues.
+ * @locks rd_kafka_q_lock(rkq)
+ */
+static void rd_kafka_q_consumer_cnt_propagate(rd_kafka_q_t *rkq,
```
pranavrth, line 132: improve doc

src/rdkafka_queue.c
```diff
@@ -118,6 +124,33 @@ rd_kafka_q_t *rd_kafka_q_new0(rd_kafka_t *rk, const char *func, int line) {
...
+                                              int cnt_delta) {
+        rd_assert(rkq);
```
pranavrth, line 134: remove this

src/rdkafka_queue.c
```diff
@@ -118,6 +124,33 @@ rd_kafka_q_t *rd_kafka_q_new0(rd_kafka_t *rk, const char *func, int line) {
...
+        if (cnt_delta == 0)
+                return;
+
+        mtx_lock(&rkq->rkq_lock);
+        rkq->rkq_consumer_cnt += cnt_delta;
+        rd_assert(rkq->rkq_consumer_cnt >= 0);
```
pranavrth, line 140: Use warn log instead of assert

pranavrth (changes requested):

pranavrth (commented):
As we are locking on polling, please check performance once.

src/rdkafka_queue.c
```diff
@@ -118,6 +124,33 @@ rd_kafka_q_t *rd_kafka_q_new0(rd_kafka_t *rk, const char *func, int line) {
...
+        if (!rkq->rkq_fwdq) {
+                mtx_unlock(&rkq->rkq_lock);
+                return;
+        }
+
+        /* Recursively propagate the change in the count. There is a chance of a
+         * deadlock here but only if the user is forwarding queues in a circular
+         * manner, which already is an incorrect mode of operation.*/
+        rd_kafka_q_consumer_cnt_propagate(rkq->rkq_fwdq, cnt_delta);
+        mtx_unlock(&rkq->rkq_lock);
```
pranavrth, line 151: Shouldn't we release this lock earlier than above call? I think circular dependency case will cause a deadlock which will be very hard to identify even if it is a very rare scenario.

pranavrth (changes requested):

src/rdkafka_queue.c
milindl: That is not possible to do, I think
Consider the case that A, B are queues.. Someone calls forward(A,B). propagate(A) is called. After processing, suppose we unlock the mutex and then call propagate(B). 
Before propagate(B) starts to run, someone does forward(A, NULL) on another thread. If A->rkq_fwdq is the only reference to B, then B will get destroyed now, and we might encounter a segmentation fault on trying to access fields of B when propagate(B) finally runs.

Both these cases (circular dependency and concurrent calling of forward(A,B) and forward(A,NULL)) are rare, but while causing a circular dependency is an error case which the user should take care to avoid, concurrent forwarding is legitimate use case if someone is trying something with dynamic queue forwarding. So better to support the legitimate use case since we have to choose between them.

Changed the documentation for this.

src/rdkafka_queue.c
```diff
@@ -118,6 +124,33 @@ rd_kafka_q_t *rd_kafka_q_new0(rd_kafka_t *rk, const char *func, int line) {
...
 
+/*
+ * Handles the logic for increment/decrement of rkq_consumer_cnt.
+ * Only used internally when forwarding queues.
+ * @locks rd_kafka_q_lock(rkq)
+ */
+static void rd_kafka_q_consumer_cnt_propagate(rd_kafka_q_t *rkq,
+                                              int cnt_delta) {
+        rd_assert(rkq);
+        if (cnt_delta == 0)
```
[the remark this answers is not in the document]
milindl: Resolving this as we are moving to flag and never unsetting it.

src/rdkafka_queue.c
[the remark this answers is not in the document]
milindl: Changed name to rd_kafka_q_consumer_propagate , since we only set the flag and there is no count anymore

src/rdkafka_queue.c
milindl: Remove this statement entirely because we're now using the flag

src/rdkafka.h
```diff
@@ -3427,10 +3427,16 @@ rd_kafka_error_t *rd_kafka_sasl_set_credentials(rd_kafka_t *rk,
  * @returns a reference to the librdkafka consumer queue.
  * This is the queue served by rd_kafka_consumer_poll().
  *
- * Use rd_kafka_queue_destroy() to loose the reference.
+ * Use rd_kafka_queue_destroy() to lose the reference.
```
emasab, line 3430: Here "loose" was correct, to mean "release"

src/rdkafka_queue.h
```diff
@@ -1164,6 +1177,22 @@ rd_kafka_enq_once_disable(rd_kafka_enq_once_t *eonce) {
         return rko;
 }
 
+/**
+ * @brief Returns true if the queue can contain fetched messages.
+ *
+ * @locks rd_kafka_q_lock(rkq) if do_lock is set.
+ */
+static RD_INLINE RD_UNUSED rd_bool_t
+rd_kafka_q_can_contain_fetched_msgs(rd_kafka_q_t *rkq, rd_bool_t do_lock) {
```
pranavrth, line 1186: We can live without the lock if the `rkq` is a consumer queue because if the flag is already set, it is never unset. So if we first check the flag and it is already set, we don't need to lock the queue to get the value.

This benefits the performance as we are only concerned about the performance while consuming messages.

tests/0089-max_poll_interval.c
```diff
@@ -351,8 +351,81 @@ static void do_test_with_log_queue(void) {
...
+        test_consumer_close(rk);
+        rd_kafka_destroy(rk);
+
+        SUB_TEST_PASS();
+}
+
 int main_0089_max_poll_interval(int argc, char **argv) {
         do_test();
         do_test_with_log_queue();
+        do_test_rejoin_after_interval_expire();
```
pranavrth, line 429: Can we also write some test case for q forwarding as we are doing that as well? Maybe after the merge if it is going to take time and may affect the release.

pranavrth (changes requested):

tests/0089-max_poll_interval.c
milindl: Done

src/rdkafka.h
milindl: Done, I got confused as "lose" can also mean get rid of

src/rdkafka_queue.h
milindl: We can't do this for correctness reasons even if the flag is only set once. 

Two reasons for this is. First is that that even if the poll is always guaranteed to be called by the user after setting of the flag is complete, it might still be on different threads, and without a memory barrier (like a lock) or unless flags is marked volatile, there is no guarantee that the two threads see the same value of the variable.

Second is that, while I'm reasonably certain that writes and reads from `int` are atomic in most cases, there isn't any requirement for them to be so.

pranavrth: As discussed, First is not a problem. We need to check second one though.

pranavrth: I was suggesting something like _[Double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking)_

milindl: Resolving for now, will address later to see what we can trade-off wrt performance/correctness.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ librdkafka v2.1.1 is a maintenance release:
    calling `rd_kafka_message_leader_epoch()` on the polled `rkmessage` (#4245).
  * Fix a segmentation fault when fetching from follower and the partition lease
    expires while waiting for the result of a list offsets operation (#4254).
+ * Fixes certain cases where polling would not rejoin the consumer group
```
emasab, line 11: Not setting `polled` would also make it leave the consumer group I think.
```suggestion
 * Fixes certain cases where polling would not keep the consumer
   in the group or make it rejoin it (#4256).
```

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ librdkafka v2.1.1 is a maintenance release:
...
+   when the cause of leaving the group was exceeding `max.poll.interval.ms`.
```
emasab, line 12:
```suggestion
```

CHANGELOG.md
```diff
@@ -22,7 +24,11 @@ librdkafka v2.1.1 is a maintenance release:
    allowing threads different from the main one to call
    the `rd_kafka_toppar_set_fetch_state` function, given they hold
    the lock on the `rktp`.
-
+ * In v2.1.0, a bug was fixed which caused polling any queue to reset the
+   `max.poll.interval.ms`. Only certain functions were made to reset the timer,
+   but it is possible for the user to obtain the queue with messages from
+   the broker, skipping these functions. This was fixed by encoding information
+   in a queue itself, that whether polling it should reset the timer.
```
emasab, line 31:
```suggestion
   in a queue itself, that, whether polling, resets the timer.
```

emasab (changes requested):

tests/0089-max_poll_interval.c
```diff
@@ -351,8 +351,96 @@ static void do_test_with_log_queue(void) {
...
+        rd_kafka_destroy(rk);
+
+        SUB_TEST_PASS();
+}
+
 int main_0089_max_poll_interval(int argc, char **argv) {
         do_test();
         do_test_with_log_queue();
+        do_test_rejoin_after_interval_expire(rd_false);
+        do_test_rejoin_after_interval_expire(rd_true);
```
pranavrth, line 444: I think we are just trying to squeeze multiple scenarios in one test case. We can make it different test specific to the case. For the forward queue test, we just need to check if the flag is getting set.

milindl: Resolving, as per discussion. We can't access internals to check the flag

emasab (approved):
Great Milind, it was difficult to find the correct combination of changes that solves both problems by resetting the timer only when needed and fits with the existing code.

milindl:
Thanks Emanuele. Will merge after #4262 for CI

pranavrth (approved):
LGTM!

PR #4264, opened 2023-04-21
Continuation of #4161

emasab:
introduced offset validation position to avoid
touching the next fetch pos that has a different
purpose.


milindl (approved):
LGTM!

Issue #2879, opened 2020-05-09
Consumer fetching once a second


sachanr:
Description
===========
Currently we have a low value for queued.max.messages.kbytes for our consumers. It is set to 25. When we were on version v0.9.5 (without change: https://github.com/edenhill/librdkafka/commit/8321e37bb4715396824bb1b9a60180dfcd520c77), our consumer had high message rates for consumption. After moving to v1.3.0 (and with this backoff change), we see that the that when the toppar reaches  queued.max.messages.kbytes, it applies an immediate backoff of 1000ms. This causes the consumer to slow down to fetch from the kafka broker only once (25KB) and then wait until the next second. In the previous case, v0.9.5, it would do the opposite, fetch more aggressively to keep the queue full, more frequently than 100ms (appears so from the logs, have not checked the code for this). (100ms, being the standard fetch timeout). We have fetch.error.backoff.ms set to 0, but that does not help in this scenario. The only way out of this scenario is to increase queued.max.messages.kbytes to a value that the app can consume in 1 second to make enough data available for the app until kafka consumer goes back the next second or to remove the property all together. Having a larger queued.max.messages.kbytes would increase the memory footprint of the app but in this case where throughput is more important, it works alright for us. 

I was wondering if you had any suggestions around this please? Ideally shouldn't setting fetch.error.backoff.ms to 0 work?

Checklist
=========

Please provide the following information:

 - [x] librdkafka version (release number or git tag): v1.3.0
 - [x] Apache Kafka version: 0.10.0.0_2
 - [x] librdkafka client configuration: 
                         [fetch.error.backoff.ms] : [0]
                         [enable.partition.eof] : [true]
                         [queued.max.messages.kbytes] : [25]
                         [fetch.message.max.bytes] : [25600]
                         [debug] : [all]
                         [statistics.interval.ms] : [1000]
                         [queue.buffering.max.ms] : [1]
                         [socket.timeout.ms] : [300000]
 - [x] Operating system: RHEL 7.2
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
The general recommendation is to set `queued.max.messages.kbytes` large enough to buffer at least a second of data, preferably more, to even out message rate passed to the application in case there are network congestion, bursty producers, etc, e.g., there should always be messages for the application to be read.

shanson7:
[...]

PR #4275, opened 2023-05-09
Updated Magnus' Add 'fetch.queue.backoff.ms' to the consumer (#2879) patch

bitemyapp:
I wanted to continue the conversation here so that others can benefit from the context and what drove this.

Bottom-line up front: patch makes individual consumers go faster and consumer groups go mucho faster. Individual consumer has ~2x higher throughput with light modifications to config, consumer group scales linearly instead of sub-linearly and I can run ~7-8 consumers instead of 36 to stay ahead of my topic's production rate. Downside: uses more CPU and the default 1000 ms setting for `fetch.queue.backoff.ms` in the patch doesn't produce higher throughput than the baseline.

I have an application that consumes data from a Kafka cluster hosted in AWS. The production rate of data is ~350-550 MB/s, 550-600 MB/s being the peak during the workday. This is highly compressed, something we achieve with a heavily customized version of MirrorMaker. The decompressed data rate peaks at 5.5–6 GB/s. All data rates in bytes, not bits. My application is written in Rust and uses [rdkafka](https://github.com/fede1024/rust-rdkafka), a Rust library that uses `librdkafka`.

Running a single consumer on a single application instance, I was getting ~60 MB/s over the wire (~550 MB/s decompressed) as a baseline figure. This is in line with localhost benchmarks that were noted as hitting about 85 MB/s. However, my worker threads weren't saturated and I wasn't using my CPU core limit for the K8S deployment. I have extensive custom metrics in addition to the rdkafka statistics callback figures tracking things like the bytes processed, messages processed, and time spent in each phase of the application.

Worker thread utilization was averaging 0.6 in this single consumer scenario, meaning it was spending 60% of its time doing useful work. About ~10 microseconds per message to process the data, but it was spending around 7 microseconds waiting on `consumer.poll()` typically. CPU utilization was averaging ~7 cores out of 24 allocated. RAM was vastly over-provisioned too, it was using ~500-900 MB out of 2.5 gb provisioned.

I would prefer to not recap the entire sequence of events from the diagnostic phase of this process, so I'll try to summarize. I spent some portion of a couple of weeks investigating and digging into the metrics and librdkafka's Consumer behavior and I had the assistance of a Confluent employee for a good part of that which was very helpful. Here are some observations:

- Bytes / fetch under optimal conditions capped at 1 MB. Virtually never higher. This is related to the interaction of two constraints. Constraint one: 1 MB / fetch / partition. Constraint two: when fetching from object storage (you've lagged behind too far), you can only get one partition per fetch.
- We tried tuning and experimenting with literally every single perf-relevant Consumer configuration setting available in librdkafka ( https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md everything marked "C" )
- No setting ever increased throughput, it only changed the efficiency and consistency of the Consumer, sometimes making throughput worse.
- Changing one of the configuration settings generally increased fetch rate and reduced bytes/fetch or decreased fetch rate and increase bytes/fetch (up to 1 MB, no more). The former would generally reduce throughput and increase CPU utilization. The latter didn't change throughput much up to a point, but it made memory and CPU usage _very_ swingy and spiky. This is less efficient and makes it hard to allocate resources for the application instances efficiently.
- The optimal settings perf and efficiency wise were the defaults, but that meant we had no way to increase throughput without running more consumers on the application.
- Running more consumers did increase throughput up to a point but this came with a good deal of overhead and had the same effect as running out the buffer limits: CPU and memory usage became volatile and swingy.

My feeling, admittedly based as much on gut as on reading the code and observing the metrics, was that what we needed was a way to increase the fetch rate while keeping the fetch bytes at 1 MB. After a lot of digging into the GitHub issues and threads for `librdkafka` among other websites that had anyone discussing `librdkafka`, I found this issue: #2879 

@edenhill wrote a patch that added a fetch backoff configuration parameter named `fetch.queue.backoff.ms`. This was essentially a counterpart to `queued.max.messages.kbytes` in the sense that both are intended to enable increasing throughput. `fetch.queue.backoff.ms` can be decreased to spend CPU resources on driving more throughput and increasing `queued.max.messages.kbytes` to ensure enough data is available to consume between fetches at the expense of increased memory utilization.

We actually tried increasing `queued.max.messages.kbytes` and some other related configuration parameters and in our use-case I don't think it helped. We seemed to be limited by the 1 MB / fetch combined with the fixed fetch rate. Increasing it increased RAM usage but I don't recall getting more than a very marginal throughput limit.

I did update rdkafka a couple of times before testing this patch. It did improve throughput.

- Updating rdkafka from 0.28 to 0.29 (librdkafka 1.9.2) didn't increase decompressed throughput. ~675 MB/s
- Updating rdkafka to use the latest (HEAD) of librdkafka (~2.1.1) increase decompressed throughput from 675 MB/s to ~705-715 MB/s. ~5-7% improvement I think.

I think @emasab suggested that an improvement to a clock thing in the hot path could help, maybe that's where the improvement came from.

Then I updated (rebased, basically) @edenhill's original patch for #2879 that never got merged.

- Updating rdkafka from librdkafka 2.1.1 to 2.1.1 w/ the patch in this PR overlaid dropped throughput from ~715 MB/s to ~525 MB/s. ~27% drop.
- Thinking about it for a moment, I decided the problem was probably that the new code w/ the default `fetch.queue.backoff.ms` setting of 1000 ms was probably making the consumer run a lower fetch rate than the baseline (2.1.1 HEAD) behavior.
- Decreasing `fetch.queue.backoff.ms` to 500 and changing nothing else increased throughput to ~820 MB/s.
- Decreasing `fetch.queue.backoff.ms` to 250 decreased throughput to ~710-740 MB/s. Thought about it, figured I probably needed to increase `fetch.min.bytes` from the default of 1.
- Increased `fetch.min.bytes` to 2 MB, throughput was now ~930 MB/s.
- Settled on `fetch.min.bytes` 2 MB and `fetch.queue.backoff.ms` 150 ms, got ~1.15 GB/s with that. All defaults otherwise.

I was satisfied with this for single consumer performance as I'd nearly doubled it and I suspected that this change might solve an even bigger throughput problem that I had: consumer group sub-linear scaling.

So I was getting ~60-70 MB/s on a single consumer originally right? The problem is my throughput per consumer would drop as I added more consumers to the group and the group rebalanced. The problem seemed to be partition skew combined with fetch rates being more or less fixed. This problem exacerbates itself: as you need to run more consumers, your skewness per consumer gets worse as the partitions do not get distributed in a way that respects workload balance. You could mitigate this by running multiple consumers per application instance, but it's an imperfect solution that makes other things worse.

Our cluster has 72 brokers and the topic I'm consuming from has 512 partitions. Every consumer is effectively 74 (72 + 2) threads which lends itself to spiky CPU usage and some amount of context switching overhead. I didn't want to run more consumers than necessary.

With the baseline version of the library, even consuming from the front of the topic, if I ran 24 instances of the application (one consumer each), the consumer group would start lagging behind the ~550 MB/s (compressed) / 5.5 GB/s (decompressed) production rate. The balance point was about 30, I ended up running 36 to give it breathing room to catch up on the offset lag.

550 / 60 is 9.16, not 36. With the scaled up consumer group I was averaging 14-20 MB/s/consumer. This is 33% of what a single consumer got.

With this `fetch.queue.backoff.ms` patch, here is what I observed in scaling up the replicas of the deployment:

1: 1.15 GB/s (decompressed) per instance
2: 1.05 GB/s per instance
3: 900 MB/s per instance
6: 900 MB/s per instance
10: 900 MB/s per instance

After a small falloff from 1 to 3 consumers, I was getting almost _exactly_ linear scaling from 3 to 10 consumers. This is with the 150 ms / 2 MB min fetch parameters I mentioned earlier. I believe the scaling was better behaved due to the combination of more partitions per consumer due to fewer consumers and the higher fetch rate per partition.

Running fewer consumers like this should create less load on the brokers although that isn't a problem in our case, at 2.6 GB/s (compressed, believe it or not) consumption rate we were at 15% load on the cluster.

If I was to potentially change one thing in this patch, I think I'd want to consider changing the default `fetch.queue.backoff.ms` to a somewhat lower figure, maybe around 600 - 750 ms, so that users don't get hit with the lower data rate / lower resource utilization behavioral profile as contrasted with baseline. However, this is what I observed with _my_ application, it's possible the fetching behavior could manifest differently with differently shaped clusters, partitioning, and traffic patterns.

@emasab asked some questions and I may have inadvertently answered them above but I wanted to ensure they were directly addressed:

>if these tests are with or without load

If I understood your question correctly, with load. We aren't tossing the data aside that we poll from the consumer. We _did_ do this much much earlier on in the production testing process but that was weeks ago. We're doing most of the production workload and as far as the worker threads polling the consumer are concerned, this is 90% of the production workload. The remaining pieces are a cheap look-up and update in a data structure (less than 500 nanos, definitely less than 1 microsecond) and in some cases shoving an update event into an mpsc channel (~350-600 nanos). The rest of the work is done in a background thread. The delta between the current production workload and the final one is ~1-2 microseconds per message whereas the current processing time averages 8-10 microseconds per message.

>the consumer configuration properties

Skipping all the testing on the baseline rdkafka we did which never improved throughput, I'd say the point of comparison is between all defaults on baseline vs. a lightly tuned config w/ this patch. 150 ms `fetch.queue.backoff.ms` and 2 MB `fetch.min.bytes`. All other config was not perf relevant. Auto commit true, auto offset true, `enable.partition.eof` false, `group.id`, sasl username and password, `statistics.interval.ms` `30000`, yadda yadda you get the idea.

>the avg message size

~1.1-1.3 KB in production. Average time to process message bounces between 7.5 and 9 microseconds recently. It takes us ~7.1 microseconds per kilobyte of data.

>That because when testing locally without load you can get more than 2 M msg/s, but with very small messages.

Message processing rate started at 433k/s yesterday and increased to ~1 million per second when I arrived at the 150 ms / 2 MB configuration w/ this patch yesterday.

>With larger ones it's limited by `queued.max.messages.kbytes` and that backoff. In this case if you're consuming more than queued.max.messages.kbytes / s and the queue is emptied in less than 1 second. But you're also fetching at more than the consume rate so the queue is filled when it starts again. Probably this could be solved also by increasing queued.max.messages.kbytes to more than the consume rate, but it'll require more memory, and can be trade off with more frequent backoff when the queue fills up.

I mentioned this earlier I think but I'll reiterate here for reader's convenience: increasing `queued.max.messages.kbytes` increased memory usage and made resource utilization more volatile (CPU and RAM). Throughput never increased from baseline defaults more than a little (5-15%) and at great cost.

Remaining issue I'm noticing: the CPU usage is still kinda spiky, currently the behavior is that I have 6 cores more or less constantly burning in the worker threads processing the messages. The problem is the library jumps between 0.5 cores and 6-10 cores pulling data down all at once. I'd prefer it stay at a steady 3-5'ish cores and stay ahead of my throughput needs. Increase buffers does not even this out, it just makes bigger fetch spikes and increases resource utilization. If you have any suggestions for evening out the fetch spikes I'd happily pursue it in another patch.

Please let me know if you have any other questions. I'm happy to help get this patch integrated into the library if there's anything I can do.


emasab:
Thanks @bitemyapp ! We'll address this to avoid having to increase too much `queued.max.messages.kbytes`. What was the max value of that property during your tests? Because real improvement happens when the queue isn't completely emptied during  `fetch.queue.backoff.ms`. Given your fetch rate it should be at least 1 GiB, if possible try the max value of 2 GiB. It's a single queue, not per partition, when using the high-level consumer.

bitemyapp:
>What was the max value of that property during your tests?

Based on our git history (gitops, all our deployment/config changes are in the app repo), it looks like we tried the default, 130 MB, and 2 GB. Here's a snippet from the git log:

```
- name: "LIBRDKAFKA_VAR_QUEUED_MAX_MESSAGES_KBYTES"
    value: "2097151" # little or no impact yet      
```

However, I can only find instances where it's commented out. I'm quite sure @abraham-leal and I tried 2gb but I think I rebase-squashed some of the history out of existence. I remember it not making much difference at the time like the comment says. Since I believe we're both curious about how much improvement it could make in the current state of play, I will drop the deployment to a single replica, back off the fetch rate to the default of 1000 ms, and set `queued.max.messages.kbytes` to 2 GB again. I will try rolling fetch min bytes back to the minimum just to see what it does but try setting it to 1 MB after that too.

Is there a chance that the older version of rdkafka (0.28) that I was testing with @abraham-leal was using a version of librdkafka that didn't have the correct behavior WRT the configuration settings for the consumer? https://github.com/fede1024/rust-rdkafka/tree/v0.28.0 looks like it has https://github.com/benesch/librdkafka/tree/30633e59a9072b7b9c76411c281578e0125d5f65 which is `v1.8.2`.

>It's a single queue, not per partition, when using the high-level consumer.

Yeah this is all high-level consumer because I didn't want to putz around with the re-balancing. I'm still hoping to find time to test manual rebalancing + splitting out partition queues but it wasn't something I wanted to be forced into at all.

bitemyapp:
Okay so with:

- the librdkafka patch in this PR
- queued.max.messages.kbytes at 2 GB
- everything else default
- Increased RAM allocation from 2.5 GB to 5 GB due to OOM.

I'm getting:

- ~730 MB/s (decompressed), baseline for this patch was 525 MB/s
- ~70-80 MB/s (compressed) over the wire
- Efficiency/utilization factor: 0.65 initially, then settled down to ~0.52
- ~540k msgs/s
- Weird P95 message wait times of ~50-100 microseconds where it's ~4.5 microseconds when it first started Cf.

Screenshot 2023-05-10 at 10 19 54

It hasn't done it since, but it is settling at a higher figure of ~8.5 microseconds now.
- CPU utilization ~11-12 cores nominally but there's weird spikes and the CPU "total" looks weird even though it's one pod replica. I think this is the usual early pre-loading spike. After a bit it settled at 10.5 cores.
- Periodic CPU throttling even though it's got 24 cores allocated.
- Data rate efficiency w/ the higher fetch rate and default message queue size was 80 MB/s/core. It's getting ~27-29MB/s/core right now. 80 MB/s/core efficiency is pretty normal and what it's been when it wasn't churning pods or running a degenerate configuration for the last month. Cf.

Screenshot 2023-05-10 at 10 28 10

The 36 replica deployment was only slightly less efficient client-side, bouncing between ~74 and 80 MB/s/core. Makes a bigger difference broker-side I would imagine but you'd know better than I.

- Bytes/fetch went from 676 kb to a peak of 4.76 MB/fetch and is now trending down from that. I might be about to lag out of instance storage into object storage, not sure.

Screenshot 2023-05-10 at 10 33 13

It's at 1.65 MB in the latest timestamp of that screenshot and settling downward. Lag's at 26 minutes.

- RAM spiked early to 4.69 GB, it's ping-ponging between 425 MB and 1.2 GB now.

CPU throttling stopped as it settled. Data rate efficiency is oscillating between 47 MB and 115 MB/s/core. EWMA_20 (moving average) is ~67-76. Periodicity of the data rate efficiency oscillation is ~20 seconds peak to trough, ~40 seconds trough to peak.

Screenshot 2023-05-10 at 10 39 00

Efficiency factor seems settled at 0.52. I haven't digested these results yet but I'm not sure I'd want to continue running with this config because of how volatile it is. I'm going to test increasing the fetch rate next since fetch min bytes isn't an issue.

bitemyapp:
New config:

- fetch backoff: 250 ms
- queued max messages kbytes: 1 GB
- all else default

Results:

- ~830 MB/s (decompressed), ~740k messages/second
- efficiency factor ~0.60 - 0.72
- 42-91 MB/s/core data rate efficiency, same peaky behavior. EWMA 60 MB/s/core
- Bytes/fetch 1.05 MB, basically the same.
- 10-16 cores, total average 12-13
- 450 MB - 1.2 GB RAM

After that I tried 150 ms fetch backoff, 512MB queued max messages kbytes, boring results broadly comparable to the previous test. Slightly higher efficiency factor. ~95 MB/s (compressed) over the wire. ~890 MB/s (decompressed). CPU usage 7-23 cores but no throttling. RAM 600 - 1.2 GB.

bitemyapp:
I think the wistful desire from an earlier comment still remains. I'd love a way to even out the fetch rate so it isn't quite so bursty and contentious.

bitemyapp:
Config: 10 replicas, reduced worker count per instance from 12 to 10, 150 ms backoff, 1 MB fetch min bytes, 1.4 GB queued max messages kbytes.

Result:
- 8.8 GB/s (decompressed), ~7 million messages/second
- ~2 GB/s (compressed) over the wire and settling down to 950 MB/s (compressed)
- Utilization factor is higher, ~0.78-0.80, probably due to lower worker count
- ~12 cores average used per instance, ~6-18 cores oscillation
- 7 gb of RAM used fleetwide, averaging ~700 MB. It varies per pod from 400 MB to 1.2 GB.

Example of what I mean about fetch rate:

Screenshot 2023-05-10 at 12 10 28

Per partition lag

Screenshot 2023-05-10 at 12 14 11

I did have a couple 12 GB/s spikes at the beginning followed by throttling that caused one of the deployments to get murked by the K8S cluster for some reason. It settled on the second attempt.

emasab:
Hello @bitemyapp. Thanks for all your tests and detailed reports. Checking related code I think you're limited not only by `queued.max.messages.kbytes` but also by `queued.min.messages` that stops fetching if that number of messages are in queue and the backoff corresponds to `rd_kafka_max_block_ms=1000` too. And this PR configures the backoff in that case too.
The default is 100000 and with your average msg size corresponds roughly to 133 MB.

```
        } else if (rd_kafka_q_len(rktp->rktp_fetchq) >=
                   rkb->rkb_rk->rk_conf.queued_min_msgs) {
                /* Skip toppars who's local message queue is already above
                 * the lower threshold. */
                reason       = "queued.min.messages exceeded";
                should_fetch = 0;

        } else if ((int64_t)rd_kafka_q_size(rktp->rktp_fetchq) >=
``` 

Could you do just a test by setting

-  queued.max.messages.kbytes: 2097151
-  queued.min.messages: 10000000

with the default fetch backoff of 1000 ms or with a lower one of 150 ms.

bitemyapp:
Config:

- Single replica to put it in sink or swim mode
- Fetch backoff 1000 ms
- queued.max.messages.kbytes: 2097151
- queued.min.messages: 10000000

Results:

1000 ms fetch backoff baseline was 550 MB/s without the increased queue min.

- ~830 MB/s (decompressed)
- ~670k messages/s
- ~80-90 MB/s (compressed) over the wire
- Worker utilization factor: ~0.74
- Data rate efficiency (CPU efficiency): ~83MB/s/core <--- this is a differentiator for increasing queue allocation, a CPU efficiency vs. RAM tradeoff. I am going to add a metric for memory-efficiency.
- Data rate per gigabyte of RAM (memory efficiency): before the pod got popped it was 115 MB/s/gb RAM. Previous tests w/ different configs yielded 200 - 800 MB/s/gb RAM. Usually 150 - 250 MB/s/GB prior to the fetch rate patch. Closer to 300-450 MB/s/GB when I started pushing fetch rate. Much more play in the joints here than data rate efficiency for CPU.

Resource util:

- CPU averages 11 cores but oscillates between 5 and 18
- ~6.4 GB-6.7 GB of RAM used, this seems to be in excess of what it should nominally be so I'm not sure what the overhead is. My app by itself doesn't use hardly any RAM when it isn't stressed. QA has a low data rate and is using 237 MB on average, same exact config and code. I auto-deploy QA and Prod in lock-step.

I have 10 GB allocated to the pod, while I was recording observations the original pod got OOM'd and replaced due to a memory alloc spike. This more or less never happens when the queue is smaller.

Resource utilization in general but esp. CPU is still very spiky and uneven. I think efficiency and stability could be greatly improved by figuring out how to pace/meter fetch requests so they don't all happen at once. Is this behavior I could potentially induce by splitting out the partition queues instead of using the high level consumer interface?

Screenshot 2023-05-11 at 11 43 51

Screenshot 2023-05-11 at 11 44 00

Gonna push fetch rate backoff down to 500 ms now.

bitemyapp:
>Gonna push fetch rate backoff down to 500 ms now.

Samesies as previous test really. Worker utilization very _slightly_ higher, similar CPU usage, similar throughput, same memory usage albeit a little more even. Same memory efficiency for data rate: 115 MB/s/gb RAM. CPU data rate efficiency moves around a little more, 80-89 MB/s/core. Workers are waiting an average of ~2.6 microseconds per message polled from the consumer.

Seems like increasing buffer and increasing fetch rate are two paths to roughly the same outcome with the main difference being CPU/memory trade-offs and utilization volatility. I was hoping padding out the queue would smooth the fetches out but it hasn't. I will say that this behavior is more consistent and less volatile with a similar configuration than baseline by a fair bit. I don't favor the higher memory usage profiles still.

Anything else you'd like me to try or investigate? Padding out the queue so far has increased throughput up to a point and you can go further driving fetch rate albeit at the expense of CPU. I'm going to try pushing fetch rate to what it was at the high watermark (when I was getting 1.2 GB/s on one consumer) and see if it's any different with the chunky buffer.

bitemyapp:
Pushed fetch backoff down to 100 ms w/ the bigqueue config we've been testing today, no real improvement if not a little worse overall. I'm going to change it back to a smaller queue and moderate fetch rate until you have something you'd like me to try.

emasab:
Very good, I asked for these last tests to verify if problem could be solved in a different way in previous and current version, not that I wanted to suggest you to increase memory so much. In your case the best option is to decrease the fetch backoff. I'm going to review and merge this PR.

PR #4276, opened 2023-05-10
Add small note for installation on Apple Silicone

rizkyarlin:
We code mainly in Python, we don't usually use docker in development, especially for a small codebase. I believe this is also the case for a lot of Pythonistas. 

This small text will help some people out there save a lot of time figuring out what went wrong since we just follow the current instruction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ On Mac OSX, install librdkafka with homebrew:
 $ brew install librdkafka
 ```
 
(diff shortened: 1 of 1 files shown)


emasab:
Hello, why do you  need this if librdkafka comes prepackaged in confluent-kafka wheel now?

rizkyarlin:
> Hello, why do you need this if librdkafka comes prepackaged in confluent-kafka wheel (OSX/x64 and OSX/arm64) now?

That wasn't the case for me. I'm running on virtualenv, python3.7, and confluent-kafka2.1.0. But I don't think I have time to reproduce the issue right now. So I'm closing the PR, thanks!

PR #4278, opened 2023-05-11
Fix number or bytes skipped when receiving tags.

emasab:
Not being for a nullable field those unsigned varints don't start from 1


milindl (approved):
Commit as : Fix number *of* bytes skipped when receiving tags

PR #4279, opened 2023-05-12
Metadata refactor to add internal fields

emasab:
The additional memory is allocated after the public struct and freed together with it, along with nested array memory, through the usage of `rd_tmpabuf_t`.

`mdit->brokers` are sorted by `mdit->brokers[i].id`
`mdit->topics[i].partitions` are sorted by `mdit->topics[i].partitions[j].id`.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -990,7 +990,7 @@ void rd_kafka_destroy_final(rd_kafka_t *rk) {
         mtx_destroy(&rk->rk_init_lock);
 
         if (rk->rk_full_metadata)
-                rd_kafka_metadata_destroy(rk->rk_full_metadata);
+                rd_kafka_metadata_destroy(&rk->rk_full_metadata->metadata);
         rd_kafkap_str_destroy(rk->rk_client_id);
         rd_kafkap_str_destroy(rk->rk_group_id);
         rd_kafkap_str_destroy(rk->rk_eos.transactional_id);
--- a/src/rdkafka_int.h
+++ b/src/rdkafka_int.h
@@ -353,7 +353,7 @@ struct rd_kafka_s {
 
         rd_kafka_metadata_internal_t
             *rk_full_metadata;       /* Last full metadata. */
-        rd_ts_t rk_ts_full_metadata; /* Timesstamp of .. */
+        rd_ts_t rk_ts_full_metadata; /* Timestamp of .. */
         struct rd_kafka_metadata_cache rk_metadata_cache; /* Metadata cache */
 
         char *rk_clusterid;      /* ClusterId from metadata */
--- a/src/rdkafka_int.h
+++ b/src/rdkafka_int.h
@@ -78,7 +78,8 @@ struct rd_kafka_topic_s;
 struct rd_kafka_msg_s;
 struct rd_kafka_broker_s;
 struct rd_kafka_toppar_s;
-
+typedef struct rd_kafka_metadata_internal_s rd_kafka_metadata_internal_t;
+typedef struct rd_kafka_toppar_s rd_kafka_toppar_t;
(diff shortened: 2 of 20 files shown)


src/rdkafka_int.h
```diff
@@ -350,8 +351,9 @@ struct rd_kafka_s {
         rd_ts_t rk_ts_metadata; /* Timestamp of most recent
                                  * metadata. */
 
-        struct rd_kafka_metadata *rk_full_metadata; /* Last full metadata. */
-        rd_ts_t rk_ts_full_metadata;                /* Timesstamp of .. */
+        rd_kafka_metadata_internal_t
+            *rk_full_metadata;       /* Last full metadata. */
+        rd_ts_t rk_ts_full_metadata; /* Timesstamp of .. */
```
milindl, line 356: nit: correct spelling of timestamp since there's a diff anyway

src/rdkafka_metadata.h
```diff
@@ -31,17 +31,58 @@
...
+ */
+typedef struct rd_kafka_metadata_topic_internal_s {
+        /** Internal metadata partition structs.
+         *  same count as metadata.topics[i].partition_cnt.
+         *  Sorted by Partition Id. */
+        rd_kafka_metadata_partition_internal_t *partitions;
+} rd_kafka_metadata_topic_internal_t;
+
+
+typedef struct rd_kafka_metadata_broker_internal_s {
```
milindl, line 55: nit: Documentation/comment for this struct and internal fields

src/rdkafka_range_assignor.c
```diff
@@ -55,12 +55,11 @@ rd_kafka_resp_err_t rd_kafka_range_assignor_assign_cb(
     const rd_kafka_assignor_t *rkas,
     const char *member_id,
     const rd_kafka_metadata_t *metadata,
+    const rd_kafka_metadata_internal_t *metadata_internal,
```
milindl, line 58: Question for here and elsewhere, why pass both metadata and metadata_internal? Doesn't it suffice to pass just metadata_internal, as it contains a reference to metadata?

Also, another question: wouldn't it make sense (from a pointer casting point of view) for partition_internal to have a reference to partition, and broker_internal to have a reference to broker (rather than the Ids),The sorting, if required, could still work based on the Ids, since they are still accessible.

If we do that, we only need to pass _internal fields around in librdkafka private code. 
One drawback may be, this change will become larger since everywhere we will need an extra "->".

And same for topic_internal and topic - though I notice that internal_topic does not have any way to map it to topic yet.

milindl (commented):
Thanks for making this change!

src/rdkafka_range_assignor.c
emasab: I was thinking to change this, also because there are requests for a custom assignor and we can't pass rd_kafka_metadata_internal_t there. Instead it could use accessor functions to get required data.

milindl: Oh, right, that would make it way more difficult anyway. If we follow this comment and mark it as experimental: https://github.com/confluentinc/librdkafka/issues/2284#issuecomment-774961322
But then, will you have a typedef for rd_kafka_metadata_internal_t in rdkafka.h along with all accessor methods, and treat it like any other private struct?

emasab: We keep `rd_kafka_metadata_internal_t `, we only have accessor methods in `rdkafka.h`. For rack_id for example.

src/rdkafka.c
```diff
@@ -990,7 +990,7 @@ void rd_kafka_destroy_final(rd_kafka_t *rk) {
         mtx_destroy(&rk->rk_init_lock);
 
         if (rk->rk_full_metadata)
```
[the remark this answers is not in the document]
emasab: They point to the same address

PR #4284, opened 2023-05-16
Updated Magnus' Add 'fetch.queue.backoff.ms' to the consumer (#2879) patch

emasab:
continuation of #4275 from a local branch.


tests/0127-fetch_queue_backoff.cpp
```diff
@@ -0,0 +1,154 @@
...
+  std::vector<RdKafka::TopicPartition *> parts;
+  parts.push_back(RdKafka::TopicPartition::create(topic, 0));
+
+  RdKafka::ErrorCode err;
+  if ((err = c->assign(parts)))
+    Test::Fail("assigned failed: " + RdKafka::err2str(err));
+
+  int received       = 0;
+  int in_profile_cnt = 0;
+  int dmax           = (int)(/*max overhead*/ 100.0 + (double)backoff_ms * 1.2);
```
milindl, line 85: It's probably good to multiply the max overhead with the `test_timeout_multiplier` to prevent flakiness in CI or on other machines.

tests/0127-fetch_queue_backoff.cpp
```diff
@@ -0,0 +1,154 @@
...
+  while (received < 5) {
+    RdKafka::Message *msg = c->consume(3000 + backoff_ms);
+
+    rd_ts_t now     = test_clock();
+    int latency     = (test_clock() - ts_consume) / 1000;
+    ts_consume      = now;
+    bool in_profile = latency <= dmax;
+
+    if (!msg)
+      Test::Fail(tostr() << "No message for " << (1000 + backoff_ms) << "ms");
```
milindl, line 98: The message seems incorrect with the `consume` timeout above, right?

tests/0127-fetch_queue_backoff.cpp
```diff
@@ -0,0 +1,154 @@
...
+    Test::Fail("assigned failed: " + RdKafka::err2str(err));
+
+  int received       = 0;
+  int in_profile_cnt = 0;
+  int dmax           = (int)(/*max overhead*/ 100.0 + (double)backoff_ms * 1.2);
+
+  int64_t ts_consume = test_clock();
+
+  while (received < 5) {
+    RdKafka::Message *msg = c->consume(3000 + backoff_ms);
```
milindl, line 90: Comment for why we're using 3000 + x as the consume timeout

is it to make sure we get a message (even if there's some delay)?

tests/0127-fetch_queue_backoff.cpp
```diff
@@ -0,0 +1,154 @@
...
+    delete msg;
+  }
+
+  Test::Say(tostr() << in_profile_cnt << "/" << received << " messages were "
+                    << "in profile (<= " << dmax
+                    << ") for backoff_ms=" << backoff_ms << "\n");
+
+  /* first message isn't counted*/
+  const int expected_in_profile = received - 1;
+  TEST_ASSERT(expected_in_profile - in_profile_cnt >= (test_on_ci ? 1 : 0),
```
milindl, line 120: Adding the timeout to the overhead should ideally fix this, right? No need for `test_on_ci ? 1 : 0` ?

tests/0127-fetch_queue_backoff.cpp
```diff
@@ -0,0 +1,154 @@
...
+  RdKafka::Conf *conf;
+  Test::conf_init(&conf, NULL, 60);
+  Test::conf_set(conf, "group.id", topic);
+  Test::conf_set(conf, "enable.auto.commit", "false");
+  Test::conf_set(conf, "auto.offset.reset", "beginning");
+  Test::conf_set(conf, "queued.min.messages", "1");
+  if (backoff_ms >= 0) {
+    Test::conf_set(conf, "fetch.queue.backoff.ms", tostr() << backoff_ms);
+  }
+  Test::conf_set(conf, "fetch.message.max.bytes", "12000");
```
milindl, line 63: If possible, a comment here for why this is 12000

I think it's because the msgsize is 10000 so this effectively makes sure that the queue only has one msg at a time, right?

tests/0127-fetch_queue_backoff.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/**
+ * Test consumer fetch.queue.backoff.ms behaviour.
+ *
+ * @param backoff_ms Backoff ms to configure, -1 to rely on default one.
+ *
+ * 1. Produce N messages, 1 message per batch.
+ * 2. Configure consumer with queued.min.messages=1 and
+ *    fetch.queue.backoff.ms=A|B.
```
milindl, line 45: What's meant by A|B?

milindl (commented):
Mostly, it looks good to me, just some comments on the test

tests/0127-fetch_queue_backoff.cpp
emasab: Not sure, because that should be like  `backoff_ms <= dmax < 2 * backoff_ms`. I could change it to 
```c
  int dmax           = (int)((double)backoff_ms * (test_timeout_multiplier > 1 ? 1.5 : 1.2));
```

tests/0127-fetch_queue_backoff.cpp
emasab: Yeah, also the condition should have been `<=`. Changed it.

tests/0127-fetch_queue_backoff.cpp
milindl: Ah, yep, that makes sense

milindl (approved):
LGTM now!

PR #4298, opened 2023-05-27
Disable Nagle algorithm by default

aivarsk:
Set TCP_NODELAY by default. That is what most libraries do by default as well (HTTP clients and servers, database connections, etc.) This library has high-level buffering through queue.buffering.max.ms, I don't see a need for it on the socket level.
Having Nagle algorithm on by default leads to poor performance and it's not apparent or documented that this setting should be changed, performance tuning tips on the internet do not mention it.


emasab:
@aivarsk thanks for this PR

emasab (approved):

emasab:
merged with https://github.com/confluentinc/librdkafka/pull/4986

PR #4306, opened 2023-06-05
Scram Config API in Admin Client [KIP-554] <- various changes

emasab:
I've included a set of commits with some changes, mainly code style, naming.
There's also the support for COMPACT BYTES, as passing a string would exclude the NULL byte from being used as salt or password.
Other things:

- random initialization if salt isn't provided
- Doxygen style documentation
- ifdefs to configure without OpenSSL
- always return an error using the queue for async functions
- use const when necessary
- allow multiple upserts and deletes for the same user as in Java. For example two upserts with the same user and SHA-256 or SHA-512 as scram mechanism


tests/0081-admin.c
```diff
@@ -3652,239 +3652,335 @@ static void do_test_ListConsumerGroupOffsets(const char *what,
...
+                            num_credentials);
         }
         rd_kafka_event_destroy(event);
 
+        /* Create a credential for user 0 */
+
         rd_kafka_UserScramCredentialAlteration_t *alterations[1];
-        char *salt = "salt";
-        char *password = "password";
+        char *salt                          = "salt";
```
milindl, line 3734: we should change the testcase to deliberately have a zero-byte between the salt, that way, we get an extra check on the serialization to bytes as well.

tests/0081-admin.c
```diff
@@ -3652,239 +3652,335 @@ static void do_test_ListConsumerGroupOffsets(const char *what,
...
+                    "Expected NO_ERROR, not %s", rd_kafka_err2name(err));
+
+        alter_result = rd_kafka_event_AlterUserScramCredentials_result(event);
+        alter_responses = rd_kafka_AlterUserScramCredentials_result_responses(
+            alter_result, &response_cnt);
+        /* response_cnt should be 1*/
+        TEST_ASSERT(response_cnt == 1,
+                    "There should be exactly 1 response, got %" PRIusz,
+                    response_cnt);
+        for (i = 0; i < response_cnt; i++) {
```
milindl, line 3773: nit: For loop is redundant if we're asserting `response_cnt == 1`, it's clearer to just use [0] and proceed with that.

milindl: Applicable here and at other places

tests/0081-admin.c
```diff
@@ -3652,239 +3652,335 @@ static void do_test_ListConsumerGroupOffsets(const char *what,
...
+                            "Username should be %s, got %s", users[0],
+                            username);
+                TEST_ASSERT(err == RD_KAFKA_RESP_ERR_NO_ERROR,
+                            "Error code should be NO_ERROR, got %s",
+                            rd_kafka_err2name(err));
         }
+        rd_kafka_event_destroy(event);
 
-        options = rd_kafka_AdminOptions_new(rk, RD_KAFKA_ADMIN_OP_DESCRIBEUSERSCRAMCREDENTIALS);  
+#if !WITH_SSL
```
milindl, line 3925: nit: not needed to wrap this in ifdef, since having the label here isn't a problem for the non-ifdef code also.

tests/0081-admin.c
```diff
@@ -4009,7 +4105,9 @@ static void do_test_apis(rd_kafka_type_t cltype) {
                     rd_true /*with subscribing consumer*/, rd_true);
         }
 
-        do_test_ScramConfigAdmin(rk,mainq);
+        if (test_broker_version >= TEST_BRKVER(2, 7, 0, 0)) {
+                do_test_UserScramCredentials(rk, mainq);
```
milindl, line 4109: For parity with other tests add test with NULL queue (using temp queue), and a "what" field in the same way.

milindl (changes requested):
Thank you for this change.

tests/0081-admin.c
emasab: there is a warning in that case

src/rdkafka.h
```diff
@@ -5446,10 +5363,10 @@ typedef int rd_kafka_event_type_t;
 #define RD_KAFKA_EVENT_LISTCONSUMERGROUPOFFSETS_RESULT 0x8000
 /** AlterConsumerGroupOffsets_result_t */
 #define RD_KAFKA_EVENT_ALTERCONSUMERGROUPOFFSETS_RESULT 0x10000
-/* DescribeUserScramCredentials_result_t */
-#define RD_KAFKA_EVENT_DESCRIBEUSERSCRAMCREDENTIALS_RESULT 0x20000
-/* AlterUserScramCredentials_result_t */
-#define RD_KAFKA_EVENT_ALTERUSERSCRAMCREDENTIALS_RESULT 0x40000
+/** DescribeUserScramCredentials_result_t */
+#define RD_KAFKA_EVENT_DESCRIBEUSERSCRAMCREDENTIALS_RESULT 0x40000
```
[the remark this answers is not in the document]
emasab: It was taken from INCREMENTALALTERCONFIGS

src/rdkafka_admin.c
```diff
@@ -5300,63 +5456,116 @@ rd_kafka_resp_err_t rd_kafka_AlterUserScramCredentialsResponse_parse(rd_kafka_op
...
+            rd_kafka_AlterUserScramCredentialsRequest,
+            rd_kafka_AlterUserScramCredentialsResponse_parse,
+        };
+
+        rko = rd_kafka_admin_request_op_new(
+            rk, RD_KAFKA_OP_ALTERUSERSCRAMCREDENTIALS,
+            RD_KAFKA_EVENT_ALTERUSERSCRAMCREDENTIALS_RESULT, &cbs, options,
+            rkqu->rkqu_q);
+
+        if (alteration_cnt > 0) {
```
milindl, line 5487: We should throw a local Invalid Argument error in case alteration_cnt == 0. What do you think?

emasab: yeah, without letting if fail on the broker

milindl (approved):

PR #4307, opened 2023-06-06
Dev kip235

anchitj:


src/rdkafka_broker.c
```diff
@@ -5238,20 +5267,36 @@ int rd_kafka_brokers_add0(rd_kafka_t *rk, const char *brokerlist) {
...
-
-                /* If rd_kafka_broker_find returned a broker its
-                 * reference needs to be released
-                 * See issue #193 */
-                if (rkb)
-                        rd_kafka_broker_destroy(rkb);
+                        RD_SOCKADDR_LIST_FOREACH(sinx, sockaddrList) {
+                                resolvedFQDN = rd_sockaddr2str(
+                                    sinx, RD_SOCKADDR2STR_F_RESOLVE);
+                                rd_kafka_find_or_add_broker(
```
pranavrth, line 5291: Can we add a debug statement here as well for resolvedFQDN?

src/rdkafka_broker.c
```diff
@@ -5219,13 +5245,16 @@ int rd_kafka_brokers_add0(rd_kafka_t *rk, const char *brokerlist) {
         char *s_copy = rd_strdup(brokerlist);
         char *s      = s_copy;
         int cnt      = 0;
-        rd_kafka_broker_t *rkb;
-        int pre_cnt = rd_atomic32_get(&rk->rk_broker_cnt);
+        int pre_cnt  = rd_atomic32_get(&rk->rk_broker_cnt);
+        rd_sockaddr_inx_t *sinx;
+        rd_sockaddr_list_t *sockaddrList;
```
pranavrth, line 5250: lets not use camelCase here. Better to use same convention which is being used in librdkafka. `sockaddr_list`

src/rdkafka_broker.c
```diff
@@ -5219,13 +5245,16 @@ int rd_kafka_brokers_add0(rd_kafka_t *rk, const char *brokerlist) {
...
 
         /* Parse comma-separated list of brokers. */
         while (*s) {
                 uint16_t port;
                 const char *host;
+                const char *errstr;
+                const char *resolvedFQDN;
```
pranavrth, line 5257: same. use snake_case.

src/rdkafka_broker.c
```diff
@@ -5238,20 +5267,36 @@ int rd_kafka_brokers_add0(rd_kafka_t *rk, const char *brokerlist) {
                         break;
 
                 rd_kafka_wrlock(rk);
+                if (rk->rk_conf.enable_bootstrap_servers_canonical_resolve) {
+                        rd_kafka_dbg(rk, ALL, "INIT",
+                                     "Canonicalizing bootstrap broker %s:%d",
+                                     host, port);
+                        sockaddrList = rd_getaddrinfo(
+                            host, RD_KAFKA_PORT_STR, AI_ADDRCONFIG,
```
pranavrth, line 5275: Don't we need to use `AI_CANONNAME` flag here? I think we are only concerned about canonical host name. Other should be discarded. https://man7.org/linux/man-pages/man3/getaddrinfo.3.html

src/rdkafka_broker.c
```diff
@@ -5238,20 +5267,36 @@ int rd_kafka_brokers_add0(rd_kafka_t *rk, const char *brokerlist) {
                         break;
 
                 rd_kafka_wrlock(rk);
+                if (rk->rk_conf.enable_bootstrap_servers_canonical_resolve) {
```
pranavrth, line 5270: Shouldn't this logic should be in resolve function? We are just adding host names here.

pranavrth (changes requested):

Issue #4214, opened 2023-03-09
kafka handles keeps topic handles alive and refreshes their metadata beyond subscription


notnoop:
Description
===========
When an `rd_kafka_t` consumer handle is reused for consuming multiple topics, it keeps references to all topics it ever listened to and keeps refreshing their metadata even after app stopped consuming them and potentially after topics have been deleted.

In my use case, we reuse kafka consumer handle to consume multiple topics, some of which are short lived. After few days, kafka consumer repeated polls metadata for thousands of already deleted topics a second. We have been rolling our consumer processes to reset consumer state to workaround this issue.

Might not be the root case, but the logs shows some `rd_kafka_toppar_t` instances that keep the topic ref counter positive, but they never get deleted. In my reproduction, there were two toppar created: one for `RD_KAFKA_PARTITION_UA` and one for the partition we consumer. They don't get freed until the consumer handle is destroyed

How to reproduce
================

I've used the `rdkafka_example` to reproduce the problem. The code is in https://github.com/notnoop/librdkafka/commit/ef06a2c0454c20f7c50debc63bdb7e80573b6e61 . In short, it blocks between topic and kafka handle destruction to allow examining state in between.

Reproduction steps:

```
# prepare code, and provision kafka instance 
./configure --debug --enable-refcnt-debug  && make
docker-compose up -d
docker-compose exec kafka kafka-topics --bootstrap-server localhost:29092 --create --topic test1

# publish sample messages
echo message1 | ./examples/rdkafka_example -P -t test1 -b localhost:29092 -p 0
echo message2 | ./examples/rdkafka_example -P -t test1 -b localhost:29092 -p 0

# start consuming (100 ms to force metadata refresh - not our production value)
./examples/rdkafka_example -C -t test1 -b localhost:29092 -p 0 -X topic.metadata.refresh.interval.ms=100 -Xdebug=all 2>verboselog

# in another terminal
# after awhile
pkill -SIGUSR1 rdkafka_example
# delete topic
docker-compose exec kafka kafka-topics --bootstrap-server localhost:29092 --delete --topic test1
# eventually kill process before handle destruction
pkill -9 rdkafka_example
```

We observe the following:
* Destroying the topic decreases refcnt from 3 to 2 (and app_refcnt from 1 to 0) but toppar isn't removed
```
#### After destroying topic
rd_kafka_t 0x5612dfea2e90: rdkafka#consumer-1
 producer.msg_cnt 0 (0 bytes)
[...]

PR #4310, opened 2023-06-08
Skip refreshing not existing (a.k.a. deleted) topics

notnoop:
Optionally, stop fetching metadata of not-existing topics in periodic refreshes.

When a Kafka topic is deleted before consumer stops, librdkafka queries for its metadata every second until the rdkafka handle is destroyed!
Every second, `rd_kafka_topic_scan_all` queries metadata of all topics with `rkt_partition_cnt == 0`, which includes all topics that got deleted on broker side: https://github.com/confluentinc/librdkafka/blob/f608e345a72f8f04ffc328b3ebd52bc741237773/src/rdkafka_topic.c#L1185 .

This addresses the egregious side-effect of https://github.com/confluentinc/librdkafka/issues/4214 . Metadata request rate increases over time as topics get deleted.


src/rdkafka_topic.c
```diff
@@ -1518,6 +1519,12 @@ void rd_kafka_topic_scan_all(rd_kafka_t *rk, rd_ts_t now) {
 
                 rd_kafka_topic_wrlock(rkt);
 
+                if (ignore_notexists &&
+                    rkt->rkt_state == RD_KAFKA_TOPIC_S_NOTEXISTS) {
+                        rd_kafka_topic_wrunlock(rkt);
+                        continue;
+                }
```
notnoop, line 1526: We may want to check app_refcnt here. If a deleted topic gets recreated, or if the application creates a new topic handle, it may make sense to keep fetching state. Though ideally, not every second.

Issue #1834, opened 2018-06-04
Confused about topic.blacklist setting


cvuillemez:
Description
===========

This setting is described as 

> Topic blacklist, a comma-separated list of regular expressions for matching topic names that should be ignored in broker metadata information as if the topics did not exist. 

When we use a consumer group, is the setting applied to all consumers within the same group ?
It 's seems not, so I need to explain why I ask this question:

Suppose you have:
* a consumer group "mygroup" with a consumer C1 and consumer C2
* C1 subscribes to "topicA, topicB, topicC" 
* C2 subscribes to "^topic.*" and has a `topic.blacklist = "topicC"`

Everything works fine, C1 consume 3 topics and C2 consumer only 2.

But when I restarted a consumer (C1), I don't know why, the `topic.blacklist` was suddently applied to each consumer, so the topicC has no more consumers...

I can't reproduce it easily... can it be a bug, or is it a bad use of this setting ?

Checklist
=========

 - [x] librdkafka version : `0.11.3`
 - [x] Apache Kafka version: `1.0.1`
 - [x] Operating system: `Ubuntu 16.04.3 LTS`

edenhill:
topic.blacklist only affects the local client, but if the local client is elected group leader then it is responsible for assigning topics and partitions to the members in the group, with a blacklist in place that client will not see topiC and will thus not assign it to any member, regardless if they have blacklist configured or not.

cvuillemez:
So clear now.. :)
Is there a doc or code who shows which setting is local or group-wide ?

edenhill:
Pretty much all client configuration properties only have local effect, but they may have secondary effects like this one.

PR #4315, opened 2023-06-12
Fixed a bug where topic leader is not refreshed in the same metadata call even if the broker is present.

pranavrth:
Fixes https://github.com/confluentinc/confluent-kafka-python/issues/1581

### DIFF
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -722,14 +722,6 @@ rd_kafka_parse_Metadata(rd_kafka_broker_t *rkb,
                 }
 
                 rd_kafka_buf_skip_tags(rkbuf);
-
-                /* Sort partitions by partition id */
-                qsort(md->topics[i].partitions, md->topics[i].partition_cnt,
-                      sizeof(*md->topics[i].partitions),
-                      rd_kafka_metadata_partition_id_cmp);
-                qsort(mdi->topics[i].partitions, md->topics[i].partition_cnt,
-                      sizeof(*mdi->topics[i].partitions),
-                      rd_kafka_metadata_partition_internal_cmp);
         }
 
         if (ApiVersion >= 8 && ApiVersion <= 10) {
@@ -773,6 +765,14 @@ rd_kafka_parse_Metadata(rd_kafka_broker_t *rkb,
                                    md->topics[i].topic);
                         continue;
                 }
+
+                /* Sort partitions by partition id */
+                qsort(md->topics[i].partitions, md->topics[i].partition_cnt,
+                      sizeof(*md->topics[i].partitions),
+                      rd_kafka_metadata_partition_id_cmp);
+                qsort(mdi->topics[i].partitions, md->topics[i].partition_cnt,
+                      sizeof(*mdi->topics[i].partitions),
+                      rd_kafka_metadata_partition_internal_cmp);
 
                 if (compute_racks)
                         rd_kafka_populate_metadata_topic_racks(&tbuf, i, mdi);
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -723,18 +723,6 @@ rd_kafka_parse_Metadata(rd_kafka_broker_t *rkb,
 
                 rd_kafka_buf_skip_tags(rkbuf);
 
-                /* Ignore topics in blacklist */
-                if (rkb->rkb_rk->rk_conf.topic_blacklist &&
-                    rd_kafka_pattern_match(rkb->rkb_rk->rk_conf.topic_blacklist,
-                                           md->topics[i].topic)) {
-                        rd_rkb_dbg(rkb, TOPIC | RD_KAFKA_DBG_METADATA,
-                                   "BLACKLIST",
-                                   "Ignoring blacklisted topic \"%s\" "
-                                   "in metadata",
-                                   md->topics[i].topic);
-                        continue;
-                }
-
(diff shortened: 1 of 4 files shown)


tests/0125-immediate_flush.c
```diff
@@ -73,6 +73,64 @@ int main_0125_immediate_flush(int argc, char **argv) {
...
+ *
+ * Number of brokers in the cluster should be more than the number of
+ * brokers in the bootstrap.servers list for this test case to work correctly
+ *
+ */
+void do_test_first_flush_immediate() {
+        rd_kafka_mock_cluster_t *mock_cluster;
+        rd_kafka_t *produce_rk;
+        const char *brokers;
+        const char *bootstrap_server;
```
emasab, line 92: remove the `const` and free it after `test_conf_set`

tests/0125-immediate_flush.c
```diff
@@ -73,6 +73,64 @@ int main_0125_immediate_flush(int argc, char **argv) {
...
+        rd_kafka_destroy(produce_rk);
+        test_mock_cluster_destroy(mock_cluster);
+        free(bootstrap_server);
+}
+
+int main_0125_immediate_flush(int argc, char **argv) {
+
+        do_test_flush_overrides_linger_ms_time();
+
+        do_test_first_flush_immediate();
```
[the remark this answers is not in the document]
pranavrth: Done

tests/0125-immediate_flush.c
pranavrth: Done.

src/rdkafka_metadata.c
```diff
@@ -735,22 +774,13 @@ rd_kafka_parse_Metadata(rd_kafka_broker_t *rkb,
...
 
-                /* Sort partitions by partition id */
-                qsort(md->topics[i].partitions, md->topics[i].partition_cnt,
-                      sizeof(*md->topics[i].partitions),
-                      rd_kafka_metadata_partition_id_cmp);
-                qsort(mdi->topics[i].partitions, md->topics[i].partition_cnt,
-                      sizeof(*mdi->topics[i].partitions),
-                      rd_kafka_metadata_partition_internal_cmp);
-
                 if (compute_racks)
```
pranavrth, line 777: I have done this change but I think this should be part of the first topic loop itself as `tbuf` is getting populated in that loop only. @milindl - Can you confirm this part once?

emasab: @pranavrth It seems it works in both places because it doesn't use `rd_kafka_broker_find_by_nodeid` that was giving problems, putting it there means we do it also for blacklisted topics. Fair, it's not using it but has the same guarantees that other topics in the metadata struct.

tests/test.c
```diff
@@ -484,6 +485,7 @@ struct test tests[] = {
     _TEST(0123_connections_max_idle, 0),
     _TEST(0124_openssl_invalid_engine, TEST_F_LOCAL),
     _TEST(0125_immediate_flush, 0),
+    _TEST(0125_immediate_flush_mock, 0),
```
emasab: this should have TEST_F_LOCAL as other mock tests

src/rdkafka_metadata.c
milindl: Looks good now to me, with blacklisted topics being ignored

emasab: If we chose previous way for performance we should also avoid sorting in the first loop. I see blacklisted topics can be much because they're regular expressions.

milindl: We need to sort this one in the first loop

```              qsort(md->topics[i].partitions, md->topics[i].partition_cnt,
                      sizeof(*md->topics[i].partitions),
                      rd_kafka_metadata_partition_id_cmp);```
                      
                      Since it's being returned to user. The internal one can be sorted anywhere.

milindl: But I think just for our own sanity we should keep the internal and non-internal topics consistent

milindl: Looking at https://github.com/confluentinc/librdkafka/issues/1834 - it seems like topic blacklist already causes some secondary effects like the topic not being assigned to other clients if the leader has them in its blacklist.  So computation of racks being below the blacklist if-clause should be okay. 
But it's weird, that we are returning the blacklisted topics in metadata pointer even though we say that "Topic blacklist, a comma-separated list of regular expressions for matching topic names that should be ignored in broker metadata information as if the topics did not exist. "

milindl: On the whole, I think moving both the sorts to just before the compute racks if clause is the best way, preserves how things were for pre-2.2.0 and also has no performance impact. What do you think, @emasab , @pranavrth ?

emasab: You mean everything starting from the blacklist with continue in the second loop? It's ok for me

milindl: Yes, that's exactly what I meant, thansk

pranavrth: I have updated the PR. Moved sorting in second loop as well.

milindl (approved):
LGTM after the changes we discussed about yesterday

r-owen:
Was this supposed to make it into librdkafka 2.2.0? Because I am still seeing the same issue there.

jzvelc:
Any news on this one? I have the same problem with librdkafka 2.2.0.

PR #4321, opened 2023-06-15
[internal branch] [mock]: Fix passing null pointer to strndup, which is UB

milindl:
Thanks @phallot for this, originally #4319. Moving to internal branch so CI can run on this req.  

---

UBSan diagnostic:
librdkafka/override/src/rd.h:166:27: runtime error: null pointer passed as argument 1, which is declared to never be null

Root cause:
 - rdkafka_mock_cgrp.c:570, `GroupInstanceId` is passed to `RD_KAFKAP_STR_DUP` while potentially NULL
 - l568, we have the check `if (GroupInstanceId)` which only checks whether the pointer is null (it can never be null based on construction). It does not check whether the string inside is null.
 - rdkafka_mock_handlers.c:1149, `GroupInstanceId` is set only if `rkbuf->rkbuf_reqhdr.ApiVersion >= 5`, otherwise it is NULL.

Fix:
 - check `if (RD_KAFKAP_STR_LEN(GroupInstanceId))` instead


pranavrth (approved):
LGTM!

PR #4322, opened 2023-06-15
Add a random key mode

travisdowns:
Currently rdkafka_performance offers a mode where you use a fixed key (-kfoobar) or no key at all, in which case the partitioner logic will choose the partition based on its own logic.

It is also useful to have a random key mode: this tends to spread messages evently across partitions (and hence brokers) which can have quite differnt characteristics as compared to the no-key partitioner behavior.

This change adds the -KN mode which uses a random key of length N, with a new key for each message.

### DIFF
--- a/examples/rdkafka_performance.c
+++ b/examples/rdkafka_performance.c
@@ -825,12 +825,21 @@ static void do_sleep(int sleep_us) {
         }
 }
 
+static void make_random_key(char *buf, int buflen) {
+        for (int i = 0; i < buflen; i++) {
+                buf[i] = rand() >> 7; // throw away low bits which are worse
+        }
+}
+
 
 int main(int argc, char **argv) {
         char *brokers   = NULL;
         char mode       = 'C';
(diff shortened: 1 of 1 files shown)


examples/rdkafka_performance.c
```diff
@@ -823,12 +823,21 @@ static void do_sleep(int sleep_us) {
...
+}
+
 
 int main(int argc, char **argv) {
         char *brokers   = NULL;
         char mode       = 'C';
         char *topic     = NULL;
-        const char *key = NULL;
+        // if non-null, use this same key for every message
+        const char *fixedkey = NULL;
```
travisdowns, line 838: The name change `key` -> `fixedkey` reflects the fact that this is set only if a fixed key is to be used.

travisdowns:
Details in commit message.

travisdowns:
force push a550ee21c9993137b826078038205847a0eff86d was a pure rebase

PR #4329, opened 2023-06-26
[admin] KIP-430: Add authorized operations to describe API

milindl:
Started in #4240 by @jainruchir . A large majority of the work is retained, updated with certain review suggestions/comments from me. 

It's based on master for easier review.


src/rdkafka_request.c
```diff
@@ -2354,12 +2373,13 @@ rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
...
-                                        * but forward parsed result to
-                                        * rko's replyq when done. */
-                                       RD_KAFKA_REPLYQ(rkb->rkb_rk->rk_ops, 0),
-                                       rd_kafka_handle_Metadata, rko);
+        rd_kafka_broker_buf_enq_replyq(
+            rkb, rkbuf,
+            /* Handle response thru rk_ops,
+             * but forward parsed result to
+             * rko's replyq when done. */
+            RD_KAFKA_REPLYQ(rkb->rkb_rk->rk_ops, 0),
```
pranavrth, line 2381: I would suggest to add more information regarding `rko` and `resp_cb`. When what is used so that in future its easier to read.

src/rdkafka_request.c
```diff
@@ -2187,26 +2190,39 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
...
+                         const rd_list_t *topics,
+                         const char *reason,
+                         rd_bool_t allow_auto_create_topics,
+                         rd_bool_t include_cluster_authorized_operations,
+                         rd_bool_t include_topic_authorized_operations,
+                         rd_bool_t cgrp_update,
+                         rd_bool_t force_racks,
+                         rd_kafka_op_t *rko,
+                         rd_kafka_resp_cb_t *resp_cb,
+                         int force,
```
pranavrth, line 2224: `rd_bool_t`

src/rdkafka_request.c
```diff
@@ -2187,26 +2190,39 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
...
  * @param cgrp_update - Update cgrp in parse_Metadata (see comment there).
  * @param force_racks - Force partition to rack mapping computation in
  *                      parse_Metadata (see comment there).
  * @param rko       - (optional) rko with replyq for handling response.
  *                    Specifying an rko forces a metadata request even if
  *                    there is already a matching one in-transit.
- *
+ * @param resp_cb - callback to be used for handling response. If NULL, then
+ *                     rd_kafka_handle_Metadata() is used.
+ * @param force - 1: force a full request.
```
pranavrth, line 2205: Can we explain more? like full request including all the topics and broker information? What happens if there is already a full request going on.

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+        rd_list_t dup_list;
+        size_t i;
+
+        static const struct rd_kafka_admin_fanout_worker_cbs fanout_cbs = {
+            rd_kafka_DescribeTopics_response_merge,
+            rd_kafka_TopicDescription_copy_opaque};
+
+        rd_assert(rkqu);
+
+        rko_fanout =
```
pranavrth, line 7165: Do we really need a fanout request here? I think we can live with request to controller for the metadata. Generally, all the brokers have metadata information present.

pranavrth (changes requested):
Rebase the PR.

First round of comments from few files. Still reviewing.

src/rdkafka_admin.h
```diff
@@ -473,10 +481,49 @@ struct rd_kafka_ConsumerGroupDescription_s {
...
+ * @{
+ */
+/**
+ * @struct DescribeCluster result
+ */
+struct rd_kafka_ClusterDescription_s {
+        char *cluster_id;       /**< current cluster id in \p cluster*/
+        int controller_id;      /**< current controller id in \p cluster*/
+        int node_cnt;           /**< Number of brokers in \p cluster*/
+        rd_kafka_Node_t *Nodes; /**< Nodes */
```
pranavrth, line 522: use `rd_list_t` instead of `rd_kafka_Node_t *`

src/rdkafka_admin.h
```diff
@@ -473,10 +481,49 @@ struct rd_kafka_ConsumerGroupDescription_s {
         rd_kafka_consumer_group_state_t state;
         /** Consumer group coordinator. */
         rd_kafka_Node_t *coordinator;
+        /** List of authorized operations.
+         * Type: rd_kafka_AclOperation_t* */
+        rd_list_t *authorized_operations;
```
pranavrth, line 486: No need to use pointer here and other places as well.

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+                    rd_list_cnt(cluster_authorized_operations), rd_free);
+                for (i = 0; i < rd_list_cnt(cluster_authorized_operations);
+                     i++) {
+                        int *entry =
+                            rd_list_elem(cluster_authorized_operations, i);
+                        int *oper = malloc(sizeof(int));
+                        *oper     = *entry;
+                        rd_list_add(clusterdesc->cluster_authorized_operations,
+                                    oper);
+                }
```
pranavrth, line 7344: Same

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+
+static void rd_kafka_ClusterDescription_free(void *ptr) {
+        rd_kafka_ClusterDescription_destroy(ptr);
+}
+/**
+ * @brief Send DescribeClusterRequest. Admin worker compatible callback.
+ */
+static rd_kafka_resp_err_t
+rd_kafka_admin_DescribeClusterRequest(rd_kafka_broker_t *rkb,
+                                      const rd_list_t *topics /*(char*)*/,
```
pranavrth, line 7383: I think this is null always in our case. Let's name and mark it correctly.

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+        rd_kafka_resp_err_t err;
+        int include_cluster_authorized_operations;
+
+        include_cluster_authorized_operations =
+            rd_kafka_confval_get_int(&options->include_authorized_operations);
+
+        err = rd_kafka_MetadataRequest(
+            rkb, NULL, "describe cluster", rd_false /*no auto create*/,
+            include_cluster_authorized_operations,
+            rd_false /*!include topic authorized operations */,
```
pranavrth, line 7399:
```suggestion
            rd_false /* don't include topic authorized operations */,
```

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+                goto err;
+        rko_result = rd_kafka_admin_result_new(rko_req);
+        rd_list_init(&rko_result->rko_u.admin_result.results, 1,
+                     rd_kafka_ClusterDescription_free);
+
+        rd_list_t *authorized_operations;
+        authorized_operations =
+            rd_kafka_AuthorizedOperations_parse(cluster_authorized_operations);
+
+        clusterdesc = rd_kafka_ClusterDescription_new(
```
pranavrth, line 7444: I think we should have a function `rd_kafka_ClusterDescription_new(rd_kafka_metadata_internal_t  *mdi)`. The current function can be `new0` but not with list of nodes instead of metadata.

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+        cluster_id = mdi->cluster_id;
+        controller_id                 = mdi->controller_id;
+        cluster_authorized_operations = mdi->cluster_authorized_operations;
+        if (err)
+                goto err;
+        rko_result = rd_kafka_admin_result_new(rko_req);
+        rd_list_init(&rko_result->rko_u.admin_result.results, 1,
+                     rd_kafka_ClusterDescription_free);
+
+        rd_list_t *authorized_operations;
```
pranavrth, line 7440: Move all the declarations to the top of the block. Check in other files as well.

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+
+        rd_list_add(&rko_result->rko_u.admin_result.results, clusterdesc);
+        *rko_resultp = rko_result;
+        return RD_KAFKA_RESP_ERR_NO_ERROR;
+
+err:
+        if (rko_result)
+                rd_kafka_op_destroy(rko_result);
+        rd_snprintf(errstr, errstr_size,
+                    "DescribeTopics response protocol parse failure: %s",
```
pranavrth, line 7457: DescribeClusters

src/rdkafka_cgrp.c
```diff
@@ -2240,9 +2242,11 @@ static int rd_kafka_cgrp_metadata_refresh(rd_kafka_cgrp_t *rkcg,
                                  rd_kafka_cgrp_handle_Metadata_op);
         rd_kafka_op_set_replyq(rko, rkcg->rkcg_ops, 0);
 
-        err = rd_kafka_metadata_request(rkcg->rkcg_rk, NULL, &topics,
-                                        rd_false /*!allow auto create */,
-                                        rd_true /*cgrp_update*/, reason, rko);
+        err = rd_kafka_metadata_request(
+            rkcg->rkcg_rk, NULL, &topics, rd_false /*!allow auto create */,
+            rd_false /*!include cluster authorized operations */,
+            rd_false /*!include topic authorized operations */,
```
pranavrth, line 2248:
```suggestion
            rd_false /*!include cluster authorized operations */,
            rd_false /*!include topic authorized operations */,
```
```suggestion
            rd_false /*don't include cluster authorized operations */,
            rd_false /*don't include topic authorized operations */,
```

Do this at other places as well.

src/rdkafka.c
```diff
@@ -4794,7 +4794,9 @@ static void rd_kafka_ListGroups_resp_cb(rd_kafka_t *rk,
 
                 state->wait_cnt++;
                 error = rd_kafka_DescribeGroupsRequest(
-                    rkb, 0, grps, i, RD_KAFKA_REPLYQ(state->q, 0),
+                    rkb, 0, grps, i,
+                    rd_false /* include_authorized_operations */,
```
pranavrth, line 4798: I think we should come up with a convention for these comments. Somewhere we try to explain what we are doing and somewhere we are just adding the field name. Sometimes its confusing what we are telling.

pranavrth:
Fix build issues and go through the PR checklist as well.

src/rdkafka.h
```diff
@@ -7730,6 +7821,345 @@ rd_kafka_DeleteRecords_result_offsets(
...
+ *
+ * @param topicdesc The topic description.
+ * @param idx Index for the partitions.
+ *
+ * @return The partition id.
+ */
+RD_EXPORT
+const int rd_kafka_TopicDescription_partition_id(
+    const rd_kafka_TopicDescription_t *topicdesc,
+    int idx);
```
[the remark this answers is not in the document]
milindl: Given that metadata_partition is a public type in rdkafka.h, what do you think of removing these functions entirely?
We can just have `rd_kafka_TopicDescription_partition(topicdesc, partition_idx)` which returns a `rd_kafka_metadata_partition` ?
Same for all such functions

src/rdkafka.h
```diff
@@ -7730,6 +7821,345 @@ rd_kafka_DeleteRecords_result_offsets(
...
+ * @param topicdesc The topic description.
+ * @param idx The index for which element is needed.
+ *
+ * @return Authorized operation at given index.
+ */
+RD_EXPORT
+const rd_kafka_AclOperation_t rd_kafka_TopicDescription_authorized_operation(
+    const rd_kafka_TopicDescription_t *topicdesc,
+    size_t idx);
+
```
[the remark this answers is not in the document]
milindl: About this: we discussed it on the call, it made sense, but I realized now that we are not using rd_list_t anywhere in the public API. 
Do you think we should add it regardless of that?

The other option is to convert to rd_kafka_AclOperation_t* and then return it, but then the user is responsible for freeing it, which also diverges from the current pattern of things.

I guess this was inspired from rd_kafka_ConsumerGroupDescription_member

src/rdkafka_admin.c
```diff
@@ -6217,6 +6285,23 @@ const char *rd_kafka_ConsumerGroupDescription_partition_assignor(
...
+        return 0;
+}
+
+rd_kafka_AclOperation_t rd_kafka_ConsumerGroupDescription_authorized_operation(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc,
+    size_t idx) {
+        if (grpdesc->authorized_operations) {
+                rd_kafka_AclOperation_t *entry =
+                    rd_list_elem(grpdesc->authorized_operations, idx);
+                return *entry;
```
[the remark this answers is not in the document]
milindl: We should not do this for any case, since the user knows already what the count is (it's similar to the idea of 'Don’t check for NULLs, if parameters are required it’s caller responsibility to pass them'). 
We should even remove the `if (grpdesc->authorized_operations)` in my opinion, what do you think?
See for example, rd_kafka_ConsumerGroupDescription_member

src/rdkafka_admin.c
```diff
@@ -6217,6 +6285,23 @@ const char *rd_kafka_ConsumerGroupDescription_partition_assignor(
         return grpdesc->partition_assignor;
 }
 
+size_t rd_kafka_ConsumerGroupDescription_authorized_operation_count(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc) {
+        if (grpdesc->authorized_operations)
```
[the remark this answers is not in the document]
milindl: Will address this after discussion on above comment

src/rdkafka_admin.c
milindl: Since it's an internal function used only in one place, I am changing it to just accept rd_kafka_metadata_internal_t.

src/rdkafka.c
milindl: I think, in this case, it's because the value doesn't make it obvious what the parameter is exactly, so for easy reading, we're adding the field name.
I think it's worth just always using the field name rather than describing what we're doing unless our behaviour is unusual in some way. I can make this change elsewhere also. What do you think?

src/rdkafka.h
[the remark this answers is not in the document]
jainruchir: we are not using rd_list_t anywhere in the public API. This was the reason and yes it is kind of inspired from rd_kafka_ConsumerGroupDescription_member

src/rdkafka_admin.c
```diff
@@ -6489,33 +6581,46 @@ rd_kafka_DescribeConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
                 }
 
                 if (api_version >= 3) {
-                        /* TODO: implement KIP-430 */
-                        int32_t authorized_operations;
                         rd_kafka_buf_read_i32(reply, &authorized_operations);
+                        /* The last 3 bits are never set. */
+                        rd_dassert(!((authorized_operations >> 0) & 1));
+                        rd_dassert(!((authorized_operations >> 1) & 1));
+                        rd_dassert(!((authorized_operations >> 2) & 1));
```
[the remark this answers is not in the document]
jainruchir: Referring to the [doc](https://cwiki.apache.org/confluence/display/KAFKA/KIP-430+-+Return+Authorized+Operations+in+Describe+Responses), the broker will set the bitfields corresponding to actual supported operations on the resource and the bits corresponding to AclOperation.ALL, AclOperation.ANY and AclOperation.UNKNOWN will never be set by the broker.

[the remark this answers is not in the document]
milindl: Yep, I updated it as per this^

src/rdkafka_admin.c
```diff
@@ -6666,3 +6773,705 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+        if (unlikely(topicdesc->error != NULL))
+                rd_kafka_error_destroy(topicdesc->error);
+        for (i = 0; i < topicdesc->partition_cnt; i++) {
+                if (likely(topicdesc->partitions[i].isrs != NULL))
+                        rd_free(topicdesc->partitions[i].isrs);
+                if (likely(topicdesc->partitions[i].replicas != NULL))
+                        rd_free(topicdesc->partitions[i].replicas);
+        }
+        if (likely(topicdesc->partitions != NULL))
+                rd_free(topicdesc->partitions);
```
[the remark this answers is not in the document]
milindl: I think the reason we don't have this is because metadata_partition is generally always allocated in the tmpabuf.

src/rdkafka.h
```diff
@@ -7952,6 +8382,31 @@ RD_EXPORT
...
+ *
+ * @param grpdesc The group description.
+ * @param idx The index for which element is needed.
+ *
+ * @return Authorized operation at given index.
+ */
+RD_EXPORT
+rd_kafka_AclOperation_t rd_kafka_ConsumerGroupDescription_authorized_operation(
+    const rd_kafka_ConsumerGroupDescription_t *grpdesc,
+    size_t idx);
```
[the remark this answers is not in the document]
milindl: Will address after discussion on above comments

src/rdkafka.h
```diff
@@ -7730,6 +7821,345 @@ rd_kafka_DeleteRecords_result_offsets(
...
+ * @param clusterdesc The cluster description.
+ * @param idx The index for which element is needed.
+ *
+ * @return Authorized operation at given index.
+ */
+RD_EXPORT
+const rd_kafka_AclOperation_t rd_kafka_ClusterDescription_authorized_operation(
+    const rd_kafka_ClusterDescription_t *clusterdesc,
+    size_t idx);
+
```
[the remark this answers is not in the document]
milindl: Will address after discussion on above comments

milindl:
I replied to, or addressed the comments. Build issues should be fixed, too (hopefully, latest CI run will confirm that).

src/rdkafka.h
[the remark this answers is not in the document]
pranavrth: Shall we typecast `rl_elems` to  `rd_kafka_AclOperation_t` and return it.

src/rdkafka_admin.c
pranavrth: Discuss with team to decide on a convention.

src/rdkafka_admin.c
```diff
@@ -7734,3 +7848,569 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+                }
+                rd_list_add(&rko_result->rko_u.admin_result.results, topicdesc);
+        }
+
+        *rko_resultp = rko_result;
+        return RD_KAFKA_RESP_ERR_NO_ERROR;
+
+err:
+        if (rko_result)
+                rd_kafka_op_destroy(rko_result);
```
pranavrth, line 8158: RD_IF_FREE

src/rdkafka_metadata.c
```diff
@@ -1425,7 +1443,8 @@ rd_kafka_resp_err_t rd_kafka_metadata_refresh_all(rd_kafka_t *rk,
         rd_list_init(&topics, 0, NULL); /* empty list = all topics */
         rd_kafka_MetadataRequest(
             rkb, &topics, reason, rd_false /*no auto create*/,
-            rd_true /*cgrp update*/, rd_false /* force_rack */, NULL);
+            rd_true /*cgrp update*/, rd_false /* force_rack */, NULL,
+            0 /* force */, NULL);
```
pranavrth, line 1447: rd_false instead of 0

src/rdkafka_request.c
```diff
@@ -2174,40 +2177,20 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
...
+                          const rd_list_t *topics,
+                          const char *reason,
+                          rd_bool_t allow_auto_create_topics,
+                          rd_bool_t include_cluster_authorized_operations,
+                          rd_bool_t include_topic_authorized_operations,
+                          rd_bool_t cgrp_update,
+                          rd_bool_t force_racks,
+                          rd_kafka_op_t *rko,
+                          rd_kafka_resp_cb_t *resp_cb,
+                          int force,
```
pranavrth, line 2192: Is this needed anymore?

pranavrth (changes requested):
Metadata and tests remaining to review. Almost there.

src/rdkafka_request.c
```diff
@@ -2279,11 +2261,12 @@ rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
         if (full_incr) {
```
pranavrth, line 2261: What is `full_incr`?

src/rdkafka_request.c
```diff
@@ -2355,17 +2338,118 @@ rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
...
+             * rko's replyq when done. */
+            RD_KAFKA_REPLYQ(rkb->rkb_rk->rk_ops, 0),
+            /* The default response handler is rd_kafka_handle_Metadata, but we
+               allow alternate handlers to be configured. */
+            resp_cb ? resp_cb : rd_kafka_handle_Metadata,
+            /* rd_kafka_handle_Metadata (the default handler) makes use of the
+               rko (if set) to forward the response to the rko's replyq. So, in
+               case the opaque value hasn't been overrriden, we set it to the
+               rko. */
+            rko ? rko : opaque);
```
pranavrth, line 2354: Need to understand more about this part.

src/rdkafka_request.c
```diff
@@ -2355,17 +2338,118 @@ rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
...
+ */
+rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
+                                             const rd_list_t *topics,
+                                             const char *reason,
+                                             rd_bool_t allow_auto_create_topics,
+                                             rd_bool_t cgrp_update,
+                                             rd_bool_t force_racks,
+                                             rd_kafka_op_t *rko,
+                                             rd_bool_t force,
+                                             void *opaque) {
```
pranavrth, line 2400: shall we really expose these two in this part? I think we should stick to the old one and add it when it is really required.

pranavrth (changes requested):

src/rdkafka_admin.c
```diff
@@ -7734,3 +7848,569 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+            rko->rko_u.admin_result.results.rl_elems;
+}
+
+/**
+ * @brief Helper function to clear a rd_kafka_metadata_partition.
+ *
+ * @note Does not deallocate the rd_kafka_metadata_partition itself.
+ */
+static void
+rd_kafka_metadata_partition_clear(struct rd_kafka_metadata_partition *rkmp) {
```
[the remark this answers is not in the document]
milindl: There are no _new or _destroys for this struct (since we almost always use it inside a pre-allocated tmpabuf).
I'm moving this function and the copy function to the metadata.c file where it might make most sense.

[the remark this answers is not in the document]
milindl: Will check for other new types and destroy/new functions

src/rdkafka_metadata.c
```diff
@@ -448,29 +450,30 @@ rd_kafka_populate_metadata_topic_racks(rd_tmpabuf_t *tbuf,
...
+        const int log_decode_errors = LOG_ERR;
+        rd_list_t *missing_topics   = NULL;
+        const rd_list_t *requested_topics =
+            request ? request->rkbuf_u.Metadata.topics : request_topics;
+        rd_bool_t all_topics =
+            request ? request->rkbuf_u.Metadata.all_topics : rd_false;
         rd_bool_t cgrp_update =
-            request->rkbuf_u.Metadata.cgrp_update && rk->rk_cgrp;
+            request ? (request->rkbuf_u.Metadata.cgrp_update && rk->rk_cgrp)
+                    : rd_false;
```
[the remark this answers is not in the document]
milindl: Single level ternary operators are okay in my view, what do you think?
Especially if the condition in the ternary operator is a simple null check

src/rdkafka_metadata.c
```diff
@@ -560,11 +569,15 @@ rd_kafka_parse_Metadata(rd_kafka_broker_t *rkb,
                 rd_kafka_buf_skip_tags(rkbuf);
         }
 
-        if (ApiVersion >= 2)
+        if (ApiVersion >= 2) {
                 rd_kafka_buf_read_str(rkbuf, &cluster_id);
+                mdi->cluster_id = rd_tmpabuf_write_str(&tbuf, cluster_id.str);
```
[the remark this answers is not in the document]
milindl: Good catch, added if guard before. Earlier we were reading it but not writing it, but while writing to tmpabuf, we call strlen(cluster_id.str) which only works for non-null cluster_id.str.

examples/describe_cluster.c
```diff
@@ -0,0 +1,315 @@
...
+
+        if (rd_kafka_conf_set(conf, name, val, errstr, sizeof(errstr)) !=
+            RD_KAFKA_CONF_OK)
+                fatal("Failed to set %s=%s: %s", name, val, errstr);
+}
+
+/**
+ * @brief Parse an integer or fail.
+ */
+int64_t parse_int(const char *what, const char *str) {
```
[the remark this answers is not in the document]
milindl: About this, should we make the change across examples? It's the same in each example that we have this method in

milindl:
Addresses most comments, but CI will fail until #4389 is merged into this branch.

pranavrth (changes requested):
- Almost There. Maybe may last review before approval. Added some feedback.
- Fix build issues.
- Some old comments are still pending. Please check them. Some of them might need discussion with @emasab so I haven't resolved them.
- Tests look fine but I would request @emasab to look more into them.

src/rdkafka.h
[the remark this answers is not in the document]
milindl: After discussion: We can remove accessor functions and just return the partitions

src/rdkafka.h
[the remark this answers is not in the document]
milindl: After discussion: investigate turning authorized_operations into array and maintaining size, return size and array (size ptr)

src/rdkafka_request.c
```diff
@@ -2187,26 +2190,39 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
...
  */
-rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
-                                             const rd_list_t *topics,
-                                             const char *reason,
-                                             rd_bool_t allow_auto_create_topics,
-                                             rd_bool_t cgrp_update,
-                                             rd_bool_t force_racks,
-                                             rd_kafka_op_t *rko) {
+rd_kafka_resp_err_t
+rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
```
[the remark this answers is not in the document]
milindl: After discussion: add _op, and _cb methods, as intermediates to the current split. Move rd_kafka_MetadataRequest_admin to admin.c

src/rdkafka_admin.c
milindl: After discussion: behaviour

src/rdkafka.h
```diff
@@ -7889,6 +7979,349 @@ rd_kafka_DeleteRecords_result_offsets(
...
+ *                Valid options:
+ *                 - include_authorized_operations
+ * @param rkqu Queue to emit result on.
+ *
+ * @remark The result event type emitted on the supplied queue is of type
+ *         \c RD_KAFKA_EVENT_DESCRIBETOPICS_RESULT
+ */
+RD_EXPORT
+void rd_kafka_DescribeTopics(rd_kafka_t *rk,
+                             const char **topics,
```
emasab, line 8010: Can receive a `rd_kafka_TopicCollection_t` that currently is created only from topic names. In future with topic ids too. See J.

emasab (changes requested):
Reviewed the public API and proposed changes to make it more similar to J. and future proof. Let's discuss about them.

src/rdkafka_metadata.c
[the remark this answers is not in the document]
emasab: Agree with Pranav because there are multiple checks with the same condition, feels more readable.

src/rdkafka.h
milindl: Currently it contains only one field as char **topics and init method rd_kafka_TopicCollection_new_from_names,  later we can make it into a union.

src/rdkafka.h
```diff
@@ -7889,6 +7979,349 @@ rd_kafka_DeleteRecords_result_offsets(
...
+ * at the index \p partition_idx for the topic \p topicdesc.
+ *
+ * @param topicdesc The topic description.
+ * @param partition_idx Index for the partitions.
+ * @param isr_idx Index for the in-sync replica.
+ *
+ * @return The partition in-sync replica.
+ */
+RD_EXPORT
+const int rd_kafka_TopicDescription_partition_isr(
```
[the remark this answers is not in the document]
milindl: returning `const rd_kafka_Node_t **` and populating the size passed as a pointer, eliminating count methods entirely, here and below.

src/rdkafka.h
```diff
@@ -7889,6 +7979,349 @@ rd_kafka_DeleteRecords_result_offsets(
...
+ * @brief Gets the partition error for partition at index position for the \p
+ * topicdesc topic.
+ *
+ * @param topicdesc The topic description.
+ * @param idx Index for the partitions.
+ *
+ * @return The partition error.
+ */
+RD_EXPORT
+const rd_kafka_resp_err_t rd_kafka_TopicDescription_partition_error(
```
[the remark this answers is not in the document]
milindl: J. API doesn't expose this properly. So I removed it. Should I reintroduce it?

src/rdkafka.h
```diff
@@ -7889,6 +7979,349 @@ rd_kafka_DeleteRecords_result_offsets(
...
+
+/**
+ * @brief Gets the node count for the \p clusterdesc cluster.
+ *
+ * @param clusterdesc The cluster description.
+ *
+ * @return The node count.
+ */
+RD_EXPORT
+const int rd_kafka_ClusterDescription_node_count(
```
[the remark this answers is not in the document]
milindl: As above, I eliminated count methods and return Node_t ** instead, populating size as an argument.

src/rdkafka.h
```diff
@@ -7889,6 +7979,349 @@ rd_kafka_DeleteRecords_result_offsets(
...
+
+/**
+ * @brief Gets the cluster current controller id for the \p clusterdesc cluster.
+ *
+ * @param clusterdesc The cluster description.
+ *
+ * @return The cluster current controller id.
+ */
+RD_EXPORT
+const int rd_kafka_ClusterDescription_controller_id(
```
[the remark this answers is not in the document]
milindl: Controller count is not needed since there is just one controller. But I have changed the function to return const Node_t*

emasab (changes requested):
Comments about the internals

src/rdkafka_admin.c
```diff
@@ -7049,6 +7071,34 @@ const rd_kafka_error_t **rd_kafka_ListConsumerGroups_result_errors(
...
  */
 
+/**
+ * @brief Parse authorized_operations returned in
+ * - DescribeConsumerGroups
+ * - DescribeTopics
+ * - DescribeCluster
+ * @returns list of authorized operations (rd_kafka_AclOperation_t *)
+ */
+rd_list_t *rd_kafka_AuthorizedOperations_parse(int32_t authorized_operations) {
```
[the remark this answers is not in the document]
emasab: Make static

src/rdkafka_admin.h
```diff
@@ -473,10 +481,79 @@ struct rd_kafka_ConsumerGroupDescription_s {
...
+/**
+ * @struct DescribeCluster result - internal type.
+ */
+typedef struct rd_kafka_ClusterDescription_s {
+        char *cluster_id;            /**< Cluster id */
+        rd_kafka_Node_t *controller; /**< Current controller. */
+        size_t node_cnt;             /**< Count of brokers in the cluster. */
+        rd_kafka_Node_t **nodes;     /**< Brokers in the cluster. */
+        size_t authorized_operations_cnt; /**< Count of operations allowed for
+                                     cluster.*/
```
emasab, line 551: Put a `*` in comments after newlines, it helps autoformat indent correctly

emasab (changes requested):

src/rdkafka_admin.c
```diff
@@ -7734,3 +7848,569 @@ rd_kafka_DescribeConsumerGroups_result_groups(
...
+        clusterdesc->controller_id = mdi->controller_id;
+
+        if (authorized_operations == NULL)
+                rd_list_init(&clusterdesc->authorized_operations, 0, rd_free);
+        else {
+                rd_list_init_copy(&clusterdesc->authorized_operations,
+                                  authorized_operations);
+                rd_list_copy_to(&clusterdesc->authorized_operations,
+                                authorized_operations, copy_AclOperation, NULL);
+        }
```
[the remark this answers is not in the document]
milindl: Resolving since the code has changed a lot from this so it doesn't apply

src/rdkafka_metadata.c
```diff
@@ -58,6 +58,46 @@ static int rd_kafka_metadata_partition_internal_cmp(const void *_a,
...
+        dst->err     = src->err;
+        dst->id      = src->id;
+        dst->isr_cnt = src->isr_cnt;
+        dst->isrs    = rd_calloc(sizeof(int32_t), dst->isr_cnt);
+        for (i = 0; i < dst->isr_cnt; i++) {
+                dst->isrs[i] = src->isrs[i];
+        }
+        dst->leader      = src->leader;
+        dst->replica_cnt = src->replica_cnt;
+        dst->replicas    = rd_calloc(sizeof(int32_t), dst->replica_cnt);
```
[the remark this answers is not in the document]
milindl: Removing this function, unused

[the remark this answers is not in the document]
milindl: Removing this function but making change elsewhere

src/rdkafka_metadata.c
```diff
@@ -58,6 +58,46 @@ static int rd_kafka_metadata_partition_internal_cmp(const void *_a,
...
+ * @note The dst should not be contained inside a metadata struct allocated with
+ * tmpabuf.
+ */
+void rd_kafka_copy_metadata_partition(struct rd_kafka_metadata_partition *src,
+                                      struct rd_kafka_metadata_partition *dst) {
+        int i;
+        dst->err     = src->err;
+        dst->id      = src->id;
+        dst->isr_cnt = src->isr_cnt;
+        dst->isrs    = rd_calloc(sizeof(int32_t), dst->isr_cnt);
```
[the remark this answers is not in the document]
milindl: Removing this function, unused

src/rdkafka_request.c
```diff
@@ -2362,17 +2364,134 @@ rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
...
+         * overriden to use a custom handler. */
+        if (resp_cb)
+                handler_cb = resp_cb;
+
+        /* If a custom handler is provided, we also allow the caller to set a
+         * custom argument which is passed as the opaque argument to the
+         * handler. However, if we're using the default handler, it expects
+         * either rko or NULL as its opaque argument (it forwards the response
+         * to rko's replyq if it's non-NULL). */
+        if (resp_cb && opaque)
```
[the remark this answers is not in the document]
milindl: If resp_cb is set then rko cannot be set, so it will be NULL. 
So it doesn't matter what we sent, rko or opaque, since they'll both me NULL, right?

src/rdkafka_aux.c
```diff
@@ -248,6 +248,54 @@ rd_kafka_Node_t *rd_kafka_Node_new(int32_t id,
...
+ * @brief Create a new Node object given a node id, and use broker information
+ * to populate other fields.
+ *
+ * @return A new allocated Node object.
+ *         Use rd_kafka_Node_destroy() to free when done.
+ * @remark The \p brokers_internal array is asumed to be sorted by node id,
+ *         while the \p brokers array is not. This is the current default from
+ *         Metadata response parsing.
+ */
+rd_kafka_Node_t *rd_kafka_Node_new_from_brokers(
```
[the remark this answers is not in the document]
milindl: The only reservation I have about this is, about the assertion

```
        /* We assume that the marshalled representation is
         * no more than 4 times larger than the wire representation.
```

Since we're allocating the tmpabuf using this. 
I'm making the change by adding the sorted one, let me know if you want me to change the factor 4

src/rdkafka.h
```diff
@@ -7889,6 +7989,297 @@ rd_kafka_DeleteRecords_result_offsets(
...
+ * @param cntp is updated with authorized ACL operations count.
+ *
+ * @return The cluster authorized operations.
+ * @remark The lifetime of the returned memory is the same
+ *         as the lifetime of the \p result object.
+ */
+RD_EXPORT
+const rd_kafka_AclOperation_t *
+rd_kafka_DescribeCluster_result_authorized_operations(
+    const rd_kafka_DescribeTopics_result_t *result,
```
pranavrth, line 8253: Why are we using "DescribeTopics" here?

milindl: Changed that now

src/rdkafka_aux.c
```diff
@@ -277,3 +328,7 @@ const char *rd_kafka_Node_host(const rd_kafka_Node_t *node) {
 uint16_t rd_kafka_Node_port(const rd_kafka_Node_t *node) {
         return node->port;
 }
+
+const char *rd_kafka_Node_rack_id(const rd_kafka_Node_t *node) {
```
emasab, line 332: Better rack only as in J.
```suggestion
const char *rd_kafka_Node_rack(const rd_kafka_Node_t *node) {
```

milindl: Done

src/rdkafka_admin.c
```diff
@@ -7049,6 +7099,51 @@ const rd_kafka_error_t **rd_kafka_ListConsumerGroups_result_errors(
...
+
+        /* Count number of bits set. ALL, ANY and UNKNOWN bits are skipped as
+         * they are always unset as per KIP-430. */
+        for (i = RD_KAFKA_ACL_OPERATION_READ; i < RD_KAFKA_ACL_OPERATION__CNT;
+             i++)
+                count += ((authorized_operations >> i) & 1);
+
+        j          = 0;
+        operations = rd_malloc(sizeof(rd_kafka_AclOperation_t) * count);
```
emasab, line 7134: Count can be zero and in that case `malloc(0)` is implementation defined.
We can `rd_malloc(1) `if it's zero while keeping count as 0, so we have a pointer instead of implementation defined behavior. Or we can set count to 0 if it's empty or to -1 if it was NULL.

src/rdkafka_admin.c
```diff
@@ -7448,6 +7579,7 @@ rd_kafka_DescribeConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
         node = rd_kafka_Node_new(nodeid, host, port, NULL);
         while (cnt-- > 0) {
                 int16_t error_code;
+                int32_t authorized_operations = 0;
```
emasab, line 7582: initialize to -1 as in other places

src/rdkafka_admin.c
```diff
@@ -7427,6 +7585,8 @@ rd_kafka_DescribeConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
         rd_kafka_error_t *error   = NULL;
         char *group_id = NULL, *group_state = NULL, *proto_type = NULL,
              *proto = NULL, *host = NULL;
+        rd_kafka_AclOperation_t *operations = NULL;
+        int operation_cnt;
```
emasab, line 7589: initialize to 0 in case authorized operations list isn't supported

milindl: I've initialized to -1 for that case. It is further handled by rd_kafka_ConsumerGroupDescription_new, that's why I've done so

pranavrth (approved):
LGTM. Great work guys!..

emasab (approved):
LGTM! Amazing work!

PR #4336, opened 2023-06-28
Add support for setting TLSv1.3 ciphers

morzelx:
Currently library supports setting ciphers only for TLSv1.2 and below. Due to major differences between the way that ciphersuites for TLSv1.2 and below and ciphersuites for TLSv1.3 work, they are configured by a different OpenSSL function: SSL_CTX_set_ciphersuites. Because of that it is not possible to configure TLSv1.3 ciphers in option "ssl.cipher.suites", as it calls old OpenSSL function (SSL_CTX_set_cipher_list) under the hood.

This patch adds possibility to set ciphersuites for TLSv1.3 via new configuration option: "ssl.tls13.ciphersuites". I suspect it is the most convinient way to resolve this for now, as older TLS ciphers still expect to be configured using older OpenSSL function.

This also should resolve question raised in discussion thread #4286.


aambroze:
@milindl @edenhill - I'm also interested in the changes. Is there something which can be done to bring this PR to your attention?

hubertmowinski:
I am also in need for this support.

PR #4347, opened 2023-07-11
remove semaphore cache as it will not be available on public clusters

jkuhnashconfluent:


jkuhnashconfluent:
The modified semaphore phases were windows related which passed on relevant build runs. Merging without a passing check.

PR #4365, opened 2023-07-24
Add boilerplate for the entire flow

milindl:
Contains some initial boilerplate for the entire codeflow, please feel free to add more push specific items, I've added the bare minimum for both the protocols. 

I created a new file entirely, as I think it's justified, seeing how we have different ones for things like fetcher, cgrp, idempotence, etc.

The code compiles, but with warnings.

### DIFF
--- a/src/rdkafka_int.h
+++ b/src/rdkafka_int.h
@@ -234,6 +234,14 @@ rd_kafka_txn_state2str(rd_kafka_txn_state_t state) {
         return names[state];
 }
 
+typedef enum {
+        RD_KAFKA_TELEMETRY_AWAIT_BROKER,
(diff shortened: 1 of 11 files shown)


src/rdkafka_int.h
```diff
@@ -234,6 +234,14 @@ rd_kafka_txn_state2str(rd_kafka_txn_state_t state) {
         return names[state];
 }
 
+typedef enum {
+        RD_KAFKA_TELEMETRY_AWAIT_BROKER,
```
anchitj, line 238: Don't we need to add state for GET_THROTTLED, PUSH_THROTTLED and also for the error cases?

anchitj (approved):

PR #4379, opened 2023-08-02
Include openssl/rand.h to fix build error

nkostoulas:


emasab:
Strage I'm not getting this with gcc, which compiler and version?
```
gcc -MD -MP -g -O2 -fPIC -Wall -Wsign-compare -Wfloat-equal -Wpointer-arith -Wcast-align -Werror -DLIBRDKAFKA_GIT_VERSION="\"v2.2.0-7-g47bb41\"" -std=c99 -Wimplicit-function-declaration     -I/usr/include/x86_64-linux-gnu -c rdkafka_admin.c -o rdkafka_admin.o
```

nkostoulas:
@emasab we use clang 7.1

PR #4382, opened 2023-08-03
Add broker selection and client termination [KIP-714]

milindl:
Main changes:
1. rd_kafka_get_preferred_broker now handles the case when the previous broker was lost
2. termination is handled gracefully, both in terms of clearing resources, and in terms of sending a Push request with the last pieces of information.

To accommodate this:
Makes changes so that the entire state machine runs on the main thread.
Communication with the broker thread (when preferred broker has to be set) and with the app thread (during termination) is done via ops or conditional variables.
Also updates some documentation comments.

### DIFF
--- a/src/rdkafka_telemetry.c
+++ b/src/rdkafka_telemetry.c
@@ -33,21 +33,21 @@
 #include "rdkafka_request.h"
 
 /**
- * @brief Returns the preferred metrics broker or NULL if unavailable.
+ * @brief Filters broker by availability of GetTelemetrySubscription.
  *
- * @locks_acquired rk_telemetry.lock
+ * @return 0 if GetTelemetrySubscription is supported, 1 otherwise.
+ *
+ * @locks rd_kafka_broker_lock()
  */
-static rd_kafka_broker_t *rd_kafka_get_preferred_broker(rd_kafka_t *rk) {
-        rd_kafka_broker_t *rkb = NULL;
-
-        mtx_lock(&rk->rk_telemetry.lock);
-        if (rk->rk_telemetry.preferred_broker) {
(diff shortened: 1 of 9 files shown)


src/rdkafka_telemetry.c
```diff
@@ -192,32 +249,164 @@ void rd_kafka_handle_push_telemetry(rd_kafka_t *rk, rd_kafka_resp_err_t err) {
...
+
+        rko         = rd_kafka_op_new(RD_KAFKA_OP_TERMINATE_TELEMETRY);
+        rko->rko_rk = rk;
+        rd_kafka_q_enq(rk->rk_ops, rko);
+
+        /* Await termination sequence completion. */
+        rd_kafka_dbg(rk, TELEMETRY, "TELTERM",
+                     "Awaiting termination of telemetry.");
+        mtx_lock(&rk->rk_telemetry.lock);
+        cnd_wait(&rk->rk_telemetry.termination_cnd, &rk->rk_telemetry.lock);
```
anchitj, line 280: Will it make sense to do a timed_wait if for some case RD_KAFKA_TELEMETRY_TERMINATING_PUSH_SCHEDULED gets blocked somewhere?

milindl: Agreed

How much should we wait for? 1s?

anchitj: Hmm not sure, but I think 1s should be good. We can probably tweak it later if the terminating push takes more time.

milindl: Done, 1s

anchitj (approved):
Thanks, LGTM!

PR #4404, opened 2023-08-23
[KIP-848] topic_id in topic_partition_internal, some new error codes

emasab:
and `rd_kafka_consumer_group_metadata_member_id` to get the member id


src/rdkafka_partition.c
```diff
@@ -2707,6 +2713,42 @@ int32_t rd_kafka_topic_partition_get_current_leader_epoch(
...
+ * @param topic_id Topic id to set.
+ */
+void rd_kafka_topic_partition_set_topic_id(rd_kafka_topic_partition_t *rktpar,
+                                           rd_kafka_Uuid_t topic_id) {
+        rd_kafka_topic_partition_private_t *parpriv;
+
+        /* Avoid allocating private_t if clearing the epoch */
+        if (!rd_kafka_Uuid_cmp(topic_id, RD_KAFKA_UUID_ZERO) &&
+            !rktpar->_private)
+                return;
```
pranavrth, line 2729: Why this? Need more understanding on this.

src/rdkafka_partition.c
```diff
@@ -2855,6 +2900,20 @@ rd_kafka_topic_partition_list_add(rd_kafka_topic_partition_list_t *rktparlist,
...
+rd_kafka_topic_partition_t *rd_kafka_topic_partition_list_add_with_topic_id(
+    rd_kafka_topic_partition_list_t *rktparlist,
+    rd_kafka_Uuid_t topic_id,
+    int32_t partition) {
+        rd_kafka_topic_partition_t *rktpar;
+        rktpar = rd_kafka_topic_partition_list_add0(
+            __FUNCTION__, __LINE__, rktparlist, NULL, partition, NULL, NULL);
+        rd_kafka_topic_partition_private_t *parpriv =
+            rd_kafka_topic_partition_get_private(rktpar);
+        parpriv->topic_id = topic_id;
```
pranavrth, line 2912: Shall we check `RD_KAFKA_UUID_ZERO` case here like we are doing in `set` and `get` topic_id? First let's understand why do we need this check there in the first place.

src/rdkafka_partition.c
```diff
@@ -3013,6 +3072,22 @@ int rd_kafka_topic_partition_cmp(const void *_a, const void *_b) {
...
+int rd_kafka_topic_partition_by_id_cmp(const void *_a, const void *_b) {
+        const rd_kafka_topic_partition_t *a = _a;
+        const rd_kafka_topic_partition_t *b = _b;
+        rd_kafka_Uuid_t topic_id_a = rd_kafka_topic_partition_get_topic_id(a);
+        rd_kafka_Uuid_t topic_id_b = rd_kafka_topic_partition_get_topic_id(b);
+        int r                      = rd_kafka_Uuid_cmp(topic_id_a, topic_id_b);
+        if (r)
+                return r;
+        else
+                return RD_CMP(a->partition, b->partition);
```
pranavrth, line 3088:
```suggestion
       return r || RD_CMP(a->partition, b->partition)
```

src/rdkafka_partition.c
```diff
@@ -3013,6 +3072,22 @@ int rd_kafka_topic_partition_cmp(const void *_a, const void *_b) {
...
+/**
+ * @brief Compare topic partitions \p a and \p b by topic id first
+ *        and then by partition.
+ */
+int rd_kafka_topic_partition_by_id_cmp(const void *_a, const void *_b) {
+        const rd_kafka_topic_partition_t *a = _a;
+        const rd_kafka_topic_partition_t *b = _b;
+        rd_kafka_Uuid_t topic_id_a = rd_kafka_topic_partition_get_topic_id(a);
+        rd_kafka_Uuid_t topic_id_b = rd_kafka_topic_partition_get_topic_id(b);
+        int r                      = rd_kafka_Uuid_cmp(topic_id_a, topic_id_b);
```
pranavrth, line 3084: Better name for `r` so that it is more readable in next statements. Maybe `are_topic_ids_equal`?

src/rdkafka_partition.c
```diff
@@ -3058,6 +3133,33 @@ static int rd_kafka_topic_partition_list_find0(
...
+ *        \p cmp.
+ * @returns the elems[] index or -1 on miss.
+ */
+static int rd_kafka_topic_partition_list_find_by_id0(
+    const rd_kafka_topic_partition_list_t *rktparlist,
+    rd_kafka_Uuid_t topic_id,
+    int32_t partition,
+    int (*cmp)(const void *, const void *)) {
+        rd_kafka_topic_partition_t *rktpar =
+            rd_kafka_topic_partition_new("", partition);
```
pranavrth, line 3147: I think we should have a new method to create `topic_partition` with `topic_id` and `partition`

src/rdkafka_proto.h
```diff
@@ -584,11 +584,30 @@ typedef struct rd_kafka_Uuid_s {
...
+
 
+/**
+ * Initialize given UUID to zero UUID.
+ *
+ * @param uuid UUID to initialize.
+ */
+static RD_INLINE RD_UNUSED void rd_kafka_Uuid_init(rd_kafka_Uuid_t *uuid) {
+        memset(uuid, 0, sizeof(*uuid));
+}
```
pranavrth, line 604: Shall we remove this function if we are not using it? Just a thought.

src/rdkafka_request.c
```diff
@@ -211,6 +211,7 @@ int rd_kafka_err_action(rd_kafka_broker_t *rkb,
  */
 rd_kafka_topic_partition_list_t *rd_kafka_buf_read_topic_partitions(
     rd_kafka_buf_t *rkbuf,
+    rd_bool_t use_topic_id,
```
pranavrth, line 214: Just for information that `use_topic_name` is added in the later PR.

src/rdkafka_request.c
```diff
@@ -320,6 +334,7 @@ int rd_kafka_buf_write_topic_partitions(
     const rd_kafka_topic_partition_list_t *parts,
     rd_bool_t skip_invalid_offsets,
     rd_bool_t only_invalid_offsets,
+    rd_bool_t use_topic_id,
```
pranavrth, line 337: Just for the information - `use_topic_name` is added in the later PR.

src/rdkafka_request.c
```diff
@@ -1004,7 +1027,8 @@ void rd_kafka_OffsetForLeaderEpochRequest(
             RD_KAFKA_TOPIC_PARTITION_FIELD_END};
         rd_kafka_buf_write_topic_partitions(
             rkbuf, parts, rd_false /*include invalid offsets*/,
-            rd_false /*skip valid offsets */, fields);
+            rd_false /*skip valid offsets */, rd_false /* use_topic name */,
```
pranavrth, line 1030:
```suggestion
            rd_false /*skip valid offsets */, rd_false /* don't use topic_id */,
```

pranavrth: Replace this in multiple places.

pranavrth (changes requested):
First Pass. Looks fine mostly.

src/rdkafka.h
```diff
@@ -4546,6 +4552,20 @@ rd_kafka_consumer_group_metadata_new_with_genid(const char *group_id,
...
+ * @brief Get member id of a group metadata.
+ *
+ * @param group_metadata The group metadata
+ *
+ * @returns The member id contained in the passed \p group_metadata.
+ *
+ * @remark The returned pointer has the same lifetime as \p group_metadata.
+ */
+RD_EXPORT
+const char *rd_kafka_consumer_group_metadata_member_id(
```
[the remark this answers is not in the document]
emasab: Needed for the mock broker implementation, but not only there, it's a functionality that is available in J. [ConsumerGroupMetadata](https://kafka.apache.org/35/javadoc/org/apache/kafka/clients/consumer/ConsumerGroupMetadata.html#memberId()) too

src/rdkafka_partition.c
emasab: `RD_KAFKA_UUID_ZERO` is returned if `parpriv` is missing, so there's no need to create the `parpriv` to set that value. But if it's already present we need to overwrite its `topic_id` value

src/rdkafka_partition.c
emasab: I can add an optimization if it's `RD_KAFKA_UUID_ZERO`

src/rdkafka_proto.h
emasab: I added it in case we had to do something more that just setting to zero when initializing a Uuid on the stack, but probably it won't be needed and we can just use `= RD_KAFKA_UUID_ZERO`

src/rdkafka_partition.c
```diff
@@ -2568,7 +2568,21 @@ rd_kafka_topic_partition_list_t *rd_kafka_topic_partition_list_new(int size) {
...
+rd_kafka_topic_partition_t *
+rd_kafka_topic_partition_new_with_topic_id(rd_kafka_Uuid_t topic_id,
+                                           int32_t partition) {
+        rd_kafka_topic_partition_private_t *parpriv;
+        rd_kafka_topic_partition_t *rktpar = rd_calloc(1, sizeof(*rktpar));
+
+        rktpar->partition = partition;
 
+        if (!rd_kafka_Uuid_cmp(topic_id, RD_KAFKA_UUID_ZERO))
+                return rktpar;
```
pranavrth, line 2580: Can we remove this hard coding so that we don't have to remember doing this everytime later? Creating parpriv always when topic_id is needed even with UUID_ZERO?

pranavrth (commented):
Can build be fixed?

pranavrth (approved):
LGTM!

PR #4408, opened 2023-08-25
Added Pragmas for the Debug Libs from Openssl

KevDi:
If Openssl is build from sources, the Debug Version of the Library ends with an `d`. Added an additional check to set the Pragma to the Correct Library Files.

### DIFF
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -40,6 +40,10 @@
 #ifdef _WIN32
 #include <wincrypt.h>
 #pragma comment(lib, "crypt32.lib")
+#ifdef _DEBUG
+#pragma comment(lib, "libcryptod.lib")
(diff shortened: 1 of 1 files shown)


KevDi:
I'm not sure if this is the best way to solve this, or if this breaks some of the Packaging Tools (VCPKG) because i was only able to test it with conan and there it works with that fix. So i'm open to any recommendations on this.

KevDi:
@emasab any news on that? I also added a PR on the Conan Side to fix this directly in Conan. ([Conan PR](https://github.com/conan-io/conan-center-index/pull/19496)) the question is what would be the best solution here?

KevDi:
@emasab are there any news on that?

KevDi:
PR Could be closed Issue was solved on Conan Side by changing the Pragmas

PR #4419, opened 2023-08-30
Use rd_kafka_compression_t in PushTelemetry

anchitj:
Also bump numbers for GetTelemetry and PushTelemetry reqs to align with the broker


milindl (approved):
lgtm, thanks for this

PR #4431, opened 2023-09-13
Move can_q_contain_fetched_msgs inside q_serve

### LINKED ISSUE #4421: max.poll.interval.ms is not reset while using consume_cb
Illustrative code to reproduce,

```C
        // Set max.poll.interval.ms and session.timeout.ms to 10000
        rd_kafka_conf_set_consume_cb(conf, consume_cb);
        rk = rd_kafka_new(RD_KAFKA_CONSUMER, conf, errstr, sizeof(errstr));
        rd_kafka_poll_set_consumer(rk);
        rd_kafka_subscribe(rk, subscription);
        while (1) { rd_kafka_poll(rk, 1000); } /* wait 10 s */
```

Why does it happen? 

We have the check, `can_q_contain_fetched_msgs`, which checks if `rk->rk_rep` can contain consumer messages. We've done this since `rk_rep` is passed to the argument inside `rd_kafka_q_serve`, so we only check that queue. However, internally, `rd_kafka_q_serve` checks if its argument queue is forwarded to any other queue (`fwdq`), and calls `rd_kafka_q_serve` on the  `fwdq`. 

In this case, `fwdq` is the `rkcg_q` which does contain consumer messages. 

A possible solution is to move the check `can_q_contain_fetched_msgs` inside `rd_kafka_q_serve`. That way, the `fwdq` will also be checked.

Also affects `rd_kafka_queue_poll`, `rd_kafka_queue_poll_callback`.

anchitj:
Fixes #4421 

Extended the existing test case to test if polling a queue forwarding to the consumer queue resets the max.poll.interval.ms. This test fails without the changes, but works with it.


src/rdkafka_queue.c
```diff
@@ -520,8 +537,14 @@ int rd_kafka_q_serve(rd_kafka_q_t *rkq,
                 return ret;
         }
 
+        const rd_bool_t can_q_contain_fetched_msgs =
```
milindl, line 540: Variable declaration at start of block

tests/0089-max_poll_interval.c
```diff
@@ -440,7 +452,8 @@ do_test_rejoin_after_interval_expire(rd_bool_t forward_to_another_q) {
 int main_0089_max_poll_interval(int argc, char **argv) {
         do_test();
         do_test_with_log_queue();
-        do_test_rejoin_after_interval_expire(rd_false);
-        do_test_rejoin_after_interval_expire(rd_true);
+        do_test_rejoin_after_interval_expire(rd_false, rd_false);
+        do_test_rejoin_after_interval_expire(rd_true, rd_false);
+        do_test_rejoin_after_interval_expire(rd_false, rd_true);
```
milindl, line 457: I think we should add a test for the specific case that brought this bug to light, too, which uses `rd_kafka_conf_set_consume_cb` to set a consume_cb, and the calls `rd_kafka_poll`.

milindl (commented):
Implementation looks okay, left some comments for the test.
Please add changelog.md entry too

src/rdkafka_queue.c
anchitj: Done

tests/0089-max_poll_interval.c
anchitj: Added a new test, though `test_wait_event` is unable to capture RD_KAFKA_EVENT_REBALANCE not sure why, so haven't added any assert in the test. But added a poll, which should fail if the consumer has left the group.

tests/0089-max_poll_interval.c
```diff
@@ -430,17 +440,69 @@ do_test_rejoin_after_interval_expire(rd_bool_t forward_to_another_q) {
...
+
+        test_consumer_subscribe(rk, topic);
+        TEST_SAY("Subscribed to %s and sleeping for 5 s\n", topic);
+        rd_sleep(5);
+        rd_kafka_poll(rk, 10);
+        TEST_SAY(
+            "Polled and sleeping again for 6s. Max poll should be reset\n");
+        rd_sleep(6);
+
+        // Poll should work
```
milindl, line 491: Comment in the /* */ format

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ librdkafka v2.3.0 is a feature release:
  * Fix for stored offsets not being committed if they lacked the leader epoch (#4442).
  * Fix to ensure permanent errors during offset validation continue being retried and
    don't cause an offset reset (#4447).
+ * Fix to ensure max.poll.interval.ms is reset when rd_kafka_poll is called with consume_cb (#4431).
```
milindl, line 28: Add below in the fixes also, inside the Consumer Fixes section

tests/0089-max_poll_interval.c
```diff
@@ -430,17 +440,67 @@ do_test_rejoin_after_interval_expire(rd_bool_t forward_to_another_q) {
...
+static void consume_cb(rd_kafka_message_t *rkmessage, void *opaque) {
+        TEST_SAY("Consume callback\n");
+}
+
+static void do_test_max_poll_reset_with_consumer_cb(void) {
+        const char *topic = test_mk_topic_name("0089_max_poll_interval", 1);
+        rd_kafka_conf_t *conf;
+        char groupid[64];
+        rd_kafka_t *rk                    = NULL;
+        rd_kafka_queue_t *consumer_queue  = NULL;
```
milindl, line 461: Unused queue variables, can be removed

tests/0089-max_poll_interval.c
```diff
@@ -430,17 +440,67 @@ do_test_rejoin_after_interval_expire(rd_bool_t forward_to_another_q) {
...
+        rd_kafka_queue_t *forwarder_queue = NULL;
+        rd_kafka_event_t *event           = NULL;
```
milindl, line 463: Same for event, it's not set anywhere, only freed

tests/0089-max_poll_interval.c
```diff
@@ -430,17 +440,67 @@ do_test_rejoin_after_interval_expire(rd_bool_t forward_to_another_q) {
...
         rd_kafka_destroy(rk);
 
         SUB_TEST_PASS();
 }
 
+static void consume_cb(rd_kafka_message_t *rkmessage, void *opaque) {
+        TEST_SAY("Consume callback\n");
+}
+
+static void do_test_max_poll_reset_with_consumer_cb(void) {
```
milindl, line 456: Documentation comment above this test
We can also mention issue # for the original case

milindl (approved):
Approved with minor comments

CHANGELOG.md
anchitj: Added

tests/0089-max_poll_interval.c
anchitj: Thanks, added.

PR #4433, opened 2023-09-14
Fix loop of OffsetForLeaderEpoch requests on quick leader changes

### LINKED ISSUE #4425: OffsetForLeaderEpoch loop of failed requests with multiple leader changes
Description
===========
Loop of OffsetForLeaderEpoch calls when multiple leader changes happen and one of them is being retried because of an error.

How to reproduce
================
To reproduce it there should be an initial partition leader change that triggers an OffsetForLeaderEpoch request. The request should fail while a second leader change happens. The corresponding current leader epoch isn't updated and there's a loop of failing requests.

Checklist
=========

 - [x] librdkafka version (release number or git tag): 2.1.0+
 - [x] Apache Kafka version: 3.5.1
 - [x] librdkafka client configuration: any
 - [x] Operating system: any
 - [ ] Critical issue

milindl:
Fixes #4425


mjd95:
Thanks for the PR! We have a pretty reliable scenario for triggering https://github.com/confluentinc/librdkafka/issues/4425 (we do a reassignment on a small topic so that there are at least two partition state changes in short succession, the first and then all subsequent OffsetForLeaderEpoch requests fail with FencedLeaderEpoch).

We ran the consumers from this branch and confirmed that we don't see the issue with this patch.

milindl:
Thanks for confirming it with some independent testing @mjd95!

tests/0139-offset_validation_mock.c
```diff
@@ -212,6 +212,95 @@ static void do_test_ssl_error_retried(void) {
...
+         * be sent to broker 2. We need to first fail it with
+         * an error, and then give enough time to change the leader before
+         * returning a success. */
+        rd_kafka_mock_broker_push_request_error_rtts(
+            mcluster, 2, RD_KAFKAP_OffsetForLeaderEpoch, 1,
+            RD_KAFKA_RESP_ERR_KAFKA_STORAGE_ERROR, 900);
+
+        rd_kafka_mock_broker_push_request_error_rtts(
+            mcluster, 2, RD_KAFKAP_OffsetForLeaderEpoch, 1,
+            RD_KAFKA_RESP_ERR_NO_ERROR, 1000);
```
pranavrth, line 272: We can merge these two.

tests/0139-offset_validation_mock.c
```diff
@@ -212,6 +212,95 @@ static void do_test_ssl_error_retried(void) {
...
+ * a. Leader Change - causes OffsetForLeaderEpoch
+ *     request 'A'.
+ * b. Request 'A' fails with a retriable error, and we retry it.
+ * c. While waiting for Request 'A', the leader changes again, and we send a
+ *    Request 'B', but the leader epoch is not updated correctly in this
+ *    request, causing a loop.
+ *
+ * See #4425.
+ */
+static void do_test_two_leader_changes(void) {
```
pranavrth, line 229: Can you confirm that this test fails with the old code?

pranavrth (approved):
LGTM! Just a question and a nit.

tests/0139-offset_validation_mock.c
milindl: Done

tests/0139-offset_validation_mock.c
milindl: Yep, it fails with a timeout after we enter the infinite loop

ktrance:
@emasab Do you have an ETA on when 2.3 will be released? We believe this might the issue we are experiencing.

Issue #4384, opened 2023-08-04
Manual offset incrementation/auto commiting stopped working on Azure eventhub since version 2.1.0


pbizon:
Read the FAQ first: https://github.com/confluentinc/librdkafka/wiki/FAQ

Do NOT create issues for questions, use the discussion forum: https://github.com/confluentinc/librdkafka/discussions

Description
===========
Changes introduced in this PR:
https://github.com/confluentinc/librdkafka/commit/a0871a1a8ff435a9342f1a60dca4c696a54c7ae2#diff-14f07c2626b2863a69e4b9b625720d25a860298da6d8f3dd81289b617ad7a9b5
cause that before setting a new offset in `rd_kafka_topic_partition_list_set_offsets` we compare not only offsets but also `leader_epochs`.

```
    if (rd_kafka_fetch_pos_cmp(&rktp->rktp_stored_pos,
                                &rktp->rktp_committed_pos) >
        0) {
        ...
        }
```
```
rd_kafka_fetch_pos_cmp(const rd_kafka_fetch_pos_t *a,
                       const rd_kafka_fetch_pos_t *b) {
        if (a->leader_epoch < b->leader_epoch)
                return -1;
        else if (a->leader_epoch > b->leader_epoch)
                return 1;
        else if (a->offset < b->offset)
                return -1;
        else if (a->offset > b->offset)
                return 1;
        else
                return 0;
}
```
Everything works fine with Apache Kafka implementation which send `-1` as leaderEpoch in `OffsetFetchResponse`.
```
{"severity":7,"fac":"OFFSETFETCH","message":"[thrd:main]: GroupCoordinator/1: OffsetFetchResponse: test-topic [0] offset 1, leader epoch -1, metadata 0 byte(s): NO_ERROR"}
```
Unfortunatelly Azure eventhub sends `0`.
```
{"severity":7,"fac":"OFFSETFETCH","message":"[thrd:main]: GroupCoordinator/0: OffsetFetchResponse: test-topic [0] offset -1, leader epoch 0, metadata 0 byte(s): NO_ERROR"}
[KAFKA] {"severity":7,"fac":"OFFSET","message":"[thrd:main]: Topic test-topic [2]: stored offset 316172 (leader epoch -1), committed offset INVALID (leader epoch 0): not including in commit"}
[...]

PR #4434, opened 2023-09-14
Add leader epoch to control messages

emasab:
to make sure they're stored for committing even without a subsequent fetch message

See #4384.


tests/0103-transactions.c
```diff
@@ -1102,6 +1103,89 @@ static void do_test_empty_txn(rd_bool_t send_offsets, rd_bool_t do_commit) {
         SUB_TEST_PASS();
 }
 
+
+/**
+ * @brief A control message should increase stored offset and
+ *        that stored offset should have correct leader epoch
+ *        an be included in commit.
```
milindl, line 1110:
```suggestion
 *        and be included in commit. (See #4384.)
```

milindl (approved):
Thanks for the fix, great PR

PR #4438, opened 2023-09-18
Fix for idempotent producer fatal errors, triggered after a possibly persisted message state

emasab:
After a possibly persisted error, such as a disconnection or a timeout, next expected sequence was increased, leading to a fatal error if the message wasn't persisted and the second one in queue failed with an `OUT_OF_ORDER_SEQUENCE_NUMBER`. 

The error could contain the message "sequence desynchronization" with just one possibly persisted error or "rewound sequence number" in case of multiple errored messages.

Solved by treating the possible persisted message as _not_ persisted, and expecting a `DUPLICATE_SEQUENCE_NUMBER` error in case it was or `NO_ERROR` in case it wasn't, in both cases the message will be considered delivered.


CHANGELOG.md
```diff
@@ -4,6 +4,23 @@ librdkafka v2.2.1 is a maintenance release:
...
  * Fixed ListConsumerGroupOffsets not fetching offsets for all the topics in a group with Apache Kafka version below 2.4.0.
+ * Fix for idempotent producer fatal errors, triggered after a possibly persisted message state (#4438).
+
+
+## Fixes
+
+### Idempotent producer fixes
+
+ * After a possibly persisted error, such as a disconnection or a timeout, next expected sequence
+   was increased, leading to a fatal error if the message wasn't persisted and
```
pranavrth, line 15:
```suggestion
   used to increase, leading to a fatal error if the message wasn't persisted and
```

src/rdkafka_request.c
```diff
@@ -3343,17 +3343,15 @@ static int rd_kafka_handle_Produce_error(rd_kafka_broker_t *rkb,
...
-                                perr->update_next_ack = rd_false;
+                         * In case the message is possibly persisted
+                         * we still treat it as not persisted,
+                         * expecting DUPLICATE_SEQUENCE_NUMBER
+                         * in case it was persisted or NO_ERROR in case
+                         * it wasn't. Increasing next expected sequence
+                         * prematurely can lead to fatal errors in case
+                         * it wasn't persisted and the second one to retry
+                         * fails with OUT_OF_ORDER_SEQUENCE_NUMBER. */
+                        perr->update_next_ack = rd_false;
```
pranavrth, line 3354: I think we should remove the part from "Increasing next expected....". This paragraph is explaining the current issue that we are fixing not what next line of the code is doing. It is confusing the statement.

tests/0143-idempotence_mock.c
```diff
@@ -0,0 +1,339 @@
...
+                }
+        }
+
+        /* Produce n messages that will be retried, msgids 6->(6+n-1). */
+        test_produce_msgs2_nowait(rk, "mytopic", 0, 0, 0, n, NULL, 64,
+                                  &remains);
+
+        /* Wait that messages are sent, then set it down and up again.
+         * Causing a "possibly persisted" error that increased next_ack,
+         * with subsequent fatal error when retrying the same sequences. */
```
pranavrth, line 252: Let's write whats happening instead of the issue that we are fixing. This is confusing.

tests/0143-idempotence_mock.c
```diff
@@ -0,0 +1,339 @@
...
+        test_curr->is_fatal_cb   = error_is_fatal_cb;
+        /* Only allow an error from the disconnection below. */
+        allowed_error = RD_KAFKA_RESP_ERR__TRANSPORT;
+
+        /* Produce 5 messages without error first, msgids 1->5. */
+        test_produce_msgs2(rk, "mytopic", 0, 0, 0, 5, NULL, 64);
+        rd_kafka_flush(rk, -1);
+
+        /* First sequence is for the immediately produced reply,
+         * never delivered because of the disconnection. */
```
pranavrth, line 226: Better description.

pranavrth (approved):
LGTM!

Minor comments.

pranavrth (approved):
The new test and changes look fine.

PR #4440, opened 2023-09-20
Unit tests push telemetry encode decode

anchitj:


milindl (approved):
mostly looks good, a few comments.

also later on, we should remove the fprintfs and use RD_UT_SAY or something similar, but leaving that in as the function is used for debugging now

milindl (approved):

dtheodor:
When do you plan to have KIP 714 working? Is there a roadmap published somewhere?

milindl:
@dtheodor , KIP-714 is not yet accepted by Apache Kafka: https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability

So, there's no roadmap yet. We need to wait for acceptance and for it to make it to Apache Kafka before we can put it into librdkafka.

PR #4442, opened 2023-09-21
Fix for stored offsets not being committed if they lacked the leader epoch

emasab:
Stored offsets where excluded from the commit if the leader epoch was less than committed epoch, as it's possible if leader epoch is the default -1.
This didn't happen in Python, Go and .NET bindings when stored position was taken from the message. Solved by only checking that offset is greater than the committed one.

See #4384.


CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ librdkafka v2.2.1 is a maintenance release:
...
+
+
+## Fixes
+
+### Consumer fixes
+
+  * Stored offsets where excluded from the commit if the leader epoch was
+    less than committed epoch, as it's possible if leader epoch is the default -1.
+    This didn't happen in Python, Go and .NET bindings when stored position was
+    taken from the message. Solved by only checking that offset is greater
```
milindl, line 17: > Solved by only checking that offset is greater
    than committed one.

to something like this?

> Solved by checking only that the stored offset is greater
    than committed one, if either stored or committed leader epoch is -1.

milindl (approved):
lgtm, some minor stuff

PR #4303, opened 2023-05-31
Upgraded OpenSSL to 3.0.8

migarc1:

PR #4454, opened 2023-09-29
OpenSSl partial bump to 3.0.11

milindl:
See #4303. Moving to internal branch for CI.

We will stick to 3.0.11 with this PR (LTS version)

vcpkg doesn't yet have anything greater than 3.0.8, so this is only a partial upgrade.


emasab (approved):

pranavrth:
Can we check on the changes between 3.0.8 and 3.0.11. We should not have different version for different OS. If there is no major change then we should wait for vcpkg to upgrade to 3.0.11

emasab:
I see but while it's not available it's better to most up to date version on each OS, to avoid releasing with CVEs that are already fixed.

PR #4455, opened 2023-09-29
librdkafka v2.3.0

emasab:


mensfeld:
@emasab is this the "release" PR? That is, will there be a beta or something? I have "my" repros that I could use to thoughtfully test it prior to stable

emasab:
No, this is for creating a release candidate, there will be more commits later

emasab:
@mensfeld RC3 should be the last RC before the final release we plan before next Monday

mensfeld:
@emasab thank you. I will do my best to run RC3 on my test suites "just in case" :)

PR #4458, opened 2023-10-06
Fix public API for mock requests

milindl:
We exposed accidentally rd_ts_t in the publicly exposed API for mock broker as a part of the KIP-580 commit.
Fixes that.
We didn't do a release for this, so public API guarantee is not broker. 

Additionally marks some file-local methods static.

### DIFF
--- a/src/rdkafka_mock.c
+++ b/src/rdkafka_mock.c
@@ -41,7 +41,11 @@
 #include "rdkafka_mock.h"
 #include <stdarg.h>
 
+typedef struct rd_kafka_mock_request_s rd_kafka_mock_request_t;
+
 static void rd_kafka_mock_cluster_destroy0(rd_kafka_mock_cluster_t *mcluster);
+static rd_kafka_mock_request_t *
+rd_kafka_mock_request_new(int32_t id, int16_t api_key, rd_ts_t timestamp);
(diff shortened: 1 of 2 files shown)


src/rdkafka_mock.c
```diff
@@ -41,7 +41,11 @@
 #include "rdkafka_mock.h"
 #include <stdarg.h>
 
+typedef struct rd_kafka_mock_request_s rd_kafka_mock_request_t;
+
 static void rd_kafka_mock_cluster_destroy0(rd_kafka_mock_cluster_t *mcluster);
+static rd_kafka_mock_request_t *
+rd_kafka_mock_request_new(int32_t id, int16_t api_key, rd_ts_t timestamp);
```
mahajanadhitya, line 48: Instead of rd_ts_t pass int64_t as argument and convert it internally as in future this might need to be exposed publicly and even if not the casting can be done internally in the function rather than externally.

mahajanadhitya (changes requested):

src/rdkafka_mock.c
milindl: Done, addressed

emasab (approved):
LGTM

PR #4462, opened 2023-10-10
Respect timeout in rd_kafka_query_watermark_offsets.

rescrv:
The query_watermark_offsets function will loop forever, never timing out.  Thread ts_end through the rest of the function.


milindl:
Closing with #4460 being merged already.
Thanks for the PR anyway @rescrv !

PR #4479, opened 2023-10-23
Docker login before pull

emasab:


.semaphore/semaphore.yml
```diff
@@ -110,6 +110,10 @@ blocks:
       env_vars:
         - name: CFLAGS
           value: -std=gnu90 # Test minimum C standard, default in CentOS 7
+      prologue:
+        commands:
+          - . vault-sem-get-secret dockerhub-semaphore-cred
```
milindl, line 115: Don't need to fetch the credentials here as a pre-job script is executed on every semaphore machine and exports the dockerhub api keys

milindl (approved):
lgtm after CI

PR #4480, opened 2023-10-23
Fix calls with empty list

emasab:
DescribeTopics and ListOffsets should return an empty array if and empty array is passed as parameter.
Consistent with Java and user friendly for users using filter and map.


milindl (approved):
lgtm, just one minor thing

Issue #343, opened 2015-08-05
rdkafka doesn't honor the same broker ID when moved to new host


liljenback:
If migrating an existing broker to a new host and re-using its broker ID (from the host that is decommissioned) rdkafka still remember the old host name broker ID combination even when it is no longer available (brought offline). The client does not start using the new host. This results in messages being lost for partitions that are reassigned to the new host. The Kafka documentation seems to hint that the clients should be able to start using the new host with the same ID without having to be restarted.

edenhill:
By messages lost you mean they are succesfully produced to the old broker, right?

liljenback:
Right, the messages will be produced to the old broker but in this case it's offline.

edenhill:
Should be fixed now.

edenhill:
If possible please try verifying on your end

PR #4494, opened 2023-11-03
Fix for brokers with different Ids but same host:port

mfleming:
The Kafka protocol allows for brokers to have multiple host:port pairs for a given node Id, e.g. see UpdateMetadata request which contains a live_brokers list where each broker Id has a list of host:port pairs. It follows from this that the thing that uniquely identifies a broker is its Id, and not the host:port.

The behaviour right now is that if we have multiple brokers with the same host:port but different Ids, the first broker in the list will be updated to have the Id of whatever broker we're looking at as we iterate through the brokers in the Metadata response in
rd_kafka_parse_Metadata0(), e.g.

 Step 1. Broker[0].id = Metadata.brokers[0].id
 Step 2. Broker[0].id = Metadata.brokers[1].id
 Step 3. Broker[0].id = Metadata.brokers[2].id

A typical situation where brokers have the same host:port pair but differ in their Id is if the brokers are behind a load balancer.

The NODE_UPDATE mechanism responsible for this was originally added in b09ff60c ("Handle broker name and nodeid updates (issue #343)") as a way to forcibly update a broker hostname if an Id is reused with a new host after the original one was decommissioned. But this isn't how the Java Kafka client works, so let's use the Metadata response as the source of truth instead of updating brokers if we can only match by their host:port.

Here's a screenshot of the traffic going to brokers sat behind a load balancer before (left) and after (right) this patch which illustrates that new connections are opened and the load is balanced correctly.
Screenshot 2023-11-03 at 13 50 20

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -5236,6 +5236,8 @@ void rd_kafka_group_list_destroy(const struct rd_kafka_group_list *grplist);
 RD_EXPORT
 int rd_kafka_brokers_add(rd_kafka_t *rk, const char *brokerlist);
 
+RD_EXPORT
+int32_t *rd_kafka_broker_get_learned_ids(rd_kafka_t *rk, size_t *cntp);
 
 
 /**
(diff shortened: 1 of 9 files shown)


mfleming:
@emasab would you mind taking a look at this PR?

PR #4500, opened 2023-11-06
add initial malware scan support

janjwerner-confluent:


janjwerner-confluent:
@emasab 
Can you please look at the build failures related to package downloads. I believe that the failures are not related to the addition of the malware scans.

PR #4507, opened 2023-11-15
Add max telemetry bytes

anchitj:


src/rdkafka_telemetry.c
```diff
@@ -225,6 +226,14 @@ static void rd_kafka_send_push_telemetry(rd_kafka_t *rk,
         // TODO: Cycle through compression types
         rd_kafka_compression_t compression_type = RD_KAFKA_COMPRESSION_GZIP;
 
+        if (metrics_payload_size > rk->rk_telemetry.telemetry_max_bytes) {
+                rd_kafka_log(
+                    rk, LOG_WARNING, "TELEMETRY",
+                    "Metrics payload size %" PRIdsz
+                    " exceeds telemetry_max_bytes %d specified by the broker.",
```
milindl, line 233: The format string will PRId32 for int32

src/rdkafka_telemetry.c
```diff
@@ -271,6 +280,19 @@ void rd_kafka_handle_push_telemetry(rd_kafka_t *rk, rd_kafka_resp_err_t err) {
                 rd_kafka_dbg(rk, TELEMETRY, "PUSHERR",
                              "PushTelemetryRequest failed: %s",
                              rd_kafka_err2str(err));
+                // Non-retriable errors
+                // TODO: What to do for TELEMETRY_TOO_LARGE?
```
milindl, line 284: I guess we can just add a comment. Or at max a log (even though we will have logged a warning already), based on the discussion with the Java clients team, we need to do nothing special.

milindl (approved):
Looks okay - just a bunch of minor stuff

src/rdkafka_telemetry.c
anchitj: Updated

src/rdkafka_telemetry.c
anchitj: Removed the comment

milindl:
LGTM

PR #4513, opened 2023-11-18
Fix BoringSSL by supplying string length to `X509_VERIFY_PARAM_set1_host` explicitly

felixschlegel:
BoringSSL requires the calls to `X509_VERIFY_PARAM_set1_host` to provide explicit string length, see:

```c
  if (name == NULL || namelen == 0) {
    // Unlike OpenSSL, we reject trying to set or add an empty name.
    return 0;
  }
```
 
https://github.com/google/boringssl/blob/39cc892c73d6c3faf2e604c44509f132c232f24c/crypto/x509/x509_vpm.c#L83-L86

This (trivial) PR fixes this issue, and allows BoringSSL to be used with librdkafka.


felixschlegel:
@emasab could you please take a look at this?

emasab:
@felixschlegel Thanks for your PR. Are you sure this is the only change that's required? I've tried compiling and linking with BoringSSL and there are these different things:
```
rdkafka_cert.c: In function ‘rd_kafka_cert_new’:
rdkafka_cert.c:228:39: error: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Werror=sign-compare]
  228 |                         for (i = 0; i < sk_X509_num(cas); i++) {
      |                                       ^
In file included from rdkafka_ssl.c:36:
rdkafka_ssl.c: In function ‘rd_kafka_ssl_ctx_init’:
rdkafka_ssl.c:1622:22: error: format ‘%lx’ expects argument of type ‘long unsigned int’, but argument 11 has type ‘int’ [-Werror=format=]
 1622 |                      "Using %sOpenSSL version %s "
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
rdkafka_int.h:886:66: note: in definition of macro ‘rd_kafka_dbg’
  886 |                                       (RD_KAFKA_DBG_##ctx), fac, __VA_ARGS__); \
      |                                                                  ^~~~~~~~~~~
rdkafka_ssl.c: In function ‘rd_kafka_ssl_hmac’:
rdkafka_ssl.c:1872:14: error: implicit declaration of function ‘HMAC’ [-Werror=implicit-function-declaration]
 1872 |         if (!HMAC(evp, (const unsigned char *)in->ptr, (int)in->size, saltplus,
      |              ^~~~
cc1: all warnings being treated as errors
make[1]: *** [../mklove/Makefile.base:97: rdkafka_cert.o] Error 1
make[1]: *** Waiting for unfinished jobs....
rdkafka_admin.c: In function ‘rd_kafka_UserScramCredentialUpsertion_new’:
rdkafka_admin.c:5911:21: error: implicit declaration of function ‘RAND_priv_bytes’ [-Werror=implicit-function-declaration]
 5911 |                 if (RAND_priv_bytes(random_salt, sizeof(random_salt)) == 1) {
      |                     ^~~~~~~~~~~~~~~
``` 

Was checking and `RAND_priv_bytes` doesn't seem to exists in BoringSSL and sk_X509_num (OPENSSL_sk_num) returns a `size_t` instead of an `int`. Is there something I'm missing?

gmilos:
@emasab our use case is: https://github.com/swift-server/swift-kafka-client and we're already capable of building successfully with BoringSSL, as of: https://github.com/swift-server/swift-kafka-client/pull/135. It doesn't work at runtime, because of the issue in this PR.

BoringSSL version comes from swift-nio-ssl, with the latest version using: https://github.com/apple/swift-nio-ssl/commit/13460776e7f072596d1cd730f56d807511d11998. I'm suspecting you hit issues because of using a different (newer?) version of BoringSSL.

To summarise, this PR is needed to unblock us. It's necessary, but might not be sufficient once we start testing more thoroughly. We don't run into the specific compilation problems you've listed though.

gmilos:
@emasab yes, we're compiling with 2.3.0 just fine, see:
https://github.com/swift-server/swift-kafka-client/tree/20faefb06a0c184f8c7cfa9c11731816ea6b0ff3/Sources/Crdkafka

Regarding the specific compilation issues you hit, we've got the important ones resolved. Including:
1. `RAND_priv_bytes` provided through this: https://github.com/swift-server/swift-kafka-client/pull/148
2. `HMAC` is defined here: https://github.com/apple/swift-nio-ssl/blob/7613886ed9230ec9e77b7ae63470d808967856e2/Sources/CNIOBoringSSL/crypto/fipsmodule/hmac/hmac.c#L69

The patch proposed in the PR is benign for OpenSSL in general, so would you mind merging it. It'd be super important for us, and others shouldn't care.

emasab:
To clarify this PR doesn't give support for BoringSSL, it still needs these defines that are in your `swift-kafka-client`.
We're open to receive a patch for complete, even if unofficial, BoringSSL support.
Given that as you said the proposed patch is benign for OpenSSL. I approve it but will need approval from a different team member that's our policy for community PRs.

emasab (dismissed):

felixschlegel:
Thank you so much for your feedback @emasab!

emasab (approved):

felixschlegel:
@emasab I think this is ready to merge, thanks again for your collaboration here!

gmilos:
@emasab what's the process for merging? We don't have sufficient perms.

milindl (approved):

PR #4524, opened 2023-11-28
Uniform test code across scripts and KRaft mode

emasab:


tests/broker_version_tests.py
```diff
@@ -46,11 +48,18 @@ def test_it(version, deploy=True, conf={}, rdkconf={}, tests=None,
...
-        cmd = 'bash --rcfile <(cat ~/.bashrc; echo \'PS1="[TRIVUP:%s@%s] \\u@\\h:\\w$ "\')' % (  # noqa: E501
-            cluster.name, version)
+        if inherit_env:
+            env = dict(os.environ, **rdkafka.env)
+        else:
+            env = dict(rdkafka.env)
+        trivup = f'[TRIVUP:{cluster.name}@{version}] '
+        PS1 = ((trivup + env['PS1']) if 'PS1' in env
+               else trivup + '\\u@\\h:\\w$ ')\
+            .translate(str.maketrans({'\'': '\\\''}))
```
pranavrth, line 58: What does this translate do and why do we need it?

tests/interactive_broker_version.py
```diff
@@ -33,229 +25,16 @@ def version_as_number(version):
...
+    Then run librdkafka's regression tests. Use inherited environment.
     """
-
-    print('## Test version %s' % version)
-
-    cluster = Cluster('LibrdkafkaTestCluster', root_path, debug=debug)
-
-    if conf.get('sasl_oauthbearer_method') == 'OIDC':
-        oidc = OauthbearerOIDCApp(cluster)
-
```
pranavrth, line 48: I think we should move whole deleted part to test_it instead of removing them. If we use other version in future then we should have fail fast working fine. Same for the other file as well.

tests/interactive_broker_version.py
```diff
@@ -33,229 +25,16 @@ def version_as_number(version):
...
-    try:
-        os.remove(test_conf_file)
-    except BaseException:
-        pass
-
-    cluster.stop(force=True)
-
-    cluster.cleanup(keeptypes=['log'])
-    return ret
+    conf['test_mode'] = 'bash'
```
pranavrth, line 34: This should be part of the test config instead of hardcoding.

pranavrth (changes requested):

tests/LibrdkafkaTestApp.py
```diff
@@ -199,13 +199,16 @@ def start_cmd(self):
...
             False)
-        self.env_add(
-            'ZK_ADDRESS',
-            self.cluster.get_all(
-                'address',
-                '',
-                ZookeeperApp)[0],
-            False)
+
+        zookeeper = self.cluster.get_all(
```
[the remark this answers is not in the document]
emasab: No, `get_all `always returns a list

tests/broker_version_tests.py
emasab: It quotes any single quotes that you have in your existing `env['PS1']`

tests/interactive_broker_version.py
emasab: in interactive_broker_version it was always in bash mode, I've kept existing behavior. Because it's interactive, for other modes there is `broker_version_tests.py`

tests/interactive_broker_version.py
emasab: The `OauthbearerOIDCApp `instance is now created by `LibrdkafkaTestCluster` in `test_it`

pranavrth: I am talking about the whole deleted part. Are we testing them already in the test_it function?

tests/interactive_broker_version.py
pranavrth: But bash might not be available in MacOS. We should give option to the user to choose it.

tests/broker_version_tests.py
pranavrth: Was it a bug earlier?

emasab: no, before it was replacing the PS1, not adding a prefix

tests/interactive_broker_version.py
emasab: this part is replaced by the `LibrdkafkaTestCluster` in `test_it` the code was very similar

tests/interactive_broker_version.py
emasab: Before was calling bash too
`        retcode = subprocess.call(
            cmd,
            env=cmd_env,
            shell=True,
            executable='/bin/bash')`

emasab: `run-test.sh` is requiring bash too, it's requisite for developing with librdkafka already

tests/broker_version_tests.py
pranavrth: I saw that this is not replacing 2 backslashs. Is this expected?

tests/interactive_broker_version.py
pranavrth: Let it be like this for now. We should have zsh support as well in the future though.

tests/broker_version_tests.py
emasab: Here it's only replacing single quotes because later it's written inside single quotes. In:
```c
cmd = f'bash --rcfile <(cat ~/.bashrc; echo \'PS1="{PS1}"\')'
```

pranavrth (dismissed):
LGTM!. Fix build.

pranavrth (approved):
LGTM!

PR #4555, opened 2023-12-12
[KIP-848] Changes related to config, error codes and topic id

pranavrth:
This PR adds the following changes

- New config changes required for the KIP-848 implementation (Hidden for now as not being used)
- Some of the new error codes required with KIP-848
- Added topic id to topic partition with various accessors and operation functions
- Added Getter for member id
- UUID comparator method
- vcpkg version change

Co-authored-by: Emanuele Sabellico <esabellico@confluent.io>


emasab (approved):
LGTM

PR #4556, opened 2023-12-12
Fix a "return by const value" warning on Clang

Quuxplusone:
Commit efbb96681 (which was apparently massive) introduced this function returning "by const value". It's possible this indicates a deeper bug, but for now let's just fix the compiler diagnostic by removing the useless "const".


gullradriel (approved):
That PR also fixes the warning on gcc.

Quuxplusone:
Ping!

Quuxplusone:
@emasab perhaps: Ping!

Quuxplusone:
Duplicates #4504, as it turns out!

PR #4557, opened 2023-12-12
Purge brokers no longer reported in metadata

### LINKED ISSUE #238: Purge brokers no longer reported in metadata
Scenario:

* Start up a topic where some partitions are on broker 0, some are on broker 1
* Move the partitions to a single broker (0)
* Kill broker 1, as it is no longer relevant to the topics currently in service

You will get this sort of log (from the example consumer in this case):

* 1428602624.709 RDKAFKA-3-FAIL: rdkafka#consumer-0: 10.42.0.144:9092/1: Failed to connect to broker at 10.42.0.144:9092: Operation timed out
* 1428602624.709 RDKAFKA-3-ERROR: rdkafka#consumer-0: 10.42.0.144:9092/1: Failed to connect to broker at 10.42.0.144:9092: Operation timed out
* 1428602625.715 RDKAFKA-3-FAIL: rdkafka#consumer-0: 10.42.0.144:9092/1: Failed to connect to broker at 10.42.0.144:9092: Host is down
* 1428602625.715 RDKAFKA-3-ERROR: rdkafka#consumer-0: 10.42.0.144:9092/1: Failed to connect to broker at 10.42.0.144:9092: Host is down
* 1428602718.698 RDKAFKA-3-FAIL: rdkafka#consumer-0: 10.42.0.144:9092/1: Failed to connect to broker at 10.42.0.144:9092: Operation timed out
* 1428602718.698 RDKAFKA-3-ERROR: rdkafka#consumer-0: 10.42.0.144:9092/1: Failed to connect to broker at 10.42.0.144:9092: Operation timed out

Producers also do this.

mfleming:
Brokers that are not in the metadata should be purged from the internal client lists. This helps to avoid annoying "No route to host" and other connection failure messages.

Fixes #238


tests/0145-broker-same-host-port.c
emasab, line 1: Please change this test number to 0149

tests/0146-purge-brokers.c
emasab, line 1: Please change this test number to 0150

src/rdkafka_broker.c
emasab, line 1: There are a few things to remove for updating brokers by matching just the id:

- in `rd_kafka_broker_op_serve`, `updated` becomes a `rd_bool_t`
- condition that checks if nodeid has changed has to be removed
- corresponding action and `rd_kafka_broker_map_partitions` needs to be removed.
- `RD_KAFKA_BROKER_STATE_UPDATE` is to be removed

emasab (changes requested):
Hi @mfleming thanks a lot for this contribution and sorry for letting it wait for long. We want to include these fixes in next version. Both fixes are good, just on broker decommission we want to do some additional checks.

Here are some comments mainly for the first fix:

emasab:
Hi @mfleming can I apply those changes or do you want to continue the PR? Thanks!

mfleming:
Hey sorry for the delay — yeah you can apply those changes if you have the
time. If not I’ll get to them sometime this week.

emasab:
@mfleming thanks, sorry for the delay too, I'm checking it again

emasab:
We're not going to merge this for 2.5.0 that is due in July as we need to do more checks on possible regressions but we want to merge it for a maintenance release in September

mfleming:
> We're not going to merge this for 2.5.0 that is due in July as we need to do more checks on possible regressions but we want to merge it for a maintenance release in September

Thanks for fixing things :)

emasab:
> Thanks for fixing things :)
Thank you for this PR!

benesch:
Just wanted to say a big thank you to both of you—@mfleming for writing this and @emasab for reviewing. We just ran into a slow thread leak in a Kafka consumer at @MaterializeInc that will be fixed by this patch.

pranavrth:
While going through the PR, I found that there is a few issues in the PR.

1. The number of threads are increased as part of this PR. Currently, all the bootstrap servers threads are updated to the real broker thread if they appear in the metadata response. This PR creates new broker threads for the new brokers which are seen in the metadata. The bootstrap servers threads are not reused for this purpose in the PR. Due to this, we generally have 'n*2' brokers threads for 'n' bootstrap servers. Earlier there were only 'n' threads.
2. In the current librdkafka implementation, we always keep bootstrap broker threads even if they are not reported in the metadata. This ensured that we always have bootstrap broker threads if required to rebootstrap. This PR removed this functionality.
3. To fix the issue introduced in point number 2, KIP-899 needs to be implemented.

We need to fix the above issues before releasing this PR and need of more testing. As a result, it won't be part of the upcoming 2.6 release.

emasab:
Closes #4881

emasab:
We agreed KIP-899 and KIP-1102 aren't strictly necessary, in Java client they were disabled by default until recently.
The reason is that when a change in broker set happens slow enough to be detected by the periodic (5 min) metadata refresh, the client cannot remain without brokers.

In any case the metadata response contains at least the majority of KRaft elegible brokers. If majority changes, it must contain at least one broker from previous set.
So it's usually not possible the client remains without brokers, unless the set of brokers changes to fast it cannot reach any of them and that's the case KIP-899 and KIP-1102 are addressing. But this last case is also a problem with current librdkafka code so we leave the improvement for a later PR.

There are a few things left to check, at least:

- removing bootstrap brokers without id (-1) and adding brokers with real ids and advertised hostnames.
- check some fields in rd_kafka_s: `no_idemp_brokers, rk_telemetry.preferred_broker, rk_broker_down_cnt, rk_broker_up_cnt, rk_broker_cnt, ...` when brokers are removed. It could help changing the state to DOWN before removing them.
- maybe there's something to change in the stats_cb, but probably nothing to do
- brokers configured with `rd_kafka_brokers_add` must be removed after usage too
- currently, decommissioned threads are still joined on destroy, is it possible to join them from time to time without accumulating them but also without blocking the main thread because of that? At specific intervals maybe, so they're already terminated when joined
- check thoroughly for memory leaks and use-after-free cases

emasab:
> - check some fields in rd_kafka_s: `no_idemp_brokers, rk_telemetry.preferred_broker, rk_broker_down_cnt, rk_broker_up_cnt, rk_broker_cnt, ...` when brokers are removed. It could help changing the state to DOWN before removing them.

`no_idemp_brokers` seems not being used, `rk_broker_cnt` is decremented in `rd_kafka_broker_thread_main`.  `rk_broker_down_cnt` and `rk_broker_up_cnt` are changed when broker thread receives `RD_KAFKA_OP_TERMINATE`: it calls `rd_kafka_broker_fail` that sets the state to DOWN and changes those counters.

src/rdkafka_broker.c
```diff
@@ -2337,6 +2352,7 @@ void rd_kafka_broker_connect_up(rd_kafka_broker_t *rkb) {
         if (rd_kafka_broker_ApiVersion_supported(
                 rkb, RD_KAFKAP_GetTelemetrySubscriptions, 0, 0, &features) !=
                 -1 &&
+            rkb->rkb_source == RD_KAFKA_LEARNED &&
```
pranavrth, line 2355: Why learned broker is needed for Telemetry subscription call?

src/rdkafka_broker.c
```diff
@@ -3525,10 +3465,13 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
...
                  * are triggered. */
-                rd_kafka_broker_fail(rkb, LOG_DEBUG, RD_KAFKA_RESP_ERR__DESTROY,
-                                     "Client is terminating");
+                rd_kafka_broker_fail(rkb, LOG_DEBUG,
+                                     rd_kafka_broker_destroy_error(rkb->rkb_rk),
+                                     "Decommissioning this broker");
 
                 rd_kafka_broker_prepare_destroy(rkb);
+                /* Release main thread reference here */
+                rd_kafka_broker_destroy(rkb);
```
pranavrth, line 3474: Why this is needed now and was not needed earlier?

src/rdkafka_broker.c
```diff
@@ -4749,18 +4685,25 @@ static int rd_kafka_broker_thread_main(void *arg) {
                 }
         }
 
+        rd_kafka_broker_fail(rkb, LOG_DEBUG, rd_kafka_broker_destroy_error(rk),
+                             "Broker handle is terminating");
+
+        rd_dassert(rkb->rkb_state == RD_KAFKA_BROKER_STATE_DOWN);
```
pranavrth, line 4691: Is this necessary and correct? Let's not force this and add this later after this patch. Might create a problem if we are reaching here somehow in some other state.

pranavrth (changes requested):
Went through the implementation on high level. Going to check deeper. Transaction and telemetry needs more understanding. Will connect with you offline on this. Tests and mock broker remaining as well.

pranavrth (changes requested):

pranavrth:
Please check admin client fanout and target all cases as well where we are not using DESTROY_BROKER error.

emasab:
@mfleming could you please sign the Contributor License Agreement . We'll need it before merging the PR.

mfleming:
> @mfleming could you please sign the Contributor License Agreement [here](https://ironcladapp.com/public-launch/66bcfa630dec9c9aeb2fc0da) . We'll need it before merging the PR.

Signed.

src/rdkafka_broker.c
```diff
@@ -5234,11 +5176,6 @@ void rd_kafka_broker_set_nodename(rd_kafka_broker_t *rkb,
         if (!changed)
                 return;
 
-        if (!rd_kafka_broker_is_addrless(rkb))
-                rd_atomic32_sub(&rkb->rkb_rk->rk_broker_addrless_cnt, 1);
-        else
-                rd_atomic32_add(&rkb->rkb_rk->rk_broker_addrless_cnt, 1);
-
```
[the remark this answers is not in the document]
emasab: We just count the number of logical brokers now (with or without address) and exclude them from broker selection and don't count them for the ALL_BROKERS_DOWN (it'll be one of the learned brokers anyway so it fails at the same time)

src/rdkafka.c
```diff
@@ -925,7 +929,29 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
...
+                                                void *rkb_decommissioned) {
+        thrd_t *thrd;
+        int i;
+        RD_LIST_FOREACH(thrd, &rk->wait_decommissioned_thrds, i) {
+                void *rkb = rd_list_elem(&rk->wait_decommissioned_brokers, i);
+                if (rkb == rkb_decommissioned) {
+                        rd_list_remove_elem(&rk->wait_decommissioned_thrds, i);
+                        rd_list_remove_elem(&rk->wait_decommissioned_brokers,
+                                            i);
+                        thrd_join(*thrd, NULL);
```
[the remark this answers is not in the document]
emasab: It's the last operation sent before terminating the thread and when executed on the main thread it'll be already terminated. Moved sending the operation after the interceptors to avoid block, thanks for spotting it

src/rdkafka.c
```diff
@@ -1212,32 +1238,32 @@ static void rd_kafka_destroy_internal(rd_kafka_t *rk) {
...
+                }
+                if (do_decommission)
+                        rd_list_add(&brokers_to_decommission, rkb);
+        }
 
-                rd_kafka_wrlock(rk);
+        RD_LIST_FOREACH(rkb, &brokers_to_decommission, i) {
+                rd_kafka_broker_decommission(rk, rkb, &wait_thrds);
         }
+        rd_list_destroy(&brokers_to_decommission);
```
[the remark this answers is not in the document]
emasab: It's needed: I've found segmentation faults because of this. Because the decommissioning brokers are removing themselves from the `rk->rk_brokers` queue (rd_kafka_broker_decommission needs to release the lock), even if using `TAILQ_FOREACH_SAFE` some of the decommissioning broker could be destroyed while looping and generate a segfault.

src/rdkafka.c
```diff
@@ -2290,6 +2323,11 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
 
         rd_atomic64_init(&rk->rk_ts_last_poll, rk->rk_ts_created);
         rd_atomic32_init(&rk->rk_flushing, 0);
+        rd_atomic32_init(&rk->rk_broker_cnt, 0);
+        rd_atomic32_init(&rk->rk_logical_broker_cnt, 0);
+        rd_atomic32_init(&rk->rk_broker_up_cnt, 0);
+        rd_atomic32_init(&rk->rk_logical_broker_up_cnt, 0);
+        rd_atomic32_init(&rk->rk_broker_down_cnt, 0);
```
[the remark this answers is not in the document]
emasab: Because we're not counting the logical broker for the down count, instead it's using the up count in the `connect_any` to subtract it from the total up count.
Given we're only interested to this count for that, we could remove the `rk_logical_broker_up_cnt` as well and only count the learned and configured brokers that are up, but still trigger the monitors for the logical brokers.

src/rdkafka.h
```diff
@@ -5371,6 +5373,16 @@ void rd_kafka_group_list_destroy(const struct rd_kafka_group_list *grplist);
...
+/**
+ * @brief Retrieve and return the learned broker ids.
+ *
+ * @param rk Instance to use.
+ * @param cntp Will be updated to the number of brokers returned.
+ *
+ * @returns a malloc:ed list of int32_t broker ids.
+ */
+RD_EXPORT
+int32_t *rd_kafka_brokers_learned_ids(rd_kafka_t *rk, size_t *cntp);
```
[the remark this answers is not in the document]
emasab: For testing and it's useful for users to check current brokers available in the instance

src/rdkafka_broker.c
```diff
@@ -342,16 +342,16 @@ void rd_kafka_broker_set_state(rd_kafka_broker_t *rkb, int state) {
                 if (rd_atomic32_add(&rkb->rkb_rk->rk_broker_down_cnt, 1) ==
                         rd_atomic32_get(&rkb->rkb_rk->rk_broker_cnt) -
                             rd_atomic32_get(
-                                &rkb->rkb_rk->rk_broker_addrless_cnt) &&
-                    !rd_kafka_broker_is_addrless(rkb) &&
-                    !rd_kafka_terminating(rkb->rkb_rk))
+                                &rkb->rkb_rk->rk_logical_broker_cnt) &&
+                    !rd_kafka_terminating(rkb->rkb_rk)) {
```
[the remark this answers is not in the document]
emasab: Even if this broker is terminating we should report all brokers down, given it's still part of `rk_broker_cnt`

src/rdkafka_broker.c
emasab: It's removed from after sending the TERMINATE op so it's done in a single place and done all the times.

src/rdkafka_broker.c
emasab: It's `rd_dassert` so will only happen when we configure it with `--enable-devel` so we detect the failure. The state DOWN is (should be always) set just before in `rd_kafka_broker_fail`

src/rdkafka_buf.h
```diff
@@ -379,8 +379,6 @@ struct rd_kafka_buf_s { /* rd_kafka_buf_t */
                         rd_list_t *
                             topic_ids; /* Requested topic ids rd_kafka_Uuid_t */
                         char *reason;  /* Textual reason */
-                        rd_kafka_op_t *rko;    /* Originating rko with replyq
-                                                * (if any) */
                         rd_bool_t all_topics;  /**< Full/All topics requested */
```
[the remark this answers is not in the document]
emasab: It not used anymore, it's currently passed as opaque

src/rdkafka_metadata.c
```diff
@@ -92,69 +92,81 @@ rd_kafka_metadata(rd_kafka_t *rk,
...
 
-        /* Error */
-        if (rko->rko_err) {
-                rd_kafka_resp_err_t err = rko->rko_err;
-                rd_kafka_op_destroy(rko);
-                return err;
-        }
+                /* In case selected broker was decommissioned,
+                 * try again with a different broker. */
+        } while (err == RD_KAFKA_RESP_ERR__DESTROY_BROKER);
```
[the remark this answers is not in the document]
emasab: Yes because if received a `_DESTROY_BROKER` error from previous broker that won't be returned again from `rd_kafka_broker_any_usable` I can add it to the comment.

src/rdkafka_op.h
```diff
@@ -700,6 +704,18 @@ struct rd_kafka_op_s {
                         rd_kafka_broker_t *rkb;
                 } telemetry_broker;
 
+                struct {
+                        /**
+                         * Terminated and freed broker pointer,
+                         * can only be used for pointer comparison.
+                         */
+                        void *rkb;
```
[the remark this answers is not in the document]
emasab: To enforce that `can only be used for pointer comparison.`

src/rdkafka_topic.c
```diff
@@ -666,6 +667,10 @@ static int rd_kafka_toppar_leader_update(rd_kafka_topic_t *rkt,
         rd_bool_t fetching_from_follower;
         int r = 0;
 
+        if (leader && rd_kafka_broker_termination_in_progress(leader)) {
+                return 0;
+        }
```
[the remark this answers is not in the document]
emasab: In this case we're not changing the delegation after a metadata update so if it was delegated it will be delegated to the internal broker wiht `rd_kafka_toppar_undelegate` in `rd_kafka_broker_fail` when receiving the TERMINATE op.
It may be that after I've removed the decommissioning broker from the `brokers_by_id` to avoid duplications this isn't needed anymore I'll try without it.

src/rdkafka_metadata.c
```diff
@@ -460,6 +472,76 @@ rd_kafka_populate_metadata_topic_racks(rd_tmpabuf_t *tbuf,
...
+                        has_learned_brokers = rd_true;
+        }
+        rd_list_init(&brokers_to_decommission,
+                     rd_atomic32_get(&rk->rk_broker_cnt), NULL);
+        TAILQ_FOREACH_SAFE(rkb, &rk->rk_brokers, rkb_link, rkb_next) {
+                void *rkb_decommissioning;
+                rd_bool_t remove_configure_broker =
+                    has_learned_brokers &&
+                    rkb->rkb_source == RD_KAFKA_CONFIGURED &&
+                    rkb != rkb_current;
```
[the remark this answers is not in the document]
emasab: It's removed now all configured brokers are removed immediately after getting learned ones.

src/rdkafka.c
```diff
@@ -925,7 +929,29 @@ rd_kafka_resp_err_t rd_kafka_test_fatal_error(rd_kafka_t *rk,
...
 
-
+/**
+ * @brief Called when a broker thread is decommissioned.
+ *        on the main thread to join the corresponding thread
+ *        and remove it from the wait lists.
+ *
+ * @locality main thread
+ */
+void rd_kafka_decommissioned_broker_thread_join(rd_kafka_t *rk,
```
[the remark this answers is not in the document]
emasab: I think it belong here because:

1. it takes `rk` as first parameter
2. it accesses private fields `wait_decommissioned_thrds` and `wait_decommissioned_brokers` from `rk`
3. it's executed in the main thread
4. doesn't touch `rkb` fields as it was already freed.
5. instead of touching the two private fields in `broker.c` we expose this function

src/rdkafka_broker.c
```diff
@@ -4770,6 +4713,12 @@ static int rd_kafka_broker_thread_main(void *arg) {
         while (rd_kafka_broker_ops_serve(rkb, RD_POLL_NOWAIT))
                 ;
 
+        terminate_op = rd_kafka_op_new(RD_KAFKA_OP_TERMINATE);
+        terminate_op->rko_u.terminated.rkb = rkb;
+        terminate_op->rko_u.terminated.cb =
+            rd_kafka_decommissioned_broker_thread_join;
+        rd_kafka_q_enq(rk->rk_ops, terminate_op);
+        /* Release broker thread reference here and call destroy final. */
```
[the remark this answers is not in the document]
emasab: This is the last thing it's doing before terminating the thread and there are no additional refcnts that prevent calling `rd_kafka_broker_destroy_final` as for exiting the while loop above they need to be 1 and the additional ones that could be added are removed by purging the ops (in the recent fix).

src/rdkafka_broker.c
```diff
@@ -6170,6 +6091,78 @@ void rd_kafka_broker_start_reauth_cb(rd_kafka_timers_t *rkts, void *_rkb) {
...
+        TAILQ_FOREACH(rkb, &rk->rk_brokers, rkb_link) {
+                if (rkb->rkb_source != RD_KAFKA_LEARNED)
+                        continue;
+
+                *p++ = rkb->rkb_nodeid;
+                (*cntp)++;
+        }
+        rd_kafka_rdunlock(rk);
+
+        return ids;
```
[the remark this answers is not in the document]
emasab: Do you mean a TAILQ_FILTER?

src/rdkafka_cgrp.c
```diff
@@ -622,7 +622,7 @@ static int rd_kafka_cgrp_coord_update(rd_kafka_cgrp_t *rkcg, int32_t coord_id) {
                  * querying metadata and querying for the coordinator
                  * until a match is found. */
 
-                if (rkb) {
+                if (rkb && rkb->rkb_source == RD_KAFKA_LEARNED) {
```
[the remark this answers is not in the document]
emasab: This isn't needed I think the problem was that `rd_kafka_broker_find_by_nodeid` was returning a decommissioning learned broker but now it's removed already at decommissioning start.

src/rdkafka_metadata.c
```diff
@@ -460,6 +472,76 @@ rd_kafka_populate_metadata_topic_racks(rd_tmpabuf_t *tbuf,
...
+
+                if (rkb->rkb_source != RD_KAFKA_LEARNED &&
+                    !remove_configure_broker)
+                        continue;
+
```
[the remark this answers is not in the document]
emasab: Code has changed so now all configured brokers are removed immediately (a consumer metadata refresh fix was needed for it)

pranavrth (changes requested):

src/rdkafka_metadata.c
```diff
@@ -460,6 +475,72 @@ rd_kafka_populate_metadata_topic_racks(rd_tmpabuf_t *tbuf,
...
+        }
+
+        rd_list_init(&brokers_to_decommission,
+                     rd_atomic32_get(&rk->rk_broker_cnt), NULL);
+        TAILQ_FOREACH(rkb, &rk->rk_brokers, rkb_link) {
+                rd_bool_t purge_broker;
+
+                if (rkb->rkb_source != RD_KAFKA_LEARNED &&
+                    rkb->rkb_source != RD_KAFKA_CONFIGURED)
+                        continue;
```
pranavrth, line 509: As discussed, better to change it to 
```suggestion
                if (rkb->rkb_source == RD_KAFKA_LOGICAL)
                        continue;
```

src/rdkafka_metadata.c
```diff
@@ -460,6 +472,76 @@ rd_kafka_populate_metadata_topic_racks(rd_tmpabuf_t *tbuf,
...
+                        for (i = 0; i < md->broker_cnt; i++) {
+                                if (md->brokers[i].id == rkb->rkb_nodeid) {
+                                        purge_broker = rd_false;
+                                        break;
+                                }
+                        }
+                }
+
+                if (!purge_broker)
+                        continue;
```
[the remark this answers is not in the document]
pranavrth: fine both ways.

tests/0149-broker-same-host-port.c
```diff
@@ -0,0 +1,88 @@
...
+
+        TEST_ASSERT(cnt == num_brokers,
+                    "expected %" PRIusz " brokers in cache, not %" PRIusz,
+                    num_brokers, cnt);
+
+        for (i = 0; i < cnt; i++) {
+                /* Brokers are added at the head of the list. */
+                int32_t expected_id = cnt - i;
+
+                TEST_ASSERT(ids[i] == expected_id,
```
pranavrth, line 77: We should also check host and port.

pranavrth (changes requested):
Checking test part.

src/rdkafka_mock.c
```diff
@@ -1808,6 +1859,20 @@ rd_kafka_mock_coord_set(rd_kafka_mock_cluster_t *mcluster,
...
+/**
+ * @brief Remove coordinator by broker id.
+ */
+void rd_kafka_mock_coord_remove(rd_kafka_mock_cluster_t *mcluster,
+                                int32_t broker_id) {
+        rd_kafka_mock_coord_t *mcoord, *tmp;
+
+        TAILQ_FOREACH_SAFE(mcoord, &mcluster->coords, link, tmp) {
+                if (mcoord->broker_id == broker_id) {
+                        rd_kafka_mock_coord_destroy(mcluster, mcoord);
```
pranavrth, line 1871: `break;` inside `if`.

src/rdkafka_mock.c
```diff
@@ -2207,6 +2272,30 @@ rd_kafka_mock_broker_set_rack(rd_kafka_mock_cluster_t *mcluster,
...
+                        rd_kafka_dbg(
+                            cluster->rk, MOCK, "MOCK",
+                            "Broker %" PRId32
+                            ": Setting advertised listener from %s:%d to %s:%d",
+                            broker_id, mrkb->advertised_listener, mrkb->port,
+                            host, port);
+                        rd_snprintf(mrkb->advertised_listener,
+                                    sizeof(mrkb->advertised_listener), "%s",
+                                    host);
+                        mrkb->port = port;
```
pranavrth, line 2293: There can be only 1 broker with broker_id. So we can `break;` the loop inside `if`

src/rdkafka_mock.c
```diff
@@ -2612,16 +2712,16 @@ static void rd_kafka_mock_cluster_destroy0(rd_kafka_mock_cluster_t *mcluster) {
...
+         * Destroy dummy broker.
+         * WARNING: This is last time we can read
+         * from dummy_rkb in this thread!
          */
         rd_kafka_q_enq(mcluster->dummy_rkb->rkb_ops,
                        rd_kafka_op_new(RD_KAFKA_OP_TERMINATE));
 
-        dummy_rkb_thread = mcluster->dummy_rkb->rkb_thread;
-
-        rd_kafka_broker_destroy(mcluster->dummy_rkb);
```
pranavrth, line 2623: Why removing this destroy?

src/rdkafka_mock.h
```diff
@@ -358,6 +365,29 @@ rd_kafka_mock_broker_set_rack(rd_kafka_mock_cluster_t *mcluster,
 
 
 
+/**
+ * @brief Remove and delete a mock broker from a cluster.
```
pranavrth, line 369: We can add that it triggers reassignment of the partitions as well.

src/rdkafka_mock.h
```diff
@@ -358,6 +365,29 @@ rd_kafka_mock_broker_set_rack(rd_kafka_mock_cluster_t *mcluster,
...
+ * @param cluster The mock cluster containing the broker
+ * @param broker_id The broker to delete
+ * @returns 0 on success or -1 on error
+ */
+RD_EXPORT int
+rd_kafka_mock_broker_decommission(rd_kafka_mock_cluster_t *cluster,
+                                  int32_t broker_id);
+
+/**
+ * @brief Add a new broker to the cluster.
```
pranavrth, line 380: We can add that it triggers reassignment of the partitions as well.

src/rdkafka_mock.c
```diff
@@ -920,6 +928,23 @@ static void rd_kafka_mock_cluster_io_add(rd_kafka_mock_cluster_t *mcluster,
         mcluster->fd_cnt++;
 }
 
+/**
+ * @brief Reassign partition replicas to broker, after deleting or
+ *        adding a new one.
+ */
+static void
+rd_kafka_mock_cluster_reassing_partitions(rd_kafka_mock_cluster_t *mcluster) {
```
pranavrth, line 936:
```suggestion
rd_kafka_mock_cluster_reassigning_partitions(rd_kafka_mock_cluster_t *mcluster) {
```

tests/0075-retry.c
```diff
@@ -106,6 +108,14 @@ static int ctrl_thrd_main(void *arg) {
...
+        if (ctrl.num_sockets < 1) {
+                /* Since librdkafka is decommissioning brokers,
+                 * first connection is with a bootstrap broker,
+                 * next one is with a learned one. */
+                ctrl.num_sockets++;
+                mtx_unlock(&ctrl.lock);
+                return 0;
+        }
         if (ctrl.skm) {
                 /* Reject all but the first connect */
```
pranavrth, line 120: Update this comment.

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+ */
+
+#include "test.h"
+
+static rd_atomic32_t do_test_remove_then_add_received_terminate;
+static rd_atomic32_t verification_complete;
+
+/** @brief Verify that \p expected_broker_ids
+ *         and \p actual_broker_ids correspond in
+ *         count and value. Sorts \p actual_broker_ids .
```
pranavrth, line 36: It doesn't sort `actual_broker_ids`

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+                        TEST_ASSERT(!err, "%s", rd_kafka_err2str(err));
+
+                rd_usleep(100 * 1000, 0);
+
+                RD_IF_FREE(actual_broker_ids, rd_free);
+                actual_broker_ids =
+                    rd_kafka_brokers_learned_ids(rk, &actual_broker_id_cnt);
+                continue_requesting_metadata = test_clock() <= abs_timeout_us;
+
+                continue_requesting_metadata &= !fetch_metadata_verify_brokers(
```
pranavrth, line 85: Use logical AND operator instead of bitwise AND operator. Though the result is going to be same here.

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+                    expected_broker_ids, expected_broker_id_cnt,
+                    actual_broker_ids, actual_broker_id_cnt);
+
+                if (await_after_action_cb)
+                        continue_requesting_metadata |=
```
pranavrth, line 90: Same change it to logical operator instead of bitwise.

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+                            await_after_action_cb(action);
+
+        } while (continue_requesting_metadata);
+
+        TEST_ASSERT(actual_broker_id_cnt == expected_broker_id_cnt,
+                    "expected %" PRIusz " brokers in cache, got %" PRIusz,
+                    expected_broker_id_cnt, actual_broker_id_cnt);
+
+        for (i = 0; i < actual_broker_id_cnt; i++) {
+                TEST_ASSERT(actual_broker_ids[i] == expected_broker_ids[i],
```
pranavrth, line 100: Better to sort `expected_broker_ids` in the function instead of expecting it to be sorted.

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+#define TEST_ACTION_ADD_BROKER            1
+#define TEST_ACTION_SET_DOWN_BROKER       2
+#define TEST_ACTION_SET_UP_BROKER         3
+#define TEST_ACTION_SET_GROUP_COORDINATOR 4
+#define TEST_GROUP                        "topic1"
+static void
+do_test_add_remove_brokers(int32_t initial_cluster_size,
+                           int32_t actions[][2],
+                           int32_t expected_broker_ids[][5],
+                           size_t expected_broker_ids_cnt,
```
pranavrth, line 143: change this to `action_size` or something like that. We are using this count for other comparison as well.

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+
+        rk = test_create_handle(RD_KAFKA_CONSUMER, conf);
+        test_consumer_subscribe(rk, TEST_GROUP);
+
+        /* Create a new topic to trigger partition reassignment */
+        rd_kafka_mock_topic_create(cluster, TEST_GROUP, 3,
+                                   initial_cluster_size);
+
+        for (action = 0; action < expected_broker_ids_cnt; action++) {
+                if (action > 0) {
```
pranavrth, line 172: Let's remove this condition altogether. Add a new action to do nothing and add first action in the actions list as that action. Improves readability of the test.

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+                                            {2, 3, 4}, {3, 4},    {3, 4, 5},
+                                            {3, 4, 5}, {4, 5},    {4, 5, 6}};
+
+        int32_t actions[][2] = {
+            {TEST_ACTION_SET_GROUP_COORDINATOR, 3},
+            {TEST_ACTION_REMOVE_BROKER, 1},
+            {TEST_ACTION_ADD_BROKER, 4},
+            {TEST_ACTION_REMOVE_BROKER, 2},
+            {TEST_ACTION_ADD_BROKER, 5},
+            {TEST_ACTION_SET_GROUP_COORDINATOR, 5},
```
pranavrth, line 257: How are we verifying that the coordinator is getting updated to the correct one we wanted? This might not be in the scope of this test but then why are we even doing that part?

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+        }
+}
+
+/**
+ * @brief Await for the TERMINATE op to be received after the action
+ *        that removes the broker then proceed to
+ *        add the broker again.
+ */
+static rd_bool_t do_test_remove_then_add_await_after_action_cb(int action) {
+        if (action == 2) {
```
pranavrth, line 333:
```suggestion
        if (action == TEST_ACTION_SET_DOWN_BROKER) {
```
Is this what you mean?

tests/0151-purge-brokers.c
```diff
@@ -0,0 +1,410 @@
...
+            {TEST_ACTION_ADD_BROKER, 1},
+        };
+
+        do_test_add_remove_brokers(
+            3, actions, expected_broker_ids, RD_ARRAY_SIZE(expected_broker_ids),
+            expected_brokers_cnt, do_test_remove_then_add_edit_configuration_cb,
+            do_test_remove_then_add_await_after_action_cb);
+
+        rd_free(log_interceptor);
+        log_interceptor = NULL;
```
pranavrth, line 392: Instead of global variable, we should get this from the `rk` opaque. `do_test_add_remove_brokers` can return this.

pranavrth (changes requested):
Few comments on the test part.

src/rdkafka_mock.c
emasab: In the mock cluster you can have coordinators for multiple consumer groups and the coordinator can be the same broker for all those groups so it's possible there are more than one.

src/rdkafka_mock.c
emasab: Maybe `rd_kafka_mock_cluster_reassign_partitions` with the typo fixed? `reassigning_partitions` seems more a question than a command

tests/0151-purge-brokers.c
emasab: There's the change to return `rd_kafka_brokers_learned_ids` from `rk_broker_by_id` that is sorted. So it's ensured. I can add it to the function documentation.

tests/0151-purge-brokers.c
emasab: The count must be the same, after every action there's an `expected_broker_ids`.
```C
 *        After each action, the client is expected to have the broker
 *        ids in \p expected_broker_ids and the count to be
 *        \p expected_brokers_cnt .
```

tests/0151-purge-brokers.c
emasab: I don't think it improves readability because there are N actions it's expected there are N+1 states including the initial state. Here it's ensuring that we're in the initial state before executing the first action.

tests/0151-purge-brokers.c
emasab: Because if the coordinator doesn't change before removing a broker, that's returned by the `FindCoordinator` request and added again because it's needed as a reference in `rkcg_curr_coord`. So it continues adding that broker again and doesn't reach the desired number of brokers.
In a real cluster this doesn't happen as a new coordinator is elected before the broker shuts down so it's simulating that behaviour.
Initially I wanted to make the coordinator change automatic but that isn't the previous mock cluster contract so I added these manual commands.

tests/0151-purge-brokers.c
emasab: No, here it's the action number that is passed (starting from 1) not the action type

tests/0151-purge-brokers.c
emasab: ok, just I return it as `void *`  to avoid making it coupled with this test.

src/rdkafka_mock.c
pranavrth: Works.

tests/0151-purge-brokers.c
pranavrth: Yes but the list of `expected_broker_ids` is defined by the user which might not be sorted. Its better to sort it.

tests/0151-purge-brokers.c
pranavrth: I am talking about initial state. This seems like a hardcoding for the initial state. We can skip this if condition if we just add new action to do nothing.

src/rdkafka_metadata.c
```diff
@@ -460,6 +472,76 @@ rd_kafka_populate_metadata_topic_racks(rd_tmpabuf_t *tbuf,
...
+        rd_kafka_wrlock(rk);
+        TAILQ_FOREACH_SAFE(rkb, &rk->rk_brokers, rkb_link, rkb_next) {
+                if (rkb->rkb_source == RD_KAFKA_LEARNED)
+                        has_learned_brokers = rd_true;
+        }
+        rd_list_init(&brokers_to_decommission,
+                     rd_atomic32_get(&rk->rk_broker_cnt), NULL);
+        TAILQ_FOREACH_SAFE(rkb, &rk->rk_brokers, rkb_link, rkb_next) {
+                void *rkb_decommissioning;
+                rd_bool_t remove_configure_broker =
```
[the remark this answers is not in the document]
pranavrth: This field is removed altogether.

tests/0149-broker-same-host-port.c
```diff
@@ -48,41 +106,33 @@ int main_0149_broker_same_host_port_mock(int argc, char **argv) {
...
+        TEST_SAY("Modified nodenames metadata request\n");
+        /* Trigger Metadata request which will get initial changed hostnames. */
+        broker_same_host_port_mock_verify_broker_ids(rk);
 
-                TEST_ASSERT(ids[i] == expected_id,
-                            "expected broker %d in cache, not %d", expected_id,
-                            ids[i]);
-        }
+        while (!(broker1_changed && broker2_changed && broker3_changed))
+                rd_usleep(100000, 0);
```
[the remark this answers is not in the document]
emasab: ok

src/rdkafka_broker.c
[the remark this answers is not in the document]
pranavrth: The function is changed. Not needed.

tests/0149-broker-same-host-port.c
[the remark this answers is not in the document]
pranavrth: Anything is fine.

emasab (approved):
Approving as I was initially reviewing. Thanks @pranavrth and @mfleming !

pranavrth (approved):
LGTM! Thanks @emasab and @mfleming !!

PR #4561, opened 2023-12-14
Change retry.backoff.* to apply to both producer and consumer

milindl:
After KIP-580, this property is used for both consumer and producer related operations, sort of as a general backoff. 

The configuration needs to be updated to reflect this, else we generate a warning at 5and higher levels of logging.


emasab (approved):
LGTM

PR #4562, opened 2023-12-14
Make install dependent on libs to avoid errors

emasab:
using parallelism.
Don't use passed libdir for librdkafka static
dependencies built from source


milindl:
LGTM pending CI.
Will also fix #4532

milindl (approved):

milindl (dismissed):

milindl (approved):

PR #4582, opened 2024-01-12
remove useless const modifier

JustAGod1:
The 'const' modifier in these functions serves no purpose. However, it will cause GCC or Clang to generate a warning in a file where rdkafka.h is included, which will result in a compilation failure if -Werror is enabled .

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -8221,7 +8221,7 @@ const rd_kafka_TopicPartitionInfo_t **rd_kafka_TopicDescription_partitions(
  * @return The partition id.
  */
 RD_EXPORT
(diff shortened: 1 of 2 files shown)


wanders:
For searchability for others hitting the same error here is the complete error message generated by gcc:
```
librdkafka/rdkafka.h:8205:11: error: type qualifiers ignored on function return type [-Werror=ignored-qualifiers]
```

clang says: 
```
warning: 'const' type qualifier on return type has no effect [-Wignored-qualifiers]
```

JustAGod1:
Rebased master

Quuxplusone:
Duplicates #4556 (but yes, @emasab or someone, please merge either one of these already, and close the other!)

PR #4591, opened 2024-01-24
Uuid support in mock cluster

emasab:


src/rdkafka.c
```diff
@@ -5077,6 +5077,38 @@ rd_kafka_Uuid_t *rd_kafka_Uuid_copy(const rd_kafka_Uuid_t *uuid) {
...
+                 * a random value. */
+                rand_values_app = (unsigned char *)&rand_uint16;
+                rand_values_bytes[i] |= rand_values_app[0];
+                rand_values_bytes[i + 1] |= rand_values_app[1];
+        }
+
+        rand_values_bytes[6] &= 0x0f; /* clear version */
+        rand_values_bytes[6] |= 0x40; /* version 4 */
+        rand_values_bytes[8] &= 0x3f; /* clear variant */
+        rand_values_bytes[8] |= 0x80; /* IETF variant */
```
pranavrth, line 5105: I think we should move this to rdrand.h

src/rdkafka.c
```diff
@@ -5077,6 +5077,38 @@ rd_kafka_Uuid_t *rd_kafka_Uuid_copy(const rd_kafka_Uuid_t *uuid) {
...
 }
 
+/**
+ * Returns a new non cryptographically secure UUIDv4 (random).
+ *
+ * @return A UUIDv4.
+ *
+ * @remark Must be freed after use using rd_kafka_Uuid_destroy().
+ */
+rd_kafka_Uuid_t *rd_kafka_Uuid_random() {
```
pranavrth, line 5087: Why are we returning a pointer instead of a value here?

src/rdkafka.c
emasab: It generates a Uuid, all the code for a Uuid is here

src/rdkafka.c
```diff
@@ -5086,6 +5118,40 @@ void rd_kafka_Uuid_destroy(rd_kafka_Uuid_t *uuid) {
...
+ * @brief Computes canonical encoding for the given uuid string.
+ *        Mainly useful for testing.
+ *
+ * @param uuid UUID for which canonical encoding is required.
+ *
+ * @return canonical encoded string for the given UUID.
+ *
+ * @remark  Must be freed after use.
+ */
+const char *rd_kafka_Uuid_str(const rd_kafka_Uuid_t *uuid) {
```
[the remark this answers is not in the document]
emasab: It prints the UUID in its standard format, it's useful to check if the UUID respects the standard v4 (the 4 always in the same position and the [89ab] in next sequence.

src/rdkafka.c
emasab: We should refactor it to a `rduuid.h` + `rduuid.c` later

emasab: Maybe I can move it to a `rduuid.h` and `rduuid.c`

pranavrth: I was talking about the generic part related to random bytes generation here.

src/rdkafka.c
```diff
@@ -5077,6 +5077,38 @@ rd_kafka_Uuid_t *rd_kafka_Uuid_copy(const rd_kafka_Uuid_t *uuid) {
...
 }
 
+/**
+ * Returns a new non cryptographically secure UUIDv4 (random).
+ *
+ * @return A UUIDv4.
+ *
+ * @remark Must be freed after use using rd_kafka_Uuid_destroy().
+ */
+rd_kafka_Uuid_t rd_kafka_Uuid_random() {
```
pranavrth, line 5087: Move this to rdkafka_proto.h where all the other internal functions related to Uuid are present

src/rdkafka.c
pranavrth, line 5131: Move this to rdkafka_proto.h where all the other internal functions related to Uuid are present

src/rdkafka.c
pranavrth: Regarding rdkafka_uuid.h and rdkafka_uuid.c, we can have them but don't do it for now. Let's do it later when we we won't have much conflicts with other ongoing dev work.

pranavrth (changes requested):

src/rdkafka.c
emasab: that is just a header we cannot move all the implementation there, we can refactor to its own file

src/rdkafka.c
emasab: same

src/rdkafka.c
emasab: > I think we should move this to rdrand.h
here the random value is generated in `rd_jitter` the rest is code related to Uuid

pranavrth (approved):
LGTM!

PR #4609, opened 2024-02-06
[KIP-951] v16: parse KIP-482 Fetch tags

emasab:


pranavrth:
I think the title is incorrect with KIP-848 part. I think its KIP-482.

PR #4611, opened 2024-02-06
[KIP-848] Mock cluster handler and tests

emasab:
Add support to mock target assignment manually or to calculate a simple range assignment.
Some mocks tests are still excluded.


src/rdkafka_partition.c
```diff
@@ -3255,6 +3276,21 @@ rd_kafka_topic_partition_t *rd_kafka_topic_partition_list_find_topic(
                 return &rktparlist->elems[i];
 }
 
+/**
+ * @returns the first element that matches \p topic_id, regardless of partition.
+ */
+rd_kafka_topic_partition_t *rd_kafka_topic_partition_list_find_topic_by_id(
```
pranavrth, line 3282: Better to name this `rd_kafka_topic_partition_list_get_topic_by_topic_id`

src/rdkafka_partition.h
```diff
@@ -768,6 +768,11 @@ int rd_kafka_topic_partition_list_find_idx(
     const char *topic,
     int32_t partition);
 
+rd_kafka_topic_partition_t *rd_kafka_topic_partition_list_find_by_id(
```
pranavrth, line 771: There are many confusing functions for the same. Can we improve the readability of the names and explains what the function is going to return?

src/rdkafka_proto.h
```diff
@@ -597,19 +597,22 @@ typedef struct rd_kafka_Uuid_s {
...
-/**
- * Initialize given UUID to zero UUID.
- *
- * @param uuid UUID to initialize.
- */
-static RD_INLINE RD_UNUSED void rd_kafka_Uuid_init(rd_kafka_Uuid_t *uuid) {
-        memset(uuid, 0, sizeof(*uuid));
-}
-
 static RD_INLINE RD_UNUSED int rd_kafka_Uuid_cmp(rd_kafka_Uuid_t a,
```
pranavrth, line 600: We should change the signature of this function to use pointer instead when we move these functions to `rdkafka_uuid.c`

src/rdkafka_topic.c
pranavrth, line 1: Need more understanding on these changes.

pranavrth (changes requested):
First pass of review. More of a high level comments.

Not reviewed:
- rdkafka_metadata.h
- rdkafka_metadata_cache.c
- rdkafka_topic.c
- Mock changes
- Test changes

src/rdkafka_cgrp.h
```diff
@@ -275,18 +275,31 @@ typedef struct rd_kafka_cgrp_s {
...
+ * considered NONE, otherwise it depends on the
+ * configured partition assignors. */
+#define RD_KAFKA_CGRP_CONSUMER_F_SUBSCRIBED_ONCE 0x8
+/* Send a complete request in next heartbeat,
+ * but don't send the acknowledgement if it's not required */
+#define RD_KAFKA_CGRP_CONSUMER_F_SEND_FULL_REQUEST 0x10
+/* Member is fenced, need to rejoin */
+#define RD_KAFKA_CGRP_CONSUMER_F_WAIT_REJOIN 0x20
+/* Member is fenced, rejoining */
+#define RD_KAFKA_CGRP_CONSUMER_F_WAIT_REJOIN_TO_COMPLETE 0x40
```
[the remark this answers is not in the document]
emasab: I use the C/javadoc style with `/**` but leave the documentation before the define, otherwise it's becomes too large and needs to wrap being less readable

src/rdkafka_metadata.c
```diff
@@ -980,7 +1027,8 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
          * which may contain only a sub-set of the subscribed topics (namely
          * the effective subscription of available topics) as to not
          * propagate non-included topics as non-existent. */
-        if (cgrp_update && (requested_topics || all_topics))
+        if (cgrp_update &&
+            (requested_topics || requested_topic_ids || all_topics))
                 rd_kafka_cgrp_metadata_update_check(rkb->rkb_rk->rk_cgrp,
```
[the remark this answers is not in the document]
emasab: Right

src/rdkafka_cgrp.c
```diff
@@ -2666,21 +2695,76 @@ static rd_kafka_op_res_t rd_kafka_cgrp_consumer_handle_next_assignment(
...
+                rd_kafka_rdunlock(rk);
+
+                rktpar = rd_kafka_topic_partition_list_find_topic_by_id(
+                    rkcg->rkcg_current_assignment, request_topic_id);
+                if (rktpar) {
+                        rd_kafka_topic_partition_list_add_with_topic_name_and_id(
+                            assignment_with_metadata, request_topic_id,
+                            rktpar->topic, assignment->elems[i].partition);
+                        continue;
+                }
```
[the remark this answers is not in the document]
emasab: We need to ack the assignment even if we don't have metadata, and the protocol ensures it will be at least in current assignment. If we remove and don't ack it, the coordinator could discard the ack.

Later the partitions for that topic will be removed by the coordinator

src/rdkafka_cgrp.c
```diff
@@ -6046,7 +6103,29 @@ void rd_kafka_cgrp_consumer_expedite_next_heartbeat(rd_kafka_cgrp_t *rkcg) {
...
+                backoff = rk->rk_conf.retry_backoff_max_ms * 1000;
+
+        /* Reset the interval as it happened `rkcg_heartbeat_intvl_ms`
+         * milliseconds ago. */
+        rd_interval_reset_to_now(&rkcg->rkcg_heartbeat_intvl,
+                                 rd_clock() -
+                                     rkcg->rkcg_heartbeat_intvl_ms * 1000);
+        /* Set the exponential backoff. */
+        rd_interval_backoff(&rkcg->rkcg_heartbeat_intvl, backoff);
+        rkcg->rkcg_expedite_heartbeat_retries++;
```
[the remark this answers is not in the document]
emasab: Which other `expedite_*`  ?

src/rdkafka_partition.c
emasab: Used the same naming of the existing
`rd_kafka_topic_partition_list_find_topic`
maybe I can rename that to `rd_kafka_topic_partition_list_find_topic_by_name`

src/rdkafka_proto.h
emasab: We took the decision to store and pass Uuid as structs internally, to avoid many heap allocations. Because they're small and not going to change, there's a different cmp function inserted later that takes pointers for when it's needed, like when they're in a list.

emasab:
Closing this PR in favour of #4662

PR #4612, opened 2024-02-06
Fix double free as reported by address sanitize that can cause a crash

arches-andy:
Fix double free as reported by address sanitize that causes a crash when deleting the memory a second time if it is already being used between the deletes.


Quuxplusone (changes requested):
#4628 seems like a more appropriate fix.

arches-andy:
Updated comments.

Issue #4348, opened 2023-07-11
The trailing dot should be removed from the host name in the call to SSL_set1_host


joejulian:
Although the host name should retain the dot for SNI, the hostname expectations for SAN excludes the trailing dot. Per this comment from openssl, I believe the trailing dot should be removed from the host name when passing it to SSL_set1_host:
https://github.com/openssl/openssl/issues/11560#issuecomment-1619143823

https://github.com/confluentinc/librdkafka/blob/c23adb959192aadfcf7461371bb9f8dfe79958d2/src/rdkafka_ssl.c#L471

PR #4614, opened 2024-02-08
[WIP] Strip a trailing . when present in name - Address confluentinc/librdkafka#4348

cscherban:
This is a very poor attempt at addressing the issue mentioned here: https://github.com/confluentinc/librdkafka/issues/4348, which is caused by https://github.com/openssl/openssl/issues/11560#issuecomment-1619143823 

If this is the wrong place to be attempting to address the issue, I would love guidance on how to properly tackle the problem. I don't use librdkafka directly, but rely on a ruby wrapping library. ([karafka](https://github.com/karafka/karafka)).

What I'm attempting here is to transform
`my.kafka.domain.com.:9093` -> `my.kafka.domain.com:9093` 
by splitting on ":" and replacing trailing "." with "\0"

### DIFF
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -468,7 +468,25 @@ static int rd_kafka_transport_ssl_set_endpoint_id(rd_kafka_transport_t *rktrans,
                 return 0;
 
 #if OPENSSL_VERSION_NUMBER >= 0x10100000 && !defined(OPENSSL_IS_BORINGSSL)
-        if (!SSL_set1_host(rktrans->rktrans_ssl, name))
+        char *token, *hostname;
+        token = strtok(name, ":");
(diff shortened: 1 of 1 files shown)


Quuxplusone:
No comment on the desirability of this patch (versus waiting for OpenSSL to fix it, or discovering that it's not really a bug after all) — but unless I'm mistaken, the `:9093` part has already been stripped by line 452, and any `:` characters that remain are part of an IPv6 address. So _probably_ all you want here is:

```
         /* Remove ":9092" port suffix from nodename */
         if ((t = strrchr(name, ':')))
                 *t = '\0';
+
+        /* Remove "." trailing dot suffix from nodename */
+        if ((t = strrchr(name, '.')) && t[1] == '\0')
+                *t = '\0';
```

PR #4621, opened 2024-02-21
Topic partition and Uuid additional common functions

emasab:


src/rdkafka_partition.c
```diff
@@ -4479,3 +4499,149 @@ const char *rd_kafka_fetch_pos2str(const rd_kafka_fetch_pos_t fetchpos) {
...
+
+typedef RD_MAP_TYPE(const rd_kafka_topic_partition_t *,
+                    void *) map_toppar_void_t;
+
+/**
+ * @brief Calculates \p a ∩ \p b.
+ *        Ordered following \p a order. Elements are copied from \p a.
+ */
+static rd_kafka_topic_partition_list_t *
+rd_kafka_topic_partition_list_intersection0(
```
pranavrth, line 4511: I think we should have these functions at `rd_list` level. We can discuss more offline on this.

src/rdkafka_partition.c
```diff
@@ -4479,3 +4499,149 @@ const char *rd_kafka_fetch_pos2str(const rd_kafka_fetch_pos_t fetchpos) {
...
+    rd_kafka_topic_partition_list_t *a,
+    rd_kafka_topic_partition_list_t *b,
+    int(cmp)(const void *_a, const void *_b),
+    unsigned int(hash)(const void *_a)) {
+        rd_kafka_topic_partition_t *rktpar;
+        rd_kafka_topic_partition_list_t *ret =
+            rd_kafka_topic_partition_list_new(a->cnt);
+        map_toppar_void_t b_map =
+            RD_MAP_INITIALIZER(b->cnt, cmp, hash, NULL, NULL);
```
pranavrth, line 4520: Magnus used to discourage Map usage. We can achieve same by sorting and then finding the intersection.

pranavrth (changes requested):
First high level pass.

src/rdmap.c
```diff
@@ -237,6 +238,21 @@ unsigned int rd_map_str_hash(const void *key) {
 }
 
 
+/**
+ * @returns a djb2 hash of \p bytes.
+ *
+ * @param len \p bytes will be hashed up to \p len.
+ */
+unsigned int rd_bytes_hash(unsigned char *bytes, size_t len) {
```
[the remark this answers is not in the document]
emasab: Not sure because it's only for this function and a hash if only needed for the map at the moment

src/rdkafka_partition.c
emasab: `rd_kafka_topic_partition_list_t `is different from `rd_list `in that it only contains partitions, here I needed this functions for `rd_kafka_topic_partition_list_t` only.

pranavrth: As discussed, we can add this later.

src/rdkafka_partition.c
pranavrth: You can use it as discussed.

src/rdkafka_partition.c
```diff
@@ -4479,3 +4499,149 @@ const char *rd_kafka_fetch_pos2str(const rd_kafka_fetch_pos_t fetchpos) {
...
+        RD_MAP_DESTROY(&b_map);
+        return ret;
+}
+
+/**
+ * @brief Calculates \p a - \p b
+ *        Ordered following \p a order. Elements are copied from \p a.
+ */
+static rd_kafka_topic_partition_list_t *
+rd_kafka_topic_partition_list_difference0(rd_kafka_topic_partition_list_t *a,
```
pranavrth, line 4538: I think we should also mention what each comparator should return. Only 1 and 0 or -1, 0 and 1.

src/rdkafka_partition.c
```diff
@@ -4479,3 +4499,149 @@ const char *rd_kafka_fetch_pos2str(const rd_kafka_fetch_pos_t fetchpos) {
...
+        rd_kafka_topic_partition_list_t *ret =
+            rd_kafka_topic_partition_list_copy(a);
+        rd_kafka_topic_partition_list_grow(ret, b->cnt);
+        rd_kafka_topic_partition_list_t *b_minus_a =
+            rd_kafka_topic_partition_list_difference0(
+                b, a, rd_kafka_topic_partition_cmp,
+                rd_kafka_topic_partition_hash);
+        rd_kafka_topic_partition_list_add_list(ret, b_minus_a);
+        rd_kafka_topic_partition_list_destroy(b_minus_a);
+        return ret;
```
pranavrth, line 4600:
```suggestion
        rd_kafka_topic_partition_list_t *b_minus_a =
            rd_kafka_topic_partition_list_difference0(
                b, a, rd_kafka_topic_partition_cmp,
                rd_kafka_topic_partition_hash);
        rd_kafka_topic_partition_list_t *ret =
            rd_kafka_topic_partition_list_new(a->cnt + b_minus_a->cnt);
        rd_kafka_topic_partition_list_add_copy(ret, a);
        rd_kafka_topic_partition_list_add_copy(ret, b_minus_a);
        rd_kafka_topic_partition_list_destroy(b_minus_a);
        return ret;
```

src/rdkafka_partition.c
```diff
@@ -4479,3 +4499,149 @@ const char *rd_kafka_fetch_pos2str(const rd_kafka_fetch_pos_t fetchpos) {
...
+}
+
+/**
+ * @brief Calculates \p a ∪ \p b.
+ *        Ordered following \p a order for elements in \p a
+ *        and \p b order for elements only in \p b.
+ *        Elements are copied the same way.
+ */
+rd_kafka_topic_partition_list_t *
+rd_kafka_topic_partition_list_union(rd_kafka_topic_partition_list_t *a,
```
pranavrth, line 4589: Create a `union0` function as well which takes `cmp` and `hash` as input as well.

src/rdkafka_partition.c
```diff
@@ -4479,3 +4499,149 @@ const char *rd_kafka_fetch_pos2str(const rd_kafka_fetch_pos_t fetchpos) {
...
+}
+
+/**
+ * @brief Calculates \p a ∪ \p b using topic id.
+ *        Ordered following \p a order for elements in \p a
+ *        and \p b order for elements only in \p b.
+ *        Elements are copied the same way.
+ */
+rd_kafka_topic_partition_list_t *
+rd_kafka_topic_partition_list_union_by_id(rd_kafka_topic_partition_list_t *a,
```
pranavrth, line 4635: Use `union0` function here as well.

src/rdkafka_partition.h
```diff
@@ -1122,4 +1125,29 @@ static RD_UNUSED RD_INLINE void rd_kafka_toppar_set_offset_validation_position(
...
+rd_kafka_topic_partition_list_intersection(rd_kafka_topic_partition_list_t *a,
+                                           rd_kafka_topic_partition_list_t *b);
+
+rd_kafka_topic_partition_list_t *
+rd_kafka_topic_partition_list_difference(rd_kafka_topic_partition_list_t *a,
+                                         rd_kafka_topic_partition_list_t *b);
+
+rd_kafka_topic_partition_list_t *
+rd_kafka_topic_partition_list_union(rd_kafka_topic_partition_list_t *a,
+                                    rd_kafka_topic_partition_list_t *b);
```
pranavrth, line 1138: lets start calling these as well `_by_name`.

pranavrth (changes requested):
More review comments.

src/rdkafka.c
```diff
@@ -5178,6 +5179,14 @@ const char *rd_kafka_Uuid_base64str(const rd_kafka_Uuid_t *uuid) {
         return uuid->base64str;
 }
 
+unsigned int rd_kafka_Uuid_map_hash(const void *key) {
```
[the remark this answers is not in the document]
emasab: While the declaration is already there, the definition is here as with other Uuid related functions.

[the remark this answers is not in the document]
pranavrth: We are keeping public functions in `rdkafka.c`. `rdkafka.c` is not the right place to keep internal function imo. Let's move them.

src/rdkafka_partition.c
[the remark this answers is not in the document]
emasab: I think this way it's more predictable to always use the order and the elements of the first list.
The called can check which one is bigger and swap then if order is not important.
First item is fine, thanks

src/rdkafka_partition.c
emasab: Nice improvement, I've just kept `rd_kafka_topic_partition_list_add_list` because `rd_kafka_topic_partition_list_add_copy` adds a single partition not a whole list, but internally it uses `rd_kafka_topic_partition_list_add_copy`

src/rdkafka.c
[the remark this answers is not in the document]
emasab: We already had this conversation. If it's public is determined by the header and the declaration is there in proto.h. That is not the right place for the definition as it's an header and it's about the protocol serialization deserialization. I'll do the refactor to `rdkafka_uuid.c` as soon as KIP-848 branch is stable

src/rdkafka_partition.c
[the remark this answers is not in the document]
pranavrth: Yes. I was thinking earlier that `a` should be used for the map instead. But `b` is also fine.

src/rdkafka_partition.c
pranavrth: For this map approach, we are okay to compare equality only. If we were using sorting approach, then less than, more than and equal is required. So the comparator choice depends on the implementations. That the reason for adding information about the comparator.

src/rdkafka_partition.h
pranavrth: You forgot to update in the Header file.

pranavrth (changes requested):

src/rdkafka.c
```diff
@@ -5178,6 +5179,17 @@ const char *rd_kafka_Uuid_base64str(const rd_kafka_Uuid_t *uuid) {
...
 }
 
+unsigned int rd_kafka_Uuid_hash(const rd_kafka_Uuid_t *uuid) {
+        unsigned char bytes[16];
+        memcpy(bytes, &uuid->most_significant_bits, 8);
+        memcpy(&bytes[8], &uuid->least_significant_bits, 8);
+        return rd_bytes_hash(bytes, 16);
+}
+
+unsigned int rd_kafka_Uuid_map_hash(const void *key) {
```
pranavrth, line 5189: Change variable from `key` to `uuid` as well.

pranavrth (approved):
LGTM!. Just a minor comment that can be done in some later PR as well.

PR #4624, opened 2024-02-27
Phase 4 produce temp

anchitj:


src/rdkafka_int.h
```diff
@@ -275,6 +275,7 @@ struct rd_kafka_s {
         int rk_topic_cnt;
 
         struct rd_kafka_cgrp_s *rk_cgrp;
+        rd_kafkap_produce_reply_tags_t *produce_tags;
```
mahajanadhitya, line 278: Remove this as it is not needed.

src/rdkafka_broker.h
```diff
@@ -331,6 +331,7 @@ struct rd_kafka_broker_s { /* rd_kafka_broker_t */
 
 
         rd_kafka_timer_t rkb_sasl_reauth_tmr;
+        rd_kafkap_produce_reply_tags_t *produce_tags;
```
mahajanadhitya, line 334: Remove this as this is not needed.

PR #4636, opened 2024-03-06
Fix duplicate messages when resuming partitions not paused

emasab:
Fixed a bug causing duplicate message consumption from a stale fetch start offset in some particular cases.

In case of subscription change with a consumer using the cooperative assignor it could resume fetching from a previous position.
That could also happen if resuming a partition that wasn't paused.

Fixed by ensuring that a resume operation is completely a no-op when the partition isn't paused.

Fixes https://github.com/confluentinc/librdkafka/issues/4637

### DIFF
--- a/tests/0145-unnecessary_resume_mock.c
+++ b/tests/0145-unnecessary_resume_mock.c
@@ -1,0 +1,119 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2024, Confluent Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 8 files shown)


emasab:
Fixes https://github.com/confluentinc/librdkafka/issues/4637

tests/0145-unnecessary_resume_mock.c
```diff
@@ -0,0 +1,119 @@
...
+#include <stdarg.h>
+
+/**
+ * Verify that no duplicate message are consumed after an unnecessary
+ * resume, ensuring the fetch version isn't bumped, leading to
+ * using a stale next fetch start.
+ *
+ * @param partition_assignment_strategy Assignment strategy to test.
+ */
+static void test_no_duplicate_messages_unnecessary_resume(
```
pranavrth, line 42: Name of the test should be better. It can be `pause_resume_mock`.

tests/0145-unnecessary_resume_mock.c
pranavrth, line 1: I was thinking to define a better numbering for mock tests. Integration tests starts with '0', unit test with '8', so we can use mock tests with '4' maybe?

tests/test.c
```diff
@@ -511,7 +512,7 @@ struct test tests[] = {
     _TEST(0142_reauthentication, 0, TEST_BRKVER(2, 2, 0, 0)),
     _TEST(0143_exponential_backoff_mock, TEST_F_LOCAL),
     _TEST(0144_idempotence_mock, TEST_F_LOCAL, TEST_BRKVER(0, 11, 0, 0)),
-
```
pranavrth, line 514: Keep extra line.

tests/0145-unnecessary_resume_mock.c
```diff
@@ -0,0 +1,119 @@
...
+        rd_kafka_topic_partition_list_add(tlist, topic, RD_KAFKA_PARTITION_UA);
+
+        rk = test_create_consumer("mygroup", NULL, conf, tconf);
+        TEST_CALL_ERR__(rd_kafka_subscribe(rk, tlist));
+
+        TEST_SAY("Consume and verify messages\n");
+        test_msgver_init(&mv, testid);
+        test_consumer_poll("consume", rk, testid, -1, 0, msgcnt, &mv);
+
+        TEST_SAY("Unnecessary resume\n");
```
pranavrth, line 83: Is this test failing before the change?

pranavrth (changes requested):
Looks fine. Just few minor changes.

pranavrth (approved):
LGTM!

PR #4648, opened 2024-03-14
Add documentation for disabling nagle for socket to review for reduci…

swehner:
Add a quick note for disabling Nagle for socket to reduce overall latency.
For us this reduced broker RTT from 40ms to around 6ms


aivarsk:
See also https://github.com/confluentinc/librdkafka/pull/4298 to make it the default option, all other HTTP/database/anything clients do it by default.

Issue #4577, opened 2024-01-09
producer continually refreshing client metadata after broker disconnection


jpiper:
Description
===========
We are producing into a topic with 84 partitions over 8 brokers. We are seeing the issue where a recoverable error condition (difficult to reproduce, but an example would be transport failure causing a broker connection to fail) will trigger the underlying librdkafka library to continually refresh the metadata for this topic at the `retry.backoff.max.ms`, even though the metadata request is successful and all partitions have leaders.

e.g. here is the error condition happening which is triggered by a new connection to a broker being broken just ~8s after entering UP state

```
%7|1704816898.865|CONNECT|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: Connected to ipv4#10.255.253.3:443
%7|1704816898.865|STATE|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: Broker changed state CONNECT -> SSL_HANDSHAKE
%7|1704816898.865|TOPPAR|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: produce_topic [51] 6 message(s) queued but broker not up
%7|1704816898.886|TOPPAR|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: produce_topic [51] 6 message(s) queued but broker not up
%7|1704816898.888|TOPPAR|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: produce_topic [51] 6 message(s) queued but broker not up
%7|1704816898.907|TOPPAR|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: produce_topic [51] 6 message(s) queued but broker not up
%7|1704816898.907|CONNECTED|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: Connected (#13)
%7|1704816898.907|STATE|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: Broker changed state SSL_HANDSHAKE -> APIVERSION_QUERY
%7|1704816898.907|TOPPAR|producer.host#producer-2| [thrd:ssl://broker-6:4]: ssl://broker-6:443/6: produce_topic [51] 6 message(s) queued but broker not up
[...]

Issue #4589, opened 2024-01-23
rd_kafka_metadata can cause a UNKNOWN_TOPIC_OR_PART message


sshanks-kx:
Description
===========
Using a client with `rd_kafka_queue_io_event_enable` that is a consumer using `rd_kafka_subscribe`. 

This works fine until metadata is requested via `rd_kafka_metadata`. 

If the metadata is requested any time after making the subscription, the poll returns a message with error set to UNKNOWN_TOPIC_OR_PART (offset -1001) for every topic subscribed to (although the metadata returns correct results and the subscription continues getting messages).

If the metadata is requested prior to making any subscription, I dont get an error message from the poll.

How to reproduce
================

example client program (given topic exists called `topic1` with partition `0`, broker `localhost:9092`) purely to show the error message.

gcc main.c -lrdkafka -lz -lpthread -lssl

```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <librdkafka/rdkafka.h>

int main(int argc,char** argv){
    rd_kafka_t *rk=0;
    rd_kafka_conf_t *conf=0;
    char *brokers = "localhost:9092";
    rd_kafka_queue_t *queue=0;
    rd_kafka_resp_err_t err=0;
    rd_kafka_topic_partition_list_t *t_partition;
    rd_kafka_message_t *msg;
    char errstr[512];
    int spair[2];

    if(pipe(spair)==-1){
        fprintf(stderr, "pipe fail\n");
        exit(1);
    }

    conf = rd_kafka_conf_new();
    if (rd_kafka_conf_set(conf, "bootstrap.servers", brokers, errstr,sizeof(errstr)) != RD_KAFKA_CONF_OK) {
        fprintf(stderr, "%% %s\n", errstr);
        exit(1);
    }

    if (rd_kafka_conf_set(conf, "group.id","0", errstr,sizeof(errstr)) != RD_KAFKA_CONF_OK) {
        fprintf(stderr, "%% %s\n", errstr);
        exit(1);
    }

    if (rd_kafka_conf_set(conf, "debug","all", errstr,sizeof(errstr)) != RD_KAFKA_CONF_OK) {
        fprintf(stderr, "%% %s\n", errstr);
        exit(1);
    }

    if (!(rk = rd_kafka_new(RD_KAFKA_CONSUMER, conf, errstr,sizeof(errstr)))) {
      fprintf(stderr,"Failed to create new consumer: %s\n",errstr);
[...]

PR #4660, opened 2024-03-26
Metadata cache by topic id and fixes

emasab:
Metadata cache by topic id is added, with an AVL tree having a Uuid as key and the same struct as value.

### General fixes

 * Metadata cache was cleared on full metadata refresh, leading to unnecessary
   refreshes and occasional `UNKNOWN_TOPIC_OR_PART` errors. Solved by updating
   cache for existing or hinted entries instead of clearing them.
   Happening since 2.1.0.
 * A metadata call before member joins consumer group,
   could lead to an `UNKNOWN_TOPIC_OR_PART` error. Solved by updating
   the consumer group following a metadata refresh only in safe states.
   Happening since 2.1.0.
 * Metadata refreshes without partition leader change could lead to a loop of
   metadata calls at fixed intervals. Solved by stopping metadata refresh when
   all existing metadata is non-stale. Happening since 2.3.0.
 * A partition migration could happen, using stale metadata, when the partition
   was undergoing a validation and being retried because of an error.
   Solved by doing a partition migration only with a non-stale leader epoch.
   Happening since 2.1.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ librdkafka v2.3.1 is a maintenance release:
  * Fix pipeline inclusion of static binaries (#4666)
  * Fix to main loop timeout calculation leading to a tight loop for a
    max period of 1 ms (#4671).
+ * [KIP-516](https://cwiki.apache.org/confluence/display/KAFKA/KIP-516%3A+Topic+Identifiers)
+   Continue partial implementation by adding a metadata cache by topic id
+   and updating the topic id corresponding to the partition name (#4660)
+ * Fixes to metadata cache expiration, metadata refresh interruption and
+   to avoid usage of stale metadata (#4660).
 
 
 ## Fixes
@@ -26,6 +31,21 @@ librdkafka v2.3.1 is a maintenance release:
    before the expiration of a timeout, it was serving with a zero timeout,
    leading to increased CPU usage until the timeout was reached.
    Happening since 1.x (#4671).
+ * Metadata cache was cleared on full metadata refresh, leading to unnecessary
+   refreshes and occasional `UNKNOWN_TOPIC_OR_PART` errors. Solved by updating
+   cache for existing or hinted entries instead of clearing them.
+   Happening since 2.1.0 (#4660).
+ * A metadata call before member joins consumer group,
+   could lead to an `UNKNOWN_TOPIC_OR_PART` error. Solved by updating
+   the consumer group following a metadata refresh only in safe states.
+   Happening since 2.1.0 (#4660).
+ * Metadata refreshes without partition leader change could lead to a loop of
+   metadata calls at fixed intervals. Solved by stopping metadata refresh when
+   all existing metadata is non-stale. Happening since 2.3.0 (#4660).
+ * A partition migration could happen, using stale metadata, when the partition
+   was undergoing a validation and being retried because of an error.
+   Solved by doing a partition migration only with a non-stale leader epoch.
+   Happening since 2.1.0 (#4660).
 
 
 
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -2333,10 +2333,6 @@ static int rd_kafka_cgrp_metadata_refresh(rd_kafka_cgrp_t *rkcg,
 
         rd_list_init(&topics, 8, rd_free);
 
-        /* Insert all non-wildcard topics in cache. */
-        rd_kafka_metadata_cache_hint_rktparlist(
-            rkcg->rkcg_rk, rkcg->rkcg_subscription, NULL, 0 /*dont replace*/);
-
         if (rkcg->rkcg_flags & RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION) {
                 /* For wildcard subscriptions make sure the
                  * cached full metadata isn't too old. */
@@ -5005,6 +5001,19 @@ rd_kafka_cgrp_calculate_subscribe_revoking_partitions(
         return revoking;
 }
 
+static void
+rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
+                               rd_kafka_topic_partition_list_t *rktparlist) {
+        rkcg->rkcg_subscription = rktparlist;
+        if (rkcg->rkcg_subscription) {
+                /* Insert all non-wildcard topics in cache immediately.
+                 * to avoid problems with subsequent metadata
+                 * requests. */
+                rd_kafka_metadata_cache_hint_rktparlist(
+                    rkcg->rkcg_rk, rkcg->rkcg_subscription, NULL,
+                    0 /*dont replace*/);
+        }
+}
 
 /**
  * @brief Handle a new subscription that is modifying an existing subscription
@@ -5037,7 +5046,7 @@ rd_kafka_cgrp_modify_subscription(rd_kafka_cgrp_t *rkcg,
             rkcg, unsubscribing_topics);
 
         rd_kafka_topic_partition_list_destroy(rkcg->rkcg_subscription);
-        rkcg->rkcg_subscription = rktparlist;
+        rd_kafka_cgrp_subscription_set(rkcg, rktparlist);
 
         if (rd_kafka_cgrp_metadata_refresh(rkcg, &metadata_age,
                                            "modify subscription") == 1) {
@@ -5146,7 +5155,7 @@ static rd_kafka_resp_err_t rd_kafka_cgrp_unsubscribe(rd_kafka_cgrp_t *rkcg,
 
         if (rkcg->rkcg_subscription) {
                 rd_kafka_topic_partition_list_destroy(rkcg->rkcg_subscription);
-                rkcg->rkcg_subscription = NULL;
+                rd_kafka_cgrp_subscription_set(rkcg, NULL);
         }
 
         if (rkcg->rkcg_group_protocol == RD_KAFKA_GROUP_PROTOCOL_GENERIC)
@@ -5244,7 +5253,7 @@ rd_kafka_cgrp_subscribe(rd_kafka_cgrp_t *rkcg,
         if (rd_kafka_topic_partition_list_regex_cnt(rktparlist) > 0)
                 rkcg->rkcg_flags |= RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION;
 
-        rkcg->rkcg_subscription = rktparlist;
+        rd_kafka_cgrp_subscription_set(rkcg, rktparlist);
 
         rd_kafka_cgrp_join(rkcg);
 
@@ -5909,8 +5918,7 @@ rd_kafka_cgrp_consumer_subscribe(rd_kafka_cgrp_t *rkcg,
                     RD_KAFKA_CGRP_CONSUMER_F_SUBSCRIBED_ONCE |
                     RD_KAFKA_CGRP_CONSUMER_F_SEND_NEW_SUBSCRIPTION;
 
-                rkcg->rkcg_subscription = rktparlist;
-
+                rd_kafka_cgrp_subscription_set(rkcg, rktparlist);
                 rd_kafka_cgrp_consumer_expedite_next_heartbeat(rkcg);
         } else {
                 rd_kafka_cgrp_unsubscribe(rkcg, rd_true /*leave group*/);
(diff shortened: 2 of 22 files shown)


emasab:
Closes #4577

CHANGELOG.md
```diff
@@ -6,6 +6,36 @@ librdkafka v2.3.1 is a maintenance release:
...
+ * Fix to main loop timeout calculation leading to a tight loop for a period
+   of 1 ms max (#4660).
+
+
+## Fixes
+
+### General fixes
+
+ * Metadata cache was cleared on full metadata refresh, leading to unnecessary
+   refreshes and accasional `UNKNOWN_TOPIC_OR_PART` errors. Solved by updating
```
Quuxplusone, line 23:
```suggestion
   refreshes and occasional `UNKNOWN_TOPIC_OR_PART` errors. Solved by updating
```

CHANGELOG.md
```diff
@@ -6,6 +6,36 @@ librdkafka v2.3.1 is a maintenance release:
...
+   cache for existing or hinted entries instead of clearing them.
+   Happening since 2.1.0 (#4660).
+ * Metadata refreshes without partition leader change could lead to a loop of
+   intervaled metadata calls. Solved by stopping metadata refresh when
```
Quuxplusone, line 27: FWIW, "intervaled" isn't a word; I don't have enough context to know what this means.

CHANGELOG.md
```diff
@@ -6,6 +6,36 @@ librdkafka v2.3.1 is a maintenance release:
...
+   all existing metadata is non-stale. Happening since 2.3.0 (#4660).
+ * A partition migration could happen, using stale metadata, when the partition
+   was undergoing a validation and being retried because of an error.
+   Solved by doing a partition migration only with a non-stale leader epoch.
+   Happening since 2.1.0 (#4660).
+ * In librdkafka main thread loop, when it was awaken less that 1 ms
```
Quuxplusone, line 33:
```suggestion
 * When the main thread loop was awakened less than 1 ms
```

CHANGELOG.md
```diff
@@ -6,6 +6,36 @@ librdkafka v2.3.1 is a maintenance release:
...
+   before the expiration of a timeout, it was serving with a zero timeout,
+   leading to increased CPU usage until the timeout was reached.
+   Happening since 1.x (#4660).
```
Quuxplusone, line 36: FWIW, it's mildly confusing that all these bullets point to #4660 (this PR), instead of to the actual bug reports about the bugs being described. Do such bug reports exist?

src/rdkafka.c
```diff
@@ -2127,7 +2127,12 @@ static int rd_kafka_thread_main(void *arg) {
                                        RD_KAFKA_CGRP_STATE_TERM)))) {
                 rd_ts_t sleeptime = rd_kafka_timers_next(
                     &rk->rk_timers, 1000 * 1000 /*1s*/, 1 /*lock*/);
-                rd_kafka_q_serve(rk->rk_ops, (int)(sleeptime / 1000), 0,
+                /* Round up to avoid calling serve with timeout 0 ms
+                 * in a tight loop until 1 ms has passed. */
+                int timeout_ms = (int)(sleeptime / 1000);
+                if (sleeptime % 1000 > 0)
+                        timeout_ms++;
```
Quuxplusone, line 2134: This is the same as
```
int timeout_ms = (int)(sleeptime / 1000) + (sleeptime % 1000 > 0);
```
or perhaps even better,
```
int timeout_ms = ((sleeptime + 999) / 1000);
```
(The cast to `int` probably isn't needed, unless it's suppressing a compiler warning.)

Also, this diff seems totally unrelated to the "metadata cache" stuff in this PR. This should be a separate commit with a separate PR.

src/rdkafka_metadata_cache.c
```diff
@@ -221,6 +209,25 @@ rd_kafka_metadata_cache_find(rd_kafka_t *rk, const char *topic, int valid) {
...
+/**
+ * @brief Find cache entry by topic id
+ *
+ * @param valid: entry must be valid (not hint)
+ *
+ * @locks rd_kafka_*lock()
+ */
+struct rd_kafka_metadata_cache_entry *
+rd_kafka_metadata_cache_find_by_id(rd_kafka_t *rk,
+                                   const rd_kafka_Uuid_t topic_id,
```
Quuxplusone, line 221: `const rd_kafka_Uuid_t` passes "by const value," which is never correct. Surely the compiler warned about this?

Everywhere else in librdkafka that takes `rd_kafka_Uuid_t` either takes it by pointer to avoid a copy — `const rd_kafka_Uuid_t*` — which is probably what you want here, since you never use the parameter's own value and it's large — or takes it by "non-const value" in two places (`rd_kafka_topic_partition_new_with_topic_id` and `rd_kafka_Uuid_cmp`: the latter is inline and the former looks like a flaw to me but it's part of the public API now).

src/rdkafka_mock.c
```diff
@@ -469,7 +469,38 @@ rd_kafka_mock_partition_assign_replicas(rd_kafka_mock_partition_t *mpart,
...
+        rd_kafka_mock_partition_leader_t *leader_response;
 
+        leader_response               = rd_calloc(1, sizeof(*leader_response));
+        leader_response->leader_id    = leader_id;
+        leader_response->leader_epoch = leader_epoch;
+        TAILQ_INSERT_TAIL(&mpart->leader_responses, leader_response, link);
+}
+
+/**
+ * @brief First forced partition leader response in passed \p mpart .
```
Quuxplusone, line 488: FWIW I don't understand this grammar. Does it mean `\p mpart holds the first forced response from the partition leader`, or I guess more likely `Fetch the first partition leader response into \p mpart`?

CHANGELOG.md
emasab: Thanks, fixed it

CHANGELOG.md
emasab: intervaled means at fixed intervals, it's rarely used I think. I can put the phrase "at fixed intervals" instead

src/rdkafka.c
emasab: Thanks, I think I'll use the second one

CHANGELOG.md
emasab: We usually put the id of the PR containing the changes there, then the PR needs to link to the corresponding issues that closes. I update this PR

src/rdkafka_metadata_cache.c
emasab: Internally we pass the Uuid by struct copy, as a primitive value, to avoid more heap allocations. Here the const isn't strictly necessary, but it doesn't hurt as that parameter shouldn't be modified inside the function

src/rdkafka.c
emasab: Putting it in a different commit too

src/rdkafka_metadata.c
```diff
@@ -954,17 +1001,17 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
-                rd_kafka_metadata_cache_expiry_start(rk);
         }
-
-
-        // TODO: Should be done for requested_topic_ids as well.
         /* Remove cache hints for the originally requested topics. */
         if (requested_topics)
                 rd_kafka_metadata_cache_purge_hints(rk, requested_topics);
+        if (requested_topic_ids)
+                rd_kafka_metadata_cache_purge_hints(rk, requested_topic_ids);
```
pranavrth, line 1009: `rd_kafka_metadata_cache_purge_hints_by_id`

src/rdkafka_metadata.c
```diff
@@ -954,17 +1001,17 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
+
+        if (cache_changes) {
+                rd_kafka_metadata_cache_propagate_changes(rk);
+                rd_kafka_metadata_cache_expiry_start(rk);
```
pranavrth, line 1013: - Do you mean to include expiry start call inside the if condition? Earlier it was outside.
- What happens to the cache entries which were not updated? Do we restart the timer for them as well?

pranavrth (changes requested):
Reviewed few files which I understood. Reviewing more. Please check comments in the original PR as well.

src/rdkafka_mock.c
emasab: It's:
`Return the first forced partition leader response in \p mpart , if available`

src/rdkafka_metadata.c
```diff
@@ -474,12 +474,14 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
-
-        const rd_list_t *requested_topics = request_topics;
-        rd_bool_t all_topics              = rd_false;
-        rd_bool_t cgrp_update             = rd_false;
+        const int log_decode_errors  = LOG_ERR;
+        rd_list_t *missing_topics    = NULL;
+        rd_list_t *missing_topic_ids = NULL;
+
+        const rd_list_t *requested_topics    = request_topics;
+        const rd_list_t *requested_topic_ids = NULL;
```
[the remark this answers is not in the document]
emasab: Updated the documentation for the three versions of MetadataRequests

src/rdkafka_metadata.c
emasab: Here it's restarting the timer if some entries have been updated, the timer is calculated from the expiry time of the first element, the one that expires first. Earlier was outside just in this code, in trunk it's done only if `!all_topics` but now existing entries can be updated in that case.

pranavrth (changes requested):
Some review for Cache.

src/rdkafka_metadata.c
```diff
@@ -831,34 +839,37 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
+                if (requested_topic_ids)
+                        rd_list_free_cb(
+                            missing_topic_ids,
+                            rd_list_remove_cmp(missing_topic_ids,
+                                               &mdi->topics[i].topic_id,
+                                               (void *)rd_kafka_Uuid_ptr_cmp));
+                /* Only update cache when not asking
+                 * for all topics or cache entry
+                 * already exists. */
+                rd_kafka_wrlock(rk);
```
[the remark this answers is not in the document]
emasab: To call `rd_kafka_metadata_cache_find `and `rd_kafka_metadata_cache_find_by_id` the lock is required because it's reading the AVL tree, it could use a read lock but that cannot be upgraded to a write lock and if releasing it, after acquiring the write lock it should have to check again if the entry was deleted in the meantime.
I think the lock is rarely contended here and in that case the futex implementation will ensure no system calls are made.

src/rdkafka_metadata.c
```diff
@@ -890,6 +901,41 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
+                        rd_kafka_rdunlock(rk);
+                        if (rkt) {
+                                /* Received metadata response contained no
+                                 * information about topic 'rkt' and thus
+                                 * indicates the topic is not available in the
+                                 *  cluster.
+                                 * Mark the topic as non-existent */
+                                rd_kafka_topic_wrlock(rkt);
+                                rd_kafka_topic_set_notexists(
+                                    rkt, RD_KAFKA_RESP_ERR__UNKNOWN_TOPIC);
```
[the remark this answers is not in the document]
emasab: This is the local error `_UNKNOWN_TOPIC`, not the broker one `UNKNOWN_TOPIC_OR_PART` or `UNKNOWN_TOPIC_ID` so it's correct to use the same code to signal that the topic is unknown, independently of which parameter was used to query for it

src/rdkafka_cgrp.c
```diff
@@ -6462,11 +6462,27 @@ void rd_kafka_cgrp_metadata_update_check(rd_kafka_cgrp_t *rkcg,
...
+        if (rkcg->rkcg_group_protocol != RD_KAFKA_GROUP_PROTOCOL_GENERIC)
+                return;
+
         if (!rkcg->rkcg_subscription || rkcg->rkcg_subscription->cnt == 0)
                 return;
 
-        if (rkcg->rkcg_group_protocol != RD_KAFKA_GROUP_PROTOCOL_GENERIC)
+        if (!(rkcg->rkcg_flags & RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION) &&
+            rkcg->rkcg_join_state != RD_KAFKA_CGRP_JOIN_STATE_WAIT_METADATA &&
+            rkcg->rkcg_join_state != RD_KAFKA_CGRP_JOIN_STATE_STEADY) {
```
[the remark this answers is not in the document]
emasab: For sure we know that in states RD_KAFKA_CGRP_JOIN_STATE_WAIT_INIT and RD_KAFKA_CGRP_JOIN_STATE_WAIT_JOIN it can cause the issue #4589 . I was also considering to put `rkcg_join_state >= RD_KAFKA_CGRP_JOIN_STATE_WAIT_METADATA` but doing these updates with a stable group seems more error proof.
In RD_KAFKA_CGRP_JOIN_STATE_WAIT_METADATA it's required because there are cases where it's waiting exactly this update.
For a regex subscription it's required in state INIT too, because it has to find the topics to include in the JoinGroup request, but in that case it doesn't have the problem with metadata cache.

emasab:
Closing in favour of

#4676 
#4677
#4678
#4679
#4680

PR #4661, opened 2024-03-26
[KIP-848] Use metadata cache by topic id, fixes and improvements

emasab:
- use metadata cache by topic id
- rename 'generic' protocol to 'classic'
- consumer group serve timer to awake the loop earlier
- compare and find into topic partition list by topic id only
- fix memory leak when instance creation fails and app_conf is provided
- fix cases where HB response is received after unsubscription
- use topic name from current assignment if it's missing from metadata


src/rdkafka.c
```diff
@@ -2661,6 +2666,8 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
          * that belong to rk_conf and thus needs to be cleaned up.
          * Legacy APIs, sigh.. */
         if (app_conf) {
+                if (group_remote_assignor_override)
+                        rd_free(group_remote_assignor_override);
```
pranavrth, line 2670: This can be outside of `if (app_conf)` condition as we are already doing a NULL check here.

pranavrth (changes requested):
Checked all files except `rdkafka_cgrp.c`. Have some doubts in that so we can discuss them first.

src/rdkafka.c
emasab: if not `app_conf `the property if freed in `rd_kafka_destroy_internal `so it would be a double free, here it's setting memory to zero later
`memset(&rk->rk_conf, 0, sizeof(rk->rk_conf));`
so they're not freed by `rd_kafka_destroy_internal ` and the properties that belonged to `app_conf `are still valid pointers that are freed by the app.

src/rdkafka.c
```diff
@@ -2381,14 +2382,18 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
...
-                        if (range_assignor && range_assignor->rkas_enabled)
-                                rk->rk_conf.group_remote_assignor =
+                        if (range_assignor && range_assignor->rkas_enabled) {
+                                group_remote_assignor_override =
                                     rd_strdup("range");
+                                rk->rk_conf.group_remote_assignor =
+                                    group_remote_assignor_override;
+                        }
                 }
         }
```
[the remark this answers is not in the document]
emasab: To be able to do a transparent upgrade until next major release we want to detect the methods it's going to use in rebalance_cb, and also give the same assignment results. I'm assuming broker assignors will give the same results in terms of stickyness or rack-awareness. For roundrobin I think it's best to revert the configuration to the classic protocol, instead when `grop.remote.assignor` is set we can assume they start directly with the incremental methods.

src/rdkafka_cgrp.c
```diff
@@ -2666,21 +2692,76 @@ static rd_kafka_op_res_t rd_kafka_cgrp_consumer_handle_next_assignment(
...
+
+                if (missing_topic_ids) {
+                        rd_kafka_Uuid_t topic_id;
+                        if (unlikely(!*missing_topic_ids))
+                                *missing_topic_ids =
+                                    rd_list_new(1, rd_list_Uuid_destroy);
+                        topic_id = rd_kafka_topic_partition_get_topic_id(
+                            &assignment->elems[i]);
+                        rd_list_add(*missing_topic_ids,
+                                    rd_kafka_Uuid_copy(&topic_id));
```
pranavrth, line 2741:
```suggestion
                        if (unlikely(!*missing_topic_ids))
                                *missing_topic_ids =
                                    rd_list_new(1, rd_list_Uuid_destroy);
                        rd_list_add(*missing_topic_ids,
                                    rd_kafka_Uuid_copy(&request_topic_id));
```

src/rdkafka_cgrp.c
```diff
@@ -2666,21 +2692,76 @@ static rd_kafka_op_res_t rd_kafka_cgrp_consumer_handle_next_assignment(
...
+                }
+                rd_kafka_dbg(rkcg->rkcg_rk, CGRP, "HEARTBEAT",
+                             "Metadata not found for the "
+                             "assigned topic id - %s."
+                             " Continuing without it",
+                             rd_kafka_Uuid_base64str(&request_topic_id));
+        }
+        if (missing_topic_ids && *missing_topic_ids)
+                rd_list_deduplicate(missing_topic_ids,
+                                    (void *)rd_kafka_Uuid_ptr_cmp);
```
pranavrth, line 2751: Add a FIXME to change the name of `rd_list_deduplicate` as it sorts as well. Ideally, the function should not change the order. If we don't add sorted then it might be assumed that the order is not changed.

src/rdkafka_cgrp.c
```diff
@@ -2689,89 +2770,45 @@ rd_kafka_cgrp_consumer_handle_Metadata_op(rd_kafka_t *rk,
...
-                                        rd_kafka_dbg(
-                                            rkcg->rkcg_rk, CGRP, "HEARTBEAT",
-                                            "Metadata not found for the "
-                                            "assigned topic id - %s due to: "
-                                            "%s: "
-                                            "Continuing without it",
-                                            rd_kafka_Uuid_base64str(
-                                                &request_topic_id),
-                                            rd_kafka_err2str(
-                                                rko->rko_u.metadata.md
```
pranavrth, line 2730: Shall we capture this error when we get the metadata response?

src/rdkafka_cgrp.c
```diff
@@ -2689,89 +2770,45 @@ rd_kafka_cgrp_consumer_handle_Metadata_op(rd_kafka_t *rk,
...
-                    sizeof(new_target_assignment_str), 0);
+                    assignment_with_metadata, assignment_with_metadata_str,
+                    sizeof(assignment_with_metadata_str), 0);
 
                 rd_kafka_dbg(
                     rkcg->rkcg_rk, CGRP, "HEARTBEAT",
-                    "Metadata available for %d/%d next target assignment "
-                    "which are: \"%s\"",
-                    new_target_assignment->cnt,
+                    "Metadata available for %d/%d of next target assignment, "
```
pranavrth, line 2792: Better wording I think.

src/rdkafka_cgrp.c
```diff
@@ -2885,62 +2927,75 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
...
+                                    rkcg_current_assignment_str,
+                                    sizeof(rkcg_current_assignment_str), 0);
+
+                                rd_kafka_dbg(rkcg->rkcg_rk, CGRP, "HEARTBEAT",
+                                             "Target assignment acked, new "
+                                             "current assignment "
+                                             " \"%s\"",
+                                             rkcg_current_assignment_str);
+                        }
+                } else if (rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION) {
```
pranavrth, line 2971: Need more understanding on this change.

src/rdkafka_cgrp.c
```diff
@@ -2885,62 +2927,75 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
...
+                }
+        }
 
         rkcg->rkcg_flags &= ~RD_KAFKA_CGRP_F_HEARTBEAT_IN_TRANSIT;
         rkcg->rkcg_consumer_flags &=
-            ~RD_KAFKA_CGRP_CONSUMER_F_SENDING_NEW_SUBSCRIPTION;
-        rkcg->rkcg_last_heartbeat_err = RD_KAFKA_RESP_ERR_NO_ERROR;
+            ~RD_KAFKA_CGRP_CONSUMER_F_SENDING_NEW_SUBSCRIPTION &
+            ~RD_KAFKA_CGRP_CONSUMER_F_SEND_FULL_REQUEST &
+            ~RD_KAFKA_CGRP_CONSUMER_F_SENDING_ACK;
```
pranavrth, line 2996: What if the state is not STEADY?

src/rdkafka_cgrp.c
```diff
@@ -3033,10 +3089,12 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
                 /* Re-query for coordinator */
                 rkcg->rkcg_consumer_flags |=
                     RD_KAFKA_CGRP_CONSUMER_F_SEND_FULL_REQUEST;
+                rd_kafka_cgrp_consumer_expedite_next_heartbeat(rkcg);
```
pranavrth, line 3092: Shouldn't it be after next line?

src/rdkafka_cgrp.c
```diff
@@ -5658,20 +5720,7 @@ void rd_kafka_cgrp_consumer_group_heartbeat(rd_kafka_cgrp_t *rkcg,
...
-                        /* Don't send heartbeats if max.poll.interval.ms was
-                         * exceeded */
-                        return;
-                } else {
-                        rkcg->rkcg_flags &= ~RD_KAFKA_CGRP_F_MAX_POLL_EXCEEDED;
-                }
-        }
-
-        /* Skip heartbeat if we have one in transit */
-        if (rkcg->rkcg_flags & RD_KAFKA_CGRP_F_HEARTBEAT_IN_TRANSIT)
```
pranavrth, line 5672: Why removing this?

src/rdkafka_cgrp.c
```diff
@@ -5751,137 +5837,42 @@ void rd_kafka_cgrp_consumer_serve(rd_kafka_cgrp_t *rkcg) {
...
+                } else {
+                        next_heartbeat = -1 * next_heartbeat;
                 }
-
-                for (j = 0; j < subscription->cnt; j++) {
-                        const char *pattern = subscription->elems[j].topic;
-                        if (rd_kafka_topic_match(rkcg->rkcg_rk, pattern,
-                                                 topic)) {
-                                break;
+                if (likely(rkcg->rkcg_heartbeat_intvl_ms > 0)) {
```
pranavrth, line 5851: Isn't the next timer already set?

pranavrth (changes requested):
More comments.

src/rdkafka_cgrp.c
emasab: I can add it to the documentation of this function later, the `missing_topic_ids` is just created so it doesn't change the order of the assignment. It's just used for the Metadata request

src/rdkafka_cgrp.c
emasab: Moved to `rd_kafka_cgrp_consumer_heartbeat_preconditions_met`

pranavrth (approved):
LGTM!.. Added couple of comments to track.

PR #4666, opened 2024-04-01
Fix pipeline: inclusion of static binaries

emasab:
without gssapi dependency


milindl (approved):
Thanks for the PR, LGTM!

PR #4667, opened 2024-04-01
Revert 8e20e1ee (#4117) to fix hang in destruction of groupconsumer

Quuxplusone:
We observed that destroying a groupconsumer would often hang waiting for the broker thread to exit. We tediously bisected the problem to the specific commit 8e20e1ee (the last commit before the v2.0.0rc1 tag). Only then did we find that a lot of people on GitHub were already complaining about that commit as introducing a resource leak: the commit adds a call to `rd_kafka_toppar_keep` that bumps the refcount of the toppar, and I don't immediately see a corresponding `rd_kafka_toppar_destroy` anywhere.

Reverting 8e20e1ee (as in this commit) does fix the hang in groupconsumer destruction which we were observing, so we've applied this patch to our downstream library.

Fixes #4486.

Issue #4486, opened 2023-10-27
C++ consumer APIs have memory leaks under certain conditions


zhangyafeikimi:
Description
===========
C++ consumer APIs have memory leaks under certain conditions.

How to reproduce
================
OS: Linux CentOS 7.9.2009.
When the kafka is not started or the topic does not exist, C++ consumer APIs have memory leaks.

I verified it with [examples/rdkafka_example.cpp](https://github.com/confluentinc/librdkafka/blob/master/examples/rdkafka_example.cpp).

```shell
valgrind --leak-check=full --show-leak-kinds=all ./rdkafka_example -C -t test2 -p 0
```

The result is like:

```plain
==2446== Memcheck, a memory error detector
==2446== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==2446== Using Valgrind-3.21.0 and LibVEX; rerun with -h for copyright info
==2446== Command: ./rdkafka_example -C -t test2 -p 0
==2446==
% Created consumer rdkafka#consumer-1
^C==2446==
==2446== HEAP SUMMARY:
==2446==     in use at exit: 2,988 bytes in 10 blocks
==2446==   total heap usage: 337 allocs, 327 frees, 78,889 bytes allocated
==2446==
==2446== 20 bytes in 2 blocks are indirectly lost in loss record 1 of 9
==2446==    at 0x4C2B0D5: malloc (in /opt/p13n-valgrind/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2446==    by 0x5B08B89: strdup (in /usr/lib64/libc-2.17.so)
==2446==    by 0x4A6138: rd_strdup (rd.h:156)
==2446==    by 0x4A6138: rd_kafka_anyconf_set_prop0 (rdkafka_conf.c:1725)
==2446==    by 0x4A6138: rd_kafka_defaultconf_set (rdkafka_conf.c:2171)
==2446==    by 0x4A63D7: rd_kafka_topic_conf_new (rdkafka_conf.c:2191)
==2446==    by 0x4A68DE: rd_kafka_topic_conf_dup (rdkafka_conf.c:2623)
==2446==    by 0x40CF06: RdKafka::Topic::create(RdKafka::Handle*, std::string const&, RdKafka::Conf const*, std::string&) (TopicImpl.cpp:92)
==2446==    by 0x405FF4: main (rdkafka_example.cpp:531)
==2446==
[...]

PR #4669, opened 2024-04-03
Fix memory leak while stopping an unknown partition

emasab:
When stopping a partition that isn't known to the client a reference to that partition was kept in the fetchq of the same, leading to a memory leak.
Solved by purging the fetch and ops queue of the unknown partition when it's stopped

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,22 @@
 librdkafka v2.3.1 is a maintenance release:
 
  * Upgrade OpenSSL to v3.0.12 (while building from source) with various security fixes,
-   check the [release notes](https://www.openssl.org/news/cl30.txt).
+   check the [release notes](https://www.openssl.org/news/cl30.txt) (#4586).
  * Integration tests can be started in KRaft mode and run against any
-   GitHub Kafka branch other than the released versions.
+   GitHub Kafka branch other than the released versions (#4635).
+ * Fix memory leak while stopping an unknown partition (#).
+
+
+## Fixes
+
+### Consumer fixes
+
+  * When stopping a partition that isn't known to the client a reference
+    to that partition was kept in the fetchq of the same,
+    leading to a memory leak.
+    Solved by purging the fetch and ops queue of the unknown partition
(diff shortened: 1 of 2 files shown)


Quuxplusone:
In #4486 I wrote:
> My employer has also tentatively bisected an issue to https://github.com/confluentinc/librdkafka/commit/8e20e1ee79b188ae610aac3a2d2517f7f12dd890. We see that if we have a groupconsumer, and the broker goes down (that is, we shut down the Kafka broker), and then comes back up, and then we try to shut down the groupconsumer, it simply hangs forever. This doesn't reproduce in 1.9.3rc2 or earlier, but it does reproduce in 2.0.0rc1 through 2.3.0. In fact, it appears not to reproduce in https://github.com/confluentinc/librdkafka/commit/a83cadf5eab3a43f7f0d3dd09f5b1a3e9e88312f but to start reproducing with https://github.com/confluentinc/librdkafka/commit/8e20e1ee79b188ae610aac3a2d2517f7f12dd890.

@emasab wrote:
> @Quuxplusone in case it causes a hang, please try this fix and in case it's still happening tell me how to reproduce it.

Okay, I just tried it. My employer's test case is _red_ with librdkafka v2.3.0, _green_ with librdkafka v2.3.0-plus-#4667, but unfortunately remains _red_ with librdkafka v2.3.0-plus-#4669.

That's not to say that #4669 is useless or fails to fix a real bug; it just doesn't fix _our_ bug.

OTOH I also don't claim that #4669 is bug-free. Its new code that takes two mutex locks at once is clearly at _risk_ of deadlock. It would benefit from a code comment explaining (or linking to) the project's [lock-ordering](https://softwareengineering.stackexchange.com/questions/425016/what-is-a-lock-ordering-inversion) policy.

emasab:
@Quuxplusone thanks for testing it. The code respects the lock ordering policy, that is:

1. rk main lock
2. rkb lock
3. rkt lock
4. rktp lock

Is your failing test doing something specific, like deleting a topic or using the `cooperative-sticky` assignor?

PR #4671, opened 2024-04-03
Fix to main loop timeout calculation leading to a tight loop for a max period of 1 ms

emasab:
When the main thread loop was awakened less than 1 ms before the expiration of a timeout, it was serving with a zero timeout, leading to increased CPU usage until the timeout was reached. Happening since 1.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ librdkafka v2.3.1 is a maintenance release:
  * Integration tests can be started in KRaft mode and run against any
    GitHub Kafka branch other than the released versions.
  * Fix pipeline inclusion of static binaries (#4666)
+ * Fix to main loop timeout calculation leading to a tight loop for a
+   max period of 1 ms (#4671).
 
 
 ## Fixes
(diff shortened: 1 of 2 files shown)


src/rdkafka.c
```diff
@@ -2127,7 +2127,10 @@ static int rd_kafka_thread_main(void *arg) {
                                        RD_KAFKA_CGRP_STATE_TERM)))) {
                 rd_ts_t sleeptime = rd_kafka_timers_next(
                     &rk->rk_timers, 1000 * 1000 /*1s*/, 1 /*lock*/);
-                rd_kafka_q_serve(rk->rk_ops, (int)(sleeptime / 1000), 0,
+                /* Use ceiling division to avoid calling serve with a 0 ms
+                 * timeout in a tight loop until 1 ms has passed. */
+                int timeout_ms = (sleeptime + 999) / 1000;
```
pranavrth, line 2132: I had one concern with the overflow but that will never happen as we use `CLOCK_MONOTONIC` and hence will never reach the limit for `rd_ts_t`.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ librdkafka v2.3.1 is a maintenance release:
    and updating the topic id corresponding to the partition name (#4660)
  * Fixes to metadata cache expiration, metadata refresh interruption and
    to avoid usage of stale metadata (#4660).
+ * Fix to main loop timeout calculation leading to a tight loop for a
+   max period of 1 ms (#4671).
```
pranavrth, line 15: Let's use something easier to understand than `tight loop`. The later statement is more comprehendible than this.

pranavrth (dismissed):

pranavrth (approved):

PR #4672, opened 2024-04-04
[KIP-848] Fixes related to Offset Commit previous error and ConsumerGroupHeartbeat not updating member epoch in a case

pranavrth:


src/rdkafka_cgrp.c
```diff
@@ -3106,6 +3104,18 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
         }
 
         rd_kafka_buf_read_i32(rkbuf, &member_epoch);
+        rkcg->rkcg_generation_id = member_epoch;
```
emasab, line 3107: Remove the same assignment at line 3123

tests/test.c
```diff
@@ -531,13 +532,13 @@ struct test tests[] = {
     _TEST(0145_pause_resume_mock, TEST_F_LOCAL),
     _TEST(0146_metadata_mock, TEST_F_LOCAL),
     _TEST(0147_consumer_group_consumer_mock, TEST_F_LOCAL),
+    _TEST(0148_offset_fetch_commit_error_mock, TEST_F_LOCAL),
```
emasab: This tests is only for KIP-848 it would be better to name it so it says: this is for KIP-848 OffsetFetch and OffsetCommit , errors but also anything else. So:
```suggestion
    _TEST(0148_consumer_group_consumer_offset_fetch_commit_mock, TEST_F_LOCAL),
```

tests/test.c
```diff
@@ -531,13 +532,13 @@ struct test tests[] = {
...
     _TEST(0149_broker_same_host_port_mock, TEST_F_LOCAL),
     _TEST(0150_telemetry_mock, 0),
     _TEST(0151_purge_brokers_mock, TEST_F_LOCAL),
     _TEST(0152_rebootstrap_local, TEST_F_LOCAL),
     _TEST(0153_memberid, 0, TEST_BRKVER(0, 4, 0, 0)),
```
emasab: This tests is also for KIP-848 only, it would be better:
```suggestion
    _TEST(0153_consumer_group_consumer_memberid, 0, TEST_BRKVER(0, 4, 0, 0)),
```

win32/tests/tests.vcxproj
```diff
@@ -228,6 +228,7 @@
     <ClCompile Include="..\..\tests\0145-pause_resume_mock.c" />
     <ClCompile Include="..\..\tests\0146-metadata_mock.c" />
     <ClCompile Include="..\..\tests\0147-consumer_group_consumer_mock.c" />
+    <ClCompile Include="..\..\tests\0148-offset_fetch_commit_error_mock.c" />
```
emasab: Reflect the changes in test naming, in `CMakeLists.txt` too

tests/0148-offset_fetch_commit_error_mock.c
```diff
@@ -0,0 +1,347 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2023, Confluent Inc.
```
emasab: Should be 2025, copyright is per file and starts when file is created and updated when file is updated.

tests/0148-offset_fetch_commit_error_mock.c
```diff
@@ -0,0 +1,347 @@
...
+        if (test_consumer_group_protocol_classic()) {
+                TEST_SKIP(
+                    "Test not meaningful with Consumer Group 'Classic' "
+                    "protocol\n");
+                return 0;
+        }
+
+        mcluster = test_mock_cluster_new(3, &bootstraps);
+
+        do_test_OffsetFetch_stale_member_error(mcluster, bootstraps);
```
emasab: Could call both this and `do_test_OffsetCommit_automatic_stale_member` with `stale_member_epoch`
```suggestion
        do_test_OffsetFetch_stale_member_epoch(mcluster, bootstraps);
```
and `do_test_OffsetCommit_automatic_stale_member_epoch`

tests/0148-offset_fetch_commit_error_mock.c
```diff
@@ -0,0 +1,347 @@
...
+ *
+ *        - variation 0: during_revocation=false, session_times_out=false
+ *        - variation 1: during_revocation=false, session_times_out=true
+ *        - variation 2: during_revocation=true, session_times_out=false
+ *        - variation 3: during_revocation=true, session_times_out=true
+ */
+void do_test_OffsetCommit_automatic_stale_member(
+    rd_kafka_mock_cluster_t *mcluster,
+    const char *bootstraps,
+    int variation) {
```
emasab: Use an `enum` for variations as in other tests.

tests/0148-offset_fetch_commit_error_mock.c
```diff
@@ -0,0 +1,347 @@
...
+        rd_bool_t during_revocation = (variation / 2) == 1;
+        rd_bool_t session_times_out = (variation % 2) == 1;
+
+        SUB_TEST_QUICK("during_revocation=%s, session_times_out=%s",
+                       RD_STR_ToF(during_revocation),
+                       RD_STR_ToF(session_times_out));
+
+        rd_kafka_mock_topic_create(mcluster, topic, 1, 2);
+        rd_kafka_mock_coordinator_set(mcluster, "group", topic, 1);
+        rd_kafka_mock_set_group_consumer_session_timeout_ms(mcluster, 10000);
```
emasab: Can reduce session timeout to 3s and hearthbeat interval to 1s with `rd_kafka_mock_set_group_consumer_heartbeat_interval_ms`.

tests/0148-offset_fetch_commit_error_mock.c
```diff
@@ -0,0 +1,347 @@
...
+                                          RD_KAFKA_RESP_ERR_STALE_MEMBER_EPOCH);
+
+        if (session_times_out) {
+                /* Simulate a session timeout after that */
+                rd_kafka_mock_push_request_errors(
+                    mcluster, RD_KAFKAP_OffsetCommit, 4,
+                    RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID,
+                    RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID,
+                    RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID,
+                    RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID);
```
emasab: Just a single error should be enough for the session timeout. After that is retries the HB and that returns the error that causes the consumer leaving the group. If you delay the response for some time the session will expire and the HB will return that error as well.
Better to use some constants for session timeout and HB interval.

```suggestion
                rd_kafka_mock_broker_push_request_error_rtts(
                    mcluster, coordinator, RD_KAFKAP_OffsetCommit, 1,
                    RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID,
                    session_timeout_ms + heartbeat_interval_ms);
```

tests/0148-offset_fetch_commit_error_mock.c
```diff
@@ -0,0 +1,347 @@
...
+                 * but the commit fails. */
+                target_assignment_partitions =
+                    rd_kafka_topic_partition_list_new(2);
+                rd_kafka_topic_partition_list_add(target_assignment_partitions,
+                                                  topic, 0);
+                test_mock_cluster_member_assignment(
+                    mcluster, 1, consumer, target_assignment_partitions);
+                rd_kafka_topic_partition_list_destroy(
+                    target_assignment_partitions);
+                rd_sleep(1);
```
emasab: Instead of sleeping here you can set a log interceptor and with mutex and condition broadcast the event that partition was committed (case 2) or lost (case 3).
For case 2 you can look for the log message "cgrp auto commit timer: returned: Success" while for case 3: "current assignment of 1 partition(s) lost: member fenced"

tests/0148-offset_fetch_commit_error_mock.c
```diff
@@ -0,0 +1,347 @@
...
+        test_consumer_poll_exact("receive first batch", consumer, testid, -1, 0,
+                                 msgcnt, rd_true, &mv);
+        test_msgver_verify("verify first batch", &mv, TEST_MSGVER_ALL, 0,
+                           msgcnt);
+        test_msgver_clear(&mv);
+
+
+        rd_kafka_mock_clear_request_errors(mcluster, RD_KAFKAP_OffsetCommit);
+
+        /* First sequence of stale member epoch for 2 s */
```
emasab: These are permanent error and are retried at `auto.commit.interval.ms` if you set the configuration value to 1s here it becomes 4 seconds.
```suggestion
        /* First sequence of stale member epoch for 4 s */
```

src/rdkafka_cgrp.c
```diff
@@ -3106,6 +3104,18 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
...
+
+        rd_kafka_dbg(rk, CGRP, "HEARTBEAT",
+                     "ConsumerGroupHeartbeat response received for "
+                     "member id \"%.*s\" with epoch %d",
+                     RD_KAFKAP_STR_PR(&member_id), member_epoch);
+
+        if (rd_kafka_cgrp_will_leave(rkcg)) {
+                err = RD_KAFKA_RESP_ERR__OUTDATED;
```
emasab, line 3115: Was checking by putting an assert and this condition isn't tested. Have a strategy for writing a test, let's talk about it.

emasab (changes requested):
Thanks for the tests @pranavrth ! Have these comments.

tests/test.c
pranavrth: I don't agree that we should name tests based on the protocol. We should adjust the tests as we add more protocols like KIP-932. I would prefer to keep the name like this.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ librdkafka v2.12.0 is a feature release:
 
 ## Fixes
 
+### General fixes
+* Fix discarding consumer group heartbeat response without epoch update when leaving during inflight HB. (#4672)
+* Fix a condition where error was being raised in commit due to old error in the topic partition. (#4672)
```
```suggestion
* Fix for an error being raised after a commit due to an existing error in the topic partition (#4672).
```

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ librdkafka v2.12.0 is a feature release:
 
 ## Fixes
 
+### General fixes
+* Fix discarding consumer group heartbeat response without epoch update when leaving during inflight HB. (#4672)
```
```suggestion
* Fix for discarding the member epoch in a consumer group heartbeat response when leaving with an inflight HB (#4672).
```

emasab (changes requested):
Thanks Pranav, just suggesting some improvements for the change log entries.

emasab (approved):
Thanks for the fixes @pranavrth ! Let's later add a detail for the second fix that should be happening with the classic protocol too.

PR #4676, opened 2024-04-08
Metadata cache by topic id

emasab:
and improved documentation of Metadata functions

### DIFF
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -474,12 +474,14 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
         rd_kafka_metadata_internal_t *mdi = NULL;
         rd_kafka_metadata_t *md           = NULL;
         size_t rkb_namelen;
-        const int log_decode_errors = LOG_ERR;
-        rd_list_t *missing_topics   = NULL;
+        const int log_decode_errors  = LOG_ERR;
+        rd_list_t *missing_topics    = NULL;
+        rd_list_t *missing_topic_ids = NULL;
 
-        const rd_list_t *requested_topics = request_topics;
-        rd_bool_t all_topics              = rd_false;
-        rd_bool_t cgrp_update             = rd_false;
+        const rd_list_t *requested_topics    = request_topics;
+        const rd_list_t *requested_topic_ids = NULL;
+        rd_bool_t all_topics                 = rd_false;
+        rd_bool_t cgrp_update                = rd_false;
         rd_bool_t has_reliable_leader_epochs =
             rd_kafka_has_reliable_leader_epochs(rkb);
         int ApiVersion             = rkbuf->rkbuf_reqhdr.ApiVersion;
@@ -496,8 +498,9 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
         rd_bool_t compute_racks = has_client_rack;
 
         if (request) {
-                requested_topics = request->rkbuf_u.Metadata.topics;
-                all_topics       = request->rkbuf_u.Metadata.all_topics;
+                requested_topics    = request->rkbuf_u.Metadata.topics;
+                requested_topic_ids = request->rkbuf_u.Metadata.topic_ids;
+                all_topics          = request->rkbuf_u.Metadata.all_topics;
                 cgrp_update =
                     request->rkbuf_u.Metadata.cgrp_update && rk->rk_cgrp;
                 compute_racks |= request->rkbuf_u.Metadata.force_racks;
@@ -519,6 +522,9 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
         if (requested_topics)
                 missing_topics =
                     rd_list_copy(requested_topics, rd_list_string_copy, NULL);
+        if (requested_topic_ids)
+                missing_topic_ids =
+                    rd_list_copy(requested_topic_ids, rd_list_Uuid_copy, NULL);
 
         rd_kafka_broker_lock(rkb);
         rkb_namelen = strlen(rkb->rkb_name) + 1;
@@ -833,39 +839,42 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
                 rd_kafka_parse_Metadata_update_topic(rkb, &md->topics[i],
                                                      &mdi->topics[i]);
 
-                // TODO: Should be done for requested_topic_ids as well.
-                if (requested_topics) {
+                if (requested_topics)
                         rd_list_free_cb(missing_topics,
                                         rd_list_remove_cmp(missing_topics,
                                                            md->topics[i].topic,
                                                            (void *)strcmp));
-                        if (!all_topics) {
-                                /* Only update cache when not asking
-                                 * for all topics. */
+                if (requested_topic_ids)
+                        rd_list_free_cb(
+                            missing_topic_ids,
+                            rd_list_remove_cmp(missing_topic_ids,
+                                               &mdi->topics[i].topic_id,
+                                               (void *)rd_kafka_Uuid_ptr_cmp));
+                if (!all_topics) {
+                        /* Only update cache when not asking
+                         * for all topics. */
 
-                                rd_kafka_wrlock(rk);
-                                rd_kafka_metadata_cache_topic_update(
-                                    rk, &md->topics[i], &mdi->topics[i],
-                                    rd_false /*propagate later*/,
-                                    /* use has_client_rack rather than
-                                       compute_racks. We need cached rack ids
-                                       only in case we need to rejoin the group
-                                       if they change and client.rack is set
-                                       (KIP-881). */
-                                    has_client_rack, mdi->brokers,
-                                    md->broker_cnt);
-                                cache_changes++;
-                                rd_kafka_wrunlock(rk);
-                        }
+                        rd_kafka_wrlock(rk);
+                        rd_kafka_metadata_cache_topic_update(
+                            rk, &md->topics[i], &mdi->topics[i],
+                            rd_false /*propagate later*/,
+                            /* use has_client_rack rather than
+                            compute_racks. We need cached rack ids
+                            only in case we need to rejoin the group
+                            if they change and client.rack is set
+                            (KIP-881). */
+                            has_client_rack, mdi->brokers, md->broker_cnt);
+                        cache_changes++;
+                        rd_kafka_wrunlock(rk);
                 }
         }
 
-        // TODO: Should be done for missing_topic_ids as well.
         /* Requested topics not seen in metadata? Propogate to topic code. */
         if (missing_topics) {
                 char *topic;
                 rd_rkb_dbg(rkb, TOPIC, "METADATA",
-                           "%d/%d requested topic(s) seen in metadata",
+                           "%d/%d requested topic(s) seen in metadata"
+                           " (lookup by name)",
                            rd_list_cnt(requested_topics) -
                                rd_list_cnt(missing_topics),
                            rd_list_cnt(requested_topics));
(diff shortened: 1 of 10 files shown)


src/rdkafka_metadata.c
```diff
@@ -890,6 +901,42 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
+                                                              *topic_id);
+                        rd_kafka_rdunlock(rk);
+                        if (rkt) {
+                                /* Received metadata response contained no
+                                 * information about topic 'rkt' and thus
+                                 * indicates the topic is not available in the
+                                 *  cluster.
+                                 * Mark the topic as non-existent */
+                                rd_kafka_topic_wrlock(rkt);
+                                rd_kafka_topic_set_notexists(
```
pranavrth, line 932: There are few places in the function `rd_kafka_topic_set_notexists` which refers to `topic_name->str` which will give segfault. Even though we are not sure if this is even executed, its better to be fail safe.

src/rdkafka_metadata.c
```diff
@@ -980,7 +1027,8 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
          * which may contain only a sub-set of the subscribed topics (namely
          * the effective subscription of available topics) as to not
          * propagate non-included topics as non-existent. */
-        if (cgrp_update && (requested_topics || all_topics))
+        if (cgrp_update &&
+            (requested_topics || requested_topic_ids || all_topics))
                 rd_kafka_cgrp_metadata_update_check(rkb->rkb_rk->rk_cgrp,
```
pranavrth, line 1032: Same here. There are multiple places where we are using topic_name in the function or nested functions.. Either we fix those places or add atleast a FIXME here to check where topic name could cause segfault and fix it. Currently, topic id is used in metadata in only new protocol so the code works fine for now.

src/rdkafka_metadata.h
```diff
@@ -282,8 +286,14 @@ void rd_kafka_metadata_cache_topic_update(
 void rd_kafka_metadata_cache_propagate_changes(rd_kafka_t *rk);
 struct rd_kafka_metadata_cache_entry *
 rd_kafka_metadata_cache_find(rd_kafka_t *rk, const char *topic, int valid);
+struct rd_kafka_metadata_cache_entry *
+rd_kafka_metadata_cache_find_by_id(rd_kafka_t *rk,
+                                   const rd_kafka_Uuid_t,
```
pranavrth, line 291:
```suggestion
                                   const rd_kafka_Uuid_t topic_id,
```

src/rdkafka_metadata.h
```diff
@@ -282,8 +286,14 @@ void rd_kafka_metadata_cache_topic_update(
...
+                                   int valid);
 void rd_kafka_metadata_cache_purge_hints(rd_kafka_t *rk,
                                          const rd_list_t *topics);
+void rd_kafka_metadata_cache_purge_hints_by_id(rd_kafka_t *rk,
+                                               const rd_list_t *topics);
```
pranavrth, line 296:
```suggestion
                                               const rd_list_t *topic_ids);
```

src/rdkafka_metadata_cache.c
```diff
@@ -247,7 +287,7 @@ static struct rd_kafka_metadata_cache_entry *rd_kafka_metadata_cache_insert(
     rd_bool_t include_racks,
     rd_kafka_metadata_broker_internal_t *brokers_internal,
     size_t broker_cnt) {
-        struct rd_kafka_metadata_cache_entry *rkmce, *old;
+        struct rd_kafka_metadata_cache_entry *rkmce, *old, *old_by_id = NULL;
         rd_tmpabuf_t tbuf;
```
pranavrth, line 291: Shouldn't we allocate memory in `tbuf` for topic_id as well?

src/rdkafka_metadata_cache.c
```diff
@@ -350,8 +390,28 @@ static struct rd_kafka_metadata_cache_entry *rd_kafka_metadata_cache_insert(
...
+                RD_AVL_REMOVE_ELM(&rk->rk_metadata_cache.rkmc_avl_by_id, old);
+        }
+        if (old) {
+                /* Delete and free old cache entry */
                 rd_kafka_metadata_cache_delete(rk, old, 0);
+        }
+        if (old_by_id && old_by_id != old) {
+                /* If there was a different cache entry in this tree,
+                 * remove and free it. */
+                RD_AVL_REMOVE_ELM(&rk->rk_metadata_cache.rkmc_avl, old_by_id);
```
pranavrth, line 412: I think this is already removed while inserting.

src/rdkafka_metadata_cache.c
```diff
@@ -350,8 +390,28 @@ static struct rd_kafka_metadata_cache_entry *rd_kafka_metadata_cache_insert(
         /* Insert (and replace existing) entry. */
         old = RD_AVL_INSERT(&rk->rk_metadata_cache.rkmc_avl, rkmce,
                             rkmce_avlnode);
-        if (old)
+        /* Insert (and replace existing) entry into the AVL tree sorted
+         * by topic id. */
+        if (!RD_KAFKA_UUID_IS_ZERO(
```
pranavrth, line 395: We should check this whole logic again. This is tricky and might have some issues.

src/rdkafka_metadata_cache.c
```diff
@@ -427,24 +489,38 @@ void rd_kafka_metadata_cache_topic_update(
...
-                                               include_racks, brokers,
-                                               broker_cnt);
-        else
-                changed =
-                    rd_kafka_metadata_cache_delete_by_name(rk, mdt->topic);
+        if (!mdt->topic) {
+                /* Cache entry found but no topic name:
+                 * delete it. */
+                changed = rd_kafka_metadata_cache_delete_by_topic_id(
+                    rk, mdit->topic_id);
```
pranavrth, line 507: This case should be similar to the else case. Lets discuss this more.

pranavrth (changes requested):

src/rdkafka_request.c
```diff
@@ -2857,41 +2863,64 @@ rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
                                              rd_bool_t cgrp_update,
                                              rd_bool_t force_racks,
                                              rd_kafka_op_t *rko) {
-        return rd_kafka_MetadataRequest_op(
```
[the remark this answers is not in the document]
emasab: As discussed the MetadataRequest_op is coupled to the request that uses the default handler (rd_kafka_MetadataRequest) and only used there so it's not generic and while documenting the behaviour this appears more clearly. That's why I removed this intermediate function

src/rdkafka_metadata.c
emasab: As seen together, the str there is taken from `requested_topics` so it always available

src/rdkafka_metadata_cache.c
emasab: this is removing it from the topic name keyed tree

src/rdkafka_metadata_cache.c
emasab: In the else case, it's needed for the `rd_kafka_cgrp_metadata_update_check`, that we don't use with the new protocol

src/rdkafka_metadata_cache.c
emasab: It's allocated in `rd_tmpabuf_add_alloc(&tbuf, sizeof(*rkmce));` as it's an embedded struct

src/rdkafka_request.c
```diff
@@ -2532,31 +2532,63 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
...
 /**
- * @brief Internal implementation of MetadataRequest (does not send).
+ * @brief Internal implementation of MetadataRequest.
+ *
+ *        - !topics && !topic_ids: only request brokers (if supported by
+ *          broker, else all topics)
+ *        - topics.cnt == 0 || topic_ids.cnt == 0: all topics in cluster
+ *          are requested
+ *        - topics.cnt > 0 || topic_ids.cnt > 0: only specified topics
+ *          are requested
```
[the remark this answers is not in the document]
emasab: used a `rd_dassert` as this can happen only during development

src/rdkafka_metadata.c
pranavrth: Not `requested_topics` but `rkt`. So its fine.

src/rdkafka_request.c
```diff
@@ -2532,31 +2532,63 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
...
+ *         will return  RD_KAFKA_RESP_ERR__PREV_IN_PROGRESS,
+ *         otherwise RD_KAFKA_RESP_ERR_NO_ERROR.
  *
- * If full metadata for all topics is requested (or
- * all brokers, which results in all-topics on older brokers) and there is
- * already a full request in transit then this function will return
- * RD_KAFKA_RESP_ERR__PREV_IN_PROGRESS otherwise RD_KAFKA_RESP_ERR_NO_ERROR.
- * If \p rko is non-NULL or if \p force is true, the request is sent regardless.
+ * @remark Either \p topics or \p topic_ids must be set, but not both.
+ *         If \p rko is specified, \p resp_cb, \p replyq, \p force, \p opaque
```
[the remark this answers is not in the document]
emasab: used a `rd_dassert` as this can happen only during development `!(topics.cnt > 0 && topic_ids.cnt > 0)`

src/rdkafka_request.c
```diff
@@ -2532,31 +2532,63 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
...
+ *         all brokers, which results in all-topics on older brokers) and
+ *         there is already a full request in transit then this function
+ *         will return  RD_KAFKA_RESP_ERR__PREV_IN_PROGRESS,
+ *         otherwise RD_KAFKA_RESP_ERR_NO_ERROR.
  *
- * If full metadata for all topics is requested (or
- * all brokers, which results in all-topics on older brokers) and there is
- * already a full request in transit then this function will return
- * RD_KAFKA_RESP_ERR__PREV_IN_PROGRESS otherwise RD_KAFKA_RESP_ERR_NO_ERROR.
- * If \p rko is non-NULL or if \p force is true, the request is sent regardless.
```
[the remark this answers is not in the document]
emasab: It's already written in the documentation of params rko and force but I can add a remark too

src/rdkafka_metadata.c
emasab: We've checked all the paths and it wasn't causing problems, currently the topic name can be NULL only in parsed `rd_kafka_metadata_topic_internal_t` when the request is made by topic id, we have to make sure it's handled correctly and that it's not inserted in cache or anywhere else in that case, just that it's returned in the response of the Metadata call

pranavrth: I agree with this. If we are making sure that this is the case, we are good here. We might need to check more cases later when we implement KIP-516 more.

src/rdkafka_metadata_cache.c
```diff
@@ -427,24 +489,38 @@ void rd_kafka_metadata_cache_topic_update(
...
-        if (!mdt->err ||
-            mdt->err == RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED ||
-            mdt->err == RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART)
-                rd_kafka_metadata_cache_insert(rk, mdt, mdit, now, ts_expires,
-                                               include_racks, brokers,
-                                               broker_cnt);
-        else
-                changed =
-                    rd_kafka_metadata_cache_delete_by_name(rk, mdt->topic);
+        if (!mdt->topic) {
```
pranavrth, line 503:
```suggestion
        if (unlikely(!mdt->topic)) {
```

src/rdkafka_metadata_cache.c
```diff
@@ -427,24 +489,38 @@ void rd_kafka_metadata_cache_topic_update(
...
+
+                if (!mdt->err ||
+                    mdt->err == RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED ||
+                    mdt->err == RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART)
+                        rd_kafka_metadata_cache_insert(
+                            rk, mdt, mdit, now, ts_expires, include_racks,
+                            brokers, broker_cnt);
+                else
+                        changed = rd_kafka_metadata_cache_delete_by_name(
+                            rk, mdt->topic);
```
pranavrth, line 522: Shall we add `delete_by_topic_id` as well here? Just give it a thought if it is required.

pranavrth (commented):
Just few minor comments.

src/rdkafka_request.c
```diff
@@ -2857,41 +2868,64 @@ rd_kafka_resp_err_t rd_kafka_MetadataRequest(rd_kafka_broker_t *rkb,
...
+ *         there is already a full request in transit then this function
+ *         will return  RD_KAFKA_RESP_ERR__PREV_IN_PROGRESS,
+ *         otherwise RD_KAFKA_RESP_ERR_NO_ERROR.
+ *
+ *  @remark Either \p topics or \p topic_ids must be set, but not both.
  */
 rd_kafka_resp_err_t rd_kafka_MetadataRequest_resp_cb(
     rd_kafka_broker_t *rkb,
     const rd_list_t *topics,
+    const rd_list_t *topics_ids,
```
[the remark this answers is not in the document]
emasab: Let's keep it, it makes it generic and we'll need it later. This doesn't need to be coupled to the way the AdminClient calls it.

src/rdkafka_metadata_cache.c
emasab: No, because it finds by name and then calls `rd_kafka_metadata_cache_delete` that deletes from both

pranavrth (approved):
LGTM!.. Nice changes!

PR #4677, opened 2024-04-08
Fix to metadata cache expiration on full metadata refresh

emasab:
Metadata cache was cleared on full metadata
refresh, leading to unnecessary refreshes and
occasional `UNKNOWN_TOPIC_OR_PART` errors.
Solved by updating cache for existing or
hinted entries instead of clearing them.
Happening since 2.1.0

### DIFF
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -490,7 +490,7 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
         rd_kafka_resp_err_t err    = RD_KAFKA_RESP_ERR_NO_ERROR;
         int broker_changes         = 0;
         int cache_changes          = 0;
-        rd_ts_t ts_start           = rd_clock();
+
         /* If client rack is present, the metadata cache (topic or full) needs
          * to contain the partition to rack map. */
         rd_bool_t has_client_rack = rk->rk_conf.client_rack &&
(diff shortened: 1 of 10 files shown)


src/rdkafka_metadata_cache.c
```diff
@@ -481,23 +463,28 @@ void rd_kafka_metadata_cache_expiry_start(rd_kafka_t *rk) {
...
         struct rd_kafka_metadata_cache_entry *rkmce = NULL;
         rd_ts_t now                                 = rd_clock();
         rd_ts_t ts_expires = now + (rk->rk_conf.metadata_max_age_ms * 1000);
         int changed        = 1;
-        if (unlikely(!mdt->topic)) {
+        if (likely(mdt->topic && only_existing)) {
+                rkmce = rd_kafka_metadata_cache_find(rk, mdt->topic, 0);
+                if (!rkmce)
+                        return 0;
+        } else if (unlikely(!mdt->topic)) {
```
pranavrth, line 483: This is not checking `only_existing` part.

pranavrth (changes requested):

src/rdkafka_metadata.h
```diff
@@ -274,15 +274,16 @@ int rd_kafka_metadata_cache_delete_by_name(rd_kafka_t *rk, const char *topic);
...
+int rd_kafka_metadata_cache_topic_update(
     rd_kafka_t *rk,
     const rd_kafka_metadata_topic_t *mdt,
     const rd_kafka_metadata_topic_internal_t *mdit,
     rd_bool_t propagate,
     rd_bool_t include_metadata,
     rd_kafka_metadata_broker_internal_t *brokers,
-    size_t broker_cnt);
+    size_t broker_cnt,
+    rd_bool_t only_existing);
```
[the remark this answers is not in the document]
emasab: If you search for `only_` the code convention in this repo is to avoid repeating the action in variable names, for example in `rd_kafka_buf_write_topic_partitions` we have `only_invalid_offsets` not `write_only_invalid_offsets`

src/rdkafka_metadata_cache.c
```diff
@@ -182,45 +182,27 @@ static int rd_kafka_metadata_cache_evict(rd_kafka_t *rk) {
...
+ *        replaced hints with existing topic information, thus this will
+ *        only remove unmatched topics from the cache.
  *
- * @returns the number of entries evicted.
+ * @returns the number of purged hints
  *
  * @locks_required rd_kafka_wrlock()
  */
-int rd_kafka_metadata_cache_evict_by_age(rd_kafka_t *rk, rd_ts_t ts) {
+int rd_kafka_metadata_cache_purge_all_hints(rd_kafka_t *rk) {
```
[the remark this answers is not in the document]
emasab: We've seen that concurrency problems aren't possible as the cache is updated only by the main thread, that is currently executing this function. This is done only for full request case and in other cases only the requested topic hints are purged.

src/rdkafka_metadata.c
```diff
@@ -850,23 +850,24 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
+                        /* use has_client_rack rather than
+                        compute_racks. We need cached rack ids
+                        only in case we need to rejoin the group
+                        if they change and client.rack is set
+                        (KIP-881). */
+                        has_client_rack, mdi->brokers,
+                        md->broker_cnt,
+                        all_topics /*cache entry needs to exist
+                                    *if all_topics*/);
+                rd_kafka_wrunlock(rk);
```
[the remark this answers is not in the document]
pranavrth: As discussed, not needed right now. Will check the performance implications before doing so.

tests/cluster_testing.py
```diff
@@ -35,15 +35,6 @@ def read_scenario_conf(scenario):
         return parser.load(f)
 
 
-# FIXME: merge in trivup
-class KafkaBrokerApp(KafkaBrokerAppOrig):
```
pranavrth, line 39: This is expected right?

tests/0146-metadata_mock.c
```diff
@@ -0,0 +1,98 @@
...
+        rd_kafka_destroy(rk);
+        test_mock_cluster_destroy(mcluster);
+
+        SUB_TEST_PASS();
+}
+
+int main_0146_metadata_mock(int argc, char **argv) {
+        TEST_SKIP_MOCK_CLUSTER(0);
+
+        do_test_metadata_persists_in_cache("range");
```
pranavrth, line 93: why not round robin case?

pranavrth (approved):
LGTM!. Just a couple of minor comments.

tests/cluster_testing.py
emasab: Yeah, it was added back during a rebase

tests/0146-metadata_mock.c
emasab: The difference can be because of taking the eager path instead of the incremental one, but roundrobin also takes the eager path, as "range", so it's not important for this test.

PR #4680, opened 2024-04-08
Fix for an undesired partition migration with stale leader epoch

emasab:
A partition migration could happen,
using stale metadata, when the partition
was undergoing a validation and
being retried because of an error.
Solved by doing a partition migration
only with a non-stale leader epoch.
Happening since 2.1.0


milindl (approved):
LGTM, thanks for adding the test too

PR #4689, opened 2024-04-16
Handle overflow in rd_buf_write_remains

anchitj:
After the implementation of KIP-320, TopicCnt was sent as a varint in the metadata request. This change caused an overflow issue due to the use of `rd_buf_erase` to erase the remaining bytes. The overflow occurred in the calculation `rbuf->rbuf_size - (rbuf->rbuf_len + rbuf->rbuf_erased)` because `rbuf->rbuf_erased` was non-zero when the buffer was almost full, leading to an overflow condition.

As a result, for certain configurations that caused the buffer to be nearly full, the client was unable to send metadata requests, and it also caused crashes in the .NET client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# vNext
+
+* Fix segfault when using long client id because of erased segment when using flexver. (#4689)
+
 # librdkafka v2.4.0
 
 librdkafka v2.4.0 is a feature release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,17 +2,24 @@
 
 librdkafka v2.5.0 is a feature release.
 
+* Fix segfault when using long client id because of erased segment when using flexver. (#4689)
+
+
 ## Enhancements
 
   * Update bundled lz4 (used when `./configure --disable-lz4-ext`) to
     [v1.9.4](https://github.com/lz4/lz4/releases/tag/v1.9.4), which contains
     bugfixes and performance improvements (#4726).
 
+
 ## Fixes
 
 ### General fixes
 
-  * Fix segfault when using long client id because of erased segment when using flexver. (#4689)
+*  Issues: [confluentinc/confluent-kafka-dotnet#2084](https://github.com/confluentinc/confluent-kafka-dotnet/issues/2084)
+   Fix segfault when a segment is erased and more data is written to the buffer.
+   Happens since 1.x when a portion of the buffer (segment) is erased for flexver or compression.
+   More likely to happen since 2.1.0, because of the upgrades to flexver, with certain string sizes like a long client id (#4689).
 
 
 
--- a/src/rdbuf.h
+++ b/src/rdbuf.h
@@ -70,6 +70,8 @@ typedef struct rd_segment_s {
                                     *   beginning in the grand rd_buf_t */
         void (*seg_free)(void *p); /**< Optional free function for seg_p */
         int seg_flags;             /**< Segment flags */
+        size_t seg_erased;         /** Total number of bytes erased from
+                                    *   this segment. */
 #define RD_SEGMENT_F_RDONLY 0x1    /**< Read-only segment */
 #define RD_SEGMENT_F_FREE                                                      \
         0x2 /**< Free segment on destroy,                                      \
--- a/src/rdbuf.h
+++ b/src/rdbuf.h
@@ -147,11 +147,7 @@ static RD_INLINE RD_UNUSED size_t rd_buf_write_pos(const rd_buf_t *rbuf) {
  * @returns the number of bytes available for writing (before growing).
  */
 static RD_INLINE RD_UNUSED size_t rd_buf_write_remains(const rd_buf_t *rbuf) {
-        ssize_t remaining =
-            rbuf->rbuf_size - (rbuf->rbuf_len + rbuf->rbuf_erased);
-        if (remaining < 0)
-                return 0;
-        return remaining;
+        return rbuf->rbuf_size - (rbuf->rbuf_len + rbuf->rbuf_erased);
 }
 
 
--- a/src/rdbuf.h
+++ b/src/rdbuf.h
@@ -147,7 +147,11 @@ static RD_INLINE RD_UNUSED size_t rd_buf_write_pos(const rd_buf_t *rbuf) {
  * @returns the number of bytes available for writing (before growing).
  */
 static RD_INLINE RD_UNUSED size_t rd_buf_write_remains(const rd_buf_t *rbuf) {
-        return rbuf->rbuf_size - (rbuf->rbuf_len + rbuf->rbuf_erased);
+        ssize_t remaining =
+            rbuf->rbuf_size - (rbuf->rbuf_len + rbuf->rbuf_erased);
+        if (remaining < 0)
+                return 0;
+        return remaining;
 }
 
 
--- a/tests/0011-produce_batch.c
+++ b/tests/0011-produce_batch.c
@@ -90,20 +90,19 @@ static void test_single_partition(void) {
         int failcnt = 0;
         int i;
         rd_kafka_message_t *rkmessages;
-        const int client_id_length = 239;
-        char *client_id;
-
+        char client_id[271];
         SUB_TEST_QUICK();
 
         msgid_next = 0;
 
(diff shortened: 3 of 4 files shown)


korchak-aleksandr:
Hi @emasab,

Could you please review and approve this PR? It's critical for us as it addresses a blocking issue with the confluent-kafka-dotnet library update. Your approval is eagerly awaited.

tests/0011-produce_batch.c
```diff
@@ -90,13 +90,22 @@ static void test_single_partition(void) {
...
         msgid_next = 0;
 
         test_conf_init(&conf, &topic_conf, 20);
 
+        clientId = (char *)malloc(clientIdLength + 1 * sizeof(char));
+        for (i = 0; i < clientIdLength; i++) {
+                clientId[i] = 'c';
+        }
+        clientId[clientIdLength] = '\0';
+        rd_kafka_conf_set(conf, "client.id", clientId, NULL, 0);
```
emasab, line 107: Please use snake case for variables and add a parenthesis around `client_id_length + 1`. Cast isn't needed as `void *` can be assigned without cast.

Also add a comment that links to this issue.
```suggestion
        client_id = malloc((client_id_length + 1) * sizeof(char));
        for (i = 0; i < client_id_length; i++) {
                client_id[i] = 'c';
        }
        client_id[client_id_length] = '\0';
```

src/rdbuf.h
```diff
@@ -147,7 +147,11 @@ static RD_INLINE RD_UNUSED size_t rd_buf_write_pos(const rd_buf_t *rbuf) {
  * @returns the number of bytes available for writing (before growing).
  */
 static RD_INLINE RD_UNUSED size_t rd_buf_write_remains(const rd_buf_t *rbuf) {
-        return rbuf->rbuf_size - (rbuf->rbuf_len + rbuf->rbuf_erased);
+        ssize_t remaining =
```
emasab, line 150: This change can be reverted as this condition must never happen, the problem happens because a segment is removed without decreasing the corresponding part in `rbuf->rbuf_erased` , in `rd_buf_erase` it should be:
```c
                /* If segment is now empty, remove it */
                if (seg->seg_of == 0) {
                        rd_buf_destroy_segment(rbuf, seg);
                        rbuf->rbuf_erased -= toerase;
                }
``` 
The issue can happen with RPCs using flexver as happens with MetadataRequest since version 9.

tests/0011-produce_batch.c
```diff
@@ -106,7 +115,14 @@ static void test_single_partition(void) {
         TEST_SAY("test_single_partition: Created kafka instance %s\n",
                  rd_kafka_name(rk));
 
-        rkt = rd_kafka_topic_new(rk, test_mk_topic_name("0011", 0), topic_conf);
+        topicSuffix = (char *)malloc(topicSuffixLength + 1 * sizeof(char));
```
emasab, line 118: This change can be removed as it's not necessary for reproducing the issue

tests/0011-produce_batch.c
```diff
@@ -178,6 +194,9 @@ static void test_single_partition(void) {
         TEST_SAY("Destroying kafka instance %s\n", rd_kafka_name(rk));
         rd_kafka_destroy(rk);
 
+        free(clientId);
+        free(topicSuffix);
```
emasab, line 198: Remove `topicSuffix`

emasab (changes requested):
Thanks Anchit, the source of the problem comes from a previous bug, check the comment.

emasab:
There also a problem in `rd_buf_write_seek` because of `rd_buf_destroy_segment`, checking how to solve it

tests/0011-produce_batch.c
anchitj: We need `topicSuffix` because with the combination of both `clientID` and `topicSuffix` we were able to see this border condition in the segment.

tests/0011-produce_batch.c
anchitj: Updated

src/rdbuf.h
anchitj: Updated

anchitj:
> There also a problem in `rd_buf_write_seek` because of `rd_buf_destroy_segment`.
> 
> My proposed solution is to add a `seg_erased` to `rd_segment_s` and increase it when part of the segment is erased, then subtract `seg_erased` from `rbuf_erased` when a segment is removed in `rd_buf_write_seek`

Makes sense. I'll create a separate PR to handle that.

tests/0011-produce_batch.c
emasab: I had verified it happens with just that client id, because the segment is removed when the varint is changed without reducing the erased bytes

emasab:
@anchitj better to do it here, as it's basically the same thing that is missing

tests/0011-produce_batch.c
anchitj: Removed suffix

tests/0011-produce_batch.c
anchitj: Removed

tests/0011-produce_batch.c
```diff
@@ -90,8 +90,8 @@ static void test_single_partition(void) {
         int failcnt = 0;
         int i;
         rd_kafka_message_t *rkmessages;
-        const int topic_suffix_length = 56, client_id_length = 239;
-        char *topic_suffix, *client_id;
+        const int client_id_length = 239;
+        char *client_id;
```
emasab, line 94: Can simplify it with a memset, see below. In C I only saw it happening from 270 on.
```c
        char client_id[271];
```

tests/0011-produce_batch.c
```diff
@@ -90,13 +90,22 @@ static void test_single_partition(void) {
...
 
         msgid_next = 0;
 
         test_conf_init(&conf, &topic_conf, 20);
 
+        client_id = malloc((client_id_length + 1) * sizeof(char));
+        for (i = 0; i < client_id_length; i++) {
+                client_id[i] = 'c';
+        }
+        client_id[client_id_length] = '\0';
```
emasab, line 106:
```suggestion
        /* A long client id must not cause a segmentation fault
         * because of an erased segment when using flexver.
         * See: https://github.com/confluentinc/confluent-kafka-dotnet/issues/2084 */
        memset(client_id, 'c', sizeof(client_id) - 1);
        client_id[sizeof(client_id) - 1] = '\0';
```

tests/0011-produce_batch.c
```diff
@@ -178,6 +187,8 @@ static void test_single_partition(void) {
         TEST_SAY("Destroying kafka instance %s\n", rd_kafka_name(rk));
         rd_kafka_destroy(rk);
 
+        free(client_id);
+
```
emasab, line 191: Remove this

emasab (changes requested):
Thanks Anchit, some changes are needed

tests/0011-produce_batch.c
emasab: Please add the CHANGELOG entry too

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ librdkafka v2.5.0 is a feature release.
     [v1.9.4](https://github.com/lz4/lz4/releases/tag/v1.9.4), which contains
     bugfixes and performance improvements (#4726).
 
+## Fixes
+
+### General fixes
+
+  * Fix segfault when using long client id because of erased segment when using flexver. (#4689)
```
emasab, line 15: You can keep a long description here with the `Issues` and `Happens since` sections and put a one line description in the summary thar goes below `librdkafka v2.5.0 is a feature release`.

```suggestion
  *  Issues: confluentinc/confluent-kafka-dotnet#2084
     Fix segfault when a segment is erased and more data is written to the buffer.
     Happens since 1.x when a portion of the buffer (segment) is erased for flexver or compression.
     More likely to happen since 2.1.1, because of the upgrades to flexver, with certain string sizes like a long client id (#4689).
```

anchitj: Thanks. Updated

emasab (approved):
Thanks Anchit! Let's wait for the CI and then we can merge this.

nkostoulas:
Could this affect the confluent-kafka-go `AdminClient`?

Issue #4653, opened 2024-03-19
zlib library security vulnerability through to version 1.3


MiikaL:
Description
===========
We use the Confluent.Kafka nuget which makes use of librdkafka, and we are receiving a security warning about the version of zlib in use: 

One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '8.0': 
zlib1.dll: CVE-2023-45853(9.8), CVE-2002-0059(9.8), CVE-2022-37434(9.8)

https://nvd.nist.gov/vuln/detail/CVE-2023-45853

Checklist
=========

**IMPORTANT**: We will close issues where the checklist has not been completed.

Please provide the following information:

 - [x] librdkafka version (release number or git tag): `2.3.1`
 - [x] Apache Kafka version: `N/A`
 - [x] librdkafka client configuration: `N/A`
 - [x] Operating system: `windows`
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka
 - [x] Provide broker log excerpts
 - [x] Critical issue

janjwerner-confluent:
Resolved in https://github.com/confluentinc/librdkafka/pull/4706

MiikaL:
Not resolved, this file (https://github.com/confluentinc/librdkafka/blob/v2.5.0/vcpkg.json) packages 1.3 version of zlib for the windows package.

MiikaL:
I have continued the issue in https://github.com/confluentinc/librdkafka/issues/4813

Issue #4664, opened 2024-03-27
Critical Vulnerabilities identified in librdkafka 


glansbury:
Description
===========
Two CVE's with cvss score of 9.8 identified in this library, please help update.

curl 7.86.0
https://github.com/advisories/GHSA-75qm-2q4j-qx6g
https://github.com/confluentinc/librdkafka/blob/master/mklove/modules/configure.libcurl#L48

zlib 1.2.13
https://github.com/advisories/GHSA-mq29-j5xf-cjwr
https://github.com/confluentinc/librdkafka/blob/master/mklove/modules/configure.zlib#L45

How to reproduce
================
Review source code and links provided.  Use any SBOM vulnerability scanner to validate that the libraries are being linked into build.

Initially I discovered this in confluent-kafka-go, however, I believe the vulnerability is coming from the C base library librdkafka

Checklist
=========

 - [x] librdkafka version: `v2.3.0`
 - [x] librdkafka client configuration: `N/A`
 - [x] Operating system: `linux (any base distro)`
 - [x] Provide logs - N/A. links to source code and CVSS all that is required
 - [x] Provide broker log excerpts : N/A
 - [ ] Critical issue: 9.8 CVSS vulnerabilities.

mikajylha:
Related issue #4653

mikajylha:
[Another related issue](https://github.com/confluentinc/confluent-kafka-dotnet/issues/2203) in dotnet lib

janjwerner-confluent:
Thank you for the report. We are in the process of resolving this issue.

janjwerner-confluent:
Resolved in https://github.com/confluentinc/librdkafka/pull/4706

PR #3148, opened 2020-11-16
Import lz4 v1.9.3

edenhill:
This is a clean import, you dont need to review the code itself.

PR #4690, opened 2024-04-16
NONJAVACLI-3460: update third party dependencies

janjwerner-confluent:


src/lz4.h
```diff
@@ -100,7 +100,7 @@ extern "C" {
 /*------   Version   ------*/
 #define LZ4_VERSION_MAJOR    1    /* for breaking interface changes  */
 #define LZ4_VERSION_MINOR    9    /* for new (non-breaking) interface capabilities */
-#define LZ4_VERSION_RELEASE  3    /* for tweaks, bug-fixes, or development */
+#define LZ4_VERSION_RELEASE  4    /* for tweaks, bug-fixes, or development */
```
milindl, line 103: upgrading lz4 is trickier, as we have actually copied the entire code into our source tree. An update will look similar to https://github.com/confluentinc/librdkafka/pull/3148/files

milindl (commented):
Thanks a lot for the PR!

The windows versions also require an update. That can be done for zlib, curl, and zstd with updation of vcpkg.json.

src/lz4.h
janjwerner-confluent: I was under the impression that most of the update has already happened, it was just the version update missing:
https://github.com/confluentinc/librdkafka/pull/4232

janjwerner-confluent: oh, it was just the CVE fix, 
maybe it would be worthwhile to update lz4  as there are some significant performance gains for ARM64 platform:
https://github.com/lz4/lz4/releases

mklove/modules/configure.libcurl
```diff
@@ -45,8 +45,8 @@ void foo (void) {
 function install_source {
     local name=$1
     local destdir=$2
-    local ver=7.86.0
-    local checksum="3dfdd39ba95e18847965cd3051ea6d22586609d9011d91df7bc5521288987a82"
+    local ver=8.7.1
```
milindl, line 48: Adding a note for the team here that curl 8 has no API or ABI changes

janjwerner-confluent: @milindl 
Thank you. Can you please see this commit through completion - this is as far as I can go with it.

milindl:
Running CI, and then I'll run try running OIDC tests locally to test out the libcurl update.

janjwerner-confluent:
@milindl 
Can we bump openssl while we're here or should I create another ticket /pr ?

milindl:
I can bump it here, we're changing things anyway. However, openssl bumps are partial, linux only, because the latest versions of openssl 3.0.x are not available on vcpkg.

milindl:
Closing and reopening this to try and make sem-approve work.

milindl:
In the latest changes (which include openssl bump):
1. I compiled with `./configure --install-deps --source-deps-only --disable-lz4-ext --enable-static --enable-strip --disable-gssapi`
2. Ran all integration tests with interactive-broker-version 3.7.0
3. Ran the OIDC integration test (the test seems to be failing for an unrelated reason, but the curl request/response is working correctly, I added some log statements to check the response. The test fails after that, there's something wrong in the broker that trivup has set up).

janjwerner-confluent:
resolves:
https://github.com/confluentinc/librdkafka/issues/4653
https://github.com/confluentinc/librdkafka/issues/4664

janjwerner-confluent:
replacing the branch name to dev_janjwerner_deps_update per @milindl suggestion

PR #4691, opened 2024-04-17
[KIP-848] Feature branch major comment addressal

pranavrth:
- Fix reading error code in CGH_leave due to incorrect parsing of the response.


emasab (approved):
Thanks, have tested it locally before running the tests with the new CGRP that are in the other branch

PR #4700, opened 2024-04-26
[KIP-848] Tests for: Fix for a rapid unsubscribe while the member id is still not assigned

emasab:


tests/0147-consumer_group_consumer_mock.c
```diff
@@ -858,6 +858,79 @@ static void do_test_adherence_to_hb_interval(void) {
...
+        c = create_consumer(bootstraps, topic, rd_true);
+
+        subscription = rd_kafka_topic_partition_list_new(1);
+        rd_kafka_topic_partition_list_add(subscription, topic,
+                                          RD_KAFKA_PARTITION_UA);
+
+        for (i = 0; i < 2; i++) {
+                TEST_CALL_ERR__(rd_kafka_subscribe(c, subscription));
+                TEST_CALL_ERR__(rd_kafka_unsubscribe(c));
+        }
```
pranavrth, line 909: I think we should be adding a real broker scenario where the assignments are the expected ones.

emasab: That is included in this test:
https://github.com/confluentinc/librdkafka/pull/5026

this is for the segfault that happened locally and for checking the case where the coordinator is down.

pranavrth (approved):
LGTM!.

Issue #4362, opened 2023-07-21
Consumer hangs if closing together with deleting topic


ilejn:
Description
===========
Consumer hangs if closing together with deleting topic

How to reproduce
================
Hello, 
in ClickHouse we have an issue with an integration test if librdkafka master is more recent than 8e20e1ee79b188, IOW if librdkafka contains PR 4117.
Test scenario:
    1. Six consumers consume messages from a topic with six partitions.
    2. Delete the topic (not via librdkafka)
    3. Close the consumers one by one
One of the consumers is more or less reproducibly hangs during closing, while virtually anything helps – it is enough to add a sleep() between (2) and (3) or even try to use a ClickHouse build with a sanitizer.
I tried to create MRU not using ClickHouse, but did not succeeded.

The scenario seems a bit insane, though it is crucial for us and effectively prevents us from using recent librdkafka.
How ClickHouse closes a consumer.
    • unsubscribe
    • drain queue
    • free callbacks
    • call rdkafka_consumer_close

ClickHouse maintains rebalance callback (actually cppkafka does).

My investigations.
Problematic part of  PR 4117 is _rd_kafka_toppar_keep(rktp)_
Specifically where it is called from _rd_kafka_toppar_pause_resume_ to do resume.
In _rd_kafka_broker_thread_main_ we are waiting forever _while (!rd_kafka_broker_terminating(rkb))_ which is actually _rd_refcnt_get(&(rkb)->rkb_refcnt) <= 1_ . 
REFCNT DEBUG output https://pastila.nl/?00659d03/ee47523355fd8a694171a23c8b2a48c6
Some ClickHouse logs https://pastila.nl/?002bbb54/247b8ebbb941432451f7ae5ce10f319b

Am I right thinking that the problem is there is no suitable counterpart to read RD_KAFKA_OP_BARRIER from fetch queue?
Is it possible to resolve this problem at application side?

Checklist
=========

 - [x] librdkafka version    8e20e1ee79b188
 - [x] Apache Kafka version:`image: confluentinc/cp-kafka:5.2.0`
 - [x] librdkafka client configuration: https://pastila.nl/?07076592/a59038db454e37bda6f455f41a9e4020)](https://pastila.nl/?07076592/a59038db454e37bda6f455f41a9e4020)
 - [x] Operating system: `ubuntu:22.04`
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
[...]

PR #4724, opened 2024-05-21
Fix to remove fetch queue messages that blocked the destroy of rdkafka instances

emasab:
Circular dependencies from a partition fetch queue message to  the same partition blocked the destroy of an instance, that happened in case the partition was removed from the cluster while it was being consumed. Solved by purging internal partition queue, after being stopped and removed, to allow reference count to reach zero and trigger a destroy

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3496,6 +3496,8 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                                 : (topic_err
                                        ? topic_err
                                        : RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION));
+
+                        rd_kafka_toppar_purge_internal_fetch_queue_maybe(rktp);
                 }
 
                 rd_kafka_toppar_unlock(rktp);
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3497,7 +3497,10 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                                        ? topic_err
                                        : RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION));
 
-                        rd_kafka_toppar_purge_internal_fetch_queue_maybe(rktp);
+                        if (rkb->rkb_rk->rk_type == RD_KAFKA_CONSUMER) {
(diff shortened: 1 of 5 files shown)


emasab:
How to reproduce: happening sporadically with test `0113` subtest `n_wildcard`. Run it it with `TEST_DEBUG=all` and `until-fail.sh` to see the refcnt not reaching zero.

zuellig:
FWIW I've confirmed that this branch also fixes a non-constant but somewhat frequent issue I've been observing. I reproduced it by stressing the client node's swap while also stressing the broker node's cpu. It took a few restart cycles but within several hours the deadlocked destroy call occurred.

antaljanosbenjamin:
Is there any expected date on this to be merged?

antaljanosbenjamin:
> All Contributor License Agreements have been signed. Ready to merge.

One step closer! :tada:

src/rdkafka_broker.c
```diff
@@ -3407,6 +3407,8 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                                 : (topic_err
                                        ? topic_err
                                        : RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION));
+
+                        rd_kafka_toppar_purge_internal_fetch_queue_maybe(rktp);
```
pranavrth, line 3411: Shall we do this only for the Consumer?

emasab: Yes, better to avoid it for the producer given `rktp_fetchq` is allocated there too but the queue isn't forwarded we avoid looping through this empty queue.

emasab:
Addressed comment, updated CHANGELOG and rebased

ydsun90:
Hi, is there an estimated time to have this fix in a release? I have got the same issue when closing kafka consumer https://github.com/confluentinc/librdkafka/discussions/4885.

pranavrth (approved):
LGTM!.

ydsun90:
> Hi, is there an estimated time to have this fix in a release? I have got the same issue when closing kafka consumer #4885.

Any change to have the fix into a release?

ydsun90:
> > Hi, is there an estimated time to have this fix in a release? I have got the same issue when closing kafka consumer #4885.
> 
> Any plan to release the fix in a new version?

@pranavrth @emasab Hi, is there any expected time to have this fix in a new release?

filimonov:
Can it be related to https://github.com/confluentinc/librdkafka/issues/4362 ?

PR #4725, opened 2024-05-21
Option to enable/disable openssl engines at build time

Cropi:
Add configuration option to disable building with SSL engines. Some distros (e.g. Fedora) are planning to build openssl without engine support.


Cropi:
Hello,
would it be possible to move this forward?
Many distros are removing openssl engines support from their openssl package.

mhdawson:
@Cropi assuming that -  https://github.com/confluentinc/librdkafka/pull/3535 would also do what's needed on latest Fedora

PR #4726, opened 2024-05-22
Add files for lz4 1.9.4

milindl:
Updates lz4 from 1.9.3 -> 1.9.4

Process: 
* clean copy of all the lz4* files from 1.9.4 checkout of lz4 into the librdkafka src directory
* copy of xxhash files into rdxxhash files in the same way
* change any refs to xxhash within lz4* files into rdxxhash
* Update changelog and licenses, and configure file
* Run 0017 (compression test)

### DIFF
--- a/src/lz4.c
+++ b/src/lz4.c
@@ -124,6 +124,7 @@
 #if defined(_MSC_VER) && (_MSC_VER >= 1400)  /* Visual Studio 2005+ */
 #  include <intrin.h>               /* only present in VS2005+ */
 #  pragma warning(disable : 4127)   /* disable: C4127: conditional expression is constant */
+#  pragma warning(disable : 6237)   /* disable: C6237: conditional expression is always 0 */
 #endif  /* _MSC_VER */
 
 #ifndef LZ4_FORCE_INLINE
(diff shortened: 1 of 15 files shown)


src/lz4.c
```diff
@@ -187,7 +188,27 @@
...
+ *            LZ4_createStreamDecode, LZ4_freeStreamDecode, LZ4_create (deprecated)
+ *  - lz4hc : LZ4_createStreamHC, LZ4_freeStreamHC,
+ *            LZ4_createHC (deprecated), LZ4_freeHC  (deprecated)
+ *  - lz4frame, lz4file : All LZ4F_* functions
+ */
+#if defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
+#  define ALLOC(s)          lz4_error_memory_allocation_is_disabled
+#  define ALLOC_AND_ZERO(s) lz4_error_memory_allocation_is_disabled
+#  define FREEMEM(p)        lz4_error_memory_allocation_is_disabled
+#elif defined(LZ4_USER_MEMORY_FUNCTIONS)
```
emasab, line 211: Previous version 1.9.3 had this part of the code changed to use `rd_kafka_mem_malloc` ,.... Let's avoid changing the original code so we can upgrade it more easlily but we can define these functions `LZ4_malloc` and `LZ4_USER_MEMORY_FUNCTIONS` to use the same allocators.

milindl: Addressed this and added upgrade instructions for future developers.

pranavrth (approved):
LGTM.!

PR #4775, opened 2024-07-01
Remove CentOS 6 and CentOS 7 binaries

emasab:
keeps using manylinux_2_28 based on AlmaLinux 8 (CentOS 8)


pranavrth (changes requested):
There are few more places where we are still using centos 6/7. Please remove them as well. Following I could find.

https://github.com/confluentinc/librdkafka/blob/dev_remove_centos6_and_centos7/packaging/cp/verify-packages.sh#L20

https://github.com/confluentinc/librdkafka/blob/dev_remove_centos6_and_centos7/packaging/rpm/mock-on-docker.sh#L14

https://github.com/confluentinc/librdkafka/blob/dev_remove_centos6_and_centos7/packaging/rpm/tests/run-test.sh#L19-L28

https://github.com/confluentinc/librdkafka/blob/dev_remove_centos6_and_centos7/packaging/rpm/tests/test-on-docker.sh#L17

pranavrth (approved):
LGTM!

PR #4794, opened 2024-07-25
Add forward definition to fix compilation without ssl

emasab:


packaging/tools/build-configurations-checks.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+./configure --install-deps --disable-ssl --disable-gssapi \
+--disable-curl --disable-zlib \
+--disable-zstd --disable-lz4-ext --disable-regex-ext \
+--disable-c11threads --disable-syslog
+make -j
```
pranavrth, line 7: Please explain more on what are we trying to do here in comment.

pranavrth (approved):
LGTM!.

Issue #4649, opened 2024-03-15
Reading committed offsets where metadata contains null byte leads to read random data after null byte.


mlowicki:
Read the FAQ first: https://github.com/confluentinc/librdkafka/wiki/FAQ

Do NOT create issues for questions, use the discussion forum: https://github.com/confluentinc/librdkafka/discussions

Description
===========
I was reproducing with code in Rust using `rust-rdkafka` but that wrapper doesn't do much while reading committed offsets besides calling librdkafka itself. Because in `rust-rdkafka` it checks that metadata is a valid UTF-8 string it panics with errors like:

```
Metadata is not UTF-8: Utf8Error { valid_up_to: 3, error_len: Some(1) }
```

Verified also by implementing `OffsetFetch` and `OffsetCommit` in Rust that it's not an issue on the Kafka side - with pure Rust impl I couldn't reproduce issue with reading invalid data.

How to reproduce
================
Use byte array `[10, 20, 0, 30, 40]` as the commit metadata and commit for any partition. Then read committed offsets via `rd_kafka_committed` and in some cases metadata after \0 is just different than what was written.

Examples from other tests I've conducted where for the same metadata written we get random responses:

```
  4 |   0 |  66 |  32 |  64 |  32 |   2 |  16 |  82 | 108 |  25 |  74 | 120 |  24 |  52 |  20 |  58 |  28 |  76 |  22 |  51 |  25 |  82 |  99 |  47 |  91 |  12 |  22 | 115 |  20 | 116 | 100 |  50 |  89 |  76 |  23 |  43 |  49 | 104 |  34 |   0 |   0 | 

  4 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |  90 |  84 |  85 |  77 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 | 160 |  32 |   0 |   0 |   0 |   0 |   0 |   0 |  90 |  84 |  85 |  77 |   0 |   0 | 
```

```
[...]

PR #4795, opened 2024-07-25
Fix reading metadata with zeros

emasab:
Closes: #4649.
When retrieving offset metadata, if the binary value contained zeros
and librdkafka was configured with `strndup`, part of
the buffer after first zero contained uninitialized data
instead of rest of metadata. Solved by avoiding to use
`strndup` for copying metadata.
Happens since: 0.9.0


milindl (approved):
Looks good to me given the background. Please fix changelog

emasab:
Closing and reopening from an internal branch #4876

PR #4803, opened 2024-08-05
Adding int identifier for transactional ID

JohnPreston:
Adding the int ID for transactional ID, to enable using that as a ResourceType in the clients (python being my use-case specifically)

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -7719,11 +7719,12 @@ rd_kafka_ConfigEntry_synonyms(const rd_kafka_ConfigEntry_t *entry,
  * @brief Apache Kafka resource types
  */
 typedef enum rd_kafka_ResourceType_t {
-        RD_KAFKA_RESOURCE_UNKNOWN = 0, /**< Unknown */
-        RD_KAFKA_RESOURCE_ANY     = 1, /**< Any (used for lookups) */
-        RD_KAFKA_RESOURCE_TOPIC   = 2, /**< Topic */
-        RD_KAFKA_RESOURCE_GROUP   = 3, /**< Group */
-        RD_KAFKA_RESOURCE_BROKER  = 4, /**< Broker */
+        RD_KAFKA_RESOURCE_UNKNOWN           = 0, /**< Unknown */
+        RD_KAFKA_RESOURCE_ANY               = 1, /**< Any (used for lookups) */
+        RD_KAFKA_RESOURCE_TOPIC             = 2, /**< Topic */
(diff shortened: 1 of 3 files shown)


src/rdkafka.h
```diff
@@ -7719,11 +7719,12 @@ rd_kafka_ConfigEntry_synonyms(const rd_kafka_ConfigEntry_t *entry,
...
-        RD_KAFKA_RESOURCE_ANY     = 1, /**< Any (used for lookups) */
-        RD_KAFKA_RESOURCE_TOPIC   = 2, /**< Topic */
-        RD_KAFKA_RESOURCE_GROUP   = 3, /**< Group */
-        RD_KAFKA_RESOURCE_BROKER  = 4, /**< Broker */
+        RD_KAFKA_RESOURCE_UNKNOWN           = 0, /**< Unknown */
+        RD_KAFKA_RESOURCE_ANY               = 1, /**< Any (used for lookups) */
+        RD_KAFKA_RESOURCE_TOPIC             = 2, /**< Topic */
+        RD_KAFKA_RESOURCE_GROUP             = 3, /**< Group */
+        RD_KAFKA_RESOURCE_BROKER            = 4, /**< Broker */
+        RD_KAFKA_RESOURCE_TRANSACTIONAL_ID  = 5, /**< Transactional ID */
```
pranavrth, line 7727: Add resource name [here](https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_admin.c#L2827-L2836)

pranavrth (commented):
Style check and tests are failing. Please fix those.

To fix style check, run `make style-check-changed` for the changed code. Note that the change code is uncommitted code.

src/rdkafka.h
JohnPreston: I was trying to do that now, but I live in this decade and the earliest version of clang-formatter is 17 :D

Going to have to pull the 10 version from the archives although I have seen https://github.com/confluentinc/librdkafka/pull/4807 which might be a better candidate to merge so that we don't do that dance twice?

JohnPreston: Well, got the binaries for clang-format in version 10 so, updated it.

JohnPreston:
I mean, the style check I ran the command to fix that locally so I don't know what else to do here :/
I am not sure why the tests pass on some architecture but not with others. Given that the validation test is to check the value as per a statically expected set of returns, I am not sure how that comes to be.

Any suggestions?

pranavrth:
Closing this PR.
Continued in https://github.com/confluentinc/librdkafka/pull/4856

PR #4806, opened 2024-08-08
Fallback to fetch v12 (#10)

anchitj:
If any toppars have a zero topic id

### How to Reproduce
Set inter.broker.protocol.version to 2.4 with 3.7 and fetch would fail always.

### DIFF
--- a/src/rdkafka_fetcher.c
+++ b/src/rdkafka_fetcher.c
@@ -944,10 +944,10 @@ static void rd_kafka_broker_fetch_reply(rd_kafka_t *rk,
 }
 
 /**
- * @brief Check if any toppars have a non-zero topic id.
+ * @brief Check if any toppars have a zero topic id.
  *
  */
-rd_bool_t can_use_topic_id(rd_kafka_broker_t *rkb) {
+static rd_bool_t can_use_topic_ids(rd_kafka_broker_t *rkb) {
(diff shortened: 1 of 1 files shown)


src/rdkafka_fetcher.c
```diff
@@ -943,7 +943,21 @@ static void rd_kafka_broker_fetch_reply(rd_kafka_t *rk,
         }
 }
 
+/**
+ * @brief Check if any toppars have a non-zero topic id.
```
anchitj, line 947: Will change to `zero topic id`

src/rdkafka_fetcher.c
```diff
@@ -979,7 +993,12 @@ int rd_kafka_broker_fetch_toppars(rd_kafka_broker_t *rkb, rd_ts_t now) {
 
         ApiVersion = rd_kafka_broker_ApiVersion_supported(rkb, RD_KAFKAP_Fetch,
                                                           0, 16, NULL);
-        rkbuf      = rd_kafka_buf_new_flexver_request(
+
+        /* Fallback to version 12 if topic id is null which can happen if
+         * inter.broker.protocol.version is old */
```
anchitj, line 998: Change old to <2.8

emasab (changes requested):
Thanks for the fix! Let's make your changes and my comment and rebase it, then I'm approving

emasab (approved):
LGTM Thanks Anchit

PR #4809, opened 2024-08-09
Github Issue 4142 Patch

mahajanadhitya:


emasab:
This isn't the expected fix and it introduces a regression for example test 0004 leaks memory with this commit.

emasab (changes requested):
This isn't the expected fix and it introduces a regression for example test 0004 leaks memory with this commit. Check issue comments.

PR #4817, opened 2024-08-16
Add multiple kerberos authentication environment adaptation

d6blju:
Added a new configuration option sasl.kerberos.krb5path to adapt to the problem of using multiple krb5 files and different cache files when multiple different kerberos authentication environments are used

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -4884,7 +4884,11 @@ rd_kafka_broker_t *rd_kafka_broker_add(rd_kafka_t *rk,
         rkb->rkb_proto    = proto;
         rkb->rkb_port     = port;
         rkb->rkb_origname = rd_strdup(name);
-
+        if (rk->rk_conf.sasl.krb5_path) {
+                rkb->krb5_path  = rd_strdup(rk->rk_conf.sasl.krb5_path);
+        } else {
+                rkb->krb5_path  = NULL;
+        }
         mtx_init(&rkb->rkb_lock, mtx_plain);
         mtx_init(&rkb->rkb_logname_lock, mtx_plain);
         rkb->rkb_logname = rd_strdup(rkb->rkb_name);
--- a/src/rdkafka_broker.h
+++ b/src/rdkafka_broker.h
@@ -354,6 +354,7 @@ struct rd_kafka_broker_s { /* rd_kafka_broker_t */
 
 
         rd_kafka_timer_t rkb_sasl_reauth_tmr;
+        char *krb5_path;
 };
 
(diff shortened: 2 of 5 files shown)


EunhaTaker:
I noticed that the main purpose of your commit is to set the environment variable KRB5_CONFIG. I tried modifying the system environment variable and restarting librdkafka, but it didn't have any effect. It still uses /etc/krb5.conf. My system is Linux. May I ask if your system is also Linux?

d6blju:
@EunhaTaker Yes, this change does not work on Windows. How do you change the environment variables? Like in the commit?

EunhaTaker:
@d6blju I executed 'export KRB5_CONFIG=xxx', and then restarted the program in the same terminal.

d6blju:
@EunhaTaker Is your process running as a service? If so, you need to modify the service file instead of directly executing 'export KRB5_CONFIG=xxx' in the terminal.

EunhaTaker:
@d6blju Thanks, I'll try it after the holiday.

EunhaTaker:
@d6blju Realms are still being Interleaved, possibly due to incompleted environment variable configurations. Not a single message can be sent now. 
And machine translation was applied, which may contain technical terminology inaccuracies.

d6blju:
@EunhaTaker You didn't show how you modified it, could you try this commit directly?

EunhaTaker:
@d6blju Forget to add, I fully adopted your commit and assigned different sasl.kerberos.krb5.path values for different brokers. It works fine in my test environment, but fails in my client's environment. Apologies for forgetting to make screenshots.

d6blju:
@EunhaTaker Generally, this is related to krb5.conf. Please note the following configuration:
dns_lookup_realm = false #Disable querying domains through dns
dns_lookup_kdc = false #Disable querying kdc servers through dns
rdns=false #When there are multiple kerberos environments on a machine, such as 10.3.101.2, the kafka node address on the interface must use the host name and set this option to false
default_ccache_name = FILE:/tmp/krb5_1 #Ticket cache path. If there are multiple kerberos, the verification cache paths of different kerberos must be different, otherwise an error will be reported
dns_lookup_realm, dns_lookup_kdc When configuring the domain name and IP correspondence through the hosts file, if there is no intranet DNS server, please set it to false.
rdns is generally not common and is used when we test the intranet.
default_ccache_name is generally his problem. Do not repeat the cache file path.
Maybe I should upload these instructions to a git repository?

EunhaTaker:
@d6blju Thank you so much for the detailed configuration you provided. However, I haven’t had the chance to work on the client's equipment recently. I’ll definitely give it a try when I get the opportunity.

PR #4826, opened 2024-08-23
Fix an assert being triggered when no metrics matched on the client side during send push telemetry call

pranavrth:


emasab:
Closes https://github.com/confluentinc/librdkafka/issues/4833

tests/0150-telemetry_mock.c
```diff
@@ -202,11 +202,11 @@ void do_test_telemetry_get_subscription_push_telemetry(void) {
  *        resent after the push interval until there are subscriptions.
  *        See `requests_expected` for detailed expected flow.
  */
-void do_test_telemetry_empty_subscriptions_list(void) {
+void do_test_telemetry_empty_subscriptions_list(char *subscription_regex) {
```
emasab, line 205: Maybe add this parameter to the `SUB_TEST();` so we can tell which run it is.

emasab, line 205: Maybe add this parameter to the `SUB_TEST();` so we can tell which run it is.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# librdkafka v2.5.1
```
emasab, line 1: Change patch version because of the .NET 2.5.1 and 2.5.2 releases
```suggestion
# librdkafka v2.5.3
```

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+
+librdkafka v2.5.1 is a feature release.
+
+* Fix an assert being triggered when no metrics matched on the client side during send push telemetry call.
```
emasab, line 5: Add the `Fixes` section with references to this PR, the issue and that happens in 2.5.0, follow other fixes template.
Add a warning in 2.5.0 release saying it's not supported anymore and to upgrade to the latest patch release, referencing the issue.

emasab (changes requested):

CHANGELOG.md
pranavrth: Already did in the airlock PR. Please check.

CHANGELOG.md
pranavrth: Already did in the airlock PR. Please check.

emasab (approved):
LGTM Thanks

anchitj (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+Fix a regression introduced with [KIP-714](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability) support in which an assert is triggered during **PushTelemetry** call. This happens when no metric is matched on the client side among those requested by broker subscription.
+Happening since 2.5.0 (#4826).
+
+*Note: there were no v2.5.1 and v2.5.2 librdkafka releases*
+
+
 # librdkafka v2.5.0
 
+> [!WARNING]
+This version has introduced a regression in which an assert is triggered during **PushTelemetry** call. This happens when no metric is matched on the client side among those requested by broker subscription.
```
AndrewJSchofield, line 21: This section could be more tightly worded. I don't think it's a regression. It's a critical defect which is only exposed when a long list of conditions are met.

I wonder if the list of conditions could be worded the other way around. Essentially, the problem only occurs if all of these are true.

emasab: Change these phrases:
`4) Broker has a subscription assigned to the client`
`5) None of the subscription metrics matched on the client side.`
`any of the above conditions`

PR #4847, opened 2024-09-12
[KIP-74] Avg and max produce latency plus more tests

anchitj:
* Produce request latency metrics

* Add more tests


emasab (approved):
LGTM

PR #4851, opened 2024-09-24
Bump version for v2.6.0 release

emasab:


showuon:
@emasab @milindl , sorry to hijack this PR, since there is no maintainer reply to our PR. 
Could we consider to include this [PR](https://github.com/confluentinc/librdkafka/pull/3535) to v2.6.0 release? Without this change, it will impact some users adopt librdkafka. 
Thank you very much.

PR #4859, opened 2024-10-03
[KIP-74] Avg and max produce latency

emasab:
* Fix MinGW unit tests run
* Produce request latency metrics and more tests
* Remove unused variable and fix macOS build issue

---------


emasab:
Updating original PR #4847

Issue #4593, opened 2024-01-28
librdkafka cannot be built with LLD linker


artem:
Description
===========
`lds-gen.py` appends `rd_ut_coverage_check` for linker script unconditionally. However, LLD produces a warning if coverage is disabled. This warning becomes an error since LLVM LLD 17.0.0.

```
ld: error: version script assignment of 'global' to symbol 'rd_ut_coverage_check' failed: symbol not defined
```

How to reproduce
================

1. Install LLVM LLD 17.0.0 or newer.
2. Make sure /usr/bin/ld points to LLD binary.
3. Compile librdkafka with LDS enabled (default).
4. Linker produces an error.

Checklist
=========

Please provide the following information:

 - [x] librdkafka version: `2.3.0`

artem:
cc @emasab

artem:
FreeBSD seems to be affected as well: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=275866

nassibnassar:
#4862 fixes the problem for me on FreeBSD 14.1.

krinkinmu:
@confluenceinc/client folks, is there something we can help with to make progress on #4862 PR? Thank you!

PR #4862, opened 2024-10-04
Always define rd_ut_coverage_check

StephanDollberg:
Define `rd_ut_coverage_check` even in non-devel mode as the `lds-gen.py` script adds the symbol unconditionally.

Otherwise LLD complains about ld:

```
error: version script assignment of 'global' to symbol 'rd_ut_coverage_check' failed: symbol not defined
```

Fixes https://github.com/confluentinc/librdkafka/issues/4593


krinkinmu:
Hey @emasab and @pranavrth,

sorry for contacting you directly, but it seems like this PR and the associated bug fell through the cracks and you seem to be the only people who showed any activity in this project recently.

Is there anything I can help with to draw attention to this PR and the associated bug to make progress?

Thank you!

Issue #4778, opened 2024-07-09
Crash (null-pointer access) in rd_kafka_metadata_cache_entry_by_id_cmp() during rd_avl_insert()


GerKr:
Description
===========
After about 2 months of usage a crash happened. The crashdump-file shows that it happened in the function rd_kafka_metadata_cache_entry_by_id_cmp(const void *_a, const void *_b), where _b is 0x0000000000000000. This leads to a write access to some address 0x0000000000000088.
During the call of rd_avl_insert(ravl, elm, ran) the variable ravl contains:
   ravl->ravl_root == 0x12eaf651500 (no nullptr) and
   ravl->ravl_root.ran_height == 0x61657268 (converted to text it is "hrea" from the below found "hread_0")
   ravl->ravl_root.ran_elm == 0x0000000000000000 (nullptr!!!)

The memory where rafl->rafl_root points to looks like following:
0f 00 00 00 00 00 00 00 00 40 89 b8 2e 01 00 00 68 72 65 61 64 5f 30 00 00 00 00 00 00 00 00 00 0f 00 00 00 00 00 00 00 00 74 e2 b6 2e 01 00 00 06 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 0f 00 00 00 00 00 00 00 00 65 74 43 68 61 6e

The call-stack looks like following:

librdkafka.dll!rd_kafka_metadata_cache_entry_by_id_cmp(const void * _a, const void * _b) Line 700	C
librdkafka.dll!rd_avl_insert_node(rd_avl_s * ravl, rd_avl_node_s * parent, rd_avl_node_s * ran, rd_avl_node_s * * existing) Line 104	C
librdkafka.dll!rd_kafka_metadata_cache_insert(rd_kafka_s * rk, const rd_kafka_metadata_topic * mtopic, const rd_kafka_metadata_topic_internal_s * metadata_internal_topic, __int64 now, __int64 ts_expires, unsigned char include_racks, rd_kafka_metadata_broker_internal_s * brokers_internal, unsigned __int64 broker_cnt) Line 380	C
[...]

PR #4864, opened 2024-10-08
remove the old id from the cache

marcin-krystianc:
Fixes https://github.com/confluentinc/librdkafka/issues/4778

This PR addresses a subtle issue in handling topic IDs during metadata cache insertion. Separating this condition ensures that old entries with non-zero topic IDs are always properly removed from the metadata cache.
Removing the entity from the `rkmc_avl` tree but not removing the old entry with non-zero topic IDs from the `rkmc_avl_by_id` tree led to an inconsistent cache state. Eventually, it led to memory corruption ("use after free" errors).

### Testing
My test application re-creates (deletes and creates) three thousand topics. Then it immediately starts producer and consumer threads. Before the fix, the app crashed most of the time, after the fix crashes do not happen anymore.


adamreeve:
@emasab, could you please take a look at this? This change fixes invalid memory access.

stackedsax:
Hey there @emasab, looks like the comment in #4964 suggests you're going to merge this PR?  Trying to read the tea leaves here.  Thanks!

PR #4872, opened 2024-10-14
Upgrade msvcr140 and vcpkg dependencies

### LINKED ISSUE #4813: Incorrect version of zlib packaged with windows version
Description
===========

Continuing previous issue #4653 . The windows version of librdkafka packages the wrong version (1.3) of zlib here (https://github.com/confluentinc/librdkafka/blob/v2.5.0/vcpkg.json). The linux packaging appears to be downloading the correct version (1.3.1) here (https://github.com/confluentinc/librdkafka/blob/v2.5.0/mklove/modules/configure.zlib).

Vcpkg has the 1.3.1 version available for zlib (https://vcpkg.io/en/package/zlib) so could that be included in the build?

Checklist
=========

 - [x] librdkafka version (release number or git tag): latest
 - [ ] Apache Kafka version: 
 - [ ] librdkafka client configuration: 
 - [x] Operating system: Windows
 - [x] Critical issue

emasab:
Upgrade Windows dependencies: MSVC runtime to 14.40.338160.0, zstd 1.5.6, zlib 1.3.1, OpenSSL 3.3.2, CURL 8.10.1

Closes #4813


vcpkg.json
```diff
@@ -4,19 +4,19 @@
...
         },
         {
             "name": "zlib",
-            "version>=": "1.3"
+            "version>=": "1.3.1#0"
         },
         {
             "name": "openssl",
-            "version>=": "3.0.8"
+            "version>=": "3.3.2#1"
```
pranavrth, line 15: Directly upgrading OpenSSL from 3.0.8 to 3.3.2 might cause some issues as there are breaking changes as well. Make sure to check those breaking changes.

emasab: Checked these entries in the changelog:
* SSL 3, TLS 1.0, TLS 1.1, and DTLS 1.0 only work at security level 0.
  Shouldn't be a breaking change because TLS 1.2 is available since 2008
* The default SSL/TLS security level has been changed from 1 to 2
  Shouldn't be a breaking change because given this are the restriction for Level 2:

>  Level 2
Security level set to 112 bits of security. As a result RSA, DSA and DH keys shorter than 2048 bits and ECC keys shorter than 224 bits are prohibited. In addition to the level 1 exclusions any cipher suite using RC4 is also prohibited. SSL version 3 is also not allowed. Compression is disabled.

And the recommendation are:

> Since 2015, NIST recommends a minimum of 2048-bit keys for [RSA](https://en.wikipedia.org/wiki/RSA_(algorithm))
> The [NSA](https://en.wikipedia.org/wiki/National_Security_Agency) previously recommended 256-bit ECC for protecting classified information up to the SECRET level, and 384-bit for TOP SECRET;[[10]](https://en.wikipedia.org/wiki/Key_size#cite_note-NSASuiteBphaseout-11) In 2015 it announced plans to transition to quantum-resistant algorithms by 2024, and until then recommends 384-bit for all classified information

Given keys should be rotated much more frequently this shouldn't cause issues too.

Source: [Key size](https://en.wikipedia.org/wiki/Key_size#cite_note-keymanagement-13)

pranavrth (approved):
LGTM!

PR #4893, opened 2024-11-05
Socket options are now all set before connection

emasab:
[documentation](https://man7.org/linux/man-pages/man7/tcp.7.html)
says it's needed for socket buffers to take effect, even if in some
cases they could have effect even after connection.
Happening since v0.9.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+# librdkafka v2.6.2
+
+librdkafka v2.6.2 is a maintenance release:
+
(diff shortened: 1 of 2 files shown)


nimbuuus (approved):

pranavrth:
Changes looks good to me. Make sure that the options that we are setting are reflected in the connection. Update the branch. Build is having some transient error I think and it would pass after updating the branch.

pranavrth:
I checked on wireshark and it is getting set.

pranavrth (approved):
LGTM!

PR #4894, opened 2024-11-06
Client certificate chain is now sent

emasab:
when using `ssl.certificate.pem` or `ssl_certificate` or `ssl.keystore.location`.
Without that, broker must explicitly add any intermediate certification
authority certificate to its truststore to be able to accept client
certificate.
Happens since: 1.x


milindl (commented):
Reviewed the non-test files, looking at tests.

milindl (approved):

chrisplo:
Hi @emasab, did you test with single cert in ssl.certificate.pem (where the CA is trusted on server)? having trouble setting up a consumer with an error bubbling up through this code change in 2.8 (2.6.1 works fine).  I tried adding the CA after the cert but that failed too in 2.8 only.

> %4|1742936798.489|SSL|rdkafka#producer-1| [thrd:app]: Failed to read certificate chain from PEM. Returning NULL certificate too.

PR #4900, opened 2024-11-13
Client cert callback to check if trusted certificate authorities match with client certificate chain

emasab:
in Java this selection is happening in [X509KeyManagerImpl.getAliases]( https://github.com/openjdk/jdk/blob/d8430efb5e159b8e08d2cac66b46cb4ff1112927/src/java.base/share/classes/sun/security/ssl/X509KeyManagerImpl.java#L717)

The field that is checked is `certificate_authorities` extension in TLS 1.3, that was present in `CertificateRequest` in previous versions in TLS and SSL.


trnguyencflt (approved):
LGTM, thanks for the change

milindl (approved):
Approved after local testing.

Issue #4804, opened 2024-08-06
Fail to recover a partition which was temporarily delegated to the internal broker


bjosv:
Description
===========
When a partition is absent from metadata it is delegated to the internal broker `:0/internal`.
If the partition comes back later, with the same leader as before, it's no longer delegated back from the internal broker.
This makes the partition unusable, causing producing and consuming from said partition to stall.

The issue is likely a result of this change https://github.com/confluentinc/librdkafka/commit/6584ed7c8b00786121c07bc0df5b3d7fa8da2661  in v2.4.0 (PR #4680)

This change requires that the epoch has been changed (`leader_epoch > rktp->rktp_leader_epoch`) before calling `rd_kafka_toppar_broker_update()`, which is not the case in this situation.

This commit also affects the related testcase [107](https://github.com/confluentinc/librdkafka/blame/master/tests/0107-topic_recreate.c) which fails.
This related test scenario only seem to pass if the commit https://github.com/confluentinc/librdkafka/commit/6584ed7c8b00786121c07bc0df5b3d7fa8da2661 is reverted.

Logs where the issue can be seen:
```
2024-08-01T03:25:05.966753578 kafka:7: STATE: [thrd:main]: Topic MYTOPIC changed state unknown -> exists
2024-08-01T03:25:05.966777471 kafka:7: METADATA: [thrd:main]:   Topic MYTOPIC partition 0 Leader 2 Epoch 0
2024-08-01T03:25:05.966785841 kafka:7: BROKER: [thrd:main]: MYTOPIC [0]: leader -1 epoch -1 -> leader 2 epoch 0
2024-08-01T03:25:05.966790507 kafka:7: BRKDELGT: [thrd:main]: MYTOPIC [0]: delegate to broker ssl://kf-2.kf.svc.cluster.local:9093/2 (rktp 0x7f2d7ea4cd00, term 0, ref 3)
2024-08-01T03:25:05.966794391 kafka:7: BRKDELGT: [thrd:main]: MYTOPIC [0]: delegating to broker ssl://kf-2.kf.svc.cluster.local:9093/2 for partition with 0 messages (0 bytes) queued
2024-08-01T03:25:05.970168433 kafka:7: METADATA: [thrd:main]:   Topic MYTOPIC partition 0 Leader 2 Epoch 0
[...]

Issue #4796, opened 2024-07-26
Not switching to a leader without leader epochs while doing a cluster roll to upgrade the cluster


lvhuat:
#### Description:
After upgrading Kafka from version 2.2.1 to 3.5.1, we observed issues related to leader switching. The specific issues and reproduction steps are as follows:

#### Process:
1. During the Kafka upgrade, there were frequent leader switches.
2. In Kafka 2.2.1, there was no leader epoch in the metadata; it was set to -1.
3. When a node upgrades and the leader is on the upgrading node (e.g., node 2), the leader epoch becomes effective, let's assume it is 20.
4. If the leader switches back to node 1, sending messages at this point returns a NotLeader error, triggering a MetadataRequest.
5. Due to a weighting algorithm, all requests are received by node 1. Node 1 returns all leader epochs as -1. Since -1 < 20, no update occurs.
6. All nodes are incorrectly routed to node 2, forming a loop until node 1 starts upgrading.

#### Expected Behavior:
In such cases, the metadata should update correctly, allowing proper leader switching and load balancing.

#### Actual Behavior:
The metadata fails to update correctly, causing all requests to be incorrectly routed to a specific node, forming an error loop.

#### Additional Information:
Relevant Kafka configuration and librdkafka version are as follows:
- Kafka Version: 2.2.1 -> 3.5.1
- librdkafka Version: 2.4.0

#### partial logger 
```
[2024-07-26 02:30:16.729] [info] ConsumerEvent|my-topic-name|LOG|Success|7|METADATA|[thrd:main]:   Topic my-topic-name partition 0 Leader 1 Epoch -1|0||0|false
[2024-07-26 02:31:00.839] [info] ConsumerEvent|my-topic-name|LOG|Success|7|METADATA|[thrd:main]:   Topic my-topic-name partition 0 Leader 1 Epoch -1|0||0|false
[2024-07-26 02:31:01.334] [info] ConsumerEvent|my-topic-name|LOG|Success|7|METADATA|[thrd:main]:   Topic my-topic-name partition 0 Leader 2 Epoch -1|0||0|false
[2024-07-26 02:31:01.642] [info] ConsumerEvent|my-topic-name|LOG|Success|7|METADATA|[thrd:main]:   Topic my-topic-name partition 0 Leader 2 Epoch -1|0||0|false
[...]

PR #4901, opened 2024-11-14
Allow to migrate partitions to leaders with same leader epoch or NULL leader epoch

emasab:
* Closes: #4796.
  Fix to allow to migrate partitions to leaders with NULL leader epoch.
  NULL leader epoch can happen during a cluster roll with an upgrade to a
  version supporting KIP-320.
  Happening since v2.1.0
  
* Closes: #4804.
  Fix to allow to migrate partitions to leaders with same leader epoch.
  Same leader epoch can happen when partition is
  temporarily migrated to the internal broker (#4804), or if broker implementation
  never bumps it, as it's not needed to validate the offsets.
  Happening since v2.4.0

### DIFF
--- a/src/rdkafka_topic.c
+++ b/src/rdkafka_topic.c
@@ -682,6 +682,10 @@ static int rd_kafka_toppar_leader_update(rd_kafka_topic_t *rkt,
 
         rd_kafka_toppar_lock(rktp);
 
+        /* -1 (null) is excluded to allow to switch back to a
+         * leader not supporting KIP-320 still, for example
+         * during a cluster roll for upgrading brokers to
+         * a version supporting that KIP. */
         if (leader_epoch != -1 && leader_epoch < rktp->rktp_leader_epoch) {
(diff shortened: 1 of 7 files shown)


src/rdkafka_topic.c
```diff
@@ -681,7 +682,7 @@ static int rd_kafka_toppar_leader_update(rd_kafka_topic_t *rkt,
 
         rd_kafka_toppar_lock(rktp);
 
-        if (leader_epoch < rktp->rktp_leader_epoch) {
+        if (leader_epoch != -1 && leader_epoch < rktp->rktp_leader_epoch) {
```
milindl, line 685: Please add a comment above for why we are excluding -1, as it may not indicate outdated metadata but rather than the broker is KIP-320-unsupported.

milindl (approved):
Minor feedback

milindl (approved):

milindl (approved):

PR #4908, opened 2024-11-20
Fix assignment lost, on illegal generation, during a commit

emasab:
Closes: #4059.

Commits during a rebalance could cause to lose the assignment if the
generation id was bumped by second join group request.
Solved by not rejoining the group in case an illegal generation error happens
during a rebalance.

Happening since v1.6.0.

See this comment for the explanation of the issue:
https://github.com/confluentinc/librdkafka/issues/4059#issuecomment-2488264020


milindl (approved):
As reviewed earlier, lgtm

PR #4912, opened 2024-11-22
[KIP-714] Fix to avoid sending a NULL metrics payload

emasab:
when no metrics are matching, client side, a NULL metrics payload must not be sent, because it's not accepted by the protocol and it causes a disconnection by the broker.
An empty payload must be sent instead.

There are some additional checks on the mock implementation to match the broker one, so the test 0150 
```C
                do_test_telemetry_empty_subscriptions_list(
                    type, "non-existent-metric");
```
is failing.
The fix returns NULL only in the serialization error case and it's not sending a Push in that case, while it returns an empty buffer when there are zero metrics to send.

### DIFF
--- a/src/rdkafka_telemetry.c
+++ b/src/rdkafka_telemetry.c
@@ -266,6 +266,24 @@ rd_kafka_push_telemetry_payload_compress(rd_kafka_t *rk,
         rd_slice_t payload_slice;
         size_t i;
         rd_kafka_resp_err_t r = RD_KAFKA_RESP_ERR_NO_ERROR;
+
+        if (payload->rbuf_len == 0) {
+                /* We can only initialize the slice to compress
+                 * if not empty. */
+                rd_kafka_dbg(rk, TELEMETRY, "PUSH",
+                             "Empty payload. "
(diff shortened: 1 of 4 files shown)


src/rdkafka_telemetry.c
```diff
@@ -359,21 +377,24 @@ static void rd_kafka_send_push_telemetry(rd_kafka_t *rk,
...
+                    compression_used, compressed_metrics_payload,
+                    compressed_metrics_payload_size, NULL, 0,
+                    RD_KAFKA_REPLYQ(rk->rk_ops, 0),
+                    rd_kafka_handle_PushTelemetry, NULL);
         } else {
-                rd_kafka_dbg(rk, TELEMETRY, "PUSH",
-                             "No metrics to push. Sending empty payload.");
+                rd_kafka_log(rk, LOG_WARNING, "PUSH",
+                             "Telemetry metrics encode error, not sending "
+                             "a NULL payload");
```
anchitj, line 395: `not sending metrics` instead of `not sending a NULL payload` is better I guess.

anchitj (approved):
LGTM!

PR #4920, opened 2024-11-29
[KIP-848] Test 0147: mock tests specific for the 848 consumer group protocol

emasab:


tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+            mcluster, num, confidence_interval, is_heartbeat_request, NULL);
+}
+
+static rd_kafka_t *create_consumer(const char *bootstraps,
+                                   const char *group_id,
+                                   rd_bool_t with_rebalance_cb) {
+        rd_kafka_conf_t *conf;
+        test_conf_init(&conf, NULL, 0);
+        test_conf_set(conf, "bootstrap.servers", bootstraps);
+        test_conf_set(conf, "group.protocol", "consumer");
```
pranavrth: This is overwritten in `test_create_consumer`. We should update `test_create_consumer` to also check if it is not set already to update this. I don't think there would be much impact on other tests there.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+
+/**
+ * @brief Test heartbeat behavior with fatal errors,
+ *        ensuring:
+ *        - a fatal error is received on poll and consumer close
+ *        - no rebalance cb is called
+ *        - no final leave group heartbeat is sent
+ *
+ * @param err The error code to test.
+ * @param variation See calling code.
```
pranavrth: Explain Variations as well.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+        TEST_SAY("Subscribing to topic\n");
+        rd_kafka_mock_start_request_tracking(mcluster);
+        TEST_CALL_ERR__(rd_kafka_subscribe(c, subscription));
+        rd_kafka_topic_partition_list_destroy(subscription);
+
+        expected_heartbeats = 1;
+        if (variation == 1)
+                expected_heartbeats++;
+
+        TEST_SAY("Awaiting first HBs\n");
```
pranavrth: We are awaiting all the heartbeats.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+ *        ensuring:
+ *        - a fatal error is received on poll and consumer close
+ *        - no rebalance cb is called
+ *        - no final leave group heartbeat is sent
+ *
+ * @param err The error code to test.
+ * @param variation See calling code.
+ */
+static void
+do_test_consumer_group_heartbeat_fatal_error(rd_kafka_resp_err_t err,
```
pranavrth, line 172: I think we should wait for the consumer to wait till whole rebalance is completed and consumer is in stable state as well. It should be normally consuming and then we should send a FATAL error. We can add a new variation for this as well.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+        TEST_ASSERT(rd_kafka_consumer_close(c) == RD_KAFKA_RESP_ERR__FATAL,
+                    "Expected a _FATAL error, got %s", rd_kafka_err2name(err));
+
+        TEST_ASSERT(rebalance_cnt == expected_rebalance_cnt,
+                    "Expected %d rebalance events, got %d",
+                    expected_rebalance_cnt, rebalance_cnt);
+
+        TEST_SAY("Ensuring there are no leave group HBs\n");
+        TEST_ASSERT((found_heartbeats =
+                         wait_all_heartbeats_done(mcluster, 0, 200)) == 0,
```
pranavrth: We should at least wait (confidence_interval) for 2s as we are setting HB interval as 1s.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+ *        variation 1: errors on second HB
+ */
+static void do_test_consumer_group_heartbeat_fatal_errors(void) {
+        rd_kafka_resp_err_t fatal_errors[] = {
+            RD_KAFKA_RESP_ERR_INVALID_REQUEST,
+            RD_KAFKA_RESP_ERR_GROUP_MAX_SIZE_REACHED,
+            RD_KAFKA_RESP_ERR_UNSUPPORTED_ASSIGNOR,
+            RD_KAFKA_RESP_ERR_UNSUPPORTED_VERSION,
+            RD_KAFKA_RESP_ERR_UNRELEASED_INSTANCE_ID,
+            RD_KAFKA_RESP_ERR_GROUP_AUTHORIZATION_FAILED};
```
pranavrth, line 302: Apart from `RD_KAFKA_RESP_ERR_INVALID_REQUEST` and `RD_KAFKA_RESP_ERR_GROUP_AUTHORIZATION_FAILED`, we should receive all the other errors on the first HB call. We can change the test accordingly.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+                                                             1);
+        }
+}
+
+/**
+ * @brief Test heartbeat behavior with retriable errors,
+ *        ensuring:
+ *        - no error is received on poll and consumer close
+ *        - rebalance cb is called to assign and revoke
+ *        - final leave group heartbeat is sent
```
pranavrth: Consumer should keep on retrying the request in this case. We should also test without close in which the consumer is stuck in retry loop.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+                    "Consume from c, no message is returned, "
+                    "but assign callback is processed\n");
+                test_consumer_poll_no_msgs("after heartbeat", c, 0, 200);
+
+                /* wait 1 HB interval more */
+                expected_heartbeats += 1;
+                rebalance_exp_event = RD_KAFKA_RESP_ERR_NO_ERROR;
+        }
+
+        TEST_SAY("Awaiting first HBs\n");
```
pranavrth, line 410: Same

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+        TEST_SKIP_MOCK_CLUSTER(0);
+
+        if (test_consumer_group_protocol_classic()) {
+                TEST_SKIP("Test only for group.protocol=consumer\n");
+                return 0;
+        }
+
+        do_test_consumer_group_heartbeat_fatal_errors();
+
+        do_test_consumer_group_heartbeat_retriable_errors();
```
pranavrth, line 818: We should add REFRESH and SPECIAL action scenarios as well.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+                    retriable_errors[i], 1);
+        }
+}
+
+/**
+ * @brief Test heartbeat behavior with consumer fenced errors,
+ *        ensuring:
+ *        - no error is received on poll and consumer close
+ *        - rebalance callbacks are called, with partitions lost when
+ *          necessary
```
pranavrth: Better to explain the necessary scenarios.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+
+        TEST_SAY("Awaiting initial HBs\n");
+        TEST_ASSERT((found_heartbeats =
+                         wait_all_heartbeats_done(mcluster, expected_heartbeats,
+                                                  200)) == expected_heartbeats,
+                    "Expected %d heartbeats, got %d", expected_heartbeats,
+                    found_heartbeats);
+
+        expected_rebalance_cnt = 0;
+        /* variation 0: Second HB assigned */
```
pranavrth: Describe in better terms

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+ *        - no error is received on poll and consumer close
+ *        - rebalance callbacks are called, with partitions lost when
+ *          necessary
+ *        - a final leave group heartbeat is sent
+ *
+ * @param err The error code to test.
+ * @param variation See calling code.
+ */
+static void
+do_test_consumer_group_heartbeat_fenced_error(rd_kafka_resp_err_t err,
```
pranavrth, line 499: Explain scenarios. We should also add an ENUM for variations to explain directly. In all the test cases in this file.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+        rkmessage = rd_kafka_consumer_poll(c, 500);
+        TEST_ASSERT(!rkmessage, "No message should be returned");
+
+        TEST_ASSERT(rebalance_cnt == expected_rebalance_cnt,
+                    "Expected %d total rebalance events, got %d",
+                    expected_rebalance_cnt, rebalance_cnt);
+
+        if (variation == 0) {
+                /* Ack for assignment HB */
+                expected_heartbeats++;
```
pranavrth: I want to understand how can we count the number of heartbeats? Its periodic with 1s time period. Is our assumption that we will execute all these operations within 1s so that we only consider expedited heartbeats?

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+        } else if (variation == 1) {
+                /* First HB assigns again
+                 * Second HB acks assignment */
+                expected_heartbeats += 2;
```
pranavrth: This should be `expected_heartbeats += 3;` as for each assignment, we acknowledge as well.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+                                                  100)) == expected_heartbeats,
+                    "Expected %d heartbeats, got %d", expected_heartbeats,
+                    found_heartbeats);
+
+        expected_rebalance_cnt++;
+        rebalance_exp_event = RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS;
+
+        rd_kafka_mock_clear_requests(mcluster);
+        /* Close c, no error is returned */
+        TEST_CALL_ERR__(rd_kafka_consumer_close(c));
```
pranavrth, line 638: We should also check what happens when we receive the error in leave call as well. Lets understand this for all the errors.

pranavrth (changes requested):
First pass except - `do_test_metadata_unknown_topic_id_error`

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+ *        ensuring:
+ *        - initially a partial acknoledgement is started, with an empty list
+ *          (variation 0) or a single topic (variation 1)
+ *        - fetch doesn't start until broker returns an unknown topic id error
+ *        - when error isn't returned anymore the client finishes assigning
+ *          the partition and reads a message.
+ *
+ * @param variation See calling code.
+ */
+static void do_test_metadata_unknown_topic_id_error(int variation) {
```
pranavrth: In case of UNKNOWN_TOPIC_ID_ERROR, we should eventually reach a state where we don't have that topic in the assignment list as the broker will remove that topic eventually. We should check that as well.

tests/0147-consumer_group_consumer_mock.c
```diff
@@ -0,0 +1,682 @@
...
+                                      RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_ID);
+
+        TEST_SAY("Subscribing to topic\n");
+        TEST_CALL_ERR__(rd_kafka_subscribe(c, subscription));
+        rd_kafka_topic_partition_list_destroy(subscription);
+
+        TEST_SAY(
+            "Cannot fetch until Metadata calls replies with "
+            "UNKNOWN_TOPIC_ID\n");
+        test_consumer_poll_no_msgs("no messages", c, 0, 1000);
```
pranavrth, line 772: We should check the assignments as well here as there won't be this topic in the assignments.

tests/0147-consumer_group_consumer_mock.c
emasab: ok I avoid setting the configuration property as if a test if only for the consumer group protocol we should only execute it once and this test is already skipped for the classic protocol.

tests/0147-consumer_group_consumer_mock.c
emasab: They're already explained in `do_test_consumer_group_heartbeat_fatal_errors`, it's better to leave the comments only there I think

tests/0147-consumer_group_consumer_mock.c
emasab: Here we have already closed the consumer so in case the leave group HB was wrongly sent it should've been sent already. But in case of some weird bug it's fine I increase it.

tests/0147-consumer_group_consumer_mock.c
emasab: ok I do that in the tests with an error on leave group HB

tests/0147-consumer_group_consumer_mock.c
emasab: in this case it's later `Awaiting leave group HB` so it's why these are the first HBs

tests/0147-consumer_group_consumer_mock.c
emasab: ok I added the enums with the variations

tests/0147-consumer_group_consumer_mock.c
emasab: This corresponds to variation 1 where the error is received on second HB. What I could do is that on variation 1 it also consumes some messages before the fatal error from second HB.

tests/0147-consumer_group_consumer_mock.c
emasab: Instead of increasing the timeout I just move it after the destroy so it's sure we've counted all of them.

tests/0147-consumer_group_consumer_mock.c
emasab: Do you mean that for rest of error codes we shouldn't test variation 1 on second HB? Doing that.

tests/0147-consumer_group_consumer_mock.c
emasab: I changed it, avoided resetting the HB count in-between and counted the HBs from the beginning.

tests/0147-consumer_group_consumer_mock.c
emasab: In this case it's testing when the assignment doesn't change and the topic is eventually found. I add a third case where the topic isn't found and it's removed from the assignment while the existing one can continue with the reconciliation.

tests/0147-consumer_group_consumer_mock.c
emasab: When it happens on first HB it should be as fast as possible after the fencing. On variation 1 it's fast too as the ACK should come immediately on second HB if the callback is processed soon, on second HB there's the ACK and the fencing.
So yes they're all expedited HBs.

tests/0147-consumer_group_consumer_mock.c
pranavrth: I meant that we should test that we can get these 2 errors later and still it will throw FATAL error. Its not necessary that these two errors are thrown on the first HB.

pranavrth: But its fine for now.

emasab: those two yes they can be on second HB too, I meant that for the rest we can test only on first HB. They were working on second HB too even if this shouldn't happen with a real broker (rest of error codes).

pranavrth (approved):
LGTM!. Thanks @emasab!

PR #4931, opened 2024-12-13
[KIP-320] Validate assigned partitions before starting to consume from them

### LINKED ISSUE #5158: Assigned offsets are not validated and can cause an offset reset in case of partition truncation
When consumer starts consuming from an assigned partition, the start offset isn't validated even if it contains a leader epoch.
In this case if should be validated as the committed offset could have been truncated and the consumer needs to start from last available offset of that leader epoch.

When it's not done the risk is a different consumer connects to the new leader and by reading the committed offset it gets an out of range and an unexpected offset reset.

Happens since 1.x not solved by KIP-320 in 2.1.0

emasab:
Closes #5158 

Also: mock handler implementation for returning the committed offset leader epoch instead of current leader epoch.

### DIFF
--- a/src/rdkafka_assignment.c
+++ b/src/rdkafka_assignment.c
@@ -492,6 +492,13 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
                         rd_kafka_fetch_pos_t pos =
                             rd_kafka_topic_partition_get_fetch_pos(rktpar);
 
+                        /* Reset the (lib) pause flag which may have
+                         * been set by the cgrp when scheduling the
+                         * rebalance callback. */
+                        rd_kafka_toppar_op_pause_resume(
+                            rktp, rd_false /*resume*/,
+                            RD_KAFKA_TOPPAR_F_LIB_PAUSE, RD_KAFKA_NO_REPLYQ);
+
                         if (!RD_KAFKA_OFFSET_IS_LOGICAL(rktpar->offset) &&
                             pos.leader_epoch != -1) {
                                 rd_kafka_dbg(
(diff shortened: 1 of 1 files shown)


src/rdkafka_assignment.c
```diff
@@ -489,29 +489,55 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
...
+                                    rktp, rk->rk_consumer.q);
+                                rd_kafka_toppar_lock(rktp);
+                                rd_kafka_toppar_set_fetch_state(
+                                    rktp,
+                                    RD_KAFKA_TOPPAR_FETCH_VALIDATE_EPOCH_WAIT);
+                                rd_kafka_toppar_set_next_fetch_position(rktp,
+                                                                        pos);
+                                rd_kafka_toppar_set_offset_validation_position(
+                                    rktp, pos);
+                                rd_kafka_offset_validate(rktp, "offset fetch");
```
milindl, line 521: is it not possible that this partition is (lib) paused like what's happening below? To be resumed before consumption using `rd_kafka_toppar_op_pause_resume`? Later on, it seems that it's not paused on successful validation, just the state is changed to active

emasab: Yes agree it should be resumed too before validating. In the tests it's calling a pause and resume so it may be it's not visible because it's tried with those operations. I do some checks.

emasab: The resume is called when assigning the topic partition for a consumer group in `rd_kafka_assignment_resume`. In that case the partition can be paused by the library before calling the rebalance callback.
```
%7|1758109224.768|PAUSE|0139_offset_validation_mock#consumer-8| [thrd:main]: Resuming fetchers for 1 assigned partition(s): incremental assign called
```
and then
```%7|1758109224.769|SRVPEND|0139_offset_validation_mock#consumer-8| [thrd:main]: Validating assigned partition offset rdkafkatest_rnd3ddf79da72aba312_do_test_offset_validation_on_partition_assignment [0] at offset 7 (leader epoch 0)
%7|1758109224.769|PARTSTATE|0139_offset_validation_mock#consumer-8| [thrd:main]: Partition rdkafkatest_rnd3ddf79da72aba312_do_test_offset_validation_on_partition_assignment [0] changed fetch state none -> validate-epoch-wait
```

emasab: Tried to remove the additional resume calls and leave only those in `rd_kafka_assignment_resume` and the tests are passing. Just to be safer I keep current way for the moment and put the resume in both cases.

milindl (approved):

PR #4936, opened 2024-12-17
Fix to avoid issuing a warning in case server requests a certificate and client is using SASL authentication only

emasab:
without any client certificate set.


trnguyencflt (approved):
LGTM, thanks 👍

milindl (approved):
thanks a lot, lgtm

PR #4939, opened 2024-12-19
KIP 848:Extend DescribeConfigs and IncrementalAlterConfigs to support GROUP Config

PratRanj07:
This PR intends to add support for Group Resource type for IncrementalALterConfigs API as specified in KIP 848 and extended the DescribeConfigs support till API version 3.
Additional changed the integration test to check for Group Resource Type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -160,6 +160,7 @@ librdkafka v2.10.0 is a feature release:
 
 librdkafka v2.8.0 is a maintenance release:
 
+* Extend Config Apis to support group config (#4939).
 * Socket options are now all set before connection (#4893).
 * Client certificate chain is now sent when using `ssl.certificate.pem`
   or `ssl_certificate` or `ssl.keystore.location` (#4894).
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -2912,22 +2912,20 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
 }
 
 
-rd_kafka_ConfigResourceType_t map_from_resource_type_to_config_resource_type(
-    rd_kafka_ResourceType_t resourcetype) {
-        if (resourcetype == RD_KAFKA_RESOURCE_GROUP) {
+int8_t map_from_resource_type_to_config_resource_type(
+    rd_kafka_ResourceType_t restype) {
+        if (restype == RD_KAFKA_RESOURCE_GROUP)
                 return RD_KAFKA_CONFIG_RESOURCE_GROUP;
-        }
 
-        return resourcetype;
+        return restype;
 }
 
-rd_kafka_ResourceType_t map_from_config_resource_type_to_resource_type(
-    rd_kafka_ConfigResourceType_t internal_resourcetype) {
-        if (internal_resourcetype == RD_KAFKA_CONFIG_RESOURCE_GROUP) {
+rd_kafka_ResourceType_t
+map_from_config_resource_type_to_resource_type(int8_t config_resource_type) {
+        if (config_resource_type == RD_KAFKA_CONFIG_RESOURCE_GROUP)
                 return RD_KAFKA_RESOURCE_GROUP;
-        }
 
-        return internal_resourcetype;
+        return config_resource_type;
 }
 
 
@@ -3388,7 +3386,7 @@ rd_kafka_IncrementalAlterConfigsResponse_parse(rd_kafka_op_t *rko_req,
                 int16_t error_code;
                 rd_kafkap_str_t error_msg;
                 int8_t res_type;
-                int8_t internal_res_type;
+                int8_t config_resource_type;
                 rd_kafkap_str_t kres_name;
                 char *res_name;
                 char *this_errstr = NULL;
@@ -3398,13 +3396,13 @@ rd_kafka_IncrementalAlterConfigsResponse_parse(rd_kafka_op_t *rko_req,
 
                 rd_kafka_buf_read_i16(reply, &error_code);
                 rd_kafka_buf_read_str(reply, &error_msg);
-                rd_kafka_buf_read_i8(reply, &internal_res_type);
+                rd_kafka_buf_read_i8(reply, &config_resource_type);
                 rd_kafka_buf_read_str(reply, &kres_name);
                 RD_KAFKAP_STR_DUPA(&res_name, &kres_name);
                 rd_kafka_buf_skip_tags(reply);
 
                 res_type = map_from_config_resource_type_to_resource_type(
-                    internal_res_type);
+                    config_resource_type);
 
                 if (error_code) {
                         if (RD_KAFKAP_STR_IS_NULL(&error_msg) ||
@@ -3662,7 +3660,7 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
         for (i = 0; i < (int)res_cnt; i++) {
                 int16_t error_code;
                 rd_kafkap_str_t error_msg;
-                int8_t internal_res_type;
+                int8_t config_resource_type;
                 int8_t res_type;
                 rd_kafkap_str_t kres_name;
                 char *res_name;
@@ -3674,12 +3672,12 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
 
                 rd_kafka_buf_read_i16(reply, &error_code);
                 rd_kafka_buf_read_str(reply, &error_msg);
-                rd_kafka_buf_read_i8(reply, &internal_res_type);
+                rd_kafka_buf_read_i8(reply, &config_resource_type);
                 rd_kafka_buf_read_str(reply, &kres_name);
                 RD_KAFKAP_STR_DUPA(&res_name, &kres_name);
 
                 res_type = map_from_config_resource_type_to_resource_type(
-                    internal_res_type);
+                    config_resource_type);
 
                 if (error_code) {
                         if (RD_KAFKAP_STR_IS_NULL(&error_msg) ||
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -2850,16 +2850,6 @@ rd_kafka_ConfigEntry_synonyms(const rd_kafka_ConfigEntry_t *entry,
         return (const rd_kafka_ConfigEntry_t **)entry->synonyms.rl_elems;
 }
 
-const rd_kafka_ConfigType_t
-rd_kafka_ConfigEntry_type(const rd_kafka_ConfigEntry_t *entry) {
-        return entry->type;
-}
-
-const char *
-rd_kafka_ConfigEntry_documentation(const rd_kafka_ConfigEntry_t *entry) {
-        return entry->documentation;
-}
-
 
 /**@}*/
 
@@ -2923,7 +2913,8 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
 
 
 rd_kafka_InternalConfigResourceType_t
-map_to_internal_config_resourcetype(rd_kafka_ResourceType_t resourcetype) {
+map_from_resource_type_to_internal_config_resource_type(
+    rd_kafka_ResourceType_t resourcetype) {
         switch (resourcetype) {
         case RD_KAFKA_RESOURCE_UNKNOWN:
                 return RD_KAFKA_INTERNAL_RESOURCE_CONFIG_UNKNOWN;
@@ -2942,7 +2933,7 @@ map_to_internal_config_resourcetype(rd_kafka_ResourceType_t resourcetype) {
         }
 }
 
-rd_kafka_ResourceType_t map_from_internal_config_resourcetype(
+rd_kafka_ResourceType_t map_from_internal_config_resource_type_to_resource_type(
     rd_kafka_InternalConfigResourceType_t internal_resourcetype) {
         switch (internal_resourcetype) {
         case RD_KAFKA_INTERNAL_RESOURCE_CONFIG_UNKNOWN:
@@ -3436,7 +3427,8 @@ rd_kafka_IncrementalAlterConfigsResponse_parse(rd_kafka_op_t *rko_req,
                 rd_kafka_buf_skip_tags(reply);
 
                 res_type =
-                    map_from_internal_config_resourcetype(internal_res_type);
+                    map_from_internal_config_resource_type_to_resource_type(
+                        internal_res_type);
 
                 if (error_code) {
                         if (RD_KAFKAP_STR_IS_NULL(&error_msg) ||
@@ -3671,9 +3663,6 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
         int32_t Throttle_Time;
         rd_kafka_ConfigResource_t *config = NULL;
         rd_kafka_ConfigEntry_t *entry     = NULL;
-        int16_t api_version;
-
-        api_version = rd_kafka_buf_ApiVersion(reply);
 
         rd_kafka_buf_read_i32(reply, &Throttle_Time);
         rd_kafka_op_throttle_time(rkb, rk->rk_rep, Throttle_Time);
@@ -3714,7 +3703,8 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                 RD_KAFKAP_STR_DUPA(&res_name, &kres_name);
 
                 res_type =
-                    map_from_internal_config_resourcetype(internal_res_type);
+                    map_from_internal_config_resource_type_to_resource_type(
+                        internal_res_type);
 
                 if (error_code) {
                         if (RD_KAFKAP_STR_IS_NULL(&error_msg) ||
@@ -3743,10 +3733,8 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                 rd_kafka_buf_read_i32(reply, &entry_cnt);
 
                 for (ci = 0; ci < (int)entry_cnt; ci++) {
-                        rd_kafkap_str_t config_name, config_value,
-                            documentation;
+                        rd_kafkap_str_t config_name, config_value;
                         int32_t syn_cnt;
-                        int8_t config_type;
                         int si;
 
                         rd_kafka_buf_read_str(reply, &config_name);
@@ -3781,7 +3769,7 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                         rd_kafka_buf_read_bool(reply, &entry->a.is_sensitive);
 
 
-                        if (rd_kafka_buf_ApiVersion(reply) >= 1) {
+                        if (rd_kafka_buf_ApiVersion(reply) == 1) {
                                 /* #config_synonyms (ApiVersion 1) */
                                 rd_kafka_buf_read_i32(reply, &syn_cnt);
 
@@ -3814,7 +3802,6 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                                 rd_kafka_buf_read_str(reply, &syn_name);
                                 rd_kafka_buf_read_str(reply, &syn_value);
                                 rd_kafka_buf_read_i8(reply, &syn_source);
-                                rd_kafka_buf_skip_tags(reply);
 
                                 syn_entry = rd_kafka_ConfigEntry_new0(
                                     syn_name.str, RD_KAFKAP_STR_LEN(&syn_name),
@@ -3840,19 +3827,9 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                                 rd_list_add(&entry->synonyms, syn_entry);
                         }
 
-                        if (api_version >= 3) {
-                                rd_kafka_buf_read_i8(reply, &config_type);
-                                rd_kafka_buf_read_str(reply, &documentation);
-                                entry->type = config_type;
-                                entry->documentation =
-                                    RD_KAFKAP_STR_DUP(&documentation);
-                        }
-                        rd_kafka_buf_skip_tags(reply);
-
                         rd_kafka_ConfigResource_add_ConfigEntry(config, entry);
                         entry = NULL;
                 }
-                rd_kafka_buf_skip_tags(reply);
 
                 /* As a convenience to the application we insert result
                  * in the same order as they were requested. The broker
@@ -3881,7 +3858,6 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                             config);
                 config = NULL;
         }
-        rd_kafka_buf_skip_tags(reply);
 
         *rko_resultp = rko_result;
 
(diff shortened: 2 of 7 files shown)


src/rdkafka_admin.c
```diff
@@ -3774,9 +3839,19 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                                 rd_list_add(&entry->synonyms, syn_entry);
                         }
 
+                        if (api_version >= 3) {
+                                rd_kafka_buf_read_i8(reply, &config_type);
+                                rd_kafka_buf_read_str(reply, &documentation);
+                                entry->type = config_type;
+                                entry->documentation =
+                                    RD_KAFKAP_STR_DUP(&documentation);
+                        }
```
pranavrth, line 3848: Is this change necessary for KIP-848 changes?

src/rdkafka_admin.c
```diff
@@ -3774,9 +3839,19 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
...
+                        rd_kafka_buf_skip_tags(reply);
+
                         rd_kafka_ConfigResource_add_ConfigEntry(config, entry);
                         entry = NULL;
                 }
+                rd_kafka_buf_skip_tags(reply);
```
pranavrth, line 3854: same

src/rdkafka_admin.c
```diff
@@ -3805,6 +3880,7 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
                             config);
                 config = NULL;
         }
+        rd_kafka_buf_skip_tags(reply);
```
pranavrth, line 3883: same

src/rdkafka_request.c
```diff
@@ -5484,14 +5489,21 @@ rd_kafka_resp_err_t rd_kafka_DescribeConfigsRequest(
...
 
-        if (ApiVersion == 1) {
+        if (ApiVersion >= 1) {
                 /* include_synonyms */
                 rd_kafka_buf_write_i8(rkbuf, 1);
         }
 
+        if (ApiVersion >= 3) {
+                /* include_documentation */
+                rd_kafka_buf_write_i8(rkbuf, 1);
```
pranavrth, line 5503: Why always true?

src/rdkafka_request.c
```diff
@@ -5445,7 +5447,7 @@ rd_kafka_resp_err_t rd_kafka_DescribeConfigsRequest(
         }
 
         ApiVersion = rd_kafka_broker_ApiVersion_supported(
-            rkb, RD_KAFKAP_DescribeConfigs, 0, 1, NULL);
+            rkb, RD_KAFKAP_DescribeConfigs, 0, 3, NULL);
```
pranavrth, line 5450: I would suggest not to increase the version if it is not necessary. We would need to improve APIs as well to support that and there would be bunch of other things as well which we should be taking care of.

src/rdkafka_request.c
```diff
@@ -5484,14 +5489,21 @@ rd_kafka_resp_err_t rd_kafka_DescribeConfigsRequest(
...
+        if (ApiVersion >= 1) {
                 /* include_synonyms */
                 rd_kafka_buf_write_i8(rkbuf, 1);
         }
 
+        if (ApiVersion >= 3) {
+                /* include_documentation */
+                rd_kafka_buf_write_i8(rkbuf, 1);
+        }
+        rd_kafka_buf_write_tags_empty(rkbuf);
```
pranavrth, line 5505: No need to write tags if we are not supporting flexver.

src/rdkafka_request.c
```diff
@@ -5484,14 +5489,21 @@ rd_kafka_resp_err_t rd_kafka_DescribeConfigsRequest(
                         /* config_name */
                         rd_kafka_buf_write_str(rkbuf, entry->kv->name, -1);
                 }
+                rd_kafka_buf_write_tags_empty(rkbuf);
```
pranavrth, line 5492: same

pranavrth (changes requested):
First high level pass

src/rdkafka_request.c
PratRanj07: removed it since wasn't neccesarry as we are not upgrading the api version

src/rdkafka_admin.c
PratRanj07: No this change was for upgrading the API version. Have removed it.

src/rdkafka_admin.h
```diff
@@ -283,6 +283,33 @@ struct rd_kafka_ConfigResource_result_s {
...
 
+typedef enum rd_kafka_InternalConfigResourceType_t {
+        RD_KAFKA_INTERNAL_RESOURCE_CONFIG_UNKNOWN = 0,
+        RD_KAFKA_INTERNAL_RESOURCE_CONFIG_ANY     = 1,
+        RD_KAFKA_INTERNAL_RESOURCE_CONFIG_TOPIC   = 2,
+        RD_KAFKA_INTERNAL_RESOURCE_CONFIG_GROUP =
+            32,  // Changed value for config APIs
+        RD_KAFKA_INTERNAL_RESOURCE_CONFIG_BROKER = 4,
+        RD_KAFKA_INTERNAL_RESOURCE_CONFIG_CNT,
+} rd_kafka_InternalConfigResourceType_t;
```
pranavrth, line 294: Let's do this conversion only for GROUP right now. We can think about others later.

pranavrth (changes requested):
Please do the changes for the comments which are valid now.

src/rdkafka_admin.c
```diff
@@ -2911,6 +2911,25 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
 }
 
 
+rd_kafka_ConfigResourceType_t map_from_resource_type_to_config_resource_type(
+    rd_kafka_ResourceType_t resourcetype) {
```
pranavrth, line 2915:
```suggestion
int8_t map_from_resource_type_to_config_resource_type(
    rd_kafka_ResourceType_t restype) {
```

src/rdkafka_admin.c
```diff
@@ -2911,6 +2911,25 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
...
+        if (resourcetype == RD_KAFKA_RESOURCE_GROUP) {
+                return RD_KAFKA_CONFIG_RESOURCE_GROUP;
+        }
+
+        return resourcetype;
+}
+
+rd_kafka_ResourceType_t map_from_config_resource_type_to_resource_type(
+    rd_kafka_ConfigResourceType_t internal_resourcetype) {
```
pranavrth, line 2924:
```suggestion
rd_kafka_ResourceType_t map_from_config_resource_type_to_resource_type(
    int8_t config_resource_type) {
```

src/rdkafka_admin.c
```diff
@@ -2911,6 +2911,25 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
 }
 
 
+rd_kafka_ConfigResourceType_t map_from_resource_type_to_config_resource_type(
+    rd_kafka_ResourceType_t resourcetype) {
+        if (resourcetype == RD_KAFKA_RESOURCE_GROUP) {
+                return RD_KAFKA_CONFIG_RESOURCE_GROUP;
+        }
```
pranavrth, line 2918:
```suggestion
        if (resourcetype == RD_KAFKA_RESOURCE_GROUP)
                return RD_KAFKA_CONFIG_RESOURCE_GROUP;
```

src/rdkafka_admin.c
```diff
@@ -2911,6 +2911,25 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
...
+
+        return resourcetype;
+}
+
+rd_kafka_ResourceType_t map_from_config_resource_type_to_resource_type(
+    rd_kafka_ConfigResourceType_t internal_resourcetype) {
+        if (internal_resourcetype == RD_KAFKA_CONFIG_RESOURCE_GROUP) {
+                return RD_KAFKA_RESOURCE_GROUP;
+        }
```
pranavrth, line 2927:
```suggestion
        if (internal_resourcetype == RD_KAFKA_CONFIG_RESOURCE_GROUP)
                return RD_KAFKA_RESOURCE_GROUP;
```

src/rdkafka_admin.c
```diff
@@ -3377,11 +3397,14 @@ rd_kafka_IncrementalAlterConfigsResponse_parse(rd_kafka_op_t *rko_req,
 
                 rd_kafka_buf_read_i16(reply, &error_code);
                 rd_kafka_buf_read_str(reply, &error_msg);
-                rd_kafka_buf_read_i8(reply, &res_type);
+                rd_kafka_buf_read_i8(reply, &internal_res_type);
```
pranavrth, line 3400: `config_resource_type`

src/rdkafka_admin.c
```diff
@@ -3638,6 +3661,7 @@ rd_kafka_DescribeConfigsResponse_parse(rd_kafka_op_t *rko_req,
         for (i = 0; i < (int)res_cnt; i++) {
                 int16_t error_code;
                 rd_kafkap_str_t error_msg;
+                int8_t internal_res_type;
```
pranavrth, line 3664: `config_resource_type`

src/rdkafka_admin.h
```diff
@@ -283,6 +283,26 @@ struct rd_kafka_ConfigResource_result_s {
...
+
+/**
+ * @brief Map from rd_kafka_ConfigResourceType_t to
+ * rd_kafka_ResourceType_t
+ */
+
+rd_kafka_ResourceType_t map_from_config_resource_type_to_resource_type(
+    rd_kafka_ConfigResourceType_t internal_resourcetype);
+
+
```
pranavrth, line 305: same name, variable and type changes here as well.

src/rdkafka_admin.h
```diff
@@ -283,6 +283,26 @@ struct rd_kafka_ConfigResource_result_s {
                               *   but with response error values. */
 };
 
+typedef enum rd_kafka_ConfigResourceType_t {
+        RD_KAFKA_CONFIG_RESOURCE_GROUP = 32,  // Changed value for config APIs
```
pranavrth, line 287:
```suggestion
        RD_KAFKA_CONFIG_RESOURCE_GROUP = 32,
```

pranavrth (changes requested):
Tests next.

src/rdkafka_admin.h
```diff
@@ -283,6 +283,24 @@ struct rd_kafka_ConfigResource_result_s {
...
+/**
+ * @brief Map from rd_kafka_ResourceType_t to int8_t
+ */
+int8_t
+map_from_resource_type_to_config_resource_type(rd_kafka_ResourceType_t restype);
+
+/**
+ * @brief Map from int8_t to rd_kafka_ResourceType_t
+ */
+
```
pranavrth, line 299: Remove extra line.

src/rdkafka_admin.h
```diff
@@ -283,6 +283,24 @@ struct rd_kafka_ConfigResource_result_s {
                               *   but with response error values. */
 };
 
+typedef enum rd_kafka_ConfigResourceType_t {
```
pranavrth, line 286: Add doc.

tests/0081-admin.c
```diff
@@ -1035,6 +1051,23 @@ static void do_test_IncrementalAlterConfigs(rd_kafka_t *rk,
...
+
+        error = rd_kafka_ConfigResource_add_incremental_config(
+            configs[ci], "consumer.session.timeout.ms",
+            RD_KAFKA_ALTER_CONFIG_OP_TYPE_SET, "50000");
+        TEST_ASSERT(!error, "%s", rd_kafka_error_string(error));
+        if (!test_consumer_group_protocol_classic()) {
+                exp_err[ci] = RD_KAFKA_RESP_ERR_NO_ERROR;
+        } else {
+                exp_err[ci] = RD_KAFKA_RESP_ERR_INVALID_REQUEST;
+        }
```
pranavrth, line 1068: We should check broker version as well here.

tests/0081-admin.c
```diff
@@ -1334,6 +1367,137 @@ static void do_test_DescribeConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
...
+        /*
+         * ConfigResource #3: group config, for a non-existent group.
+         */
+        configs[ci] =
+            rd_kafka_ConfigResource_new(RD_KAFKA_RESOURCE_GROUP, group);
+        if (!test_consumer_group_protocol_classic()) {
+                exp_err[ci] = RD_KAFKA_RESP_ERR_NO_ERROR;
+        } else {
+                exp_err[ci] = RD_KAFKA_RESP_ERR_INVALID_REQUEST;
+        }
```
pranavrth, line 1405: same

tests/0081-admin.c
```diff
@@ -5258,6 +5422,7 @@ static void do_test_apis(rd_kafka_type_t cltype) {
 
         /* DescribeConfigs */
         do_test_DescribeConfigs(rk, mainq);
+        do_test_DescribeConfigs_groups(rk, mainq);
```
pranavrth, line 5425: Why not add new condition in `do_test_DescribeConfigs`?

pranavrth (changes requested):

src/rdkafka_admin.h
pranavrth, line 1: There is a problem that the error message contain 32 instead of 3. We are not exposing 32 to the user. I think we should modify the error code in the response as well though it is difficult to catch all the cases. Can be an improvement though.

`ConfigResource result: 3,my-group: error: Unknown resource type 32`

src/rdkafka_admin.h
```diff
@@ -283,6 +283,26 @@ struct rd_kafka_ConfigResource_result_s {
...
 
+/**
+ * @brief Resource type specific to config apis.
+ */
+typedef enum rd_kafka_ConfigResourceType_t {
+        RD_KAFKA_CONFIG_RESOURCE_GROUP = 32,
+} rd_kafka_ConfigResourceType_t;
+
+/**
+ * @brief Map from rd_kafka_ResourceType_t to int8_t
```
pranavrth, line 294: Better doc explaining what and why are we doing this.

src/rdkafka_admin.h
```diff
@@ -283,6 +283,26 @@ struct rd_kafka_ConfigResource_result_s {
...
+ */
+int8_t
+map_from_resource_type_to_config_resource_type(rd_kafka_ResourceType_t restype);
+
+/**
+ * @brief Map from int8_t to rd_kafka_ResourceType_t
```
pranavrth, line 300: Better doc explaining what and why are we doing this.

tests/0081-admin.c
```diff
@@ -947,6 +948,10 @@ static void do_test_IncrementalAlterConfigs(rd_kafka_t *rk,
                 const char *confs_delete_subtract_broker[] = {
                     "background.threads", "DELETE",   "",
                     "log.cleanup.policy", "SUBTRACT", "compact"};
+                const char *confs_set_append_group[] = {
```
pranavrth, line 951:
```suggestion
                const char *confs_set_group[] = {
```

tests/0081-admin.c
```diff
@@ -1334,6 +1368,138 @@ static void do_test_DescribeConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
...
+                            "got type %s name %s",
+                            i,
+                            rd_kafka_ResourceType_name(
+                                rd_kafka_ConfigResource_type(configs[i])),
+                            rd_kafka_ConfigResource_name(configs[i]),
+                            rd_kafka_ResourceType_name(
+                                rd_kafka_ConfigResource_type(rconfigs[i])),
+                            rd_kafka_ConfigResource_name(rconfigs[i]));
+                        fails++;
+                        continue;
```
pranavrth, line 1487: `continue;` not needed.

pranavrth (changes requested):

src/rdkafka_admin.c
```diff
@@ -2911,6 +2913,23 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
 }
 
 
+int8_t map_from_resource_type_to_config_resource_type(
```
emasab, line 2916: Given we're the treating the Java types let's use the camelcase for function name and the reverse function name.
It should return `rd_kafka_ConfigResourceType_t`.
```suggestion
rd_kafka_ConfigResourceType_t rd_kafka_ResourceType_to_ConfigResourceType(
```

src/rdkafka_admin.c
```diff
@@ -2911,6 +2913,23 @@ const char *rd_kafka_ResourceType_name(rd_kafka_ResourceType_t restype) {
...
+    rd_kafka_ResourceType_t restype) {
+        if (restype == RD_KAFKA_RESOURCE_GROUP)
```
emasab, line 2918: Let's use a `switch` to map all the options, in case we pass a `ResourceType` that cannot be used as a `ConfigResourceType` like `RD_KAFKA_RESOURCE_TRANSACTIONAL_ID` it should be mapped to `RD_KAFKA_CONFIG_RESOURCE_TYPE_UNKNOWN` not to 5 that isn't a valid `ConfigResourceType`.
Same for the reverse function.

src/rdkafka_admin.c
```diff
@@ -2867,6 +2867,8 @@ const char *rd_kafka_ConfigSource_name(rd_kafka_ConfigSource_t confsource) {
             "UNKNOWN_CONFIG",        "DYNAMIC_TOPIC_CONFIG",
             "DYNAMIC_BROKER_CONFIG", "DYNAMIC_DEFAULT_BROKER_CONFIG",
             "STATIC_BROKER_CONFIG",  "DEFAULT_CONFIG",
+            "UNSUPPORTED",           "UNSUPPORTED",
```
emasab, line 2870: Let's put the real name, even if we don't have the configuration values still they could be returned by the broker.
```suggestion
            "DYNAMIC_BROKER_LOGGER_CONFIG", "CLIENT_METRICS_CONFIG",
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 librdkafka v2.8.0 is a maintenance release:
 
+* Extend Config Apis to support group config (#4939).
```
emasab, line 5: Mapping should be done for AlterConfigs too otherwise there's a problem if they pass the `RD_KAFKA_RESOURCE_GROUP` there.

CHANGELOG.md
emasab, line 1: Change the `incremental_alter_configs.c` example to add support for group configuration.

emasab (changes requested):
Thanks @PratRanj07 some comments about the PR

tests/0081-admin.c
```diff
@@ -935,6 +935,7 @@ static void do_test_IncrementalAlterConfigs(rd_kafka_t *rk,
         /** Test the test helper, for use in other tests. */
         do {
                 const char *broker_id = tsprintf("%d", avail_brokers[0]);
+                const char *group_id  = "my-group";
```
emasab, line 938: Let's use the first topic as group here as in different places. It's randomly generated and avoid problem with repeated tests.
```suggestion
                const char *group_id  = topics[0];
```

tests/0081-admin.c
```diff
@@ -1334,6 +1367,142 @@ static void do_test_DescribeConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
...
+        const rd_kafka_ConfigResource_t **rconfigs;
+        char *group = rd_strdup(test_mk_topic_name(__FUNCTION__, 1));
+        size_t rconfig_cnt;
+        char errstr[128];
+        const char *errstr2;
+        int ci = 0;
+        int i;
+        int fails = 0;
+
+        SUB_TEST_QUICK();
```
emasab, line 1391: Let's initialize group after the test definition, it creates a memory leak in case the test is skipped adding noise to the checks.
```suggestion
        SUB_TEST_QUICK();
        group = rd_strdup(test_mk_topic_name(__FUNCTION__, 1));
```

tests/0081-admin.c
```diff
@@ -1334,6 +1367,142 @@ static void do_test_DescribeConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
...
+
+        /*
+         * ConfigResource #0: group config, for a non-existent group.
+         */
+        configs[ci] =
+            rd_kafka_ConfigResource_new(RD_KAFKA_RESOURCE_GROUP, group);
+        if (test_broker_version >= TEST_BRKVER(3, 8, 0, 0) &&
```
emasab, line 1398:
```suggestion
        if (group_configs_supported()) {
```

tests/0081-admin.c
```diff
@@ -1334,6 +1367,142 @@ static void do_test_DescribeConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
...
+                        fails++;
+                }
+        }
+
+        TEST_ASSERT(!fails, "See %d previous failure(s)", fails);
+
+        rd_kafka_event_destroy(rkev);
+
+        rd_kafka_ConfigResource_destroy_array(configs, ci);
+
```
emasab, line 1500: Free the options and the group
```suggestion
        rd_kafka_AdminOptions_destroy(options);

        rd_free(group);
```

tests/0081-admin.c
```diff
@@ -1334,6 +1367,142 @@ static void do_test_DescribeConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
...
+        TEST_LATER_CHECK();
```
emasab, line 1501:
```suggestion
        TEST_LATER_CHECK();
#undef MY_CONFRES_CNT
```

tests/0081-admin.c
emasab: The reason could be that when testing with a broker version where it's not supported it gives the `INVALID_REQUEST` on all the `ConfigEntry` then existing tests would need to be changed too. There's some duplicated code on config APIs, for the verification especially that maybe could be refactored to simplify, not in this PR I think.

emasab (changes requested):
Some comments on the tests

tests/0081-admin.c
```diff
@@ -1035,6 +1050,24 @@ static void do_test_IncrementalAlterConfigs(rd_kafka_t *rk,
...
+         */
+        configs[ci] =
+            rd_kafka_ConfigResource_new(RD_KAFKA_RESOURCE_GROUP, "my-group");
+
+        error = rd_kafka_ConfigResource_add_incremental_config(
+            configs[ci], "consumer.session.timeout.ms",
+            RD_KAFKA_ALTER_CONFIG_OP_TYPE_SET, "50000");
+        TEST_ASSERT(!error, "%s", rd_kafka_error_string(error));
+        if (test_broker_version >= TEST_BRKVER(3, 8, 0, 0) &&
+            !test_consumer_group_protocol_classic()) {
```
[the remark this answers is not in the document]
PratRanj07: It is indeed giving us a Invalid request in case of Classic protocol.

emasab (approved):
Great work Pratyush! Approving too

pranavrth (approved):
LGTM!.

PR #4968, opened 2025-02-18
[KIP-848] Added Regex support for the new consumer group protocol.

pranavrth:


src/rdkafka_cgrp.c
```diff
@@ -5151,6 +5177,18 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
         }
 }
 
+
+static void rd_kafka_cgrp_subscription_unset(rd_kafka_cgrp_t *rkcg) {
```
emasab, line 5181: We can unify this function with `rd_kafka_cgrp_subscription_set` which should do all the necessary changes when changing a subscription, that are:

- destroying previous `rkcg_subscription` and `rkcg_subscription_regex`
- unsetting the flags RD_KAFKA_CGRP_F_SUBSCRIPTION and RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION
- for the group protocol `consumer` setting the flags RD_KAFKA_CGRP_CONSUMER_F_SUBSCRIBED_ONCE and RD_KAFKA_CGRP_CONSUMER_F_SEND_NEW_SUBSCRIPTION
- for the group protocol `consumer` if it's a wildcard subscription combine the regexes and remove them from the subscription
- we can remove `rd_kafka_metadata_cache_hint_rktparlist` as it's fixed in a different way with the subscription version and creates some problems. I'll change the corresponding PR for that.

src/rdkafka_cgrp.c
```diff
@@ -6484,6 +6533,9 @@ static rd_kafka_op_res_t rd_kafka_cgrp_op_serve(rd_kafka_t *rk,
                 rko = NULL;
                 break;
 
+        /**
+         * TODO: Check regex for KIP-848
+         */
```
emasab, line 6538: This returns the subscription so if we remove it from `rkcg_subscription` it can return an empty subscription if we have all regexes. We can have two new fields: `rkcg_subscription_topics` and the existing `rkcg_subscription_regex`.  `rkcg_subscription` remains unchanged and `rkcg_subscription_topics` is the same with the regexes removed.

src/rdkafka_cgrp.c
```diff
@@ -5274,6 +5312,9 @@ rd_kafka_cgrp_modify_subscription(rd_kafka_cgrp_t *rkcg,
 static rd_kafka_resp_err_t rd_kafka_cgrp_unsubscribe(rd_kafka_cgrp_t *rkcg,
                                                      rd_bool_t leave_group) {
 
+        /**
+         * TODO: Update debug with Regex changes for KIP-848
+         */
```
emasab, line 5317: If we do the change below to avoid modfying `rkcg_subscription` we don't need to change the logging part.

src/rdkafka_cgrp.h
```diff
@@ -231,8 +231,22 @@ typedef struct rd_kafka_cgrp_s {
...
         rd_kafka_topic_partition_list_t *rkcg_next_subscription;
+
+        /**
+         * Subscription regex pattern. All the provided regex patterns are
+         * stored as a single string with each pattern separated by '|'.
+         *
+         * Only applicable for the consumer protocol introduced in KIP-848.
+         *
+         * TODO: Improve doc.
+         */
```
emasab, line 247: What did you want to add here?

src/rdkafka_cgrp.h
```diff
@@ -231,8 +231,22 @@ typedef struct rd_kafka_cgrp_s {
         /** If a SUBSCRIBE op is received during a COOPERATIVE rebalance,
          *  actioning this will be postponed until after the rebalance
          *  completes. The waiting subscription is stored here.
-         *  Mutually exclusive with rkcg_next_subscription. */
+         *  Mutually exclusive with rkcg_next_subscription.
+         *
+         *  For the consumer protocol, this field doesn't include regex
+         *  subscriptions. For that please refer `rkcg_subscription_regex` */
```
emasab, line 237: This doc if for `rkcg_next_subscription`, the field should include regex subscriptions.
Move it to the doc for `rkcg_subscription_topics`

src/rdkafka_partition.c
```diff
@@ -4421,6 +4421,86 @@ int rd_kafka_topic_partition_list_regex_cnt(
...
+        for (i = 0; i < (*rktparlist)->cnt; i++) {
+                const rd_kafka_topic_partition_t *rktpar =
+                    &((*rktparlist)->elems[i]);
+                if (*rktpar->topic != '^')
+                        rd_kafka_topic_partition_list_add_copy(
+                            non_regex_rktparlist, rktpar);
+        }
+
+        rd_kafka_topic_partition_list_destroy(*rktparlist);
+        *rktparlist = non_regex_rktparlist;
```
emasab, line 4446: We can use the match function:
```suggestion
        return rd_kafka_topic_partition_list_match(
            rktparlist, rd_kafka_topic_partition_not_regex, NULL);
```

src/rdkafka_partition.c
```diff
@@ -4421,6 +4421,86 @@ int rd_kafka_topic_partition_list_regex_cnt(
...
+                if (*rktpar->topic == '^') {
+                        combined_regex_len += strlen(rktpar->topic);
+                        regex_cnt++;
+                }
+        }
+
+        if (regex_cnt == 0)
+                return NULL;
+
+        combined_regex_len += regex_cnt - 1; /* 1 for each '|' separator */
```
emasab, line 4475: We need to add `3 * (regex_cnt - 1)` chars here we have to combine with `)|(` as it's possible a regex is like `^a|b` and another one is `^c` and the combined one would be `(^a|b|^c)`. There shouldn't be a problem as `b` is matched in the first one as well, but I think it's safer (and more evident) if we write `(^a|b)|(^c)`

src/rdkafka_partition.c
```diff
@@ -4421,6 +4421,86 @@ int rd_kafka_topic_partition_list_regex_cnt(
...
+
+        // Construct the combined regex string
+        combined_regex_str[0] = '(';
+        for (i = 0; i < rktparlist->cnt; i++) {
+                const rd_kafka_topic_partition_t *rktpar =
+                    &(rktparlist->elems[i]);
+                char *topic = rktpar->topic;
+                if (*topic == '^') {
+                        while (*topic) {
+                                combined_regex_str[j++] = *topic;
```
emasab, line 4489: Here it's not adding the `|` or `)|(` to `combined_regex_str`

src/rdkafka_partition.h
```diff
@@ -1010,6 +1010,12 @@ rd_kafka_resp_err_t rd_kafka_topic_partition_list_get_err(
 int rd_kafka_topic_partition_list_regex_cnt(
     const rd_kafka_topic_partition_list_t *rktparlist);
 
+void rd_kafka_topic_partition_list_remove_regex(
```
emasab, line 1013:
```suggestion
void rd_kafka_topic_partition_list_remove_regexes(
```

src/rdkafka_partition.h
```diff
@@ -1010,6 +1010,12 @@ rd_kafka_resp_err_t rd_kafka_topic_partition_list_get_err(
...
+    rd_kafka_topic_partition_list_t **rktparlist);
+
+rd_kafkap_str_t *rd_kafka_topic_partition_list_combine_regex(
```
emasab, line 1016:
```suggestion
rd_kafkap_str_t *rd_kafka_topic_partition_list_combine_regexes(
```

src/rdkafka_proto.h
```diff
@@ -620,6 +620,8 @@ static RD_INLINE RD_UNUSED int rd_kafka_Uuid_ptr_cmp(void *a, void *b) {
 
 rd_kafka_Uuid_t rd_kafka_Uuid_random();
 
+char *rd_kafka_Uuid_random_str();
+
```
emasab, line 624: same: to remove

src/rdkafka_request.c
```diff
@@ -2283,9 +2284,11 @@ void rd_kafka_ConsumerGroupHeartbeatRequest(
         int16_t ApiVersion = 0;
         int features;
         size_t rkbuf_size = 0;
+        rd_kafkap_str_t *subscribed_topic_regex_to_send =
```
emasab, line 2287: Here it's possible it sends a NULL regex. We have to make sure that when we have a subscription to send we send the topic list even if empty and an empty string as the subscription regex. We have to do this otherwise with NULL it keeps the previous topic list or the previous subscription.

src/rdkafka_request.c
```diff
@@ -2283,9 +2284,11 @@ void rd_kafka_ConsumerGroupHeartbeatRequest(
...
+            subscribed_topic_regex;
 
         ApiVersion = rd_kafka_broker_ApiVersion_supported(
-            rkb, RD_KAFKAP_ConsumerGroupHeartbeat, 0, 0, &features);
+            rkb, RD_KAFKAP_ConsumerGroupHeartbeat, 0, 1, &features);
```
emasab, line 2291: We can remove support for version 0 as we didn't have a GA version with it and avoid handling the case where we have a regex to send but the broker isn't supporting it.
```suggestion
            rkb, RD_KAFKAP_ConsumerGroupHeartbeat, 1, 1, &features);
```
In case it's not supported we can mark the coordinator as dead and return
```C
        if (ApiVersion == -1) {
                rd_kafka_cgrp_coord_dead(rkb->rkb_rk->rk_cgrp,
                                         RD_KAFKA_RESP_ERR__UNSUPPORTED_FEATURE,
                                         "ConsumerGroupHeartbeatRequest not "
                                         "supported by broker");
                return;
        }
```

emasab (changes requested):
Thanks Pranav! Sent the first comments on the implementation, next ones are for the mock cluster and the tests.

src/rdkafka_cgrp.c
```diff
@@ -5144,8 +5162,25 @@ rd_kafka_cgrp_calculate_subscribe_revoking_partitions(
 static void
 rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
                                rd_kafka_topic_partition_list_t *rktparlist) {
+
+        int is_wildcard_subscription_present = 0;
```
emasab, line 5166: This bool isn't necessary you can put the `rd_kafka_topic_partition_list_regex_cnt` call in the if condition and then use `rkcg->rkcg_flags & RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION`

src/rdkafka_cgrp.c
```diff
@@ -5154,8 +5189,42 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
...
+                if (rkcg->rkcg_subscription) {
+                        rkcg->rkcg_subscription_topics =
+                            rd_kafka_topic_partition_list_copy(
+                                rkcg->rkcg_subscription);
+
+                        if (is_wildcard_subscription_present) {
+                                rkcg->rkcg_subscription_regex =
+                                    rd_kafka_topic_partition_list_combine_regexes(
+                                        rkcg->rkcg_subscription);
+                                rd_kafka_topic_partition_list_remove_regexes(
```
emasab, line 5208: You can make `rd_kafka_topic_partition_list_remove_regexes` directly return the filtered topic partition list and assign it to `rkcg->rkcg_subscription_topics` here

src/rdkafka_cgrp.c
```diff
@@ -5154,8 +5189,42 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
...
+        if (rkcg->rkcg_group_protocol == RD_KAFKA_GROUP_PROTOCOL_CONSUMER) {
+                RD_IF_FREE(rkcg->rkcg_subscription_topics,
+                           rd_kafka_topic_partition_list_destroy);
+                RD_IF_FREE(rkcg->rkcg_subscription_regex,
+                           rd_kafkap_str_destroy);
+
+                if (rkcg->rkcg_subscription) {
+                        rkcg->rkcg_subscription_topics =
+                            rd_kafka_topic_partition_list_copy(
+                                rkcg->rkcg_subscription);
```
emasab, line 5202: This assignment can be put in the else statement below only

src/rdkafka_cgrp.c
```diff
@@ -5154,8 +5189,42 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
...
+                        }
+
+                        rkcg->rkcg_consumer_flags |=
+                            RD_KAFKA_CGRP_CONSUMER_F_SUBSCRIBED_ONCE |
+                            RD_KAFKA_CGRP_CONSUMER_F_SEND_NEW_SUBSCRIPTION;
+                } else {
+                        rkcg->rkcg_subscription_regex =
+                            rd_kafkap_str_new("", 0);
+                        rkcg->rkcg_subscription_topics =
+                            rd_kafka_topic_partition_list_new(0);
```
emasab, line 5222: These two should be NULL when we unsubscribe, as `rkcg_subscription`

src/rdkafka_cgrp.c
```diff
@@ -5186,7 +5250,6 @@ rd_kafka_cgrp_modify_subscription(rd_kafka_cgrp_t *rkcg,
         revoking = rd_kafka_cgrp_calculate_subscribe_revoking_partitions(
             rkcg, unsubscribing_topics);
 
-        rd_kafka_topic_partition_list_destroy(rkcg->rkcg_subscription);
         rd_kafka_cgrp_subscription_set(rkcg, rktparlist);
```
emasab, line 5253: After rebasing, here it should assign the returned value to `cgrp_subscription_version` and pass it to `rd_kafka_cgrp_metadata_refresh`

src/rdkafka_cgrp.c
```diff
@@ -5389,11 +5446,6 @@ rd_kafka_cgrp_subscribe(rd_kafka_cgrp_t *rkcg,
         if (!rktparlist)
                 return RD_KAFKA_RESP_ERR_NO_ERROR;
 
-        rkcg->rkcg_flags |= RD_KAFKA_CGRP_F_SUBSCRIPTION;
-
-        if (rd_kafka_topic_partition_list_regex_cnt(rktparlist) > 0)
-                rkcg->rkcg_flags |= RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION;
-
         rd_kafka_cgrp_subscription_set(rkcg, rktparlist);
```
emasab, line 5449: Receive the `subscription_version` here and pass it to `rd_kafka_cgrp_join`

src/rdkafka_partition.c
```diff
@@ -4421,6 +4421,91 @@ int rd_kafka_topic_partition_list_regex_cnt(
 }
 
 
+/**
+ * @brief Match function that returns true if topic is non regex.
```
emasab, line 4425:
```suggestion
 * @brief Match function that returns true if topic is non a regex.
```

src/rdkafka_partition.c
```diff
@@ -4421,6 +4421,91 @@ int rd_kafka_topic_partition_list_regex_cnt(
...
+ */
+static RD_UNUSED int rd_kafka_topic_partition_not_regex(const void *elem,
```
emasab, line 4427: `RD_UNUSED` is usually together with `RD_INLINE` for functions in headers that can be included and used in other files
```suggestion
static int rd_kafka_topic_partition_not_regex(const void *elem,
```

src/rdkafka_partition.c
```diff
@@ -4421,6 +4421,91 @@ int rd_kafka_topic_partition_list_regex_cnt(
...
+                                                        const void *opaque) {
+        const rd_kafka_topic_partition_t *rktpar = elem;
+        return *rktpar->topic != '^';
+}
+
+/**
+ * @brief Removes regexes from the list and returns the list without any regex
+ * in it.
+ */
+void rd_kafka_topic_partition_list_remove_regexes(
```
emasab, line 4437: should return the `rd_kafka_topic_partition_list_t *` and the `rktparlist` should be a single pointer and `const`

src/rdkafka_partition.c
```diff
@@ -4421,6 +4421,91 @@ int rd_kafka_topic_partition_list_regex_cnt(
...
+        rd_kafka_topic_partition_list_t *non_regex_rktparlist =
+            rd_kafka_topic_partition_list_match(
+                *rktparlist, rd_kafka_topic_partition_not_regex, NULL);
+        rd_kafka_topic_partition_list_destroy(*rktparlist);
+        *rktparlist = non_regex_rktparlist;
+}
+
+
+/**
+ * @brief Combine regex present in the list into a single regex.
```
emasab, line 4448:
```suggestion
 * @brief Combine regexes present in the list into a single regex.
```

src/rdkafka_request.c
```diff
@@ -2283,25 +2284,36 @@ void rd_kafka_ConsumerGroupHeartbeatRequest(
...
+        }
 
         if (rd_rkb_is_dbg(rkb, CGRP)) {
-                char current_assignments_str[512] = "NULL";
-                char subscribe_topics_str[512]    = "NULL";
-                const char *member_id_str         = "NULL";
-                const char *group_instance_id_str = "NULL";
-                const char *remote_assignor_str   = "NULL";
+                char current_assignments_str[512]              = "NULL";
+                char subscribe_topics_str[512]                 = "NULL";
```
emasab, line 2303:
```suggestion
                char subscribed_topics_str[512]                 = "NULL";
```

src/rdkafka_request.c
```diff
@@ -2359,23 +2382,26 @@ void rd_kafka_ConsumerGroupHeartbeatRequest(
...
         rd_kafka_buf_write_i32(rkbuf, rebalance_timeout_ms);
 
-        if (subscribe_topics) {
+        if (subscribed_topics) {
                 size_t of_TopicsArrayCnt;
-                int topics_cnt = subscribe_topics->cnt;
+                int topics_cnt = subscribed_topics->cnt;
 
                 /* write Topics */
                 of_TopicsArrayCnt = rd_kafka_buf_write_arraycnt_pos(rkbuf);
```
emasab, line 2390: we can remove `rd_kafka_buf_write_arraycnt_pos` and `of_TopicsArrayCnt` and use directly `rd_kafka_buf_write_arraycnt`. That is necessary only if the count is to update after some element is skipped.

src/rdkafka_partition.c
pranavrth: Its `non regex` but I am changing it to `not a regex`

src/rdkafka_cgrp.c
```diff
@@ -263,10 +263,27 @@ typedef RD_MAP_TYPE(const rd_kafka_topic_partition_t *,
 
 /**
  * @returns true if consumer has joined the group and thus requires a leave.
+ *
+ * `rkcg_member_id` is sufficient to know this with "classic" group protocol.
  */
+#define RD_KAFKA_CGRP_HAS_JOINED_CLASSIC(rkcg)                                 \
+        ((rkcg->rkcg_group_protocol == RD_KAFKA_GROUP_PROTOCOL_CLASSIC &&      \
```
emasab, line 270: These double brackets can be only one

src/rdkafka_cgrp.c
```diff
@@ -263,10 +263,27 @@ typedef RD_MAP_TYPE(const rd_kafka_topic_partition_t *,
...
+          RD_KAFKAP_STR_LEN((rkcg)->rkcg_member_id) > 0))
+
+/**
+ * @returns true if consumer has joined the group and thus requires a leave.
+ *
+ * With "consumer" group protocol we cannot rely on the `rkcg_member_id`
+ * as it's client generated.
+ */
+#define RD_KAFKA_CGRP_HAS_JOINED_CONSUMER(rkcg)                                \
+        ((rkcg->rkcg_group_protocol == RD_KAFKA_GROUP_PROTOCOL_CONSUMER &&     \
```
emasab, line 281: These double brackets can be only one

src/rdkafka_cgrp.c
```diff
@@ -5173,10 +5202,45 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
                                rd_kafka_topic_partition_list_t *rktparlist) {
         int32_t new_subscription_version =
             rd_atomic32_add(&rkcg->rkcg_subscription_version, 1);
```
emasab, line 5204: This can be done at the end of the function and directly returned without the variable

src/rdkafka_cgrp.c
```diff
@@ -5173,10 +5202,45 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
...
+
+        rkcg->rkcg_flags &= ~(RD_KAFKA_CGRP_F_SUBSCRIPTION |
+                              RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION);
+        if (rkcg->rkcg_subscription)
+                rd_kafka_topic_partition_list_destroy(rkcg->rkcg_subscription);
```
emasab, line 5209: Let's use `RD_IF_FREE` here too
```suggestion
        RD_IF_FREE(rkcg->rkcg_subscription,
                   rd_kafka_topic_partition_list_destroy);
```

src/rdkafka_partition.c
```diff
@@ -4426,6 +4426,91 @@ int rd_kafka_topic_partition_list_regex_cnt(
...
+}
+
+/**
+ * @brief Return a new list with all regex topics removed.
+ *
+ * @remark The caller is responsible for freeing the returned list.
+ */
+rd_kafka_topic_partition_list_t *rd_kafka_topic_partition_list_remove_regexes(
+    const rd_kafka_topic_partition_list_t *rktparlist) {
+        rd_kafka_topic_partition_list_t *non_regex_rktparlist =
```
emasab, line 4445: This variable can be removed and it can directly return the `rd_kafka_topic_partition_list_match` returned value

src/rdkafka_request.c
```diff
@@ -2319,18 +2334,26 @@ void rd_kafka_ConsumerGroupHeartbeatRequest(
                            ", group instance id \"%s\""
                            ", current assignment \"%s\""
                            ", subscribe topics \"%s\""
+                           ", subscribe topic regex \"%s\""
```
emasab, line 2337:
```suggestion
                           ", subscribed topics \"%s\""
                           ", subscribed topic regex \"%s\""
```

emasab (changes requested):
Comments on the implementation

.semaphore/semaphore.yml
```diff
@@ -145,6 +145,7 @@ blocks:
             - (cd tests && python3 -m trivup.clusters.KafkaCluster --version 3.9.0 --cpversion 7.9.0 --cmd 'make quick')
         - name: 'Build and integration tests with "consumer" protocol'
           commands:
+            - sem-version java 17
```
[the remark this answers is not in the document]
pranavrth: Yes, it was added in prologue with the latest commit in master.

src/rdkafka_cgrp.c
pranavrth: Was already there. Doing the change.

src/rdkafka_mock_handlers.c
```diff
@@ -3014,7 +3018,7 @@ const struct rd_kafka_mock_api_handler
         [RD_KAFKAP_OffsetForLeaderEpoch] =
             {2, 2, -1, rd_kafka_mock_handle_OffsetForLeaderEpoch},
         [RD_KAFKAP_ConsumerGroupHeartbeat] =
-            {0, 0, 0, rd_kafka_mock_handle_ConsumerGroupHeartbeat},
+            {0, 1, 0, rd_kafka_mock_handle_ConsumerGroupHeartbeat},
```
emasab, line 3021: Minimum version an flexver should be 1

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,19 +1402,65 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
-        if (!SubscribedTopicNames)
+    int32_t SubscribedTopicNamesCnt,
+    const rd_kafkap_str_t *SubscribedTopicRegex) {
+        rd_bool_t changed           = rd_false;
+        rd_list_t *new_subscription = NULL;
+        int32_t i;
+
+        if (!SubscribedTopicNames &&
+            RD_KAFKAP_STR_IS_NULL(SubscribedTopicRegex) &&
+            !member->subscribed_topic_regex)
```
emasab, line 1413: Here it should return changed also if `subscribed_topic_regex` is an empty string

emasab: After this `if` we can have an assert that both values must be non-NULL and either one or the other must be non-empty. We ensure this in the subscribe call, we return invalid `_INVALID_ARG` if passing an empty list or if it's containing invalid regexes

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,19 +1402,65 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
                 /* When client is sending NULL,
                  * its subscription didn't change */
                 return changed;
 
-        int32_t i;
-        rd_list_t *new_subscription = rd_list_new(
-            SubscribedTopicNamesCnt > 0 ? SubscribedTopicNamesCnt : 1, rd_free);
-        for (i = 0; i < SubscribedTopicNamesCnt; i++) {
-                rd_list_add(new_subscription,
+        if (!RD_KAFKAP_STR_IS_NULL(SubscribedTopicRegex)) {
```
emasab, line 1418: This `if` can be changed because of previous assert.
We can check `subscribed_topic_regex` here, if it's different we replace it with a copy of `SubscribedTopicRegex`. Even if `SubscribedTopicRegex` is an empty string it must be replaced.

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,19 +1402,65 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                                rd_list_add(new_subscription,
+                                            RD_KAFKAP_STR_DUP(topic_name));
+                                rd_kafkap_str_destroy(topic_name);
+                        }
+                }
+
+                rd_regex_destroy(re);
+        }
+
+        if (SubscribedTopicNames) {
```
emasab, line 1452: Remove this `if` because of previous assert
```suggestion
```

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,19 +1402,65 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                if (member->subscribed_topic_regex) {
+                        rd_kafkap_str_destroy(member->subscribed_topic_regex);
+                        member->subscribed_topic_regex = NULL;
+                }
+                if (SubscribedTopicRegex->len > 0)
+                        member->subscribed_topic_regex =
+                            rd_kafkap_str_copy(SubscribedTopicRegex);
+        }
+
+        if (member->subscribed_topic_regex) {
```
emasab, line 1428: Because of previous assert one of the two fields will be non-empty. So let's create the `new_subscription` in advance.

src/rdkafka_mock_cgrp.c
emasab: Also let's put the literal topic names before those matched by the regexes

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,19 +1402,65 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                    rd_regex_comp(member->subscribed_topic_regex->str, errstr,
+                                  sizeof(errstr));
+
+                TAILQ_FOREACH(mtopic, &mcluster->topics, link) {
+                        if (rd_regex_exec(re, mtopic->name)) {
+                                rd_kafkap_str_t *topic_name =
+                                    rd_kafkap_str_new(mtopic->name, -1);
+                                if (!new_subscription)
+                                        new_subscription =
+                                            rd_list_new(1, rd_free);
```
emasab, line 1442: we can remove this when creating it in advance

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,19 +1402,65 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                rd_regex_destroy(re);
+        }
+
+        if (SubscribedTopicNames) {
+                if (!new_subscription)
+                        new_subscription =
+                            rd_list_new(SubscribedTopicNamesCnt > 0
+                                            ? SubscribedTopicNamesCnt
+                                            : 1,
+                                        rd_free);
```
emasab, line 1458: we can remove this when creating it in advance. Use `SubscribedTopicNamesCnt` there too.

src/rdkafka_mock_cgrp.c
```diff
@@ -1464,21 +1512,19 @@ rd_kafka_mock_cgrp_consumer_member_add(rd_kafka_mock_cgrp_consumer_t *mcgrp,
...
+                         */
+                        rd_free(member->id);
+                        member->id = RD_KAFKAP_STR_DUP(MemberId);
                 }
         }
 
         if (!member) {
-                if (SubscribedTopicNamesCnt < 1)
+                if (SubscribedTopicNamesCnt < 1 &&
+                    RD_KAFKAP_STR_IS_NULL(SubscribedTopicRegex))
```
emasab, line 1527: let's transform this into an assert that both are non-NULL and at least one is non-empty. This should be always ensured by the implementation (when adding a new member or changing subscription).

src/rdkafka_mock_handlers.c
```diff
@@ -2817,6 +2818,9 @@ rd_kafka_mock_handle_ConsumerGroupHeartbeat(rd_kafka_mock_connection_t *mconn,
                 }
         }
 
+        if (rkbuf->rkbuf_reqhdr.ApiVersion >= 1)
```
emasab, line 2821: after setting the `MinVer` as 1 this `if` can be removed

emasab (changes requested):
Review on the mock cluster implementation

src/rdkafka_mock_cgrp.c
```diff
@@ -1464,21 +1512,19 @@ rd_kafka_mock_cgrp_consumer_member_add(rd_kafka_mock_cgrp_consumer_t *mcgrp,
...
-                         * It both don't hold, we cannot add the member
-                         * to the group. */
-                        return NULL;
+                if (member && RD_KAFKAP_STR_LEN(MemberId) > 0) {
+                        /**
+                         * If the member is found by instance id and the
+                         * member id is provided, we update the member id.
+                         */
+                        rd_free(member->id);
+                        member->id = RD_KAFKAP_STR_DUP(MemberId);
```
emasab, line 1521: I have an improved PR for static membership, for the moment you can revert the changes and skip the relative failing test

src/rdkafka_mock_cgrp.c
emasab: Below when creating a new member you can always set the `member->id` to what's passed.
```C
                /* KIP 1082: MemberId is generated by the client */
                member->id = RD_KAFKAP_STR_DUP(MemberId);
```

tests/0045-subscribe_update.c
```diff
@@ -674,17 +694,90 @@ static void do_test_replica_rack_change_leader_no_rack_mock(
...
+        /* Wait for assignment */
+        await_assignment("Assignment for topic2", rk, queue, 1, topic2, 4);
+
+        /* Unsubscribe from topic2 */
+        TEST_SAY("Unsubscribing from %s\n", topic2);
+        rd_kafka_unsubscribe(rk);
+        /* Wait for revocation */
+        await_revoke("Revocation after unsubscribing", rk, queue);
+
+        rd_sleep(1);
```
emasab, line 745: This `rd_sleep` can be removed

tests/test.c
```diff
@@ -153,6 +153,7 @@ _TEST_DECL(0042_many_topics);
 _TEST_DECL(0043_no_connection);
 _TEST_DECL(0044_partition_cnt);
 _TEST_DECL(0045_subscribe_update);
+_TEST_DECL(0045_resubscribe_with_regex);
```
emasab, line 156: Put it after the `subscribe_update*` tests

tests/0102-static_group_rebalance.c
```diff
@@ -786,7 +1007,9 @@ int main_0102_static_group_rebalance(int argc, char **argv) {
         /* TODO: check again when regexes
          * will be supported by KIP-848 */
         if (test_consumer_group_protocol_classic()) {
-                do_test_static_group_rebalance();
+                do_test_static_group_rebalance_classic();
+        } else {
+                do_test_static_group_rebalance_consumer();
         }
```
emasab, line 1014: merge these with the condition below

tests/0102-static_group_rebalance.c
```diff
@@ -118,13 +118,21 @@ static void rebalance_cb(rd_kafka_t *rk,
 
                 c->partition_cnt = parts->cnt;
                 c->assigned_at   = test_clock();
-                rd_kafka_assign(rk, parts);
+                if (test_consumer_group_protocol_classic()) {
+                        rd_kafka_assign(rk, parts);
+                } else {
+                        rd_kafka_incremental_assign(rk, parts);
+                }
```
emasab, line 125: Use `test_consumer_assign_by_rebalance_protocol`

tests/0102-static_group_rebalance.c
```diff
@@ -118,13 +118,21 @@ static void rebalance_cb(rd_kafka_t *rk,
...
                 break;
 
         case RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS:
                 c->revoked_at = test_clock();
-                rd_kafka_assign(rk, NULL);
+                if (test_consumer_group_protocol_classic()) {
+                        rd_kafka_assign(rk, NULL);
+                } else {
+                        rd_kafka_incremental_unassign(rk, parts);
+                }
```
emasab, line 135: Use `test_consumer_unassign_by_rebalance_protocol`

emasab (changes requested):
Comments about the mock cluster and the tests. Test 0102 `do_test_static_group_rebalance_consumer` is to finish reviewing because there are some mock cluster  changes to do for static group membership let's talk about them.

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,20 +1402,73 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                if (member->subscribed_topic_regex) {
+                        rd_kafkap_str_destroy(member->subscribed_topic_regex);
+                        member->subscribed_topic_regex = NULL;
+                }
+                if (RD_KAFKAP_STR_LEN(SubscribedTopicRegex) > 0)
+                        member->subscribed_topic_regex =
+                            rd_kafkap_str_copy(SubscribedTopicRegex);
+        }
+
+        if (!SubscribedTopicNames) {
```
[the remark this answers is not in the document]
pranavrth:
```
new_subscription =
            rd_list_new(member->subscribed_topic_names && rd_list_cnt(member->subscribed_topic_names) > 0
                            ? rd_list_cnt(member->subscribed_topic_names)
                            : 1,
                        rd_free);
```

Need to do NULL check as well as there can be only regex subscription since the starting.

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,20 +1402,73 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+    int32_t SubscribedTopicNamesCnt,
+    const rd_kafkap_str_t *SubscribedTopicRegex) {
         rd_bool_t changed = rd_false;
-        if (!SubscribedTopicNames)
+        rd_list_t *new_subscription;
+        int32_t i;
+
+        if (!SubscribedTopicNames &&
+            RD_KAFKAP_STR_IS_NULL(SubscribedTopicRegex) &&
+            !member->subscribed_topic_regex)
```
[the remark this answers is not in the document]
pranavrth: I prefer this to be the way it is right now. Its much more readable with just one extra condition. Other developer will have to understand `validate_subscription` function first to understand this condition otherwise.

I can add a comment to describe the current condition.

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,20 +1414,75 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+        }
+
+        if (SubscribedTopicNames) {
+                if (member->subscribed_topic_names)
+                        rd_list_destroy(member->subscribed_topic_names);
+                member->subscribed_topic_names =
+                    rd_list_new(SubscribedTopicNamesCnt, rd_free);
+                for (i = 0; i < SubscribedTopicNamesCnt; i++) {
+                        rd_list_add(member->subscribed_topic_names,
+                                    rd_strdup(SubscribedTopicNames[i].str));
```
emasab, line 1452: Given it comes from the read buffer it's possible it's not null-terminated. Better to use:
```suggestion
                                    RD_KAFKAP_STR_DUP(&SubscribedTopicNames[i]));
```

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,20 +1414,75 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                }
+        }
+
+        new_subscription =
+            rd_list_new(member->subscribed_topic_names &&
```
emasab, line 1457: We don't need to check that `member->subscribed_topic_names` is non-NULL, it should fail in that case. When a member is created if it's NULL it returns. When subscription is updated it's also non-NULL.

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,20 +1414,75 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                                rd_list_cnt(member->subscribed_topic_names) > 0
+                            ? rd_list_cnt(member->subscribed_topic_names)
+                            : 1,
+                        rd_free);
+
+        if (member->subscribed_topic_names) {
```
emasab, line 1463: Same here, you can directly copy it

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,20 +1414,75 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                rd_list_copy_to(new_subscription,
+                                member->subscribed_topic_names,
+                                rd_list_string_copy, NULL);
+        }
+
+        if (member->subscribed_topic_regex) {
```
emasab, line 1469: Given NULL is stored when an empty regex is passed this condition is needed otherwise it would have been 
```suggestion
        if (member->subscribed_topic_regex[0]) {
```

src/rdkafka_mock.c
```diff
@@ -3133,6 +3133,7 @@ static int ut_cgrp_consumer_member_next_assignment0(
         rd_kafkap_str_t MemberId        = {.str = "A", .len = 1};
         rd_kafkap_str_t InstanceId      = {.len = -1};
         rd_kafkap_str_t SubscribedTopic = {.str = topic, .len = strlen(topic)};
+        rd_kafkap_str_t SubscribedTopicRegex = RD_KAFKAP_STR_INITIALIZER;
```
emasab, line 3136: As the unit test is failing, please add and use here:
```C
#define RD_KAFKAP_STR_INITIALIZER_EMPTY                                        \
        { .len = 0, .str = "" }
```

tests/0113-cooperative_rebalance.cpp
```diff
@@ -1998,21 +1994,24 @@ static void n_wildcard() {
...
-                    rebalance_cb1.revoke_call_cnt);
-        TEST_ASSERT(rebalance_cb2.revoke_call_cnt == 1,
-                    "Expecting C_2's revoke_call_cnt to be 1 not %d",
-                    rebalance_cb2.revoke_call_cnt);
+        expected_lost_cnt++;
       }
 
-      TEST_ASSERT(rebalance_cb1.lost_call_cnt == 1,
+      TEST_ASSERT(rebalance_cb1.lost_call_cnt == expected_lost_cnt,
                   "Expecting C_1's lost_call_cnt to be 1 not %d",
```
emasab, line 2012: Here write `expected_lost_cnt`
```suggestion
                  "Expecting C_1's lost_call_cnt to be %d not %d",
```

tests/0113-cooperative_rebalance.cpp
```diff
@@ -1998,21 +1994,24 @@ static void n_wildcard() {
...
                   rebalance_cb1.lost_call_cnt);
-      TEST_ASSERT(rebalance_cb2.lost_call_cnt == 1,
+      TEST_ASSERT(rebalance_cb2.lost_call_cnt == expected_lost_cnt,
                   "Expecting C_2's lost_call_cnt to be 1 not %d",
```
emasab, line 2015: same: `expected_lost_cnt`

tests/0113-cooperative_rebalance.cpp
```diff
@@ -2053,25 +2049,22 @@ static void n_wildcard() {
...
+  TEST_ASSERT(rebalance_cb1.revoke_call_cnt == 2,
+              "Expecting C_1's revoke_call_cnt to be 2 not %d",
+              rebalance_cb1.revoke_call_cnt);
+  TEST_ASSERT(rebalance_cb2.revoke_call_cnt == 2,
+              "Expecting C_2's revoke_call_cnt to be 2 not %d",
+              rebalance_cb2.revoke_call_cnt);
 
-  TEST_ASSERT(rebalance_cb1.lost_call_cnt == 1,
+  TEST_ASSERT(rebalance_cb1.lost_call_cnt == expected_lost_cnt,
               "Expecting C_1's lost_call_cnt to be 1, not %d",
```
emasab, line 2065: same: `expected_lost_cnt`

src/rdkafka_mock_cgrp.c
```diff
@@ -1402,20 +1414,75 @@ rd_kafka_mock_cgrp_consumer_member_find_by_instance_id(
...
+                rd_kafka_mock_cluster_t *mcluster = member->mcgrp->cluster;
+                rd_kafka_mock_topic_t *mtopic;
+                char errstr[1];
+                rd_regex_t *re =
+                    rd_regex_comp(member->subscribed_topic_regex->str, errstr,
```
emasab, line 1474: Here it's ok to use `->str` as it was copied from the protocol string, so it's null terminated.

emasab (changes requested):
Last few things

emasab (approved):
LGTM! Thanks Pranav!

Issue #4907, opened 2024-11-19
segmentation fault from rdkafka_topic.c 


ojktx:
Read the FAQ first: https://github.com/confluentinc/librdkafka/wiki/FAQ

Do NOT create issues for questions, use the discussion forum: https://github.com/confluentinc/librdkafka/discussions

Description
===========
Hello!

I am reporting a segmentation fault.
I will explain briefly because the symptoms and code are clear.
In the rd_kafka_topic_metadata_update() function of rdkafka_topic.c, a NULL reference exception occurs on a variable rktp.

In v2.6.0, the rktp variable is referenced at line 1390.

After analyzing the code for a short time, I found that a null check was missing.

In simple terms,
if (unlikely(!rktp)) {
rd_kafka_dbg(~~~);
return;
}
This code was missing.

I found this problem when I tried to test the problem when the broker was restarted repeatedly.

It is not always the case, but about once in dozens of times, the rktp pointer becomes NULL.
However, since I cannot write an issue at work, I am writing it simply at home without a call stack and screenshots.

Since my company uses librdkafka on at least a thousand servers, I need to fix the issue before I can upgrade the version.

I hope for a quick fix. Thanks.

How to reproduce
================
With librdkafka producer running,

repeat start-stop of brokers.

An issue occurred in 2.6.0.

Checklist
=========

Please provide the following information:

 - [x] librdkafka version (release number or git tag): v2.6.0
 - [x] Apache Kafka version: 3.6.0
 - [ ] librdkafka client configuration: `<REPLACE with e.g., message.timeout.ms=123, auto.reset.offset=earliest, ..>`
 - [x] Operating system: win10 , ubuntu 22.04
 - [ ] Provide logs (with `debug=..` as necessary) from librdkafka
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

thmic:
I encountered the same issue. When I stop the Kafka service, delete all Kafka logs, and then restart the service, many applications using Kafka generate core dumps. When debugging the core files, I found that `rktp` is `0x0`.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000000004a6efb in rd_kafka_topic_metadata_update (rkt=rkt@entry=0x14e474004650, mdt=mdt@entry=0x14e4740041e8, mdit=mdit@entry=0x14e474004208, 
    ts_age=<optimized out>) at rdkafka_topic.c:1384
1384    rdkafka_topic.c: No such file or directory.
[...]

Issue #4884, opened 2024-10-25
rd_kafka_assign() hangs forever when closing consumer


aiquestion:
Description
===========
we enconter a issue when closing the consumer in Rust SDK, it hangs calling rd_kafka_assign() in rebalance_cb.
code is like (using C code)
```
rd_kafka_conf_set_events(conf, RD_KAFKA_EVENT_REBALANCE|RD_KAFKA_EVENT_OFFSET_COMMIT|RD_KAFKA_EVENT_ERROR|RD_KAFKA_EVENT_OAUTHBEARER_TOKEN_REFRESH);
rd_kafka_poll_set_consumer(rk);
rd_kafka_queue_t * queue = rd_kafka_queue_get_consumer(rk);
// should poll here, but we don't do poll to reproduce the bug

// close consumer
rd_kafka_consumer_close_queue(rk, queue);
while (!rd_kafka_consumer_closed(rk)) {
                rd_kafka_event_t *rkev;
                rkev = rd_kafka_queue_poll(queue, 60 * 1000);
                fprintf(stderr, "%% Get event %s \n", rd_kafka_event_name(rkev));
                if (rkev) {
                        if (rd_kafka_event_type(rkev) == RD_KAFKA_EVENT_REBALANCE) {
                                rebalance_cb_event(rk, rkev);
                        }
                }
}

static void rebalance_cb_event(rd_kafka_t *rk,
                               rd_kafka_event_t* rkev) {
        rd_kafka_error_t *error     = NULL;
        rd_kafka_resp_err_t ret_err = RD_KAFKA_RESP_ERR_NO_ERROR;
        switch (rd_kafka_event_error(rkev)) {
        case RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS:
                print_partition_list(stderr, rd_kafka_topic_partition_list_new(0));
                if (!strcmp(rd_kafka_rebalance_protocol(rk), "COOPERATIVE")) {
                        error = rd_kafka_incremental_assign(
                            rk, rd_kafka_topic_partition_list_new(0));
                }
                else {
                        ret_err = rd_kafka_assign(
                            rk, rd_kafka_topic_partition_list_new(0));
                        printf(stderr, "assign ret: %s\n",
                               rd_kafka_err2str(ret_err));
                        sleep(3);
                }
                break;
        case RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS:
[...]

PR #4970, opened 2025-02-18
Code and tests fixes to make the full test suite pass

### LINKED ISSUE #4964: Metadata cache corruption / crash when updating with same topic, different topic ID
When `rd_kafka_metadata_cache_topic_update()` is called with the same topic but a different internal topic ID as an existing metadata cache entry, the metadata cache AVL indexed by topic_id will retain a stale pointer to a now-freed metadata cache entry. This causes corruption and, usually in short order, crashes.

Suppose we start with the metadata cache containing a single entry at address `a1`, with {topic, topic_id} being `{"t1", 1}`. `rd_kafka_metadata_cache_insert()` is called with topic `"t1"` and topic_id `2`. Follow the code path through `rd_kafka_metadata_cache_insert()`:

```
        rkmce = rd_tmpabuf_alloc(&tbuf, sizeof(*rkmce));

        rkmce->rkmce_mtopic = *mtopic;

        rkmce->rkmce_metadata_internal_topic = *metadata_internal_topic;
```

`rkmce` is the new metadata cache entry, with {topic, topic_id} being `{"t1", 2}`.

```
        /* Insert (and replace existing) entry. */
        old = RD_AVL_INSERT(&rk->rk_metadata_cache.rkmc_avl, rkmce,
                            rkmce_avlnode);
```

`old` is `a1`, containing `{"t1", 1}`. `old` has been removed from the AVL indexed by topic; `rkmce` is inserted in the AVL indexed by topic.

```
        /* Insert (and replace existing) entry into the AVL tree sorted
         * by topic id. */
        if (!RD_KAFKA_UUID_IS_ZERO(
                rkmce->rkmce_metadata_internal_topic.topic_id)) {
                /* If topic id isn't zero insert cache entry into this tree */
                old_by_id = RD_AVL_INSERT(&rk->rk_metadata_cache.rkmc_avl_by_id,
                                          rkmce, rkmce_avlnode_by_id);
```
`rkmce` is inserted in the AVL indexed by topic ID. Since there was no previous metadata cache entry with ID `2`, `old_by_id` is null. The AVL indexed by topic ID continues to have an entry for ID `1`, still pointing to `a1`.
```
        } else if (old && !RD_KAFKA_UUID_IS_ZERO(
                              old->rkmce_metadata_internal_topic.topic_id)) {
                /* If it had a topic id, remove it from the tree */
                RD_AVL_REMOVE_ELM(&rk->rk_metadata_cache.rkmc_avl_by_id, old);
        }
```
This is the `else` clause to the preceding `if`, so it does not run.
```
        if (old) {
                /* Delete and free old cache entry */
                rd_kafka_metadata_cache_delete(rk, old, 0);
        }
```

The last parameter to `rd_kafka_metadata_cache_delete()` is `unlink_avl`, whether it is to unlink the cache entry from the AVLs. Since `unlink_avl` is 0, `rd_kafka_metadata_cache_delete()` doesn't update the AVLs and just frees `old` / `a1`.

At this point, the landmine has been armed: the AVL indexed by topic ID has a stale pointer to freed memory `a1`. Depending on when and how that memory is reallocated, a variety of ill-effects can occur. Most of these ill-effects are crashes at inconsistent locations, but usually related to metadata cache operations.

```
        if (old_by_id && old_by_id != old) {
                /* If there was a different cache entry in this tree,
                 * remove and free it. */
                RD_AVL_REMOVE_ELM(&rk->rk_metadata_cache.rkmc_avl, old_by_id);
                rd_kafka_metadata_cache_delete(rk, old_by_id, 0);
        }
```
Since `old_by_id` is null, this code doesn't run.

I believe that the bug is that this:
```
        } else if (old && !RD_KAFKA_UUID_IS_ZERO(
                              old->rkmce_metadata_internal_topic.topic_id)) {
                /* If it had a topic id, remove it from the tree */
                RD_AVL_REMOVE_ELM(&rk->rk_metadata_cache.rkmc_avl_by_id, old);
        }
```
should not be an `else`. Instead it should be:
```
        if (old && !RD_KAFKA_UUID_IS_ZERO(
                old->rkmce_metadata_internal_topic.topic_id) &&
            rd_kafka_Uuid_cmp(rkmce->rkmce_metadata_internal_topic.topic_id,
                old->rkmce_metadata_internal_topic.topic_id) != 0) {
                /* If it had a different topic id, remove it from the tree */
                RD_AVL_REMOVE_ELM(&rk->rk_metadata_cache.rkmc_avl_by_id, old);
        }
```
That is, if `old` had a topic ID which was different from `rkmce`, remove `old` from the AVL indexed by topic ID. If `old` has the same topic ID as `rkmce`, then from the preceding `if` body, `old_by_id` will be equal to `old` and properly handled by the existing code.

This seems to solve the problem in my testing.

This was found using librdkafka v2.6.0, but, as far as I can see, it continues to exist on the most recent main.

I suspect this is also the root cause of issues #4904 and #4907.

emasab:
Includes a task to run the test suite on demand on Semaphore CI.

A description can be found in each commit message.

Closes #4964, #4778, #4907, #4884


tests/run-test-batches.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
```
milindl, line 1: Please add a documentation comment to the top of the file with the different allowed env variables and arguments.

I would also say to move all the globals and the functions above, mix of scripting/functions defs/globals makes it a bit harder to read, especially as this is a useful script to use ourselves too, not just for the CI

milindl (commented):
Leaving partial review comments.

src/rdkafka_int.h
```diff
@@ -409,9 +409,7 @@ struct rd_kafka_s {
         rd_ts_t rk_ts_metadata; /* Timestamp of most recent
                                  * metadata. */
 
-        rd_kafka_metadata_internal_t
-            *rk_full_metadata;       /* Last full metadata. */
-        rd_ts_t rk_ts_full_metadata; /* Timestamp of .. */
+        rd_ts_t rk_ts_full_metadata;                      /* Timestamp of .. */
```
milindl, line 412: nit
```
/* Timestamp of most recent full metadata */
```

src/rdkafka_request.c
```diff
@@ -2619,7 +2619,12 @@ static void rd_kafka_handle_Metadata(rd_kafka_t *rk,
                                       RD_KAFKA_ERR_ACTION_END);
 
         if (actions & RD_KAFKA_ERR_ACTION_RETRY) {
-                if (rd_kafka_buf_retry(rkb, request))
+                /* In case it's a brokers full refresh call,
+                 * avoid retrying it on this same broker.
+                 * This is to prevent client is hung
+                 * until it can connect to this broker again. */
+                if (!request->rkbuf_u.Metadata.decr &&
```
milindl, line 2626: Wouldn't we need to hold 
```
Metadata.decr_lock
```
while reading it?

milindl (commented):
Another partial review, 2 commits are still pending review.

tests/0082-fetch_max_bytes.cpp
```diff
@@ -87,6 +87,9 @@ static void do_test_fetch_max_bytes(void) {
     Test::Fail("Failed to create KafkaConsumer: " + errstr);
   delete conf;
 
+  /* For next consumer */
+  test_wait_topic_exists(c->c_ptr(), topic.c_str(), 5000);
```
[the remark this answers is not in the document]
emasab: It similar, at the moment, we may change how we wait a topic exists by using JMX requests.

src/rdkafka_request.c
emasab: No, it's only needed if we need to decrease the integer it points to, that can be `rkmc_full_brokers_sent` or `rkmc_full_topics_sent`. Lock for it will be acquired when the buffer is destroyed.

tests/0093-holb.c
```diff
@@ -116,7 +116,12 @@ int main_0093_holb_consumer(int argc, char **argv) {
 
         test_conf_set(conf, "session.timeout.ms", "6000");
         test_conf_set(conf, "max.poll.interval.ms", "20000");
-        test_conf_set(conf, "socket.timeout.ms", "3000");
+        /* Socket timeout must be greater than
```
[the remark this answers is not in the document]
emasab: I cannot reproduce this test failure anymore. I'll remove the test change.

milindl (approved):
Reviewed individual commits and fixups. Please merge with all commits intact. Thank you for the fixes!

milindl (approved):
I reviewed the newly added commits.

Issue #4953, opened 2025-01-28
Sporadic message duplication during leader transition with idempotent producer


marcin-krystianc:
# Description
It is possible to get duplicated messages in a single topic-partition. It happens despite configuring the producer for exactly-once delivery. It occurs sporadicly during the partition leader change, where for some reason the producer epoch is also being incremented. It is not trivial to reproduce, but it is fairly doable. (I can provide repro-steps, but I think it is not helpful as the attached pcap file contains enough details to analyze this issue).
The essential configuration bits for the producer are:
```
--replication-factor=3
--min-isr=2
--config request.required.acks=-1
--config enable.idempotence=true
--config max.in.flight.requests.per.connection=5
```

# Details
Below is the excerpt from the [captured packets](https://github.com/user-attachments/files/18571344/capture_topic0040_0188_filtered.zip) demonstrating the problem.
The producer app was writing a sequence of numbers (strictly incrementing) into `my-topic-0188`. Because of the topic-partition leadership change, the values `0x0a90` and `0x0a91` were written twice. This breaks exactly-once delivery guarantees and can cause data inconsistencies in applications relying on message deduplication.

It is unclear why these values were written twice as there is only one successful confirmation of writing them.
What is also interesting, the response from frame 1566024, has a valid offset but also a non-zero error code.

### Confluent.Kafka(Librdkafka) version
```
Confluent.Kafka, Version=2.6.1.0, Culture=neutral, PublicKeyToken=12c514ca49093d1em
librdkafka Version: 2.6.1 (20601FF)
```

### Produce requests

<details>
  <summary><code>1560411, 15:37:11.568610, 192.168.65.3:65494->3.70.194.188:40001 - Kafka Produce v10 Request</code></summary>

```
Kafka (Produce v10 Request)
    Length: 129
    API Key: Produce (0)
    API Version: 10
    Correlation ID: 3561
    Client ID: rdkafka
    Tagged fields
    Transactional ID: [ Null ]
    Required Acks: Full ISR (-1)
    Timeout: 30000
    Topic (Name=my-topic-0188)
        Topic Name: my-topic-0188
        Partition (ID=0)
            Partition ID: 0
            Message Set
                Record Batch
                    Offset: 0
                    Message Size: 68
                    Leader Epoch: 0
                    Magic Byte: 2
                    CRC32: 0x1ee6e86f
                    .... .... .... .000 = Compression Codec: None (0)
[...]

PR #4972, opened 2025-02-19
Fix: Avoid unnecessary producer epoch bumps

marcin-krystianc:
Fixes: #4953
 
 When the producer is idempotent and `max.in.flight.requests.per.connection` is set to a value between 2 and 5, it's normal to receive `OUT_OF_ORDER_SEQUENCE_NUMBER` produce responses for requests R2 through R5 when the R1 failed for any other reason.
 
Bumping the producer epoch in this scenario violates the "exactly-once" guarantees. Therefore, we believe that it's unnecessary to bump the producer's epoch; re-enqueuing the messages is sufficient.
 
The same "retry, but don't bump producer epoch" behavior is implemented in the Java client: https://github.com/apache/kafka/blob/a6a588fbed9982598377060c63f94ee6184b4295/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L1015-L1016

### DIFF
--- a/src/rdkafka_request.c
+++ b/src/rdkafka_request.c
@@ -3902,16 +3902,15 @@ rd_kafka_handle_idempotent_Produce_error(rd_kafka_broker_t *rkb,
                         perr->update_next_err = rd_true;
 
                 } else if (r > 0) {
-                        /* R2 failed:
+                        /* R2..R5 failed:
                          * With max.in.flight > 1 we can have a situation
                          * where the first request in-flight (R1) to the broker
                          * fails, which causes the sub-sequent requests
                          * that are in-flight to have a non-sequential
                          * sequence number and thus fail.
-                         * But these sub-sequent requests (R2 to R4) are not at
-                         * the risk of being duplicated so we bump the epoch and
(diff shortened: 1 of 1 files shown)


emasab:
Thanks for the contribution @marcin-krystianc  I just need to change a test that is expecting a different behaviour. It's to note that the exactly-once guarantees are only partial for the idempotent producer, see [this comment](https://github.com/confluentinc/librdkafka/issues/4953#issuecomment-2887059722).

marcin-krystianc:
> Thanks for the contribution @marcin-krystianc I just need to change a test that is expecting a different behaviour. It's to note that the exactly-once guarantees are only partial for the idempotent producer, see [this comment](https://github.com/confluentinc/librdkafka/issues/4953#issuecomment-2887059722).

Thanks. Are there any other tests worth adding? I'm happy to do it, but I need some guidance.

emasab:
@marcin-krystianc at the moment I can't think to other tests related to this error that are different from the existing ones. In theory it has a meaning and in practice it's done by Java client as well so it seems like a safer behaviour.
I request a second review as well.
Thanks again!

PR #4986, opened 2025-03-07
Change tcp_nodelay config to be true

milindl:
Currently, we use Nagle's algorithm, which doesn't make much sense as we're already implementing batching behaviour with linger.ms. In cases the user wants to optimize the latency and sets linger.ms = 0, it will still be delayed, which is undesirable/buggy behaviour.


marcin-krystianc:
I fully support this change. I was surprised to discover that the current default setting utilizes Nagle's algorithm. I only became aware of this while investigating some unexpected latency spikes in our benchmarks, which were resolved upon disabling Nagle's algorithm (socket.nagle.disable=true).

emasab (approved):
Thanks @milindl ! We decided to ship it as a patch and not in a major release because it's only giving improvements if QUICKACK isn't enabled on the other side, and this last one is a flag that needs to be reapplied after every read to the socket. Even if any broker implementation was setting QUICKACK after every read and this change was causing increased acks, a small increase in network traffic would be better that an increase in latency.

PR #4990, opened 2025-03-11
Remove installation script with OpenSSL 1.1.1k

emasab:
as it's not used anymore. Was used for
AppVeyor CI builds.


milindl (approved):
Thanks, lgtm.

PR #4991, opened 2025-03-11
Fix SSL verification fail bug when Kafka endpoint is a FQDN with a do…

mk8:
I'm not able to connect to Kafka server when the BOOTSTRAP_BROKER is a FQDN with a dot at the end.
Different Kafka implementation provide endpoints with a dot at the end so you can use a BOOTSTRAP_BROKER without dot at the end but you ad not able to connect to the topic anyway.

### DIFF
--- a/src/rdkafka_transport.c
+++ b/src/rdkafka_transport.c
@@ -581,6 +581,11 @@ void rd_kafka_transport_post_connect_setup(rd_kafka_transport_t *rktrans) {
  * Locality: broker thread
  */
 static void rd_kafka_transport_connected(rd_kafka_transport_t *rktrans) {
+        // If the domain name is an FQDN, remove the trailing dot
+        char* dot = strrchr(rktrans->rktrans_rkb->rkb_nodename, '.');
+        if (dot != NULL && dot[1] == ':')
+                memmove(dot, &dot[1], strlen(&dot[1]) + 1);
+
         rd_kafka_broker_t *rkb = rktrans->rktrans_rkb;
 
         rd_rkb_dbg(


src/rdkafka_transport.c
```diff
@@ -581,6 +581,11 @@ void rd_kafka_transport_post_connect_setup(rd_kafka_transport_t *rktrans) {
  * Locality: broker thread
  */
 static void rd_kafka_transport_connected(rd_kafka_transport_t *rktrans) {
+        // If the domain name is an FQDN, remove the trailing dot
+        char* dot = strrchr(rktrans->rktrans_rkb->rkb_nodename, '.');
+        if (dot != NULL && dot[1] == ':')
+                memmove(dot, &dot[1], strlen(&dot[1]) + 1);
```
joejulian, line 587: What if there's no port notation? Can there be a host with no port designated? Is this a good idea?
```suggestion
        if (dot != NULL && (dot[1] == ':' || dot[1] == '\0'))
                memmove(dot, &dot[1], strlen(&dot[1]) + 1);
```

david-yu:
Hi team based on the discussion in https://github.com/openssl/openssl/issues/11560#issuecomment-1631520748 would it be reasonable to accept the change in behavior for ignoring the trailing . in the FQDN? Also see this here: https://github.com/confluentinc/librdkafka/issues/4348. Although folks have worked around this already it does seem like the proper behavior to follow.

joejulian (approved):

david-yu:
@emasab Any chance we could get your take on this PR? This seems like valid behavior to have and check for in the client and doesn't sound like would break existing behavior.

src/rdkafka_transport.c
emasab: It shouldn't modify the `rkb_nodename` but the change should be done in `rd_kafka_transport_ssl_set_endpoint_id` after copying the string as the name can be resolved again on `rd_kafka_broker_connect`

mk8:
Hi

I tried to move the dot removal in rd_kafka_transport_ssl_set_endpoint_id
function.

Using this patch it seems that everything work fine but I need someone else
test it:

diff --git a/src/rdkafka_ssl.c b/src/rdkafka_ssl.c
index 417da818..ae546af2 100644
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -461,6 +461,9 @@ static int
rd_kafka_transport_ssl_set_endpoint_id(rd_kafka_transport_t *rktrans,
         /* Remove ":9092" port suffix from nodename */
         if ((t = strrchr(name, ':')))
                 *t = '\0';
+        if (name[strlen(name) - 1] == '.') {
+                name[strlen(name) - 1] = '\0';
+        }

 #if (OPENSSL_VERSION_NUMBER >= 0x0090806fL) && !defined(OPENSSL_NO_TLSEXT)
         /* If non-numerical hostname, send it for SNI */

PR #5001, opened 2025-03-24
NONJAVACLI-3632: Non-java clients identity pool extension comma parsing

PratRanj07:
This PR intends to implement CSV format parsing for strings related to SASL OAuthbearer Extensions. Jira Ticket:  [NONJAVACLI-3632](https://confluentinc.atlassian.net/browse/NONJAVACLI-3632)

### DIFF
--- a/src/rdstring.c
+++ b/src/rdstring.c
@@ -554,6 +554,135 @@ char **rd_string_split(const char *input,
 }
 
 /**
+ * @brief Split a character-separated string into an array, supporting quoted
+ * fields.
+ *
+ * @remark This function is designed to handle cases where fields may be
+ * enclosed in double quotes (`"`), allowing separators within quoted fields to
+ * be treated as part of the field rather than as delimiters.
+ *
+ * @param input Input string to parse.
+ * @param sep The separator character (typically ',').
+ * @param skip_empty Do not include empty fields in the output array.
+ * @param cntp Will be set to the number of elements in the array.
+ *
+ * Supports quoted fields using `"`. Escaped characters (e.g., `\n`, `\t`) are
+ * also handled. Quoted fields can contain separators, and the quotes will be
+ * removed in the output. The array and the array elements will be allocated
+ * together and must be freed with a single rd_free(array) call. The array
+ * elements are copied, and any escape sequences or quotes are removed.
+ *
+ * @returns The parsed fields in an array. The number of elements in the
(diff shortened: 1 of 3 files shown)


src/rdstring.c
```diff
@@ -553,6 +553,135 @@ char **rd_string_split(const char *input,
...
+ * Supports quoted fields using `"`. Escaped characters (e.g., `\n`, `\t`) are
+ * also handled. Quoted fields can contain separators, and the quotes will be
+ * removed in the output. The array and the array elements will be allocated
+ * together and must be freed with a single rd_free(array) call. The array
+ * elements are copied, and any escape sequences or quotes are removed.
+ *
+ * @returns The parsed fields in an array. The number of elements in the
+ *          array is returned in \p cntp.
+ */
+char **rd_string_split_csv(const char *input,
```
anchitj, line 578: I think we should return error code and also handle the cases where input string is invalid, like `"` started but no ending `"` present.

src/rdstring.c
```diff
@@ -632,6 +761,68 @@ static int ut_string_split(void) {
...
+                {"extension1:extension1val",
+                 "extension2:extension2val1,extension2val2"},
+            },
+            {
+                "extension1:extension1val,extension2:extension2val1",
+                ',',
+                rd_true,
+                2,
+                {"extension1:extension1val", "extension2:extension2val1"},
+            },
```
anchitj, line 790: Can we also add other edge cases like empty string, string with just a single k:v pair, and using escape sequences. And also test for invalid strings.

anchitj (changes requested):

anchitj:
Thanks @PratRanj07. Some minor comments, but overall looks good. Let's verify manually with a cluster too.

src/rdstring.c
```diff
@@ -553,6 +553,142 @@ char **rd_string_split(const char *input,
...
+        char *p;
+        char **arr;
+        size_t inputlen;
+        size_t i            = 0;
+        size_t elen         = 0;
+        rd_bool_t in_quotes = rd_false;
+
+        *cntp = 0;
+
+        // Count approximate field count (ignore sep inside quotes)
```
anchitj, line 596: Please place comments in /* */ blocks

anchitj:
Also verify if the extensions have spaces can they be parsed correctly and interpreted by the broker. Like `identityPoolId=p1,       p2`

PR #5009, opened 2025-03-27
[KIP-848] Handle topic authorization failure on consumer group heartbeat

emasab:
in this case the heartbeat is retried with given interval or with exponential backoff if it's the first heartbeat. An error is returned
to the user in this case and isn't sent again for the given heartbeat interval or at least 3 seconds.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -3061,7 +3061,7 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
         const int log_decode_errors = LOG_ERR;
         int16_t error_code          = 0;
         int actions                 = 0;
-        rd_kafkap_str_t error_str;
+        rd_kafkap_str_t error_str   = RD_KAFKAP_STR_INITIALIZER_EMPTY;
         rd_kafkap_str_t member_id;
         int32_t member_epoch;
         int32_t heartbeat_interval_ms;
@@ -3324,15 +3324,20 @@ err:
         if (actions & RD_KAFKA_ERR_ACTION_SPECIAL &&
             rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION) {
                 rd_ts_t min_error_interval =
-                    RD_MAX(rkcg->rkcg_heartbeat_intvl_ms * 1000, 3000000);
+                    RD_MAX(rkcg->rkcg_heartbeat_intvl_ms * 1000,
+                           /* default group.consumer.heartbeat.interval.ms */
+                           5000000);
                 if (rkcg->rkcg_last_err != err ||
                     (rd_clock() >
                      rkcg->rkcg_ts_last_err + min_error_interval)) {
                         rd_kafka_cgrp_set_last_err(rkcg, err);
                         rd_kafka_consumer_err(
                             rkcg->rkcg_q, rd_kafka_broker_id(rkb), err, 0, NULL,
-                            NULL, err, "ConsumerGroupHeartbeat failed: %s",
-                            rd_kafka_err2str(err));
+                            NULL, err,
+                            "ConsumerGroupHeartbeat failed: %s%s%.*s",
+                            rd_kafka_err2str(err),
+                            RD_KAFKAP_STR_LEN(&error_str) ? ": " : "",
+                            RD_KAFKAP_STR_PR(&error_str));
                 }
         }
 
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -3321,8 +3321,7 @@ err:
                     rkcg, "coordinator query");
         }
 
-        if (actions & RD_KAFKA_ERR_ACTION_SPECIAL &&
-            rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION) {
+        if (actions & RD_KAFKA_ERR_ACTION_SPECIAL) {
                 rd_ts_t min_error_interval =
                     RD_MAX(rkcg->rkcg_heartbeat_intvl_ms * 1000,
                            /* default group.consumer.heartbeat.interval.ms */
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -370,6 +370,36 @@ static int rd_kafka_cgrp_set_state(rd_kafka_cgrp_t *rkcg, int state) {
         return 1;
 }
 
+/**
+ * @brief Set the cgrp last error and current timestamp
+ *        as last error timestamp.
+ */
+static void rd_kafka_cgrp_set_last_err(rd_kafka_cgrp_t *rkcg,
+                                       rd_kafka_resp_err_t rkcg_last_err) {
+        rkcg->rkcg_last_err    = rkcg_last_err;
+        rkcg->rkcg_ts_last_err = rd_clock();
+}
+
+/**
+ * @brief Clears cgrp last error and its timestamp.
+ */
+static void rd_kafka_cgrp_clear_last_err(rd_kafka_cgrp_t *rkcg) {
+        rkcg->rkcg_last_err    = RD_KAFKA_RESP_ERR_NO_ERROR;
+        rkcg->rkcg_ts_last_err = 0;
+}
+
+/**
+ * @brief Clears cgrp last error if it's an heartbeat related error like
+ *        a topic authorization failed one.
+ */
+static void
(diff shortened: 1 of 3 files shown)


src/rdkafka_cgrp.c
```diff
@@ -3205,8 +3206,13 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
                 actions = RD_KAFKA_ERR_ACTION_FATAL;
                 break;
         default:
-                actions = rd_kafka_err_action(rkb, err, request,
-                                              RD_KAFKA_ERR_ACTION_END);
+                actions = rd_kafka_err_action(
+                    rkb, err, request,
+
+                    RD_KAFKA_ERR_ACTION_SPECIAL,
+                    RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED,
```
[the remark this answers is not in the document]
emasab: As discussed we keep the `GROUP_AUTHORIZATION_FAILED` error fatal as in Java client, to have the same contract.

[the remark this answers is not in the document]
pranavrth: Also that `group.id` is set in the configuration but subscribed topics can be added dynamically. So its better to fail at `GROUP_AUTHORIZATION_FAILED` error and let consumer run for `TOPIC_AUTHORIZATION_ERROR`.

src/rdkafka_cgrp.c
```diff
@@ -3285,6 +3291,16 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
                     rkcg, "coordinator query");
         }
 
+        if (actions & RD_KAFKA_ERR_ACTION_SPECIAL &&
+            rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION &&
+            !(rkcg->rkcg_consumer_flags &
+              RD_KAFKA_CGRP_CONSUMER_F_ERROR_SENT)) {
+                rkcg->rkcg_consumer_flags |=
+                    RD_KAFKA_CGRP_CONSUMER_F_ERROR_SENT;
+                rd_kafka_consumer_err(
```
pranavrth, line 3300: Can we change this approach of sending error only once per subscription until the error is reset to continuously send the error for with some interval like we send in case of Disconnection.

- Its easy to miss a single log.
- User will never know if the problem is fixed or not until some message is received.

pranavrth (changes requested):

src/rdkafka_cgrp.c
```diff
@@ -5321,6 +5372,7 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
                         rkcg->rkcg_consumer_flags |=
                             RD_KAFKA_CGRP_CONSUMER_F_SUBSCRIBED_ONCE |
                             RD_KAFKA_CGRP_CONSUMER_F_SEND_NEW_SUBSCRIPTION;
+                        rd_kafka_cgrp_maybe_clear_heartbeat_failed_err(rkcg);
```
pranavrth, line 5375: This is more specific to `TopicAuthorizationError`, We should not clear it here. If we add a new error there, this is not the correct place to clear the error. Moreover, if there is a new subscription set, we send the new HB and then if there is no issue, the error will be cleared there.

src/rdkafka_cgrp.c
```diff
@@ -3285,6 +3321,21 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
...
+                rd_ts_t min_error_interval =
+                    RD_MAX(rkcg->rkcg_heartbeat_intvl_ms * 1000, 3000000);
+                if (rkcg->rkcg_last_err != err ||
+                    (rd_clock() >
+                     rkcg->rkcg_ts_last_err + min_error_interval)) {
+                        rd_kafka_cgrp_set_last_err(rkcg, err);
+                        rd_kafka_consumer_err(
+                            rkcg->rkcg_q, rd_kafka_broker_id(rkb), err, 0, NULL,
+                            NULL, err, "ConsumerGroupHeartbeat failed: %s",
+                            rd_kafka_err2str(err));
```
pranavrth, line 3335: It would be good if we can add the error message as well there. We are extracting `error_str` from the response but not using it anywhere.

src/rdkafka_cgrp.c
```diff
@@ -3285,6 +3321,21 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
                     rkcg, "coordinator query");
         }
 
+        if (actions & RD_KAFKA_ERR_ACTION_SPECIAL &&
+            rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION) {
```
pranavrth, line 3325: - We send HB only if we have this flag set, so we can skip this check. There might be a case where we unsubscribed and then sending this error once. I think that should be fine. If not, then we can add this check with error check case.
- This is also more specific to TopicAuthorizationError.
- This won't be correct if we add a new Error here.

src/rdkafka_cgrp.c
```diff
@@ -3185,6 +3215,7 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
             ~RD_KAFKA_CGRP_CONSUMER_F_SENDING_NEW_SUBSCRIPTION &
             ~RD_KAFKA_CGRP_CONSUMER_F_SEND_FULL_REQUEST &
             ~RD_KAFKA_CGRP_CONSUMER_F_SENDING_ACK;
+        rd_kafka_cgrp_maybe_clear_heartbeat_failed_err(rkcg);
```
pranavrth, line 3218: We should also clear this if we got some other error as well.

pranavrth (changes requested):

src/rdkafka_cgrp.c
```diff
@@ -3285,6 +3321,21 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
                     rkcg, "coordinator query");
         }
 
+        if (actions & RD_KAFKA_ERR_ACTION_SPECIAL &&
+            rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION) {
+                rd_ts_t min_error_interval =
+                    RD_MAX(rkcg->rkcg_heartbeat_intvl_ms * 1000, 3000000);
```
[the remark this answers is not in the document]
emasab: ok I keep it the same as the default HB interval: 5s

src/rdkafka_cgrp.c
emasab: This just ensures that when you change the subscription you receive a new error on first HB after the subscription if it's still on error and don't have to wait for at least 5 seconds. That applies to other kind of future errors as well.

src/rdkafka_cgrp.c
emasab: In case we have set a previous error in `FindCoordinator` response we should not clear it.

src/rdkafka_cgrp.c
emasab: > This is also more specific to TopicAuthorizationError.

In case we'll have more errors that set this flag and we need to do something different we can use a switch on the error code

src/rdkafka_cgrp.c
emasab: Tried but the error string is empty for this error, maybe it could be added in future, handled the case where it's empty.

tests/0109-auto_create_topics.cpp
```diff
@@ -107,34 +118,40 @@ static void do_test_consumer(bool allow_auto_create_topics,
...
+  } else {
+    topics.push_back(topic_exists);
+  }
+
+  if (test_consumer_group_protocol_classic()) {
     /* If the subscription contains at least one wildcard/regex
      * then no auto topic creation will take place (since the consumer
      * requests all topics in metadata, and not specific ones, thus
      * not triggering topic auto creation).
      * We need to handle the expected error cases accordingly. */
```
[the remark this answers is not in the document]
emasab: This is inside the `if` for the classic part. In the new protocol there's no topic autocreation on the consumer.

src/rdkafka_cgrp.c
pranavrth: Add a comment.

emasab: ok removed the condition as in that case the error in the response is converted to `OUTDATED`

pranavrth (approved):
LGTM!.

Issue #4629, opened 2024-02-29
coop sticky algo on large partition numbers


ericwuseattle:
Hey there, 

There are 2 issues, I noticed on kafka coop sticky mode.
1. The hard code on partition_cnt inside rd_kafka_sticky_assignor_assign_cb
https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_sticky_assignor.c#L1834

2. On 3K partitions, it's working without issue, but if I increase the partition to 6K with fresh topic(I mean recreate the topic as new one). Have to increase the session.timeout.ms=10000 and max.poll.interval.ms=10000 from 3s to 10s to make it working.
Otherwise will get kickout out from grp
Broker logs: 
Member XXX-6F958DDF5F-CDXRQ~-0793c679-d5ef-4753-9056-7da314e1415b in group XXX-TOPIC-NAME-XXX has failed, removing it from the group (kafka.coordinator.group.GroupCoordinator). 

I'm not sure what's causing the timeout, but I'm sure we keep calling kafka poll in a timer infinitely. So increase into 10s it's working without any issue. Then I'm trying on 15K partitions with 10s, no lucky, same error.

How to reproduce
On rdkafka 2.3.0
big partition count.
6K partitions with 3s session timeout
or
15K partitions with 10s session timeout

Additionally we were using customize assignment(consistent hash), it's working without any issue with 15K partitions and 10s session timeout.

Thanks in advance,
Eric

ericwuseattle:
Any thoughts on this problem? Are more details needed?

emasab:
@ericwuseattle could you send some logs with `debug=all` . It's possible that it's needed to increase those values for a rebalance with those many partitions, but from the logs we can see where most of the time goes.

ericwuseattle:
unfortunately I do not have the evn was set up to testing in my hand right now, have you checked the hard code of partition count in the code, if we could fix that part first then I'll find time to retry it.

/* FIXME: Let the cgrp pass the actual eligible partition count */
size_t partition_cnt = member_cnt * 10; /* FIXME */

https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_sticky_assignor.c#L1834

emasab:
@ericwuseattle 
```
/* FIXME: Let the cgrp pass the actual eligible partition count /
size_t partition_cnt = member_cnt * 10; / FIXME */
```
[...]

PR #5014, opened 2025-04-02
create new topic-partition list comparison method based on hashing

lucas-sonnabend:
the new method is much faster than the old, O(Na + Nb) compared to the older one, which is O(Na * Nb)

Background:
The old one was causing problems with partition counts ~3600 in the cooperative-sticky partition assignment.
The group leader ended up being kicked out of the group because he spent too much time calculating the partition assignment, and didn't sent any heartbeat inbetween.

reducing the complexity of this pushes the runtime down. Concrete example with 3600 partitions and 1800 consumers before: ~240s
after: ~3s

fixes problem in https://github.com/confluentinc/librdkafka/issues/4629

I have tried to stick to the styles, but ran into problems with getting the right version of clang-format (10 doesn't seem to be available via homebrew)

I have have also run the unit tests, they pass.

I am currently running 0113-cooperative-rebalance.cpp test case.

I'm not sure if I should be completely replacing the old comparison function, but happy to do so based on comments


lucas-sonnabend:
so, I'm trying to run the 0113 test, but it keeps timing out locally, both with the changes and on main.
I'm also running a kafka cluster via docker-compose for it rather than the recommended way

PR #5017, opened 2025-04-03
[KIP-714] Fix idle ratio calculation for not forwarded queues

emasab:
Fix for devel assert "rkq->rkq_ts_last_poll_end >= rkq->rkq_ts_last_poll_start" in test 0056.

Partition queues can be not forwarded to the main poll queue so these timestamps must be per-queue instead of per-instance. We avoid checking if app polled should be called in case of internal poll calls where we're sure it's not a consume call, we also avoid checking it if "app polled" was already called by a dedicated consume function.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3275,12 +3275,12 @@ static rd_kafka_op_res_t rd_kafka_consume_callback0(
         struct consume_ctx ctx = {.consume_cb = consume_cb, .opaque = opaque};
         rd_kafka_op_res_t res;
 
-        rd_kafka_app_poll_start(rkq->rkq_rk, 0, timeout_ms);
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
...
+## Fixes
+
+### Telemetry fixes
+
+* Issues: #
+  Fix for poll ratio calculation in case the queues are forwarded.
+  Poll ratio is now calculated per-queue instead of per-instance and
+  it allows to avoid calcultation problems linked to using the same
+  field.
+  Happens since 2.6.0 (#).
```
anchitj: Do we have the issue link?

anchitj (approved):

anchitj (approved):

anchitj (approved):

PR #5023, opened 2025-04-07
Script to update max RPC versions supported in AK

emasab:


packaging/tools/update_rpcs_max_versions.py
```diff
@@ -0,0 +1,87 @@
...
+        text = "".join([line for line in text
+                        if '#' not in line and '//' not in line])
+        json_object = json.loads(text)
+        max_version = json_object["validVersions"].split("-")[-1]
+        print('| ', end='')
+        print(str(api_num).ljust(max_first_column), end='')
+        print('| ', end='')
+        print(api.ljust(max_second_column), end='')
+        print('| ', end='')
+        print(str(max_version).ljust(max_third_column) + '|')
```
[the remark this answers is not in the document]
emasab: Not really, it's better to fail the whole script in that case instead of just skipping the file

packaging/tools/update_rpcs_max_versions.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+import json
+import sys
+import re
+
+# This script updates the Apache Kafka RPCs max versions.
+# It reads the input from stdin, which should be a table
+# looking like the first three columns of the table in `INTRODUCTION.md`.
+# Should look like this (without the first space after the comment):
+# | 0       | Produce                       | 12         |
```
anchitj, line 10: I think it might be better to ask to copy paste all the columns, and the script handles it by modifying only the first three. 
This would reduce manual work first in removing the last column and then again adding it back when pasting in INTRODUCTION.md

emasab: yeah, nice addition. Done

anchitj (approved):

PR #5024, opened 2025-04-08
fixup: Use same strategy for updating cache partition metadata

emasab:
remove topic id check given it's the zero UUID when topic doesn't exist and it invalidates the new check that avoids the transition exists -> not exists before `metadata.propagation.max.ms`.

For `rd_kafka_topic_set_exists` it's fine to keep it as we're comparing the previous UUID with new one to see if it's a deleted topic that is re-appearing or a completely new one.


milindl (approved):
lgtm for both the changes, discussed internally.

PR #5026, opened 2025-04-08
[KIP-848] Tests for dev_kip848_fix_fast_subscribe_or_unsubscribe

emasab:


tests/0045-subscribe_update.c
```diff
@@ -819,6 +819,94 @@ static void do_test_resubscribe_with_regex() {
...
+        TEST_CALL_ERR__(test_CreateTopics_simple(
+            rk, NULL, topics, RD_ARRAY_SIZE(topics), partition_cnt, NULL));
+        test_wait_topic_exists(rk, topics[0], 5000);
+        /* Give the cluster some more time to propagate metadata
+         * for 100 topics */
+        rd_sleep(1);
+
+        RD_ARRAY_FOREACH(topic, topics, i) {
+                int j, k;
+                rd_kafka_topic_partition_list_t *subscription =
```
pranavrth, line 876: Add a new variable to calculate `subcription_size` which is `(int)RD_MIN(max_subscription_size, RD_ARRAY_SIZE(topics) - i)`.

tests/test.c
```diff
@@ -3093,6 +3095,99 @@ void test_consumer_verify_assignment0(const char *func,
...
+                for (i = 0; i < assignment->cnt; i++) {
+                        if (strcmp(assignment->elems[i].topic,
+                                   expected_assignment->elems[i].topic) ||
+                            assignment->elems[i].partition !=
+                                expected_assignment->elems[i].partition) {
+                                ret = rd_false;
+                                goto done;
+                        }
+                }
+        }
```
pranavrth, line 3138: `goto done` or `break`. Whatever you want.

```suggestion
        if (assignment->cnt != expected_assignment->cnt) {
                ret = rd_false;
                goto done;
        }

        for (i = 0; i < assignment->cnt; i++) {
                if (strcmp(assignment->elems[i].topic,
                            expected_assignment->elems[i].topic) ||
                    assignment->elems[i].partition !=
                        expected_assignment->elems[i].partition) {
                        ret = rd_false;
                        break;
                }
        }
```

tests/0045-subscribe_update.c
```diff
@@ -862,6 +950,12 @@ int main_0045_resubscribe_with_regex(int argc, char **argv) {
         return 0;
 }
 
+int main_0045_many_updates(int argc, char **argv) {
```
pranavrth, line 953:
```suggestion
int main_0045_subscribe_many_updates(int argc, char **argv) {
```

pranavrth (changes requested):

tests/test.c
```diff
@@ -3093,6 +3095,99 @@ void test_consumer_verify_assignment0(const char *func,
...
+                verified =
+                    test_consumer_verify_assignment_topic_partition_list0(
+                        func, line, rk, expected_assignment);
+                if (!verified) {
+                        if (do_poll)
+                                test_consumer_poll_once(rk, NULL, 100);
+                        else
+                                rd_usleep(100 * 1000, NULL);
+                }
+        } while (test_clock() < end && !verified);
```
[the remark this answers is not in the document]
emasab: That way we still need to call 
```c
                verified =
	                    test_consumer_verify_assignment_topic_partition_list0(
	                        func, line, rk, expected_assignment);
``` 
before of after the loop in case we call it with 0 timeout, I use the break but keep the `do ... while`

tests/test.c
```diff
@@ -3093,6 +3095,99 @@ void test_consumer_verify_assignment0(const char *func,
...
+
+        if (!verified) {
+                TEST_FAIL(
```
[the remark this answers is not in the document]
emasab: not needed, `TEST_FAIL` either calls `thrd_exit` or `assert`

tests/0045-subscribe_update.c
```diff
@@ -819,6 +819,94 @@ static void do_test_resubscribe_with_regex() {
...
+         * for 100 topics */
+        rd_sleep(1);
+
+        RD_ARRAY_FOREACH(topic, topics, i) {
+                int j, k;
+                rd_kafka_topic_partition_list_t *subscription =
+                    rd_kafka_topic_partition_list_new(subscription_size);
+                RD_IF_FREE(expected_assignment,
+                           rd_kafka_topic_partition_list_destroy);
+                expected_assignment = rd_kafka_topic_partition_list_new(
```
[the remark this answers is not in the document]
emasab: I used a boolean to avoid adding a new loop on the topics

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,108 @@ static void do_test_resubscribe_with_regex() {
...
+        rd_kafka_t *rk;
+        rd_kafka_conf_t *conf;
+        const int max_subscription_size = 5, partition_cnt = 4;
+        rd_kafka_topic_partition_list_t *expected_assignment = NULL;
+
+        SUB_TEST("%s", with_rebalance_cb ? "with rebalance callback"
+                                         : "without rebalance callback");
+
+        RD_ARRAY_FOREACH(topic, topics, i) {
+                topics[i] = rd_strdup(test_mk_topic_name("topic", 1));
```
pranavrth: Add `i` in the topic name as well.

tests/0045-subscribe_update.c
```diff
@@ -819,6 +819,94 @@ static void do_test_resubscribe_with_regex() {
...
+                rd_kafka_topic_partition_list_t *subscription =
+                    rd_kafka_topic_partition_list_new(subscription_size);
+                RD_IF_FREE(expected_assignment,
+                           rd_kafka_topic_partition_list_destroy);
+                expected_assignment = rd_kafka_topic_partition_list_new(
+                    subscription_size * partition_cnt);
+                for (j = i; j < (int)RD_MIN(i + subscription_size,
+                                            RD_ARRAY_SIZE(topics));
+                     j++) {
+                        topic = topics[j];
```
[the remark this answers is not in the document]
pranavrth: `topic` is getting overridden which is first assigned in above FOREACH macro. We should remove the usage of this topic from here.

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,108 @@ static void do_test_resubscribe_with_regex() {
...
+                                   rd_kafka_topic_partition_list_destroy);
+                        expected_assignment = rd_kafka_topic_partition_list_new(
+                            expected_assignment_cnt);
+                }
+
+
+                for (j = i; j < i + subscription_size; j++) {
+                        topic = topics[j];
+                        rd_kafka_topic_partition_list_add(
+                            subscription, topic, RD_KAFKA_PARTITION_UA);
```
pranavrth: use `topic[j]` instead.

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,108 @@ static void do_test_resubscribe_with_regex() {
...
+
+        TEST_SAY("Creating %d topics\n", TOPIC_CNT);
+        TEST_CALL_ERR__(test_CreateTopics_simple(rk, NULL, topics, TOPIC_CNT,
+                                                 partition_cnt, NULL));
+        test_wait_topic_exists(rk, topics[TOPIC_CNT - 1], 5000);
+        /* Give the cluster some more time to propagate metadata
+         * for TOPICS_CNT topics */
+        rd_sleep(1);
+
+        RD_ARRAY_FOREACH(topic, topics, i) {
```
pranavrth: No need for FOREACH call here. We are not using topic anywhere below. We can have a simple for loop here.

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,108 @@ static void do_test_resubscribe_with_regex() {
...
+                            /* poll when we have a rebalance callback */
+                            with_rebalance_cb, expected_assignment, 10000);
+        }
+        rd_kafka_topic_partition_list_destroy(expected_assignment);
+
+        TEST_CALL_ERR__(
+            test_DeleteTopics_simple(rk, NULL, topics, TOPIC_CNT, NULL));
+        RD_ARRAY_FOREACH(topic, topics, i) {
+                rd_free(topics[i]);
+        }
```
pranavrth: No need for `FOREACH` here as well. We can just use simple forloop.

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,108 @@ static void do_test_resubscribe_with_regex() {
...
+        char *group;
+        rd_kafka_t *rk;
+        rd_kafka_conf_t *conf;
+        const int max_subscription_size = 5, partition_cnt = 4;
+        rd_kafka_topic_partition_list_t *expected_assignment = NULL;
+
+        SUB_TEST("%s", with_rebalance_cb ? "with rebalance callback"
+                                         : "without rebalance callback");
+
+        RD_ARRAY_FOREACH(topic, topics, i) {
```
pranavrth: No need for FOREACH here as simple for loop will work.

tests/test.c
```diff
@@ -3113,6 +3115,108 @@ void test_consumer_verify_assignment0(const char *func,
...
+        int i;
+        rd_ts_t end        = test_clock() + timeout_ms * 1000;
+        rd_bool_t verified = rd_false;
+        rd_kafka_topic_partition_list_t *expected_assignment_copy;
+
+        TEST_SAY("Verifying assignment\n");
+        expected_assignment_copy =
+            rd_kafka_topic_partition_list_copy(expected_assignment);
+        rd_kafka_topic_partition_list_sort(expected_assignment_copy, NULL,
+                                           NULL);
```
pranavrth: No need to copy or sort the list. It is already being done in the verify function.

src/rd.h
```diff
@@ -227,6 +227,12 @@ static RD_INLINE RD_UNUSED char *rd_strndup(const char *s, size_t len) {
 #define RD_ARRAYSIZE(A)           RD_ARRAY_SIZE(A)
 #define RD_SIZEOF(TYPE, MEMBER)   sizeof(((TYPE *)NULL)->MEMBER)
 #define RD_OFFSETOF(TYPE, MEMBER) ((size_t) & (((TYPE *)NULL)->MEMBER))
+/** Array foreach */
+#define RD_ARRAY_FOREACH(ELEM, ARRAY, INDEX)                                   \
```
pranavrth: No need for this foreach as not used in the PR.

tests/0045-subscribe_update.c
emasab: don't agree, we have it for every other iterable type, the foreach in general is less typo error prone than writing the initialization and while condition every time. Of course it works only for arrays that are on stack on in structure where we can determine the full size with `sizeof()` but it's an improvement for these cases.

tests/0045-subscribe_update.c
emasab: why they're already different names because of the randomized parameter

pranavrth: Yes but its easier to debug. Anything is fine.

tests/0045-subscribe_update.c
pranavrth: We can keep the definition of `RD_ARRAY_FOREACH` if you feel like but here the topic is being accessed as `topics[i]`.

If we are keeping `RD_ARRAY_FOREACH` then we should also have `RD_ARRAY_FOREACH_INDEX(ELEM, ARRAY, INDEX)` and have `RD_ARRAY_FOREACH(ELEM, ARRAY)` which internally uses `RD_ARRAY_FOREACH_INDEX`. We don't need index in FOREACH in general like other high level languages.

tests/test.c
```diff
@@ -3113,6 +3115,106 @@ void test_consumer_verify_assignment0(const char *func,
...
+    const rd_kafka_topic_partition_list_t *expected_assignment,
+    int timeout_ms) {
+        int i;
+        rd_ts_t end        = test_clock() + timeout_ms * 1000;
+        rd_bool_t verified = rd_false;
+        rd_kafka_topic_partition_list_t *expected_assignment_copy;
+
+        TEST_SAY("Verifying assignment\n");
+        expected_assignment_copy =
+            rd_kafka_topic_partition_list_copy(expected_assignment);
```
pranavrth: We don't need this copy as well. It is already being copied in `test_consumer_verify_assignment_topic_partition_list0`. It is expected that the called function doesn't modify the list.

pranavrth (changes requested):
Few more comments.

tests/0045-subscribe_update.c
emasab: It's fine I add them. The index is only for when we need it so it can be a different directive.

emasab: > if you feel like but here the topic is being accessed as topics[i].

right, I use topic here

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,107 @@ static void do_test_resubscribe_with_regex() {
...
+                int k,
+                    subscription_size =
+                        RD_MIN(max_subscription_size, TOPIC_CNT - i),
+                    expected_assignment_cnt = subscription_size * partition_cnt;
+
+                rd_kafka_topic_partition_list_t *subscription =
+                    rd_kafka_topic_partition_list_new(subscription_size);
+                rd_bool_t check_expected_assignment =
+                              (i % 5 == 0 || i == TOPIC_CNT - 1),
+                          do_unsubscribe = i % 7 == 0;
```
pranavrth: Its more readable if we create each variable in a separate line.
```suggestion
                size_t j;
                int k;
                int subscription_size =
                        RD_MIN(max_subscription_size, TOPIC_CNT - i);
                int expected_assignment_cnt = subscription_size * partition_cnt;
                rd_kafka_topic_partition_list_t *subscription =
                    rd_kafka_topic_partition_list_new(subscription_size);
                rd_bool_t check_expected_assignment =
                              (i % 5 == 0 || i == TOPIC_CNT - 1);
                rd_bool_t do_unsubscribe = i % 7 == 0;
```

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,107 @@ static void do_test_resubscribe_with_regex() {
...
+
+                if (do_unsubscribe)
+                        TEST_CALL_ERR__(rd_kafka_unsubscribe(rk));
+                if (check_expected_assignment)
+                        test_consumer_wait_assignment_topic_partition_list(
+                            rk,
+                            /* poll when we have a rebalance callback */
+                            with_rebalance_cb, expected_assignment, 10000);
+        }
+        rd_kafka_topic_partition_list_destroy(expected_assignment);
```
pranavrth: This should be inside the loop in the `if (check_expected_assignment)` before the loop ends. Both the conditions can be merged to a single one.

Use `RD_IF_FREE` as well here.

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,107 @@ static void do_test_resubscribe_with_regex() {
...
+
+                rd_kafka_topic_partition_list_t *subscription =
+                    rd_kafka_topic_partition_list_new(subscription_size);
+                rd_bool_t check_expected_assignment =
+                              (i % 5 == 0 || i == TOPIC_CNT - 1),
+                          do_unsubscribe = i % 7 == 0;
+
+                if (check_expected_assignment) {
+                        RD_IF_FREE(expected_assignment,
+                                   rd_kafka_topic_partition_list_destroy);
```
pranavrth: This can be moved to the `if (check_expected_assignment)` condition before the loop ends.

tests/0045-subscribe_update.c
```diff
@@ -821,6 +821,107 @@ static void do_test_resubscribe_with_regex() {
...
+#define TOPIC_CNT 100
+        char *topics[TOPIC_CNT] = {0};
+        char *topic;
+        size_t i;
+        char *group;
+        rd_kafka_t *rk;
+        rd_kafka_conf_t *conf;
+        const int max_subscription_size = 5, partition_cnt = 4;
+        rd_kafka_topic_partition_list_t *expected_assignment = NULL;
+
```
pranavrth: Should be move inside the topic for loop as this is not needed outside.

pranavrth (changes requested):

src/rd.h
```diff
@@ -228,11 +228,18 @@ static RD_INLINE RD_UNUSED char *rd_strndup(const char *s, size_t len) {
 #define RD_SIZEOF(TYPE, MEMBER)   sizeof(((TYPE *)NULL)->MEMBER)
 #define RD_OFFSETOF(TYPE, MEMBER) ((size_t) & (((TYPE *)NULL)->MEMBER))
 /** Array foreach */
-#define RD_ARRAY_FOREACH(ELEM, ARRAY, INDEX)                                   \
-        for ((INDEX = 0, (ELEM) = (ARRAY)[INDEX]);                             \
-             INDEX < RD_ARRAY_SIZE(ARRAY);                                     \
-             (ELEM) =                                                          \
-                 (++INDEX < RD_ARRAY_SIZE(ARRAY) ? (ARRAY)[INDEX] : (ELEM)))
+#define RD_ARRAY_FOREACH_INDEX(ELEM, ARRAY, INDEX, BLOCK)                      \
```
pranavrth: Let's not make it complex to understand. Remove BLOCK from here. It should be a simple FOREACH.

tests/0045-subscribe_update.c
emasab: ok, done

src/rd.h
emasab: that way we need to provide the index variable externally instead of defining it here and we need to use always `RD_ARRAY_FOREACH_INDEX`

tests/0045-subscribe_update.c
emasab: no because that way it's reinitialized to NULL at every iteration.

src/rd.h
pranavrth: I think we should keep it simple to understand the functions. We should remove this BLOCK from the macro and make it work it without it.  Something like below.

```c
/** Array foreach */
#define RD_ARRAY_FOREACH_INDEX(ELEM, ARRAY, INDEX)                             \
        for ((INDEX = 0, (ELEM) = (ARRAY)[INDEX]);                             \
             INDEX < RD_ARRAY_SIZE(ARRAY);                                     \
             (ELEM) =                                                          \
                 (++INDEX < RD_ARRAY_SIZE(ARRAY) ? (ARRAY)[INDEX] : (ELEM)))

#define RD_ARRAY_FOREACH(ELEM, ARRAY)                                          \
        RD_ARRAY_FOREACH_INDEX(ELEM, ARRAY, __i)
```

pranavrth: There are places in librdkafka where we expect a variable to be defined already. We can expect __i defined already.

emasab: > There are places in librdkafka where we expect a variable to be defined already. We can expect __i defined already.

I don't like it because you've to define the variable outside it and the name is fixed, in that case it's better if we leave only the original `RD_ARRAY_FOREACH` passing the variable name

pranavrth: How about using C99 standard of declaring variable inside the `for` loop. Like `for(int i=0;i<n;i++)` ?

emasab: That way it had problem with older versions of MSVC that didn't implement the C99 standard fully until some time ago.

pranavrth: Understood. I would prefer to remove this `FOREACH` loop altogether. But if you want to keep it, keep only `RD_ARRAY_FOREACH_INDEX`. Remove `RD_ARRAY_FOREACH`. So the implementation of original `RD_ARRAY_FOREACH` but with changed name `RD_ARRAY_FOREACH_INDEX`.

pranavrth (approved):
LGTM!.

PR #5030, opened 2025-04-09
[KIP-848] [mock cluster] Improved static group membership implementation

emasab:
When member is leaving with -2 a new member that is joining can replace the previous member id with the new one. If previous member didn't leave the new member joining with same group.instance.id is fenced in new protocol.


src/rdkafka_mock_cgrp.c
```diff
@@ -1521,17 +1530,24 @@ rd_kafka_mock_cgrp_consumer_member_t *rd_kafka_mock_cgrp_consumer_member_add(
...
-                        return NULL;
+
+                if (member) {
+                        if (!member->left_static_membership) {
+                                /* Old member still active,
+                                 * fence this one */
+                                return NULL;
+                        }
+
+                        if (rd_kafkap_str_cmp_str(MemberId, member->id) != 0) {
```
pranavrth, line 1541: nit: 
```suggestion
                        if (!rd_kafkap_str_cmp_str(MemberId, member->id)) {
```

pranavrth (approved):
LGTM!.

PR #5036, opened 2025-04-13
Handle the DESTROY_BROKER error on SaslAuthenticate RPC

emasab:
by not sending an error message to the user as the authentication will be done on the new broker connection.
Similar to what done on SaslHandshake


pranavrth (approved):
LGTM!.

PR #5039, opened 2025-04-15
Add an interval to retry connection to any random broker after interval has passed

emasab:
Remove the override to schedule a connection when there's no existing one as it causes too frequent connection retries in case cluster isn't reachable.

Remove scheduled connections count as the `sparse_connect_random` interval is again effective in any case.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -5710,6 +5710,7 @@ void rd_kafka_connect_any(rd_kafka_t *rk, const char *reason) {
 
         suppr = rd_interval(&rk->rk_suppress.sparse_connect_random,
                             rk->rk_conf.sparse_connect_intvl * 1000, 0);
+        mtx_unlock(&rk->rk_suppress.sparse_connect_lock);
 
         if (suppr <= 0) {
                 rd_kafka_dbg(rk, BROKER | RD_KAFKA_DBG_GENERIC, "CONNECT",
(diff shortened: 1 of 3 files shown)


src/rdkafka_broker.c
```diff
@@ -5707,32 +5707,20 @@ void rd_kafka_connect_any(rd_kafka_t *rk, const char *reason) {
...
-                             scheduled_connections, reason);
+                             "still suppressed for %" PRId64 "ms: %s",
+                             -suppr / 1000, reason);
+                /* Retry after interval + 1ms has passed */
+                rd_kafka_timer_start_oneshot(
+                    &rk->rk_timers, &rk->rk_suppress.sparse_connect_random_tmr,
+                    rd_false, 1000LL - suppr, rd_kafka_connect_any_timer_cb,
+                    (void *)reason);
                 goto done;
         }
```
pranavrth, line 5726: unlock here as well - mtx_unlock(&rk->rk_suppress.sparse_connect_lock);

src/rdkafka_broker.c
```diff
@@ -5707,32 +5707,20 @@ void rd_kafka_connect_any(rd_kafka_t *rk, const char *reason) {
...
-                any_connecting = rd_false;
-
-        scheduled_connections =
-            rd_atomic32_get(&rk->rk_scheduled_connections_cnt);
 
-        if (!any_connecting && scheduled_connections == 0)
-                /* Skip interval */
-                rd_interval_reset(&rk->rk_suppress.sparse_connect_random);
         suppr = rd_interval(&rk->rk_suppress.sparse_connect_random,
                             rk->rk_conf.sparse_connect_intvl * 1000, 0);
```
pranavrth, line 5713: unlock here - `mtx_unlock(&rk->rk_suppress.sparse_connect_lock);`

pranavrth (changes requested):

PR #5048, opened 2025-04-24
GroupCoordinator gets nodeis "-1" instead of bootstrap (>=2.10.0)

resamaraschi:
With version 2.10.0 the GroupCoordinator get the -1 id instead of bootstraps broker, which I guess is introduced by this #4557.
Just wanted to correct the statistics docs.


resamaraschi:
I was not aware that a CLA need to be signed. Since I do not want to do so, I am going to close the PR. Hope someone else can correct the docs though :)

PR #5055, opened 2025-04-28
Add missing wrlock around rd_kafka_metadata_cache_hint call

marcin-krystianc:
Fixes #5051 

This PR addresses a heap-use-after-free error detected by AddressSanitizer within the `rd_kafka_metadata_cache_topics_to_list` function. The issue was particularly reproducible during tests involving frequent topic deletion and recreation, which heavily exercises the metadata cache logic.

The root cause was identified as missing write lock protection when invoking the rd_kafka_metadata_cache_hint. Without proper locking, concurrent threads (e.g., the broker communication thread handling metadata updates and potentially other threads accessing metadata) could race to modify or read the hint, leading to metadata cache corruption. This corruption could manifest as accessing freed memory when iterating through the cache, resulting in the observed heap-use-after-free.

With this change, the previously reported ASan error is resolved when running the test case involving rapid topic recreation.

PR #5066, opened 2025-05-05
[internal] Add missing wrlock around rd_kafka_metadata_cache_hint call

emasab:
Re-open #5055 from an internal branch


mensfeld:
@emasab once this is released I will run my stress tests for rdkafka-ruby and karafka. While I cannot reproduce it easily my specs fail randomly when running in stress mode so it should be a good indicator if that's the only issue happening.

milindl (approved):
Thanks for the fix.

PR #5067, opened 2025-05-05
Fix reboostrapping with no boostrap servers

emasab:
when brokers were added only through `rd_kafka_brokers_add`.

### DIFF
--- a/tests/0152-rebootstrap.c
+++ b/tests/0152-rebootstrap.c
@@ -1,0 +1,59 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2025, Confluent Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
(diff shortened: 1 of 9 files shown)


tests/0152-rebootstrap.c
```diff
@@ -0,0 +1,59 @@
...
+
+        /* Give it time to trigger ALL_BROKERS_DOWN */
+        rd_sleep(1);
+        rd_kafka_destroy(rk);
+        SUB_TEST_PASS();
+}
+
+int main_0152_rebootstrap_local(int argc, char **argv) {
+
+        do_test_rebootstrap_local_no_bootstrap_servers(RD_KAFKA_PRODUCER);
```
pranavrth, line 55: We should add for normal case as well as for empty string case.

pranavrth (changes requested):

pranavrth:
Update the PR and we can merge.

tests/0152-rebootstrap.c
pranavrth: I was talking about empty bootstrap brokers string case. Lets add it in that PR.

pranavrth (approved):

pranavrth:
LGTM!. Thanks.

tests/0152-rebootstrap.c
emasab: ok

PR #5075, opened 2025-05-12
Compatible versions for nuget dependencies

emasab:
this avoids frequent dependency bot commits.

### DIFF
--- a/packaging/nuget/requirements.txt
+++ b/packaging/nuget/requirements.txt
@@ -1,3 +1,3 @@
-boto3==1.38.14
-rpmfile==1.1.1
-filemagic==1.6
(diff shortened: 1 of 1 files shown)


packaging/nuget/requirements.txt
```diff
@@ -1,3 +1,3 @@
-boto3==1.38.13
-rpmfile==1.1.1
-filemagic==1.6
+boto3~=1.38
```
pranavrth, line 1: We are not using s3, so we can remove boto and related s3 code as well. Can be done later as well.

pranavrth (approved):
LGTM!.

pranavrth (approved):

PR #5110, opened 2025-06-10
Remove CURL warning and Makefile fix

emasab:


milindl (approved):
lgtm. I checked documentation to ascertain that the older one is deprecated.

PR #5126, opened 2025-06-24
Avoid returning an all brokers down error on planned disconnections

### LINKED ISSUE #5142: Periodic all brokers down error happening because of idle disconnections
A sequence of _all brokers down_ errors can be received in an idle client, a producer for example, even if the broker disconnections are just idle disconnections, that is normal broker TCP closes, and not TCP resets or connection failures.
To test it, start a cluster with a short idle disconnection time like with
```
python interactive_broker_version.py --kraft 4.0.0 --conf '{"conf": ["connections.max.idle.ms=10000"]}'
``` 
then start a ` examples/producer` with cluster bootstrap servers but don't produce anything. After some time the errors will start appearing.

It doesn't happen only when the disconnections happen all at the same time, but it's the down state that isn't reset if other broker connections aren't needed (sparse connections) contributing to raising the error even if a single connection is closed.

emasab:
Closes #5142 .

Avoid returning an all brokers down error on planned disconnections

This is done by avoiding to count planned disconnections, such as idle disconnections, broker host change and similar, as events that can cause the client to reach the "all brokers down" state, returning an error since 1.x and possibly restarting a re-bootstrap sequence since 2.10.0.

Also added jitter to the idle disconnections to avoid they happen at the same time and improved the disconnection error messages.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -4339,6 +4339,7 @@ static RD_INLINE void rd_kafka_broker_idle_check(rd_kafka_broker_t *rkb) {
         rd_ts_t ts_last_activity = RD_MAX(ts_send, ts_recv);
         int idle_ms;
         const int max_jitter_ms = 2000;
+        rd_dassert(rkb->rkb_rk->rk_conf.connections_max_idle_ms > 0);
 
         /* If nothing has been sent yet, use the connection time as
          * last activity. */
@@ -4347,16 +4348,16 @@ static RD_INLINE void rd_kafka_broker_idle_check(rd_kafka_broker_t *rkb) {
 
         idle_ms = (int)((rd_clock() - ts_last_activity) / 1000);
 
-        if (!rkb->rkb_c.connection_max_idle_ms) {
+        if (unlikely(rkb->rkb_c.connections_max_idle_ms == -1)) {
                 /* Add a different jitter for each broker. */
-                rkb->rkb_c.connection_max_idle_ms =
+                rkb->rkb_c.connections_max_idle_ms =
                     rkb->rkb_rk->rk_conf.connections_max_idle_ms;
-                if (rkb->rkb_c.connection_max_idle_ms >= 2 * max_jitter_ms)
-                        rkb->rkb_c.connection_max_idle_ms -=
+                if (rkb->rkb_c.connections_max_idle_ms >= 2 * max_jitter_ms)
+                        rkb->rkb_c.connections_max_idle_ms -=
                             rd_jitter(0, max_jitter_ms);
         }
 
-        if (likely(idle_ms < rkb->rkb_c.connection_max_idle_ms))
+        if (likely(idle_ms < rkb->rkb_c.connections_max_idle_ms))
                 return;
 
         rd_kafka_broker_planned_fail(rkb, RD_KAFKA_RESP_ERR__TRANSPORT,
@@ -4894,13 +4895,14 @@ rd_kafka_broker_t *rd_kafka_broker_add(rd_kafka_t *rk,
                 rd_snprintf(rkb->rkb_name, sizeof(rkb->rkb_name), "%s", name);
         }
 
-        rkb->rkb_source   = source;
-        rkb->rkb_rk       = rk;
-        rkb->rkb_ts_state = rd_clock();
-        rkb->rkb_nodeid   = nodeid;
-        rkb->rkb_proto    = proto;
-        rkb->rkb_port     = port;
-        rkb->rkb_origname = rd_strdup(name);
+        rkb->rkb_source                    = source;
+        rkb->rkb_rk                        = rk;
+        rkb->rkb_ts_state                  = rd_clock();
+        rkb->rkb_nodeid                    = nodeid;
+        rkb->rkb_proto                     = proto;
+        rkb->rkb_port                      = port;
+        rkb->rkb_origname                  = rd_strdup(name);
+        rkb->rkb_c.connections_max_idle_ms = -1;
 
         mtx_init(&rkb->rkb_lock, mtx_plain);
         mtx_init(&rkb->rkb_logname_lock, mtx_plain);
--- a/src/rdkafka_broker.h
+++ b/src/rdkafka_broker.h
@@ -187,20 +187,27 @@ struct rd_kafka_broker_s { /* rd_kafka_broker_t */
                 rd_atomic64_t reqtype[RD_KAFKAP__NUM]; /**< Per request-type
                                                         *   counter */
 
-                rd_atomic64_t ts_send;      /**< Timestamp of last send */
-                rd_atomic64_t ts_recv;      /**< Timestamp of last receive */
-                rd_bool_t skip_broker_down; /**< Avoid reporting the
-                                             *   broker down on next
-                                             *   state change.
-                                             *   Useful for a planned
-                                             *   disconnection to avoid
-                                             *   reaching the all
-                                             *   brokers down state. */
-                int connection_max_idle_ms; /**< Maximum idle time
-                                             *   for this broker connection.
-                                             *   jitter is different for
-                                             *   each broker to avoid the
-                                             *   ALL_BROKERS_DOWN error. */
+                rd_atomic64_t ts_send;       /**< Timestamp of last send */
+                rd_atomic64_t ts_recv;       /**< Timestamp of last receive */
+                rd_bool_t skip_broker_down;  /**< Avoid reporting the
+                                              *   broker down on next
+                                              *   state change.
+                                              *   Useful for a planned
+                                              *   disconnection to avoid
+                                              *   reaching the all
+                                              *   brokers down state. */
+                int connections_max_idle_ms; /**< Maximum idle time
+                                              *   for this broker connections.
+                                              *   jitter is different for
+                                              *   each broker to avoid the
+                                              *   ALL_BROKERS_DOWN error.
+                                              *   Initial value is `-1` as
+                                              *   `connections.max.idle.ms=0`
+                                              *   means disabled, but must
+                                              *   never be used when
+                                              *   `connections.max.idle.ms=0`
+                                              *   for the same reason
+                                              *   (an assert is present). */
         } rkb_c;
 
         struct {


CHANGELOG.md
```diff
@@ -16,6 +17,16 @@ librdkafka v2.11.1 is a maintenance release:
...
 
+* Issues: #5142.
+  Avoid returning an all brokers down error on planned disconnections.
+  This is done by avoiding to count planned disconnections, such as idle
+  disconnections, broker host change and similar as events that can cause
+  the client to reach the "all brokers down" state, returning an error and
+  since 2.10.0 possibly starting a re-bootstrap sequence.
+  Also avoids the same error just after connecting and disconnecting from
+  a single broker, rest of brokers need to be tried again in that case.
+  Happening since 1.x (#5126).
```
pranavrth: This should be a separate entry.

CHANGELOG.md
```diff
@@ -16,6 +17,16 @@ librdkafka v2.11.1 is a maintenance release:
   a range and were failing if the older version was removed.
   Happening since 1.x (#5130).
 
+* Issues: #5142.
+  Avoid returning an all brokers down error on planned disconnections.
+  This is done by avoiding to count planned disconnections, such as idle
+  disconnections, broker host change and similar as events that can cause
```
pranavrth, line 25: We should define "similar" events.

src/rdkafka_broker.c
```diff
@@ -4333,7 +4333,18 @@ static RD_INLINE void rd_kafka_broker_idle_check(rd_kafka_broker_t *rkb) {
 
         idle_ms = (int)((rd_clock() - ts_last_activity) / 1000);
 
-        if (likely(idle_ms < rkb->rkb_rk->rk_conf.connections_max_idle_ms))
+        if (!rkb->rkb_c.connection_max_idle_ms) {
```
pranavrth: I think we should not use 0 as default. We should keep -1 maybe. 0 means disabled in the main property and lets keep the semantics same here as well.

src/rdkafka_broker.c
```diff
@@ -4333,7 +4333,18 @@ static RD_INLINE void rd_kafka_broker_idle_check(rd_kafka_broker_t *rkb) {
...
+                /* Add a different jitter for each broker. */
+                rkb->rkb_c.connection_max_idle_ms =
+                    rkb->rkb_rk->rk_conf.connections_max_idle_ms;
+                if (rkb->rkb_c.connection_max_idle_ms >=
+                    2 * rkb->rkb_rk->rk_conf.socket_connection_setup_timeout_ms)
```
pranavrth: We shouldn't rely on some property that can be configured by the user. I think we should just use some % of connections_max_idle_ms itself as jitter.

pranavrth (changes requested):

src/rdkafka_transport.c
```diff
@@ -344,10 +342,9 @@ static ssize_t rd_kafka_transport_socket_recv0(rd_kafka_transport_t *rktrans,
                 } else if (unlikely(r == 0)) {
                         /* Receive 0 after POLLIN event means
                          * connection closed. */
-                        rd_rkb_dbg(rktrans->rktrans_rkb, BROKER, "SOCKET",
-                                   "Disconnected: connection closed by "
-                                   "peer");
-                        rd_snprintf(errstr, errstr_size, "Disconnected");
+                        rd_snprintf(errstr, errstr_size,
+                                    "Disconnected: connection closed by "
+                                    "peer");
```
pranavrth: We have same messages in multiple places. Let's improve this for our debugging.

Same for other messages as well if required.

src/rdkafka_broker.c
```diff
@@ -482,7 +483,7 @@ static void rd_kafka_broker_set_error(rd_kafka_broker_t *rkb,
 
         /* Provide more meaningful error messages in certain cases */
         if (err == RD_KAFKA_RESP_ERR__TRANSPORT &&
-            rd_kafka_transport_error_disconnected(errstr)) {
+            rd_kafka_transport_error_disconnected(&errstr[of_message])) {
```
pranavrth: Let's use `fmt` instead.

src/rdkafka_broker.c
```diff
@@ -580,13 +580,14 @@ static void rd_kafka_broker_set_error(rd_kafka_broker_t *rkb,
                    identical ? ": identical to last error" : "",
                    suppress ? ": error log suppressed" : "");
 
-        if (level != LOG_DEBUG && (level <= LOG_CRIT || !suppress)) {
+        if (level <= LOG_WARNING && (level <= LOG_CRIT || !suppress)) {
```
pranavrth: I think we should log LOG_INFO as well.

src/rdkafka_broker.c
```diff
@@ -3436,9 +3459,9 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                  * and trigger a state change.
                  * This makes sure any eonce dependent on state changes
                  * are triggered. */
-                rd_kafka_broker_fail(rkb, LOG_DEBUG,
```
pranavrth: Check this condition on whether this is planned fail or not.

src/rdkafka_broker.c
```diff
@@ -4669,8 +4701,8 @@ static int rd_kafka_broker_thread_main(void *arg) {
         while (rd_kafka_broker_ops_serve(rkb, RD_POLL_NOWAIT))
                 ;
 
-        rd_kafka_broker_fail(rkb, LOG_DEBUG, rd_kafka_broker_destroy_error(rk),
```
pranavrth: As discussed, check this condition on whether this is planned fail or not.

pranavrth (changes requested):

tests/0045-subscribe_update.c
```diff
@@ -814,6 +814,7 @@ static void do_test_resubscribe_with_regex() {
 
         rd_free(topic1);
         rd_free(topic2);
+        rd_free(topic_a);
```
pranavrth, line 818: We should delete `topic_a` as well.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 librdkafka v2.11.1 is a maintenance release:
 
 * Made the conditions for enabling the features future proof (#5130).
+* Avoid returning an all brokers down error on planned disconnections (#5126).
```
pranavrth, line 6: I think expected is better word here than planned. Planned gives the impression that we know at what exact time the disconnection will occur. Expected will give the impression that whatever has happened is expected behaviour. For example, idle disconnections are expected but not planned. We expect that after certain time of idleness, the connection are disconnected for the broker.

src/rdkafka_broker.c
```diff
@@ -348,15 +349,20 @@ void rd_kafka_broker_set_state(rd_kafka_broker_t *rkb, int state) {
                      rd_kafka_broker_state_names[rkb->rkb_state],
                      rd_kafka_broker_state_names[state]);
 
+        if (rd_kafka_broker_state_is_down(state) && skip_broker_down)
+                /* Reset the flag so if it disconnects again it won't be
+                 * treated as a planned disconnection. */
```
pranavrth, line 362: We should improve this comment as there can be further planned disconnections.

src/rdkafka_broker.c
```diff
@@ -371,27 +377,37 @@ void rd_kafka_broker_set_state(rd_kafka_broker_t *rkb, int state) {
...
+                                 * the down counter to try again with rest of
+                                 * brokers. Otherwise, a single broker
+                                 * connection can cause `ALL_BROKERS_DOWN` to
+                                 * get triggered even if the connection to rest
+                                 * of brokers was restored but they didn't
+                                 * attempt a re-connection because of sparse
+                                 * broker connections. */
                                 rd_atomic32_add(&rkb->rkb_rk->rk_broker_up_cnt,
                                                 1);
+                                rd_kafka_reset_any_broker_down_reported(
```
pranavrth, line 410: I don't think we should be resetting down counter if a single broker goes up again. Let's say that 3 brokers are present in the cluster. 2 goes down. Then 1 goes up. 1 is still down but we will reset that flag for that broker as well.

src/rdkafka_broker.c
pranavrth:
```suggestion
        if (level <= LOG_CRIT || (level <= LOG_INFO && !suppress)) {
```
We can modify the condition like this for better understanding.

src/rdkafka_broker.c
```diff
@@ -1022,11 +1044,11 @@ static void rd_kafka_broker_timeout_scan(rd_kafka_broker_t *rkb, rd_ts_t now) {
                                             1000.0f));
                         else
                                 rttinfo[0] = 0;
-                        rd_kafka_broker_fail(rkb, LOG_ERR,
-                                             RD_KAFKA_RESP_ERR__TIMED_OUT,
-                                             "%i request(s) timed out: "
-                                             "disconnect%s",
-                                             rkb->rkb_req_timeouts, rttinfo);
+                        rd_kafka_broker_planned_fail(
```
pranavrth, line 1054: Here we were logging ERROR. Why are we changing it now?

src/rdkafka_broker.c
```diff
@@ -4691,7 +4723,8 @@ static int rd_kafka_broker_thread_main(void *arg) {
 
                 if (RD_KAFKA_BROKER_IS_LOGICAL(rkb)) {
                         rd_atomic32_sub(&rkb->rkb_rk->rk_logical_broker_cnt, 1);
-                } else if (rkb->rkb_down_reported) {
+                } else if (rd_atomic32_set(&rkb->rkb_down_reported, 0) == 1) {
```
pranavrth, line 4734: Its confusing to change these values from multiple places. Can we try to change them from a single place? Either here or from state change?

I am bit confused here on how to reach here. We reach this condition only when `skip_broker_down` was set to true and broker handle is getting destroyed. Don't you think that we should be adding ALL_BROKERS_DOWN counter in this condition?

src/rdkafka_broker.h
```diff
@@ -331,7 +343,9 @@ struct rd_kafka_broker_s { /* rd_kafka_broker_t */
 
         rd_kafka_secproto_t rkb_proto;
 
-        int rkb_down_reported; /* Down event reported */
+        /** Down event was reported for this broker
+         *  after last connection to any broker. */
```
pranavrth, line 354: What do we mean by `after last connection to any broker`

src/rdkafka_ssl.c
```diff
@@ -237,24 +237,21 @@ rd_kafka_transport_ssl_io_update(rd_kafka_transport_t *rktrans,
                         rd_kafka_ssl_error(NULL, rktrans->rktrans_rkb, errstr,
                                            errstr_size);
                 else if (!rd_socket_errno) {
-                        rd_rkb_dbg(rktrans->rktrans_rkb, BROKER, "SOCKET",
```
pranavrth, line 240: I think we shouldn't remove these debug statements. I can see that we are logging these errstr in some case. Can we verify this? I would prefer "Disconnected" error change in a separate PR so that we can focus solely on those. I think we need to elaborate more on different kind of Disconnections which are very generic. We can brainstorm that in a separate PR.

CHANGELOG.md
emasab: I think it's better to avoid explaining everything in the changelog, there's the PR for more details. In this case they were all events that were reported as DEBUG plus the disconnection after a timeout.

src/rdkafka_broker.c
emasab: This is setting the value on first use, it's works fine given it's initialized to zero on allocation and we don't call `rd_kafka_broker_idle_check` if `rkb->rkb_rk->rk_conf.connections_max_idle_ms` is zero.

src/rdkafka_broker.c
emasab: I later changed to 2s as it's being set in AK. The important is they don't happen at the same time for client originated idle disconnections given they're already not counted for reaching "all brokers down" with the planned disconnection function in this PR.

CHANGELOG.md
emasab: Expected seems like we're reacting to a disconnection from the broker that is somehow expected, here we're not reacting we're closing the connection from the client that's why "planned".
When the idle disconnection comes from the broker we cannot know that it was for idleness and those aren't the planned disconnections in this PR.

src/rdkafka_broker.c
emasab: How, this way?
```suggestion
                /* Reset the flag so if it disconnects again and it's not a planned disconnection it won't be
                 * treated as a planned disconnection. */
```

src/rdkafka_broker.c
emasab: Because it's planned to close the connection client side in this case, it's originating from the client after timing out some buffers. In case the broker is not available it will generate a non-planned disconnection when reconnecting.

src/rdkafka_broker.c
emasab: Yes, the goal is that we need to try other brokers again after at least a broker goes up otherwise it doesn't fix the bug regarding the caching of the down state. In your example that is exactly the goal, to try other brokers again before declaring that all brokers are down.

It could be that the broker disconnected for idleness for example and it wasn't needed to connect to it for sparse connection. In this case the broker isn't unavailable at all and we need to give it a chance to connect again before considering it down, only when we cannot connect to any we'll finally reach the all brokers down state.

It could be that we connect to any broker and then we get consistently disconnected for some RPC problem or incompatibility but that goes beyond the scope of ALL_BROKERS_DOWN than is considered like "the cluster is unavailable, we cannot connect to it", and for re-bootstrapping in that case there's always the fail-safe `metadata.recovery.rebootstrap.trigger.ms` in KIP-1102.

src/rdkafka_broker.h
emasab: As I said in https://github.com/confluentinc/librdkafka/pull/5126#discussion_r2230467390: the down event was reported after any other brokers connected, when any broker connects the flag is reset for all brokers so they need to be tried again.

Given the code it could be better called `down_counted` than `down_reported` as it's reporting the `ALL_BROKERS_DOWN` error but only when all of them are down not for each of them.

src/rdkafka_ssl.c
emasab: I added these debug statement recently and only because we couldn't change the the "Disconnected" error message before this PR. If we can add it to the error than they're not needed because the error message is logged as debug here:
https://github.com/confluentinc/librdkafka/blob/11d20ac7f5c7e0bd232c1c73553e9f7230a45a10/src/rdkafka_broker.c#L590

So once we can use different error messages these debug logs become duplicate.

src/rdkafka_broker.c
emasab: No here we're _subtracting_ from `rk_broker_down_cnt` if we had counted this broker earlier. Given we're removing the broker (we're at the end of `rd_kafka_broker_thread_main`) we need not to count it otherwise there's a mismatch.

src/rdkafka_broker.c
emasab: In general it shouldn't happen that a re-bootstrap is needed in this case, because there's a different broker that is receiving the metadata response and it should be up.
But what if the decommissioned broker is the last broker up? It should cause a re-bootstrap in that case. Is it possible?
I reverted these two changes on broker decommissioning because of this edge case:

1. if we're decommissioning broker X there should be broker Y that received the metadata response and is UP as broker Y cannot remove itself unless it's a bootstrap broker
2. if broker X == Y and it's a bootstrap broker there should be another learned broker Z that needs to be tried because the counter is reset on re-bootstrap
3. broker X and Y are learned ones and metadata response is received by Y, leaving only Y
4. after sending TERMINATE to broker X broker Y is removed from the cluster too and replaced by A, it disconnects from Y and increases broker down count
5. then when broker X is decommissioned it should be the last counted broker and it should trigger a re-bootstrap given when Y tries to re-connect and cannot succeed it was already counted as down

src/rdkafka_broker.c
pranavrth: As discussed, update this value to default -1 for unset. Also, name should be `connections_max_idle_ms` instead of `connection_max_idle_ms`.

pranavrth (changes requested):

src/rdkafka_broker.c
emasab: Used a default of -1 to avoid using a different meaning for 0. Added an assert too for the case where it's zero. Changed the name.

pranavrth (approved):
LGTM!.

PR #5136, opened 2025-07-04
Fix OAuthBearer OIDC preprocessor directive when using CMake without CURL

### LINKED ISSUE #5135: librdkafka v2.11.0 can not build without curl
cmake with `-DWITH_CURL=OFF`

from file: `./generated/config.h`
```c
#define WITH_OAUTHBEARER_OIDC 0
```

from `src/rdkafka_conf.c`
```c
#ifdef WITH_OAUTHBEARER_OIDC
#include <curl/curl.h>
#endif
```

get error:
```sh
src/rdkafka_conf.c:60:10: fatal error: 'curl/curl.h' file not found
   60 | #include <curl/curl.h>
```

emasab:
## Summary

Fixes several issues that surface when building librdkafka with CMake and a reduced
feature set, and adds CI coverage so these build-configuration regressions are caught
going forward.

Closes #5135, #5282

## Changes

- **Fix OAuthBearer OIDC preprocessor directive when building with CMake without CURL**
  (#5135) — corrects the conditional so compilation succeeds when CURL is disabled.
- **Add build-configuration CI checks** that build with CMake on Alpine and manylinux
  images with the optional features disabled, to catch compile/runtime breakage in
  minimal builds. Adds `packaging/tools/build-configurations-checks.sh`,
  `run-in-docker.sh`, `Dockerfile.alpine`, `Dockerfile.manylinux`, and a Semaphore job.
- **Skip test 0066 (plugins) when `dlopen` is unsupported**, so the broker-less quick
  suite passes in builds without dynamic loading.
- **Fix `rd_atomic{32,64}_set` returning the new value instead of the previous one in
  CMake builds.** CMake defined `HAVE_ATOMICS_{32,64}` (and optionally `_SYNC`) but never
  `HAVE_ATOMICS_{32,64}_ATOMIC`, so the setters fell through to a non-atomic fallback that
  returned the new value. This broke the planned-disconnection all-brokers-down handling,
  which relies on `rd_atomic32_set(&rkb_down_reported, 1)` returning the previous value to
  detect the `0 -> 1` transition — so `ALL_BROKERS_DOWN` was never raised and test 0095
  timed out under CMake. CMake now detects/defines `HAVE_ATOMICS_*_ATOMIC` (mirroring
  mklove's `configure.atomics`), and both setters select the `__atomic` / `__sync`
  implementation from the flags with the mutex path as a correct fallback.
- **Pin specific tool/image versions** for reproducible build-configuration checks.

## Testing

The new build-configuration checks run tests 0095 and 0121 under CMake/manylinux with all
optional features disabled; both pass (0095 previously timed out before the atomics fix).

### DIFF
--- a/src/rdatomic.h
+++ b/src/rdatomic.h
@@ -239,7 +239,13 @@ static RD_INLINE int RD_UNUSED rd_atomic64_cas(rd_atomic64_t *ra,
         return InterlockedCompareExchange64((LONG64 *)&ra->val, (LONG64)desired,
                                             (LONG64)expected) ==
                (LONG64)expected;
-#elif !HAVE_ATOMICS_64
+#elif HAVE_ATOMICS_64 && HAVE_ATOMICS_64_ATOMIC
+        return __atomic_compare_exchange_n(&ra->val, &expected, desired,
+                                           0 /* strong */, __ATOMIC_SEQ_CST,
+                                           __ATOMIC_SEQ_CST);
+#elif HAVE_ATOMICS_64 && HAVE_ATOMICS_64_SYNC
+        return __sync_bool_compare_and_swap(&ra->val, expected, desired);
+#else
         int r;
         mtx_lock(&ra->lock);
         if (ra->val == expected) {
@@ -250,19 +256,6 @@ static RD_INLINE int RD_UNUSED rd_atomic64_cas(rd_atomic64_t *ra,
         }
         mtx_unlock(&ra->lock);
         return r;
-#elif HAVE_ATOMICS_64_ATOMIC
-        return __atomic_compare_exchange_n(&ra->val, &expected, desired,
-                                           0 /* strong */, __ATOMIC_SEQ_CST,
-                                           __ATOMIC_SEQ_CST);
-#elif HAVE_ATOMICS_64_SYNC
-        return __sync_bool_compare_and_swap(&ra->val, expected, desired);
-#else
-        // FIXME
-        if (ra->val == expected) {
-                ra->val = desired;
-                return 1;
-        }
-        return 0;
 #endif
 }
 
--- a/src/rdatomic.h
+++ b/src/rdatomic.h
@@ -119,19 +119,17 @@ static RD_INLINE int32_t RD_UNUSED rd_atomic32_set(rd_atomic32_t *ra,
                                                    int32_t v) {
 #ifdef _WIN32
         return InterlockedExchange((LONG *)&ra->val, v);
-#elif !HAVE_ATOMICS_32
+#elif HAVE_ATOMICS_32 && HAVE_ATOMICS_32_ATOMIC
+        return __atomic_exchange_n(&ra->val, v, __ATOMIC_SEQ_CST);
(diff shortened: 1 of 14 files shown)


src/rdatomic.h
```diff
@@ -119,19 +119,17 @@ static RD_INLINE int32_t RD_UNUSED rd_atomic32_set(rd_atomic32_t *ra,
                                                    int32_t v) {
 #ifdef _WIN32
         return InterlockedExchange((LONG *)&ra->val, v);
-#elif !HAVE_ATOMICS_32
+#elif HAVE_ATOMICS_32 && HAVE_ATOMICS_32_ATOMIC
+        return __atomic_exchange_n(&ra->val, v, __ATOMIC_SEQ_CST);
+#elif HAVE_ATOMICS_32 && HAVE_ATOMICS_32_SYNC
```
prashah-confluent, line 124: Isn't this branch a dead code in case of CMake and eventually ending up using `#else` branch which is safe though because lock would be defined given HAVE_ATOMICS_32 is 0? 

CMake ends up setting only HAVE_ATOMICS_32_SYNC [reference](https://github.com/confluentinc/librdkafka/blob/c61cf7faee607fc46cbc9bdf0c883bd32b151e15/packaging/cmake/try_compile/rdkafka_setup.cmake#L59-L79) and not HAVE_ATOMICS_32. They both seem mutually exclusive in case of CMake. So this branch would never be entered right?

Should we change to something like below which seems to handle all scenarios and isn't based on the invariant `HAVE_ATOMICS_32_ATOMIC`

```
static RD_INLINE int32_t RD_UNUSED rd_atomic32_set(rd_atomic32_t *ra,
                                                   int32_t v) {
#ifdef _WIN32
        return InterlockedExchange((LONG *)&ra->val, v);
#elif HAVE_ATOMICS_32_SYNC
        return __sync_lock_test_and_set(&ra->val, v);
#elif !HAVE_ATOMICS_32
        int32_t r;
        mtx_lock(&ra->lock);
        r       = ra->val;
        ra->val = v;
        mtx_unlock(&ra->lock);
        return r;
#else
        return __atomic_exchange_n(&ra->val, v, __ATOMIC_SEQ_CST);
#endif
}
```

The only tradeoff I see is that default assumes `__atomic` is available.

emasab: I was uniforming cmake in this PR to set the same values as mklove, so `HAVE_ATOMICS_32` for example is set to 1 in both cases and then either `HAVE_ATOMICS_32_ATOMIC` or `HAVE_ATOMICS_32_SYNC` is set to 1.
Just there's a case when using `try_compile(HAVE_ATOMICS_32_SYNC` where `HAVE_ATOMICS_32` isn't set to true. I've pushed the cmake change. In mklove instead there are cases where `HAVE_ATOMICS_32_SYNC` or `HAVE_ATOMICS_32_ATOMIC` aren't defined if `HAVE_ATOMICS_32` is false.

prashah-confluent (approved):
LGTM !! Thanks a lot @emasab

PR #5156, opened 2025-08-01
Fix deadlock in `rd_kafka_reset_any_broker_down_reported` 

emasab:
by releasing and reacquiring the broker lock, to respect lock ordering and avoid deadlocks.

Fix is for #5126 , not released still. To be merged before releasing.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -407,8 +407,12 @@ void rd_kafka_broker_set_state(rd_kafka_broker_t *rkb, int state) {
                                  * broker connections. */
                                 rd_atomic32_add(&rkb->rkb_rk->rk_broker_up_cnt,
                                                 1);
+                                rd_kafka_broker_unlock(rkb);
+                                /* Releases rkb->rkb_lock to respect
+                                 * lock ordering and avoid deadlocks */
                                 rd_kafka_reset_any_broker_down_reported(
                                     rkb->rkb_rk);
+                                rd_kafka_broker_lock(rkb);
                         }
 
                         trigger_monitors = rd_true;


src/rdkafka_broker.c
```diff
@@ -407,8 +407,12 @@ void rd_kafka_broker_set_state(rd_kafka_broker_t *rkb, int state) {
                                  * broker connections. */
                                 rd_atomic32_add(&rkb->rkb_rk->rk_broker_up_cnt,
                                                 1);
+                                rd_kafka_broker_unlock(rkb);
+                                /* Releases rkb->rkb_lock to respect
+                                 * lock ordering and avoid deadlocks */
                                 rd_kafka_reset_any_broker_down_reported(
                                     rkb->rkb_rk);
+                                rd_kafka_broker_lock(rkb);
```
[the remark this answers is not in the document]
emasab: It's correct to check this so:
the state is only changed by `rd_kafka_broker_set_state` and this function is only called by the broker thread that is currently executing this piece of code. Broker cannot be destroyed when reacquiring the lock as there's always a last reference held by the broker thread. `rkb->rkb_destroyed` doesn't exist.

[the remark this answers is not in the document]
emasab: Current broker state is only modified by this broker. `rkb_rk->rk_broker_up_cnt` is an atomic and can already be modified concurrently by multiple brokers.

src/rdkafka_broker.c
```diff
@@ -407,8 +407,12 @@ void rd_kafka_broker_set_state(rd_kafka_broker_t *rkb, int state) {
                                  * broker connections. */
                                 rd_atomic32_add(&rkb->rkb_rk->rk_broker_up_cnt,
                                                 1);
+                                rd_kafka_broker_unlock(rkb);
```
pranavrth: I think there should be only one thread setting state at any given amount of time. Unlocking in the middle might introduce the path which is not expected. For example, 

Thread A: set state and releases lock
Thread B: acquires the lock again. Doesn't enter this condition and unlocks after the function exits.
Thread A: acquires the lock again.

The state would be inconsistent in the scenario. I would avoid unlocking inside this function.

pranavrth (changes requested):

src/rdkafka_broker.c
emasab: Yes only one thread is setting the state and it's the broker thread. This function has `@locality broker thread` and the state isn't modified by other threads and when it's read with `rd_kafka_broker_get_state` a lock is acquired.
The state is modified only after re-acquiring the lock.

We need to unlock `rkb` as the lock order is `rk` -> `rkb` for example when selecting a broker to connect to. Given here it was `rkb` -> `rk` there was a deadlock because both threads had acquired the first of the two locks and were waiting to acquire the second one.

pranavrth: There are many fields getting changed in this function. If we unlock in between and then try to acquire the lock again, we   will not be certain if any other thread would have changed the value or not. I think we should avoid unlocking here in between the function. Can we move this reset part to another function or at the end of the function somehow?

emasab: ok I'll move further down in this function after setting the state and before triggering the monitors when all fields (rkb_state and rkb_ts_state) have been written and rest of fields to read after that are modified only by the broker thread that is executing the function.

pranavrth (approved):
LGTM!. Just add a changelog in later commit regarding this.

emasab:
Thanks for the review! A changelog entry isn't needed because the fix is a commit that wasn't released

PR #5157, opened 2025-08-06
Handle Integer Overflow

ritishkumarsingh:
In this PR, the cast to int64_t ensures that the left shift operation is performed using 64-bit arithmetic, and the result will not overflow even if rkbuf_retries is equal to INT_MAX. The resulting value of the left shift operation is then multiplied by retry_backoff_ms, which is an int value. The result of the multiplication will be an int64_t value, which can hold the product of the two operands without overflow.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -2970,7 +2970,8 @@ void rd_kafka_broker_buf_retry(rd_kafka_broker_t *rkb, rd_kafka_buf_t *rkbuf) {
                 int shift = rkbuf->rkbuf_retries - 1;
 
                 /* Cap shift at 34 to prevent overflow in final calculation.
-                 * Accounts for multiplication by retry_backoff_ms (max 300000) and jitter (max 1200). */
+                 * Accounts for multiplication by retry_backoff_ms (max 300000)
+                 * and jitter (max 1200). */
                 if (shift > 34)
                         shift = 34;
 
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -2967,7 +2967,7 @@ void rd_kafka_broker_buf_retry(rd_kafka_broker_t *rkb, rd_kafka_buf_t *rkbuf) {
         /* In some cases, failed Produce requests do not increment the retry
          * count, see rd_kafka_handle_Produce_error. */
         if (rkbuf->rkbuf_retries > 0)
-                backoff = (1 << (rkbuf->rkbuf_retries - 1)) *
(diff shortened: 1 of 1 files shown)


src/rdkafka_broker.c
```diff
@@ -2957,10 +2957,18 @@ void rd_kafka_broker_buf_retry(rd_kafka_broker_t *rkb, rd_kafka_buf_t *rkbuf) {
...
         /* In some cases, failed Produce requests do not increment the retry
          * count, see rd_kafka_handle_Produce_error. */
-        if (rkbuf->rkbuf_retries > 0)
-                backoff = (1 << (rkbuf->rkbuf_retries - 1)) *
+        if (rkbuf->rkbuf_retries > 0) {
+                int shift = rkbuf->rkbuf_retries - 1;
+
+                /* Cap the shift at 62 to avoid undefined behavior for
+                 * shifts >= 64 and to prevent signed overflow for a shift of 63,
+                 * which would result in a negative backoff. */
```
milindl: nit. 

> which would result in undefined behaviour.

signed int overflow is actually not allowed to be done at all, so both cases are undefined behaviour only

src/rdkafka_broker.c
```diff
@@ -2966,10 +2966,17 @@ void rd_kafka_broker_buf_retry(rd_kafka_broker_t *rkb, rd_kafka_buf_t *rkbuf) {
...
          * count, see rd_kafka_handle_Produce_error. */
-        if (rkbuf->rkbuf_retries > 0)
-                backoff = (1 << (rkbuf->rkbuf_retries - 1)) *
+        if (rkbuf->rkbuf_retries > 0) {
+                int shift = rkbuf->rkbuf_retries - 1;
+
+                /* Cap shift at 62 to avoid undefined behavior.
+                 * shift=63 sets sign bit, shift>=64 exceeds type width. */
+                if (shift > 62)
+                        shift = 62;
```
emasab: Thanks for the PR! We should limit it further to 34. As it's multiplied by `rkb->rkb_rk->rk_conf.retry_backoff_ms` (max 300000) and then in the worst case by 1200. Given binary representation of 300000 * 1200 is 29 bits: 63 - 29 = 34. If you shift it by 35 it overflows.
Example:
```c
int64_t offset = ((int64_t)1 << 34) * 300000 * 120 * 10;
``` 
After that it's limited to `retry_backoff_max_ms` * 1000 (3600000000)

ritishkumarsingh: Nice catch!

emasab (approved):
LGTM. Thanks @ritishkumarsingh !

PR #5174, opened 2025-08-23
docs(cpp): clarify ownership of Conf::dump() return (#5166)

### LINKED ISSUE #5166: Documentation bug: Conf::dump()
The documentation of Conf::dump() method does not mention that the caller must deallocate the heap-allocated list returned by the function.
The documentation also says the function is returning "tuples". If that were true, maybe the function signature should be:

virtual std::list<std::tuple<std::string, std::string>> *dump() = 0;
OR better:
virtual std::list<std::pair<std::string, std::string>> *dump() = 0;

Anyway, the documentation should say that every odd list item is a configuration parameter name (the key) and every even list item is the value for the preceding configuration parameter name.

An enhancement idea: Create a dump() method overload that returns nothing but accepts a reference to a list.

aryanrahar:
Clarifies that RdKafka::Conf::dump() returns a heap-allocated std::list<std::string> owned by the caller.
Adds a CHANGELOG entry.
Fixes #5166.


aryanrahar:
???

aryanrahar:
Can you take a look at the PR?

PR #5175, opened 2025-08-24
Remove misleading warning log message

shimonturjeman:
The consumer rebalance protocol is GA in Apache Kafka 4.0, hence, this log message is misleading and should be removed.


shimonturjeman:
@parvathyvmohan Not sure what's the procedure for reviews... Can u maybe assist here?

PR #5177, opened 2025-08-26
Fix for KIP-1102 time based re-bootstrap condition

### LINKED ISSUE #5178: Gratuitous re-bootstrap sometimes triggered at multiples of `topic.metadata.refresh.interval.ms`
Since 2.11.0 KIP-1102 implementation triggers a re-bootstrap if no successful metadata response is received for `metadata.recovery.rebootstrap.trigger.ms`. In librdkafka this is sometimes causing that the periodic refresh every `topic.metadata.refresh.interval.ms`  (5 mins as well) overlaps with the timer for `metadata.recovery.rebootstrap.trigger.ms` and, given metadata wasn't refreshed still, a re-bootstrap is started while not being needed.

emasab:
Closes #5178 .

Re-bootstrap is now triggered only after `metadata.recovery.rebootstrap.trigger.ms`
have passed since first metadata refresh request after last successful
metadata response. The calculation was since last successful metadata response
so it's possible it did overlap with the periodic `topic.metadata.refresh.interval.ms`
and cause a re-bootstrap even if not needed.

It's now more similar to the Java implementation here:
https://github.com/apache/kafka/blob/49ee1fb4f9deb326d82b86d3a047182f7d96ddfa/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L1222

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2154,6 +2154,11 @@ static void rd_kafka_metadata_refresh_cb(rd_kafka_timers_t *rkts, void *arg) {
         rd_kafka_t *rk = rkts->rkts_rk;
         rd_kafka_resp_err_t err;
 
+        /* Restart the rebootstrap timer only if it's the first
+         * metadata refresh request after last successful response,
+         * so the timer is not reset if already scheduled. */
+        rd_kafka_rebootstrap_tmr_restart(rk);
+
         /* High-level consumer:
          * We need to query both locally known topics and subscribed topics
          * so that we can detect locally known topics changing partition
@@ -2833,6 +2838,10 @@ fail:
 
 /**
  * Schedules a rebootstrap of the cluster immediately.
+ *
+ * @locks none
+ * @locks_acquired rd_kafka_timers_lock()
+ * @locality any
  */
 void rd_kafka_rebootstrap(rd_kafka_t *rk) {
         if (rk->rk_conf.metadata_recovery_strategy ==
@@ -2851,9 +2860,11 @@ void rd_kafka_rebootstrap(rd_kafka_t *rk) {
 }
 
 /**
- * Restarts rebootstrap timer with the configured interval.
+ * Restarts rebootstrap timer with the configured interval. Only if not
+ * started or stopped.
  *
  * @locks none
+ * @locks_acquired rd_kafka_timers_lock()
  * @locality any
  */
 void rd_kafka_rebootstrap_tmr_restart(rd_kafka_t *rk) {
@@ -2862,9 +2873,27 @@ void rd_kafka_rebootstrap_tmr_restart(rd_kafka_t *rk) {
                 return;
 
         rd_kafka_timer_start_oneshot(
-            &rk->rk_timers, &rk->rebootstrap_tmr, rd_true /*restart*/,
+            &rk->rk_timers, &rk->rebootstrap_tmr, rd_false /*don't restart*/,
             rk->rk_conf.metadata_recovery_rebootstrap_trigger_ms * 1000LL,
             rd_kafka_rebootstrap_tmr_cb, NULL);
+}
+
+/**
+ * Stops rebootstrap timer, for example after a successful metadata response.
+ *
+ * @return 1 if the timer was started (before being stopped), else 0.
+ *
+ * @locks none
+ * @locks_acquired rd_kafka_timers_lock()
+ * @locality any
+ */
+int rd_kafka_rebootstrap_tmr_stop(rd_kafka_t *rk) {
+        if (rk->rk_conf.metadata_recovery_strategy ==
+            RD_KAFKA_METADATA_RECOVERY_STRATEGY_NONE)
+                return 0;
+
+        return rd_kafka_timer_stop(&rk->rk_timers, &rk->rebootstrap_tmr,
+                                   rd_true /* lock */);
 }
 
 /**
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2855,14 +2855,14 @@ void rd_kafka_rebootstrap(rd_kafka_t *rk) {
 }
 
 /**
- * Restarts rebootstrap timer with the configured interval. Only if not
+ * Starts rebootstrap timer with the configured interval. Only if not
  * started or stopped.
  *
  * @locks none
  * @locks_acquired rd_kafka_timers_lock()
  * @locality any
  */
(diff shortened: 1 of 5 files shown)


src/rdkafka.c
```diff
@@ -2154,6 +2154,11 @@ static void rd_kafka_metadata_refresh_cb(rd_kafka_timers_t *rkts, void *arg) {
         rd_kafka_t *rk = rkts->rkts_rk;
         rd_kafka_resp_err_t err;
 
+        /* Restart the rebootstrap timer only if it's the first
```
pranavrth: The `topic.metadata.refresh.interval.ms` can be -1 which means that this can be disabled. I think we should find a way to make it work in hat condition as well. As I was suggesting earlier in the call that we should restart the timer on all the internal metadata call. We should explore that  as well I think.

src/rdkafka.c
```diff
@@ -2862,11 +2873,29 @@ void rd_kafka_rebootstrap_tmr_restart(rd_kafka_t *rk) {
...
             rd_kafka_rebootstrap_tmr_cb, NULL);
 }
 
+/**
+ * Stops rebootstrap timer, for example after a successful metadata response.
+ *
+ * @return 1 if the timer was started (before being stopped), else 0.
+ *
+ * @locks none
+ * @locks_acquired rd_kafka_timers_lock()
```
pranavrth, line 2882: I don't think its necessary to acquire the lock for this timer. I get that the implementation of `rd_kafka_timer_stop` has the lock so I think we should think about removing the lock from there (with modification to the timer maybe) in some future release.

src/rdkafka.c
```diff
@@ -2862,11 +2873,29 @@ void rd_kafka_rebootstrap_tmr_restart(rd_kafka_t *rk) {
...
+ * @locality any
+ */
+int rd_kafka_rebootstrap_tmr_stop(rd_kafka_t *rk) {
+        if (rk->rk_conf.metadata_recovery_strategy ==
+            RD_KAFKA_METADATA_RECOVERY_STRATEGY_NONE)
+                return 0;
+
+        return rd_kafka_timer_stop(&rk->rk_timers, &rk->rebootstrap_tmr,
+                                   rd_true /* lock */);
```
pranavrth, line 2891: Why do we need to lock this? Everything related to this timer is being called from `main thread`

emasab: Because `rd_kafka_timer_unschedule` calls `TAILQ_REMOVE(&rkts->rkts_timers, rtmr, rtmr_link);` and it's not already locked.

src/rdkafka.c
emasab: See below

src/rdkafka.c
emasab: Changed it a bit so the timer is started when it's ensured there are retries on retriable errors.

src/rdkafka.c
pranavrth: Its fine for now but I think we should improve this in a future release. I think we can omit this lock some how. I have added this to our task list.

src/rdkafka.c
```diff
@@ -2851,9 +2855,11 @@ void rd_kafka_rebootstrap(rd_kafka_t *rk) {
...
 /**
- * Restarts rebootstrap timer with the configured interval.
+ * Restarts rebootstrap timer with the configured interval. Only if not
+ * started or stopped.
  *
  * @locks none
+ * @locks_acquired rd_kafka_timers_lock()
  * @locality any
  */
 void rd_kafka_rebootstrap_tmr_restart(rd_kafka_t *rk) {
```
pranavrth: Better to name it `rd_kafka_rebootstrap_tmr_try_start` now as we don't ever restart it.

```suggestion
void rd_kafka_rebootstrap_tmr_try_start(rd_kafka_t *rk) {
```

src/rdkafka_request.c
```diff
@@ -2877,9 +2878,23 @@ rd_kafka_MetadataRequest0(rd_kafka_broker_t *rkb,
                 rkbuf->rkbuf_u.Metadata.decr = full_incr;
                 rkbuf->rkbuf_u.Metadata.decr_lock =
                     &rkb->rkb_rk->rk_metadata_cache.rkmc_full_lock;
+        } else if (!resp_cb) {
+                /* In case it's a full request, forced or not, it won't be
+                 * retried on the same broker to avoid blocking metadata
+                 * requests, because of the lock, when that broker isn't
+                 * available. We don't start the timer as we cannot ensure the
+                 * request is retried for the duration of
+                 * `metadata.recovery.rebootstrap.trigger.ms`.
```
pranavrth: Let's reference that we are talking about the `if` case.

src/rdkafka_request.c
```diff
@@ -2877,9 +2878,23 @@ rd_kafka_MetadataRequest0(rd_kafka_broker_t *rkb,
...
+                 *
+                 * Same reasoning applies when we use a custom callback, for the
+                 * AdminClient requests for example.
```
pranavrth, line 2907: Let's reference that we are talking about `!resp_cb` case.

src/rdkafka_request.c
```diff
@@ -2877,9 +2878,23 @@ rd_kafka_MetadataRequest0(rd_kafka_broker_t *rkb,
...
+                 *
+                 * Restart the rebootstrap timer only if it's the first
+                 * metadata refresh request after last successful response,
+                 * so the timer is not reset if already scheduled. */
+                rd_kafka_rebootstrap_tmr_restart(rkb->rkb_rk);
```
pranavrth: I am fine with this. Just check the lock ordering and impact of taking locking in all the metadata calls.

pranavrth (changes requested):

src/rdkafka.c
emasab: how about `rd_kafka_rebootstrap_tmr_start_maybe`, as in other places?

pranavrth: Yes thats fine as well.

src/rdkafka_request.c
emasab: It's always called with no locks taken, and no other locks are acquired after the timers lock. Added a summary of locks required and acquired.

pranavrth (approved):
LGTM!. Thanks @emasab.

src/rdkafka.c
emasab: Timers are scheduled from different threads: main thread, broker threads. We can remove the lock later if we remove the broker threads.

PR #5189, opened 2025-09-10
Fix flakyness of tests 0014 and 0085

emasab:
Second commit solves a frequent error appearing on tests `0014-reconsume`: `Broker: Not coordinator` even if the RPC call has a permanent error it will be retried after refreshing the group coordinator so it doesn't need to be notified to the user.

### DIFF
--- a/src/rdkafka_partition.c
+++ b/src/rdkafka_partition.c
@@ -2162,6 +2162,7 @@ static rd_kafka_op_res_t rd_kafka_toppar_op_serve(rd_kafka_t *rk,
                 rd_kafka_toppar_lock(rktp);
 
                 if (rko->rko_err) {
+                        int actions;
                         rd_kafka_dbg(
                             rktp->rktp_rkt->rkt_rk, TOPIC, "OFFSET",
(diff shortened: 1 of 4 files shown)


src/rdkafka_partition.c
```diff
@@ -2176,10 +2177,15 @@ static rd_kafka_op_res_t rd_kafka_toppar_op_serve(rd_kafka_t *rk,
                         rd_kafka_toppar_unlock(rktp);
 
 
-                        /* Propagate error to application */
+                        actions = rd_kafka_handle_OffsetFetch_err_action(
+                            NULL, rko->rko_err, NULL);
+                        /* Propagate error to application. Exclude
+                         * permanent errors that caused a coordinator
```
milindl, line 2183:
```suggestion
                         * permanent errors that cause a coordinator
```

emasab: In this case they just caused it because it's getting the actions corresponding to the error from the `RD_KAFKA_OP_OFFSET_FETCH` reply

PR #5198, opened 2025-09-22
K2 testing final

Ankith-Confluent:
This PR refactors the librdkafka test suite to improve cloud compatibility by adding configurable options in `test.conf`. This enables better test portability and maintainability across different Kafka environments.

### **1. Configurable Test Framework**
- **`test.skip.idempotent`**: Skip tests requiring idempotent producer (replaces `-I` flag)
- **`test.supported.acks`**: Configure supported acknowledgment values (e.g., `-1,0,1`)
- **`test.sleep.multiplier`**: Configurable sleep multiplier for different environments. Multiplies explicit sleep_for() delays for cluster state propagation (set 0 to skip sleeps), different from test.timeout.multiplier which multiplies API operation timeouts (can't be disabled).
- **`test.timeout.multiplier`**: Unified timeout scaling (existing feature, now consistently used)
- **`test.skip.numbers`**: Skip specific tests by number (comma-separated list)

### **2. Topic Cleanup**
- Many tests were creating topics, but not deleting them.
- Added topic delete utility.

### **3. Replication Factor**
- Changed hardcoded replication factors to use broker defaults (`-1`)

### **4. Preprocessor Directives for Version Compatibility**
- Added preprocessor directives to support librdkafka versions 2.0 through 2.11

## 🔧 Configuration Examples

```ini
test.timeout.multiplier=2.0
test.sleep.multiplier=3.0
test.supported.acks=-1,1
test.skip.idempotent=true
test.skip.numbers=0011,0055,0063,0068,0081,0086

```


k-raina (changes requested):
Please update PR description, since there are large amount of changes. It will help with review

tests/0001-multiobj.c
```diff
@@ -91,9 +93,9 @@ int main_0001_multiobj(int argc, char **argv) {
                 TIMING_STOP(&t_full);
 
                 /* Topic is created on the first iteration. */
-                if (i > 0)
-                        TIMING_ASSERT(&t_full, 0, 999);
-                else
+                if (i > 0) 
+                        TIMING_ASSERT(&t_full, 0, tmout_multip(999));
```
k-raina, line 99: Instead of depending on sleep and guesstimates, can we write a function which waits on topic creation/metadata propagation?

tests/0009-mock_cluster.c
```diff
@@ -89,6 +89,7 @@ int main_0009_mock_cluster(int argc, char **argv) {
         test_consumer_poll("CONSUME", c, 0, -1, 0, msgcnt, NULL);
 
         rd_kafka_destroy(c);
+
```
k-raina: nit

tests/0008-reqacks.c
```diff
@@ -175,5 +189,6 @@ int main_0008_reqacks(int argc, char **argv) {
                 rd_kafka_destroy(rk);
         }
 
+
```
k-raina: nit

tests/0030-offset_commit.c
```diff
@@ -539,49 +539,54 @@ static void do_nonexist_commit(void) {
...
-                       0 /* enable.auto.commit */,
-                       0 /* enable.auto.offset.store */, 1 /* sync */,
-                       1 /* use subscribe */);
 
-        do_offset_test("MANUAL.COMMIT.SYNC & MANUAL.STORE",
-                       0 /* enable.auto.commit */,
-                       0 /* enable.auto.offset.store */, 0 /* sync */,
-                       1 /* use subscribe */);
 
         rd_free(topic);
```
k-raina, line 598: Do we need remove this?

tests/0013-null-msgs.c
```diff
@@ -462,6 +464,7 @@ static void test_produce_consume(void) {
         consume_messages_with_queues(testid, topic, partition_cnt, msgcnt);
         verify_consumed_msg_check();
 
+
```
k-raina: Nit: Maybe try to remove these additional new lines from other places as well?

tests/0040-io_event.c
```diff
@@ -106,6 +106,22 @@ int main_0040_io_event(int argc, char **argv) {
         pfd.fd      = fds[0];
         pfd.events  = POLLIN;
         pfd.revents = 0;
+        
+        /* Handle initial rebalance by polling consumer queue directly */
+        for (int i = 0; i < 3; i++) {
```
k-raina: Was this needed for K1 also?

tests/0041-fetch_max_bytes.c
```diff
@@ -93,5 +94,11 @@ int main_0041_fetch_max_bytes(int argc, char **argv) {
         rd_kafka_topic_destroy(rkt);
         rd_kafka_destroy(rk);
 
+        /* Delete the topic */
+        {
```
k-raina: Why is this code needed?

tests/0042-many_topics.c
```diff
@@ -244,6 +247,14 @@ int main_0042_many_topics(int argc, char **argv) {
                 assign_consume_many(topics, topic_cnt, testid);
         }
 
+        /* Delete all topics */
+        {
+                rd_kafka_t *del_rk = test_create_handle(RD_KAFKA_PRODUCER, NULL);
+                for (i = 0; i < topic_cnt; i++) {
```
k-raina: Loop doesn't make sense?

tests/0045-subscribe_update.c
```diff
@@ -245,7 +281,10 @@ static void do_test_non_exist_and_partchange(void) {
          * - Increase the partition count
          * - Verify updated assignment
          */
-        test_kafka_topics("--alter --topic %s --partitions 4", topic_a);
+        test_create_partitions(rk, topic_a, 4);
+        
+        test_sleep(2);
```
k-raina: This file has multiple `test_sleep` . Do we need to add multiplier here also?

k-raina (changes requested):
First round of review. Left some comments!

tests/0001-multiobj.c
Ankith-Confluent: Actually this test is related to latency issues, its not related to metadata propagation.
So the time needs to be extended for K2 related environments.

tests/0030-offset_commit.c
Ankith-Confluent: Actually I just shifted the block up, I will edit it back to normal with preprocessor directive

tests/0041-fetch_max_bytes.c
Ankith-Confluent: I think some of the changes were not reverted, will revert this, since its not needed 👍

tests/0042-many_topics.c
Ankith-Confluent: This too should have been reverted, I missed it

tests/0045-subscribe_update.c
Ankith-Confluent: Yes for this test, I added it because it was very Flaky 
The flakiness was solved using these

tests/0040-io_event.c
Ankith-Confluent: Yes it is required

tests/0050-subscribe_adds.c
```diff
@@ -33,6 +33,40 @@
 
 #include <stdarg.h>
 
+/**
+ * @brief Version-aware partition list printing that avoids leader epoch APIs 
+ *        on older versions
+ */
+static void safe_print_partition_list(
```
k-raina: This function seems to be defined at multiple places, should we move it to utility?

tests/0054-offset_time.cpp
```diff
@@ -61,13 +61,16 @@ static int verify_offset(const RdKafka::TopicPartition *tp,
...
+  int64_t current_time = (int64_t)ts.tv_sec * 1000 + ts.tv_usec / 1000;
   std::string topic = Test::mk_topic_name("0054-offset_time", 1);
   RdKafka::Conf *conf, *tconf;
   int64_t timestamps[] = {
       /* timestamp, expected offset */
-      1234,
+      current_time,
       0,
-      999999999999,
+      current_time + 500,
```
k-raina, line 73: why this change was needed?

tests/0076-produce_retry.c
```diff
@@ -409,32 +409,46 @@ static void do_test_produce_retry_invalid_msg(rd_kafka_mock_cluster_t *mcluster,
...
-                /* Idempotence, try fail, should succeed. */
-                do_test_produce_retries_disconnect(topic, 1, 1, 0);
-        }
         /* No idempotence, try fail, should fail. */
         do_test_produce_retries_disconnect(topic, 0, 1, 1);
 
         return 0;
 }
 
+int main_0076_produce_retry_idempotent(int argc, char **argv) {
```
k-raina, line 424: Is this function used?

tests/0081-admin.c
```diff
@@ -787,10 +863,8 @@ static void do_test_AlterConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
             configs[ci], "max.compaction.lag.ms", "3600000");
         TEST_ASSERT(!err, "%s", rd_kafka_err2str(err));
 
-        if (test_broker_version >= TEST_BRKVER(2, 7, 0, 0))
-                exp_err[ci] = RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART;
-        else
-                exp_err[ci] = RD_KAFKA_RESP_ERR_UNKNOWN;
+        /* Cloud/managed brokers typically return UNKNOWN_TOPIC_OR_PART regardless of version */
+        exp_err[ci] = RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART;
```
k-raina, line 874: Why is this change needed?

tests/0081-admin.c
```diff
@@ -1047,10 +1148,8 @@ static void do_test_IncrementalAlterConfigs(rd_kafka_t *rk,
             RD_KAFKA_ALTER_CONFIG_OP_TYPE_SET, "3600000");
         TEST_ASSERT(!error, "%s", rd_kafka_error_string(error));
 
-        if (test_broker_version >= TEST_BRKVER(2, 7, 0, 0))
-                exp_err[ci] = RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART;
-        else
-                exp_err[ci] = RD_KAFKA_RESP_ERR_UNKNOWN;
+        /* Cloud/managed brokers typically return UNKNOWN_TOPIC_OR_PART regardless of version */
+        exp_err[ci] = RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART;
```
k-raina, line 1195: Same: why is this needed?

tests/0081-admin.c
```diff
@@ -1344,7 +1466,7 @@ static void do_test_DescribeConfigs(rd_kafka_t *rk, rd_kafka_queue_t *rkqu) {
                                     exp_err[i], rd_kafka_err2name(err),
                                     errstr2 ? errstr2 : "");
                                 rd_kafka_event_destroy(rkev);
-                                rd_sleep(1);
+                                test_sleep(1);
```
k-raina: Why is this change needed?

tests/0081-admin.c
```diff
@@ -1693,7 +1817,7 @@ do_test_DescribeAcls(rd_kafka_t *rk, rd_kafka_queue_t *useq, int version) {
             test_CreateAcls_simple(rk, NULL, acl_bindings_create, 2, NULL);
 
         /* Wait for ACL propagation. */
-        rd_sleep(tmout_multip(2));
+        test_sleep(2);
```
k-raina: tmout_multip will assure that test waits for longer. Why is it removed?

tests/0082-fetch_max_bytes.cpp
```diff
@@ -78,10 +88,23 @@ static void do_test_fetch_max_bytes(void) {
...
    * larger than fetch.max.bytes.
+   * 
+   * K2 clusters have a higher minimum requirement for receive.message.max.bytes
+   * (4MB vs 1MB), so we adjust all fetch limits proportionally for K2 clusters.
    */
+  /* K2 clusters require higher receive.message.max.bytes minimum (4MB vs 1MB) */
   Test::conf_set(conf, "max.partition.fetch.bytes", "20000000"); /* ~20MB */
-  Test::conf_set(conf, "fetch.max.bytes", "1000000");            /* ~1MB */
-  Test::conf_set(conf, "receive.message.max.bytes", "1000512");  /* ~1MB+512 */
+  if (test_k2_cluster) {
```
k-raina: test_k2_cluster -> This should be removed as discussed

tests/0082-fetch_max_bytes.cpp
```diff
@@ -117,7 +148,7 @@ static void do_test_fetch_max_bytes(void) {
 
     delete msg;
   }
-  Test::Say("Done\n");
+  Test::Say(tostr() << "Done - consumed " << cnt << " messages successfully");
```
k-raina, line 131: i liked detailed logging!

tests/testshared.h
```diff
@@ -59,6 +60,12 @@ extern int tmout_multip(int msecs);
 /** @brief true if tests should run in quick-mode (faster, less data) */
 extern int test_quick;
 
+/** @brief true if tests should run in K2 cluster mode (acks=-1, higher limits) */
+extern int test_k2_cluster;
```
k-raina: test_k2_cluster -> This was added as a quick way to change configuration when running with k2 cluster, it should be removed now.

k-raina (changes requested):
Thanks for changes! 
Few comments for rest of files.

tests/0050-subscribe_adds.c
Ankith-Confluent: Yes will make the changes 👍

tests/0076-produce_retry.c
Ankith-Confluent: Actually this change was made my Milind I think
Basically the idempotent part was shifted to another function so that it can be skipped through a config

Basically creating two new tests, 76 and 76_idempotent

tests/0081-admin.c
Ankith-Confluent: This change was to make the sleep configurable, so test_sleep will take the sleep multiplier config and multiply the value, this part needs more sleep time

tests/0081-admin.c
Ankith-Confluent: test_sleep is the new function which does the same job, we have a new config test.sleep.multiplier which is used in test_sleep instead of test.timeout.multiplier

tests/0082-fetch_max_bytes.cpp
Ankith-Confluent: Yes I missed this, will make the changes

tests/0081-admin.c
k-raina: As discussed
- Add comment for need of`test.sleep.multiplier` in test.conf
- Maybe change test_sleep and base_sleep_ms names

tests/0054-offset_time.cpp
Ankith-Confluent: Its required for K1 and K2, and passes for AK

tests/0001-multiobj.c
```diff
@@ -60,6 +60,8 @@ int main_0001_multiobj(int argc, char **argv) {
                 if (!topic)
                         topic = test_mk_topic_name("0001", 0);
 
+                test_create_topic_if_auto_create_disabled(NULL, topic, -1);
```
milindl: Shouldn't this be done only once when we're in the `if (!topic)` block above?

tests/0004-conf.c
```diff
@@ -511,26 +511,19 @@ int main_0004_conf(int argc, char **argv) {
...
+            "compression.codec", "gzip", /* S2I property */
 #endif
 #if defined(_WIN32)
-            "ssl.ca.certificate.stores",
-            "Intermediate ,, Root ,",
+            "ssl.ca.certificate.stores", "Intermediate ,, Root ,",
 #endif
-            "client.dns.lookup",
-            "resolve_canonical_bootstrap_servers_only",
+            /* client.dns.lookup was introduced in librdkafka 2.2.0+ - skip
```
milindl, line 525: nit: Remove this comment, it's only useful for this review

tests/0007-autotopic.c
```diff
@@ -85,22 +85,25 @@ int main_0007_autotopic(int argc, char **argv) {
...
-
         /* Set delivery report callback */
         rd_kafka_conf_set_dr_cb(conf, dr_cb);
 
         /* Create kafka instance */
         rk = test_create_handle(RD_KAFKA_PRODUCER, conf);
 
-        rkt = rd_kafka_topic_new(rk, test_mk_topic_name("0007_autotopic", 1),
-                                 topic_conf);
+        const char *topic = test_mk_topic_name("0007_autotopic", 1);
```
milindl: We can't have new variables like this, we must declare `topic` at the top of the block

tests/0008-reqacks.c
```diff
@@ -96,9 +96,23 @@ int main_0008_reqacks(int argc, char **argv) {
             "all brokers!\033[0m\n");
 
         /* Try different request.required.acks settings (issue #75) */
-        for (reqacks = -1; reqacks <= 1; reqacks++) {
+        /* Test all standard acks values, but skip unsupported ones */
+        int start_acks = -1;
```
milindl: declare these with `i` at the top of the block

tests/test.c
```diff
@@ -853,12 +886,138 @@ int test_set_special_conf(const char *name, const char *val, int *timeoutp) {
...
+
+        /* Not supported - test should be skipped */
+        return NULL;
+}
+
+/**
+ * @brief Sleep with configurable multiplier (only if multiplier > 0)
+ * @param wait_time Sleep time in seconds
+ */
+void sleep_for(int wait_time) {
```
milindl: Rename this to test_<something>, see the other comments made on other test files regarding this function

tests/0011-produce_batch.c
```diff
@@ -114,7 +116,12 @@ static void test_single_partition(void) {
         TEST_SAY("test_single_partition: Created kafka instance %s\n",
                  rd_kafka_name(rk));
 
-        rkt = rd_kafka_topic_new(rk, test_mk_topic_name("0011", 0), topic_conf);
+        topic = test_mk_topic_name("0011", 0);
+        test_create_topic_if_auto_create_disabled(rk, topic, 3);
+
+        sleep_for(5);
```
milindl: Suggestion, here and everywhere else in the PR:

Instead of using sleep_for etc., create a function called test_wait_for_metadata_propagation(N) - you can probably come up with a better name.
The function can just call rd_sleep internally.

Two reasons: first, it's self-documenting, second if at some point in the future we figure out a smarter way to ensure propagation, we can just rewrite one function instead of figuring out which sleep_for is metadata propagation and which isn't.

tests/0011-produce_batch.c
```diff
@@ -628,11 +650,16 @@ static void test_message_single_partition_record_fail(int variation) {
 
         SUB_TEST_QUICK();
 
-        const char *confs_set_append[] = {"cleanup.policy", "APPEND",
-                                          "compact"};
+        // Modified for Confluent Cloud compatibility:
```
milindl: nit: in most c files, we use `/*` for comments only

milindl: Also, `Modified for Confluent Cloud compatibility` is for the reviewer of the PR, not for someone who reads the code in the future. It's better to say why you made this change and point to relevant documentation links

tests/0011-produce_batch.c
milindl: In cases where you're waiting for someone else, like a config change, you can keep using sleep_for, just add a comment about what you're waiting for.

tests/0022-consume_batch.c
```diff
@@ -133,7 +139,6 @@ static void do_test_consume_batch(void) {
                 }
 
                 rd_kafka_topic_destroy(rkts[i]);
-                rd_free(topics[i]);
```
milindl, line 136: Why was this free removed?

tests/0033-regex_subscribe.c
```diff
@@ -304,33 +308,48 @@ static int do_test(const char *assignor) {
...
+                    test_mk_topic_name(
+                        tsprintf("regex_subscribe_NONEXISTENT_0004_IV#!_%s",
+                                 test_run_id),
+                        0));
 
         /* Produce messages to topics to ensure creation. */
-        for (i = 0; i < topic_cnt; i++)
+        for (i = 0; i < topic_cnt; i++) {
+                test_create_topic_if_auto_create_disabled(NULL, topics[i], 1);
+                test_wait_topic_exists(NULL, topics[i], tmout_multip(10000));
```
milindl, line 340: This should be after produce, in the case where we're relying on autocreate. 
But instead of this why not just create the topics explicitly all the time, would that be possible and easier?

tests/0033-regex_subscribe.c
```diff
@@ -304,33 +308,48 @@ static int do_test(const char *assignor) {
...
+                sleep_for(3);
                 test_produce_msgs_easy(topics[i], testid, RD_KAFKA_PARTITION_UA,
```
milindl, line 342: Further, can we call sleep_for once after he topic creation loop ends instead of every time? Or does that cause issues?

tests/0038-performance.c
```diff
@@ -59,15 +59,23 @@ int main_0038_performance(int argc, char **argv) {
...
         msgcnt = totsize / msgsize;
 
-        TEST_SAY("Producing %d messages of size %d to %s [%d]\n", msgcnt,
-                 (int)msgsize, topic, partition);
+        /* Use acks=1 for performance test */
+        if (!test_is_acks_supported("1")) {
+                TEST_SKIP("acks=1 not supported by this cluster\n");
+                return 0;
+        }
+        const char *acks_value = "1";
```
milindl: Variable declaration at top of block

tests/0040-io_event.c
```diff
@@ -92,6 +91,7 @@ int main_0040_io_event(int argc, char **argv) {
         queue = rd_kafka_queue_get_consumer(rk_c);
 
         test_consumer_subscribe(rk_c, topic);
+        sleep_for(5);
```
milindl: What is this particular sleep for?

tests/0040-io_event.c
```diff
@@ -107,6 +107,27 @@ int main_0040_io_event(int argc, char **argv) {
         pfd.events  = POLLIN;
         pfd.revents = 0;
 
+        /* Handle initial rebalance by polling consumer queue directly */
```
milindl: Why is it necessary to use the consumer queue and not the fd (because I see rebalance handling there too)?

tests/0042-many_topics.c
```diff
@@ -234,8 +234,11 @@ int main_0042_many_topics(int argc, char **argv) {
 
         /* Generate unique topic names */
         topics = malloc(sizeof(*topics) * topic_cnt);
-        for (i = 0; i < topic_cnt; i++)
+        for (i = 0; i < topic_cnt; i++) {
                 topics[i] = rd_strdup(test_mk_topic_name(__FUNCTION__, 1));
+                test_create_topic_if_auto_create_disabled(NULL, topics[i], -1);
+                sleep_for(3);
```
milindl: I think this sleep_for could be outside the loop, right?

tests/0044-partition_cnt.c
```diff
@@ -61,17 +61,25 @@ static void test_producer_partition_cnt_change(void) {
         rd_kafka_conf_set_dr_msg_cb(conf, test_dr_msg_cb);
         rk = test_create_handle(RD_KAFKA_PRODUCER, conf);
 
-        test_create_topic_wait_exists(rk, topic, partition_cnt / 2, 1, 5000);
+        int topic_wait_timeout = tmout_multip(5000);
```
milindl: Move the declaration of ints at the top of the block here and for msg_timeout_ms

tests/0044-partition_cnt.c
```diff
@@ -61,17 +61,25 @@ static void test_producer_partition_cnt_change(void) {
...
+        sleep_for(3);
+        int msg_timeout_ms = tmout_multip(10000);
 
-        rkt =
-            test_create_topic_object(rk, topic, "message.timeout.ms",
-                                     tsprintf("%d", tmout_multip(10000)), NULL);
+
+        rkt = test_create_topic_object(
+            rk, topic, "message.timeout.ms",
+            tsprintf("%d", tmout_multip(msg_timeout_ms)), NULL);
```
milindl: Isn't msg_timeout_ms already inclusive of the multiplier?

tests/0045-subscribe_update.c
```diff
@@ -725,65 +784,84 @@ static void do_test_resubscribe_with_regex() {
...
         rk    = test_create_consumer(group, NULL, conf, NULL);
         queue = rd_kafka_queue_get_consumer(rk);
 
+        sleep_for(3);
+
         /* Subscribe to topic1 */
         TEST_SAY("Subscribing to %s\n", topic1);
         test_consumer_subscribe(rk, topic1);
+
+        sleep_for(3);
```
milindl: Question: In this test, why are we sleeping after subscribe and unsubscribe - there is no propagation step needed for subscribe/unsubscribe, so, where is the extra time requirement coming from? Are the await_assignment and await_revocations timing out? I see that test timeout multiplier is  not used in await_*, it's just 30

tests/0045-subscribe_update.c
```diff
@@ -392,23 +417,49 @@ static void do_test_topic_remove(void) {
                     rd_kafka_err2str(err));
         rd_kafka_topic_partition_list_destroy(topics);
 
+        /* Version-specific wait for assignment */
+        if (rd_kafka_version() >= 0x020100ff) {
```
milindl, line 423: Question: why is the wait version specific, what changed?

tests/0059-bsearch.cpp
```diff
@@ -99,11 +99,10 @@ class MyDeliveryReportCb : public RdKafka::DeliveryReportCb {
 
     if (!msg.msg_opaque())
       return;
-
     RdKafka::MessageTimestamp ts = msg.timestamp();
-    if (ts.type != RdKafka::MessageTimestamp::MSG_TIMESTAMP_CREATE_TIME)
+    if (ts.type != RdKafka::MessageTimestamp::MSG_TIMESTAMP_CREATE_TIME &&
```
milindl, line 105: is this a confirmed issue in K2? If so add a comment. AK behaves as expected, if we have timestamp type set as create time correctly, then it will return it correctly also.

tests/0102-static_group_rebalance.c
```diff
@@ -104,17 +104,29 @@ static void rebalance_cb(rd_kafka_t *rk,
                          void *opaque) {
         _consumer_t *c = opaque;
 
-        TEST_ASSERT(c->expected_rb_event == err,
-                    "line %d: %s: Expected rebalance event %s got %s\n",
-                    c->curr_line, rd_kafka_name(rk),
-                    rd_kafka_err2name(c->expected_rb_event),
-                    rd_kafka_err2name(err));
+        /* Accept both REVOKE and ASSIGN as valid rebalance events during
+         * unsubscribe Some clusters may send ASSIGN directly instead of REVOKE
```
milindl: nit
```suggestion
         * unsubscribe. Some clusters may send ASSIGN directly instead of REVOKE.
```

tests/0102-static_group_rebalance.c
```diff
@@ -104,17 +104,29 @@ static void rebalance_cb(rd_kafka_t *rk,
...
+         */
+        if (c->expected_rb_event == RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS &&
+            err == RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS) {
+                TEST_SAY(
+                    "line %d: %s: Got ASSIGN instead of expected REVOKE "
+                    "(acceptable behavior)\n",
+                    c->curr_line, rd_kafka_name(rk));
+                /* Accept ASSIGN as valid alternative to REVOKE */
```
milindl:
```suggestion

```

tests/0113-cooperative_rebalance.cpp
```diff
@@ -907,22 +915,28 @@ static void b_subscribe_with_cb_test(rd_bool_t close_consumer) {
...
+                         tmout_multip(10 * 1000));
+  sleep_for(5);
 
   Test::subscribe(c1, topic_name);
 
   bool c2_subscribed = false;
   while (true) {
-    Test::poll_once(c1, 500);
-    Test::poll_once(c2, 500);
+    /* Version-specific poll timeouts for cooperative rebalancing */
```
milindl: Please add the reason for doing this in the comment, instead of stating what's done in the code

tests/testshared.h
```diff
@@ -412,5 +420,30 @@ const char *test_consumer_group_protocol();
...
+
+#ifdef _RDKAFKA_H_
+rd_kafka_resp_err_t test_DeleteTopics_simple(rd_kafka_t *rk,
+                                             rd_kafka_queue_t *useq,
+                                             char **topics,
+                                             size_t topic_cnt,
+                                             void *opaque);
+
+void test_delete_topic_simple(rd_kafka_t *rk, const char *topic_name);
+#endif
```
milindl:
```suggestion
#endif // _RDKAFKA_H_
```

milindl (commented):
Mostly looks OK, left some minor comments.

tests/0022-consume_batch.c
Ankith-Confluent: I had implemented delete topics functionality after the test is over, but we found a better approach so I had to revert it. I removed this also by mistake, will revert it.

tests/0059-bsearch.cpp
Ankith-Confluent: Yes its confirmed bug in K2

tests/0045-subscribe_update.c
Ankith-Confluent: The test was flaky in the older versions, and adding this sleep resolved the issue. The newer versions no longer require this sleep.

tests/0033-regex_subscribe.c
Ankith-Confluent: Yes creating the topic explicitly all the time is better and easier I think, will change it 👍

tests/0045-subscribe_update.c
Ankith-Confluent: Actually in 2.0 version rd_kafka_destroy() was hanging, I had added these sleeps for debugging, but forgot to remove them. Also I will add multiplier to timeouts

tests/0042-many_topics.c
Ankith-Confluent: I think it was getting stuck while creating topics, so a short wait was required. We can probably reduce the sleep time from 3 to 1.

tests/0040-io_event.c
Ankith-Confluent: The test was getting stuck because we subscribe at line 93 (which triggers a rebalance event), but we don't set up the fd monitoring until line 104. So it was added to handle that first event (initial rebalance).

tests/0040-io_event.c
Ankith-Confluent: The test was flaky in older versions so I added this to fix it. (metadata propogation)

tests/0033-regex_subscribe.c
Ankith-Confluent: The test was getting stuck while creating topics, so a short wait was required. We can probably reduce the sleep time from 3 to 1.

tests/0040-io_event.c
milindl: We shouldn't get a rebalance event just by subscribing, unless someone is polling the queue or the rk. There seems to be something going wrong somewhere.

Ankith-Confluent: Oh, that’s interesting.
I did some testing and found that the test passes on AK (without the polling code) but hangs on K2. It might be causing a race condition, since it did pass once on K2 when I reduced the sleep multiplier.
I moved the subscribe code to run after enabling the IO events, and now it passes on both. Is this change valid?

PR #5202, opened 2025-09-23
Remove performance regression when we're subscribing with a regex

emasab:
and none of many topics in the cluster are matching it.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -5415,6 +5415,64 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
         return rd_atomic32_add(&rkcg->rkcg_subscription_version, 1);
 }
 
+static rd_bool_t
+rd_kafka_cgrp_stale_subscribed_topics_cache(rd_kafka_cgrp_t *rkcg) {
+        rd_bool_t stale = rd_false;
+        int32_t current_subscription_version =
+            rd_atomic32_get(&rkcg->rkcg_subscription_version);
+        rd_ts_t current_ts_metadata = rkcg->rkcg_rk->rk_ts_metadata;
+
+        stale = rkcg->rkcg_subscribed_topics_cache.rk_ts_metadata !=
+                    current_ts_metadata ||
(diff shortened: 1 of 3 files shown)


src/rdkafka_cgrp.c
```diff
@@ -5415,6 +5415,64 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
         return rd_atomic32_add(&rkcg->rkcg_subscription_version, 1);
 }
 
+static rd_bool_t
+rd_kafka_cgrp_stale_subscribed_topics_cache(rd_kafka_cgrp_t *rkcg) {
+        rd_bool_t stale = rd_false;
+        int32_t current_subscription_version =
+            rd_atomic32_get(&rkcg->rkcg_subscription_version);
+        rd_ts_t current_ts_metadata = rkcg->rkcg_rk->rk_ts_metadata;
```
k-raina, line 5423: Is it possible that cache is changed however `rk_ts_metadata` is left unchanged?

src/rdkafka_metadata.c
```diff
@@ -1236,13 +1237,16 @@ rd_kafka_metadata_topic_match(rd_kafka_t *rk,
...
         size_t cnt = 0;
         rd_kafka_topic_partition_list_t *unmatched;
-        rd_list_t cached_topics;
-        const char *topic;
+        const struct rd_kafka_metadata_cache_entry *rkmce;
+        map_str_str_t map;
 
         rd_kafka_rdlock(rk);
+        map = (map_str_str_t)RD_MAP_INITIALIZER(
+            0, rd_map_str_cmp, rd_map_str_hash, NULL /* topic list element */,
```
k-raina, line 1248: In this code snippet, 0 as EXPECTED_CNT argument of RD_MAP_INITIALIZER, flows through `rd_map_init` into `rd_map_alloc_buckets(expected_cnt)`, and selects fixed 199-bucket default. Should this bucket size be proportional as topics count?

k-raina (changes requested):
Thanks for PR @emasab 
Left some comments.

PR #5207, opened 2025-09-29
Revert "[KIP-320] Validate assigned partitions before starting to consume from them (#4931)

emasab:
This reverts commit 13a2bbae88d8be073986f974db1b853cfc91ae8f.

Reason is there are issues with the consumer group given the partition isn't started when validating it and the result is a validation success. In that case the partition can continue to be fetched even if it was revoked.


pranavrth (approved):
LGTM!.

PR #5213, opened 2025-10-06
Fix: rd_kafka_consume_batch incorrectly advances offset by 2 on EOF messages

### LINKED ISSUE #4844: Offset increased for eof pseudo messages
Description
===========
When using enable.partition.eof=true, the offset returned by rd_kafka_position() after consuming the artificial eof message is the last real message's offset + 2. This changed at some point between 2.0.2 and 2.2.0.

How to reproduce
================
Set enable.partition.eof=true, consume an eof message, and call rd_kafka_position().

**IMPORTANT**: Always try to reproduce the issue on the latest released version (see https://github.com/confluentinc/librdkafka/releases), if it can't be reproduced on the latest version the issue has been fixed.

Checklist
=========

 - [x] librdkafka version (release number or git tag): broken in >= 2.2.0; bug is not present in 2.0.2
 - [ ] Apache Kafka version: `<REPLACE with e.g., 0.10.2.3>`
 - [ ] librdkafka client configuration: `<REPLACE with e.g., message.timeout.ms=123, auto.reset.offset=earliest, ..>`
 - [ ] Operating system: `<REPLACE with e.g., Centos 5 (x64)>`
 - [ ] Critical issue

Ankith-Confluent:
Closes #4844 

### Summary

This PR fixes a bug in `rd_kafka_consume_batch()` where EOF messages incorrectly advance the consumer position by 2 instead of 1 when `enable.partition.eof=true`. The bug causes `rd_kafka_position()` to return last_offset + 2 instead of the correct last_offset + 1 after consuming an EOF message in batch mode.

### Motivation

- Addresses this bug https://github.com/confluentinc/librdkafka/issues/4844
- Prevents potential message loss or duplicate consumption due to incorrect offset positioning
- Bug has existed since v2.2.0 (introduced in commit [51397194a63a4bbf07caa757c84f8a86edf7af5b](https://github.com/confluentinc/librdkafka/commit/51397194a63a4bbf07caa757c84f8a86edf7af5b))

### Root Cause

In `src/rdkafka_queue.c`, the `rd_kafka_q_serve_rkmessages()` function updates the consumer position for all messages in a batch, including EOF messages. EOF messages have error code `RD_KAFKA_RESP_ERR__PARTITION_EOF` but are not control messages, so they were incorrectly incrementing the offset.

### Changes

1. File: `src/rdkafka_queue.c`
Added a check to prevent EOF messages from updating the consumer position.

2. Regression subtest: `tests/0137-barrier_batch_consume.c`
New Subtest `do_test_consume_batch_eof_position()` is added

### Test Flow
1. Produces 5 messages to a test topic
2. Consumes messages using `rd_kafka_consume_batch_queue()` with 
   `enable.partition.eof=true`
3. Receives EOF message after consuming all real messages
4. Verifies `rd_kafka_position()` returns `last_offset + 1` (not `last_offset + 2`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,11 @@ librdkafka v2.13.1 is a maintenance release:
   The increased CPU usage (~30%) was seen in particular when there are many topics
   in the clusters and the given subscription regex doesn't match any.
   Happening since 2.10.0 (#5324).
-* Issues: #5213.
+* Issues: #4844.
   Fix `rd_kafka_consume_batch_queue` incorrectly updating the application
   position when receiving EOF or error messages, causing the position to
-  move backward. Aligned the position update logic with
-  `rd_kafka_fetch_op_app_prepare` by filtering non-FETCH and error ops
-  and comparing positions with leader epoch awareness.
+  move forward and likely be stored and committed.
+  When storing the application offset the leader epoch is also considered for correct offset ordering in case of log truncation.
   Happening since 2.2.0 (#5213).
 
 


src/rdkafka_queue.c
```diff
@@ -829,7 +829,10 @@ int rd_kafka_q_serve_rkmessages(rd_kafka_q_t *rkq,
                 rko = (rd_kafka_op_t *)rkmessages[i]->_private;
                 rd_kafka_toppar_t *rktp = rko->rko_rktp;
                 int64_t offset          = rkmessages[i]->offset + 1;
-                if (unlikely(rktp && (rktp->rktp_app_pos.offset < offset)))
+                /* Only update position for messages that are not EOF */
+                if (unlikely(rktp && (rktp->rktp_app_pos.offset < offset) &&
+                             (rkmessages[i]->err !=
+                              RD_KAFKA_RESP_ERR__PARTITION_EOF)))
                         rd_kafka_update_app_pos(
```
emasab: Let's use for `rd_kafka_update_app_pos`: 
```C
RD_KAFKA_FETCH_POS(
     rko->rko_u.fetch.rkm.rkm_rkmessage.offset + 1,
     rko->rko_u.fetch.rkm.rkm_u.consumer.leader_epoch
)
```    
like in `rd_kafka_fetch_op_app_prepare` to avoid an additional call to `rd_kafka_message_leader_epoch`.
We need to do the strictly increasing check with the leader epoch as well using `rd_kafka_fetch_pos_cmp` instead of `rktp->rktp_app_pos.offset < offset`.

src/rdkafka_queue.c
```diff
@@ -829,7 +829,10 @@ int rd_kafka_q_serve_rkmessages(rd_kafka_q_t *rkq,
                 rko = (rd_kafka_op_t *)rkmessages[i]->_private;
                 rd_kafka_toppar_t *rktp = rko->rko_rktp;
                 int64_t offset          = rkmessages[i]->offset + 1;
-                if (unlikely(rktp && (rktp->rktp_app_pos.offset < offset)))
+                /* Only update position for messages that are not EOF */
+                if (unlikely(rktp && (rktp->rktp_app_pos.offset < offset) &&
+                             (rkmessages[i]->err !=
+                              RD_KAFKA_RESP_ERR__PARTITION_EOF)))
```
emasab: this should be like the condition that we had before and that is in `rd_kafka_fetch_op_app_prepare`:
```suggestion
                             (!rko->rko_err && rko->rko_type == RD_KAFKA_OP_FETCH)))
```

src/rdkafka_queue.c
emasab: Same at :859 about `rd_kafka_fetch_pos_cmp`, the error check can be avoided there as control messages aren't errors.

tests/0137-barrier_batch_consume.c
```diff
@@ -608,12 +608,145 @@ static void do_test_consume_batch_control_msgs(void) {
...
+        rd_kafka_message_t **rkmessages;
+        int msg_cnt, i;
+        int64_t last_real_offset = -1;
+        int64_t eof_offset       = -1;
+        int64_t position_after_eof;
+        rd_kafka_topic_partition_list_t *positions;
+        rd_kafka_resp_err_t err;
+        int eof_received = 0;
+
+        SUB_TEST("Testing EOF position with consume_batch_queue");
```
emasab: This can be considered a quick test (usually < 60s)
```suggestion
        SUB_TEST_QUICK("Testing EOF position with consume_batch_queue");
```

tests/0137-barrier_batch_consume.c
```diff
@@ -608,12 +608,145 @@ static void do_test_consume_batch_control_msgs(void) {
...
+        const char *topic;
+        rd_kafka_t *consumer;
+        rd_kafka_conf_t *conf;
+        rd_kafka_queue_t *rkq;
+        uint64_t testid;
+        const int partition_cnt             = 1;
+        const int partition                 = 0;
+        const int produce_msg_cnt           = 5;
+        const int consume_msg_cnt           = 10;
+        const int timeout_ms                = 5000;
```
emasab: 1s is enough to consume everything
```suggestion
        const int timeout_ms                = 1000;
```

CHANGELOG.md
```diff
@@ -14,6 +16,13 @@ librdkafka v2.13.1 is a maintenance release:
   The increased CPU usage (~30%) was seen in particular when there are many topics
   in the clusters and the given subscription regex doesn't match any.
   Happening since 2.10.0 (#5324).
+* Issues: #5213.
```
```suggestion
* Issues: #4844.
```

CHANGELOG.md
```diff
@@ -14,6 +16,13 @@ librdkafka v2.13.1 is a maintenance release:
...
+  Fix `rd_kafka_consume_batch_queue` incorrectly updating the application
+  position when receiving EOF or error messages, causing the position to
+  move backward. Aligned the position update logic with
+  `rd_kafka_fetch_op_app_prepare` by filtering non-FETCH and error ops
+  and comparing positions with leader epoch awareness.
```
```suggestion
  move forward and likely be stored and committed.
  When storing the application offset the leader epoch is also considered for correct offset ordering in case of log truncation.
```

emasab (changes requested):

PR #5233, opened 2025-11-07
Add configurable SASL Kerberos hostname override (Java-style kerberos.domain.name)

HAIKUOTIANKONG132:
Introduce sasl.kerberos.domain.name, mirroring the Java client’s kerberos.domain.name so librdkafka users can target a fixed hostname portion in Kerberos service principals. Document the new property and register it in the config tables.
Update both SASL paths to honor the override:
Cyrus (Linux/macOS): pass the configured hostname to sasl_client_new() when present.
Win32 SSPI: build the SPN as service/override instead of the broker-advertised hostname.
This allows environments that rely on alternative hostnames (e.g., /etc/hosts aliases) to request the same SPN that the Java clients use.


weimengxi:
I have met the same problem while adapting HuaWei Cloud MRS Kafka. 
https://support.huaweicloud.com/intl/zh-cn/trouble-mrs/mrs_03_0223.html

Issue #4897, opened 2024-11-12
Missing shared-object file "apline-librdkafka.so" under "runtimes/linux-arm64/native/" folder in librdkafka.redist package 


KaiWesterschwiensterdt:
Description
===========
When I create a Docker image using an Alpine Linux as the base image and use my .NET project which has the Confluent.Kafka package installed (which has a dependency on the librdkafka.redist package), I encounter an error while running the project. The error is due to the missing shared object file "alpine-librdkafka.so" under the "runtimes/linux-arm64/native/" folder in the librdkafka.redist package.

Logs and additional information about this error are well documented in this confluent-kafka-dotnet issue:
https://github.com/confluentinc/confluent-kafka-dotnet/issues/778

How to reproduce
================

1. Create a Dockerfile with Alpine Linux as the base image.
2. Install the Confluent.Kafka package in a .NET project.
3. Build the Docker image.
4. Run the Docker container.
5. Observe the error indicating the missing shared object file while produce/consume messages.

Checklist
=========
 - [x] librdkafka version: v2.6.0
 - [ ] Apache Kafka version: v3.9.0
 - [ ] librdkafka client configuration: any
 - [ ] Operating system: Alpine Linux ARM64
 - [ ] Provide logs 
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

Workaround in Dockerfile
=========
Just copy and rename the file manual in your Dockerfile.

```csharp
FROM mcr.microsoft.com/dotnet/aspnet:8.0-alpine-arm64v8
RUN apk add librdkafka

EXPOSE 8080

WORKDIR /app
COPY . .

# Delete librdkafka.so in build output
RUN rm -f /app/librdkafka.so

# Copy and rename shared object file
RUN cp /usr/lib/librdkafka.so.1 /app/alpine-librdkafka.so
# [...]
```

How to fix
=========
Copy _apline-librdkafka.so_ file in _runtimes/linux-arm64/native/_ folder while creating _librdkafka.redist_ package.

KaiWesterschwiensterdt:
Any news on this?

dominik-niebuhr:
will this ever be tackled?

csrakowski:
For visibility here: Due to some issues with Semaphore CI on public PRs, the changes for this were moved into https://github.com/confluentinc/librdkafka/pull/5329, which was merged about an hour ago.

PR #5237, opened 2025-11-11
Support librdkafka on Alpine arm64

mclayton7:
Currently in Confluent's .NET client, when running on Alpine platforms with musl, the .NET library looks for the native `alpine-librdkafka.so` in `./linux-x64/native/alpine-librdkafka.so`, but the ARM64 equivalent does not exist. This PR hopes to address that.

See:
https://github.com/confluentinc/confluent-kafka-dotnet/issues/2341
https://github.com/confluentinc/librdkafka/issues/4897


lucasneuwit:
Hello! I'm actively waiting for this fix to be available. Do you think it will be merged soon?

Thanks!

PR #5244, opened 2025-11-13
Fix build failure due to linting from ssl_normalize_hostname change

fangnx:
Move to unit test

PR #5253, opened 2025-11-27
Strip trailing dot of hostname to fix SSL certificate verification issue (correct setup w/ unit test)

fangnx:
Combining the original PR to address the trailing dot bug (https://github.com/confluentinc/librdkafka/pull/5231) which was reverted as it was breaking master and the PR to add unit test (https://github.com/confluentinc/librdkafka/pull/5244)

Jira: https://confluentinc.atlassian.net/browse/NONJAVACLI-4101

### DIFF
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -2240,4 +2240,3 @@ int unittest_ssl(void) {
 
         return fails;
 }
-#endif /* WITH_SSL */


src/rdkafka_ssl.c
```diff
@@ -449,6 +450,43 @@ static int rd_kafka_transport_ssl_cert_verify_cb(int preverify_ok,
...
+ * @returns The normalized hostname (same as \p normalized)
+ *
+ * @remark This function is exposed for testing via ENABLE_DEVEL.
+ */
+#if ENABLE_DEVEL
+RD_EXPORT
+#else
+static
+#endif
+const char *rd_kafka_ssl_normalize_hostname(const char *hostname,
```
emasab: This is just static, the unit test runs from the same file.
```suggestion
static const char *rd_kafka_ssl_normalize_hostname(const char *hostname,
```

src/rdkafka_ssl.c
```diff
@@ -2127,3 +2184,66 @@ int rd_kafka_ssl_hmac(rd_kafka_broker_t *rkb,
 
         return 0;
 }
+
+#if WITH_SSL
```
emasab: Remove this condition as the `rdkafka_ssl.c` file is already included only when SSL is enabled. In Makefile:
```
SRCS_$(WITH_SSL) += rdkafka_ssl.c
```

emasab (changes requested):

src/rdkafka_ssl.c
naxinf: Good catch, will remove 👍

PR #5275, opened 2025-12-16
[KIP-932] Remove offset management from the assignment path

PratRanj07:
Remove calls of OffsetCommits, OffsetFetch and List Groups and filter assignment serve for Share Groups

### DIFF
--- a/src/rdkafka_assignment.c
+++ b/src/rdkafka_assignment.c
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
         rd_kafka_topic_partition_list_destroy(offsets);
 }
 
+static int rd_kafka_share_assignment_serve_removals(rd_kafka_t *rk) {
+        rd_kafka_topic_partition_t *rktpar;
+
+        RD_KAFKA_TPLIST_FOREACH(rktpar, rk->rk_consumer.assignment.removed) {
+                rd_kafka_toppar_t *rktp =
+                    rd_kafka_topic_partition_ensure_toppar(
+                        rk, rktpar, rd_true); /* Borrow ref */
+
+                /* Remove from pending list if present */
+                rd_kafka_topic_partition_list_del(
+                    rk->rk_consumer.assignment.pending, rktpar->topic,
+                    rktpar->partition);
+
+                /* Send PARTITION_LEAVE to cgrp to trigger cleanup.
+                 * This calls rd_kafka_cgrp_partition_del() which sends
+                 * SHARE_SESSION_PARTITION_REMOVE to the broker. */
+                if (rktp->rktp_started) {
+                        /* Partition was started, detach from cgrp. */
+                        rd_assert(rk->rk_consumer.assignment.started_cnt > 0);
+
+                        rd_kafka_toppar_lock(rktp);
+                        rd_assert(rktp->rktp_cgrp);
+                        rd_kafka_cgrp_op(rktp->rktp_cgrp, rktp,
+                                         RD_KAFKA_NO_REPLYQ,
+                                         RD_KAFKA_OP_PARTITION_LEAVE, 0);
+                        rktp->rktp_cgrp = NULL;
+                        rd_kafka_toppar_unlock(rktp);
+                }
+
+                rd_kafka_toppar_lock(rktp);
+
+                /* No longer desired */
+                rd_kafka_toppar_desired_del(rktp);
+
+                rd_assert((rktp->rktp_flags & RD_KAFKA_TOPPAR_F_ASSIGNED));
+                rktp->rktp_flags &= ~RD_KAFKA_TOPPAR_F_ASSIGNED;
+
+                rd_kafka_toppar_unlock(rktp);
+
+                rd_kafka_dbg(rk, CGRP, "REMOVE",
+                             "Removing %s [%" PRId32 "] from share assignment",
+                             rktpar->topic, rktpar->partition);
+        }
+
+        rd_kafka_topic_partition_list_clear(rk->rk_consumer.assignment.removed);
+
+        return rk->rk_consumer.assignment.wait_stop_cnt;
+}
+
 
 /**
  * @brief Decommission all partitions in the removed list.
@@ -432,6 +481,39 @@ static int rd_kafka_assignment_serve_removals(rd_kafka_t *rk) {
 
         return rk->rk_consumer.assignment.wait_stop_cnt +
                rk->rk_consumer.wait_commit_cnt;
+}
+
+static int rd_kafka_share_assignment_serve_pending(rd_kafka_t *rk) {
+        int i;
+
+        for (i = rk->rk_consumer.assignment.pending->cnt - 1; i >= 0; i--) {
+                rd_kafka_topic_partition_t *rktpar =
+                    &rk->rk_consumer.assignment.pending->elems[i];
+                rd_kafka_toppar_t *rktp =
+                    rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
+
+                rd_kafka_dbg(rk, CGRP, "SRVPEND",
+                             "Adding share partition %s [%" PRId32
+                             "] to assignment",
+                             rktpar->topic, rktpar->partition);
+
+                /* Attach partition to cgrp and send PARTITION_JOIN.
+                 * This triggers SHARE_SESSION_PARTITION_ADD to the broker. */
+                rktp->rktp_started = rd_true;
+                rk->rk_consumer.assignment.started_cnt++;
+
+                rd_kafka_toppar_lock(rktp);
+                rd_assert(!rktp->rktp_cgrp);
+                rktp->rktp_cgrp = rk->rk_cgrp;
+                rd_kafka_toppar_unlock(rktp);
+
+                rd_kafka_cgrp_op(rk->rk_cgrp, rktp, RD_KAFKA_NO_REPLYQ,
+                                 RD_KAFKA_OP_PARTITION_JOIN, 0);
+                rd_kafka_topic_partition_list_del_by_idx(
+                    rk->rk_consumer.assignment.pending, i);
+        }
+
+        return rk->rk_consumer.assignment.pending->cnt;
 }
 
 
@@ -601,7 +683,43 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
                rk->rk_consumer.assignment.queried->cnt;
 }
 
+/**
+ * @brief Serve updates to the share assignment.
+ *
+ * Call on:
+ * - assignment changes
+ */
+void rd_kafka_share_assignment_serve(rd_kafka_t *rk) {
+        int inp_removals = 0;
+        int inp_pending  = 0;
 
+        rd_kafka_assignment_dump(rk);
+
+        /* Serve any partitions that should be removed */
+        if (rk->rk_consumer.assignment.removed->cnt > 0)
+                inp_removals = rd_kafka_share_assignment_serve_removals(rk);
+
+        /* Serve pending partitions */
+        if (rk->rk_consumer.assignment.wait_stop_cnt == 0 &&
+            inp_removals == 0 && rk->rk_consumer.assignment.pending->cnt > 0)
+                inp_pending = rd_kafka_share_assignment_serve_pending(rk);
+
+        if (inp_removals + inp_pending +
+                rk->rk_consumer.assignment.wait_stop_cnt ==
+            0) {
+                /* No assignment operations in progress,
+                 * signal assignment done back to cgrp */
+                rd_kafka_cgrp_assignment_done(rk->rk_cgrp);
+        } else {
+                rd_kafka_dbg(rk, CGRP, "ASSIGNMENT",
+                             "Share assignment of %d partition(s) "
+                             "with %d pending adds,%d partitions awaiting stop "
+                             "and %d pending removals",
+                             rk->rk_consumer.assignment.all->cnt, inp_pending,
+                             rk->rk_consumer.assignment.wait_stop_cnt,
+                             inp_removals);
+        }
+}
 
 /**
  * @brief Serve updates to the assignment.
@@ -612,6 +730,10 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
  * - partition fetcher is stopped
  */
 void rd_kafka_assignment_serve(rd_kafka_t *rk) {
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
+                rd_kafka_share_assignment_serve(rk);
+                return;
+        }
         int inp_removals = 0;
         int inp_pending  = 0;
 
--- a/src/rdkafka_assignment.c
+++ b/src/rdkafka_assignment.c
@@ -349,6 +349,7 @@ static void rd_kafka_share_assignment_serve_removals(rd_kafka_t *rk) {
                 /* Detach the partition from the cgrp directly. This sends
                  * SHARE_SESSION_PARTITION_REMOVE to the partition's broker
                  * so it queues a forget on the next ShareFetch. */
+                rd_dassert(rktp->rktp_cgrp);
                 rd_kafka_cgrp_partition_del(rktp->rktp_cgrp, rktp);
 
                 /* desired_del requires both topic_wrlock and toppar_lock. */
--- a/src/rdkafka_assignment.c
+++ b/src/rdkafka_assignment.c
@@ -349,7 +349,6 @@ static void rd_kafka_share_assignment_serve_removals(rd_kafka_t *rk) {
                 /* Detach the partition from the cgrp directly. This sends
                  * SHARE_SESSION_PARTITION_REMOVE to the partition's broker
                  * so it queues a forget on the next ShareFetch. */
-                rd_dassert(rktp->rktp_cgrp);
                 rd_kafka_cgrp_partition_del(rktp->rktp_cgrp, rktp);
 
                 /* desired_del requires both topic_wrlock and toppar_lock. */
@@ -490,9 +489,7 @@ static void rd_kafka_share_assignment_serve_pending(rd_kafka_t *rk) {
                              "] to assignment",
                              rktpar->topic, rktpar->partition);
 
-                rd_kafka_toppar_lock(rktp);
                 rktp->rktp_cgrp = rk->rk_cgrp;
-                rd_kafka_toppar_unlock(rktp);
 
                 rd_kafka_cgrp_partition_add(rktp->rktp_cgrp, rktp);
                 rd_kafka_topic_partition_list_del_by_idx(
@@ -872,11 +869,6 @@ rd_kafka_share_assignment_add(rd_kafka_t *rk,
                 rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
         }
 
-        for (i = 0; i < partitions->cnt; i++) {
-                rd_kafka_topic_partition_t *rktpar = &partitions->elems[i];
-                rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
-        }
-
         rd_kafka_topic_partition_list_add_list(rk->rk_consumer.assignment.all,
                                                partitions);
         if (!was_empty)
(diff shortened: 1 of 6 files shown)


src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+
+                /* Remove from pending list if present */
+                rd_kafka_topic_partition_list_del(
+                    rk->rk_consumer.assignment.pending, rktpar->topic,
+                    rktpar->partition);
+
+                /* Send PARTITION_LEAVE to cgrp to trigger cleanup.
+                 * This calls rd_kafka_cgrp_partition_del() which sends
+                 * SHARE_SESSION_PARTITION_REMOVE to the broker. */
+                if (rktp->rktp_started) {
```
pranavrth: 1. We don't need `rktp_started` handling.
2. Moreover, the below code is not part of this if condition in normal `assignment_serve_removals`
3. Remove rktp->rktp_started handling and make sure share consumer is not affected by this handling.

src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+                        /* Partition was started, detach from cgrp. */
+                        rd_assert(rk->rk_consumer.assignment.started_cnt > 0);
+
+                        rd_kafka_toppar_lock(rktp);
+                        rd_assert(rktp->rktp_cgrp);
```
pranavrth: dassert

src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+                        rd_kafka_cgrp_op(rktp->rktp_cgrp, rktp,
+                                         RD_KAFKA_NO_REPLYQ,
+                                         RD_KAFKA_OP_PARTITION_LEAVE, 0);
```
pranavrth: This op is handled by main thread and we are already in main thread. Let's directly call `rd_kafka_cgrp_partition_del` here itself.

src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+                /* Send PARTITION_LEAVE to cgrp to trigger cleanup.
+                 * This calls rd_kafka_cgrp_partition_del() which sends
+                 * SHARE_SESSION_PARTITION_REMOVE to the broker. */
+                if (rktp->rktp_started) {
+                        /* Partition was started, detach from cgrp. */
+                        rd_assert(rk->rk_consumer.assignment.started_cnt > 0);
+
+                        rd_kafka_toppar_lock(rktp);
+                        rd_assert(rktp->rktp_cgrp);
+                        rd_kafka_cgrp_op(rktp->rktp_cgrp, rktp,
```
pranavrth: I think we don't need `rd_kafka_toppar_purge_internal_fetch_queue_maybe` in `rd_kafka_cgrp_partition_del`. Verify and remove that handling.

src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+                                         RD_KAFKA_NO_REPLYQ,
+                                         RD_KAFKA_OP_PARTITION_LEAVE, 0);
+                        rktp->rktp_cgrp = NULL;
+                        rd_kafka_toppar_unlock(rktp);
+                }
+
+                rd_kafka_toppar_lock(rktp);
+
+                /* No longer desired */
+                rd_kafka_toppar_desired_del(rktp);
```
pranavrth, line 360: I think this handling is incorrect as the function asks to acquire topic lock as well. We should do what is asked by the function. This is existing issue in librdkafka. Let's fix this for share consumer and discuss internally for other places.

src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+
+                rd_assert((rktp->rktp_flags & RD_KAFKA_TOPPAR_F_ASSIGNED));
+                rktp->rktp_flags &= ~RD_KAFKA_TOPPAR_F_ASSIGNED;
```
pranavrth: We can remove this as well. Verify and check if required.

src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+
+                rd_kafka_toppar_unlock(rktp);
+
+                rd_kafka_dbg(rk, CGRP, "REMOVE",
+                             "Removing %s [%" PRId32 "] from share assignment",
+                             rktpar->topic, rktpar->partition);
+        }
+
```
pranavrth, line 369: Keep log. Remove valid offsets part.

```
rd_kafka_dbg(rk, CONSUMER | RD_KAFKA_DBG_CGRP, "REMOVE",
                     "Served %d removed partition(s), "
                     "with %d offset(s) to commit",
                     rk->rk_consumer.assignment.removed->cnt, valid_offsets);
```

src/rdkafka_assignment.c
```diff
@@ -612,6 +730,10 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
  * - partition fetcher is stopped
  */
 void rd_kafka_assignment_serve(rd_kafka_t *rk) {
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
```
pranavrth, line 705: Handle "enable.auto.commit" conf initialization. We shouldn't let anyone initialize share consumer with this property.

src/rdkafka_assignment.c
```diff
@@ -332,6 +332,55 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+                rd_kafka_toppar_unlock(rktp);
+
+                rd_kafka_dbg(rk, CGRP, "REMOVE",
+                             "Removing %s [%" PRId32 "] from share assignment",
+                             rktpar->topic, rktpar->partition);
+        }
+
+        rd_kafka_topic_partition_list_clear(rk->rk_consumer.assignment.removed);
+
+        return rk->rk_consumer.assignment.wait_stop_cnt;
```
pranavrth: Remove handling of

- queried list
- wait_stop_cnt
- wait_commit_cnt
- started_cnt
- version

We don't return anything from this function essentially.

src/rdkafka_assignment.c
```diff
@@ -612,6 +730,10 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
  * - partition fetcher is stopped
  */
 void rd_kafka_assignment_serve(rd_kafka_t *rk) {
```
pranavrth, line 704: Apart from this function, we need to make sure that we remove offset management from `assignment_add` and `assignment_subtract` part as well.

src/rdkafka_assignment.c
```diff
@@ -601,7 +683,43 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
...
+        int inp_pending  = 0;
+
+        rd_kafka_assignment_dump(rk);
+
+        /* Serve any partitions that should be removed */
+        if (rk->rk_consumer.assignment.removed->cnt > 0)
+                inp_removals = rd_kafka_share_assignment_serve_removals(rk);
 
+        /* Serve pending partitions */
+        if (rk->rk_consumer.assignment.wait_stop_cnt == 0 &&
```
pranavrth: Remove `wait_stop_cnt`

src/rdkafka_assignment.c
```diff
@@ -434,6 +483,39 @@ static int rd_kafka_assignment_serve_removals(rd_kafka_t *rk) {
...
+
+                rd_kafka_dbg(rk, CGRP, "SRVPEND",
+                             "Adding share partition %s [%" PRId32
+                             "] to assignment",
+                             rktpar->topic, rktpar->partition);
+
+                /* Attach partition to cgrp and send PARTITION_JOIN.
+                 * This triggers SHARE_SESSION_PARTITION_ADD to the broker. */
+                rktp->rktp_started = rd_true;
+                rk->rk_consumer.assignment.started_cnt++;
```
pranavrth: Remove these.

src/rdkafka_assignment.c
```diff
@@ -434,6 +483,39 @@ static int rd_kafka_assignment_serve_removals(rd_kafka_t *rk) {
...
+
+                rd_kafka_toppar_lock(rktp);
+                rd_assert(!rktp->rktp_cgrp);
+                rktp->rktp_cgrp = rk->rk_cgrp;
+                rd_kafka_toppar_unlock(rktp);
+
+                rd_kafka_cgrp_op(rk->rk_cgrp, rktp, RD_KAFKA_NO_REPLYQ,
+                                 RD_KAFKA_OP_PARTITION_JOIN, 0);
```
pranavrth: Call "rd_kafka_cgrp_partition_add" directly here.

src/rdkafka_assignment.c
```diff
@@ -601,7 +683,43 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
...
+        rd_kafka_assignment_dump(rk);
+
+        /* Serve any partitions that should be removed */
+        if (rk->rk_consumer.assignment.removed->cnt > 0)
+                inp_removals = rd_kafka_share_assignment_serve_removals(rk);
 
+        /* Serve pending partitions */
+        if (rk->rk_consumer.assignment.wait_stop_cnt == 0 &&
+            inp_removals == 0 && rk->rk_consumer.assignment.pending->cnt > 0)
+                inp_pending = rd_kafka_share_assignment_serve_pending(rk);
```
pranavrth: We don't have any use of `inp_pending`. We can remove this field.

src/rdkafka_assignment.c
```diff
@@ -601,7 +683,43 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
...
+
+        if (inp_removals + inp_pending +
+                rk->rk_consumer.assignment.wait_stop_cnt ==
```
pranavrth: This count will always be 0 in our case. We can just log that we are calling assignment done.

src/rdkafka_assignment.c
```diff
@@ -601,7 +683,43 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
...
+            0) {
+                /* No assignment operations in progress,
+                 * signal assignment done back to cgrp */
+                rd_kafka_cgrp_assignment_done(rk->rk_cgrp);
+        } else {
```
pranavrth: Remove this else case.

pranavrth (changes requested):
First Pass comments.

src/rdkafka_conf.c
```diff
@@ -4333,6 +4333,33 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                         }
                 }
 
+                /* Share-consumer-specific config validation.
+                 * Share consumers are broker-driven (ShareGroupHeartbeat);
+                 * the application does not control partition assignment, so
+                 * the rebalance callback / event has no semantics here.
+                 * Auto-commit also doesn't apply — share consumers don't
+                 * track offsets locally. */
+                if (conf->share.is_share_consumer) {
```
pranavrth, line 4347: Use -> `RD_KAFKA_IS_SHARE_CONSUMER` instead

src/rdkafka_conf.c
```diff
@@ -4333,6 +4333,33 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                        /* Reject explicit enable.auto.commit=true; silently
+                         * force the default to false otherwise. */
+                        if (rd_kafka_conf_is_modified(conf,
+                                                      "enable.auto.commit")) {
+                                if (conf->enable_auto_commit)
+                                        return "`enable.auto.commit` must be "
+                                               "false for share consumers";
+                        } else {
+                                conf->enable_auto_commit = 0;
+                        }
```
pranavrth, line 4370: We need to make sure that this is not being used in `ShareConsumer` flow.

src/rdkafka_conf.c
```diff
@@ -4333,6 +4333,33 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                 * Auto-commit also doesn't apply — share consumers don't
+                 * track offsets locally. */
+                if (conf->share.is_share_consumer) {
+                        if (conf->rebalance_cb)
+                                return "`rebalance_cb` is not supported "
+                                       "for share consumers";
+
+                        if (conf->enabled_events & RD_KAFKA_EVENT_REBALANCE)
+                                return "`RD_KAFKA_EVENT_REBALANCE` event is "
+                                       "not supported for share consumers";
```
pranavrth, line 4359: The behavior is undefined right now. We might want to enable this though not tested. Let's come back on this after understanding.

src/rdkafka_assignment.c
```diff
@@ -966,6 +1231,14 @@ void rd_kafka_assignment_pause(rd_kafka_t *rk, const char *reason) {
  */
 void rd_kafka_assignment_resume(rd_kafka_t *rk, const char *reason) {
 
+        /**
+         * TODO KIP-932: Check how can remove the invocation
+         * of this function altogether for share consumers.
+         */
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
```
pranavrth, line 1237: I think let's add same in pause as well.

src/rdkafka_assignment.c
```diff
@@ -716,9 +908,14 @@ int rd_kafka_assignment_clear(rd_kafka_t *rk) {
 rd_kafka_error_t *
 rd_kafka_assignment_add(rd_kafka_t *rk,
                         rd_kafka_topic_partition_list_t *partitions) {
-        rd_bool_t was_empty = rk->rk_consumer.assignment.all->cnt == 0;
+        rd_bool_t was_empty;
```
pranavrth: Revert this harmless change.

src/rdkafka_assignment.c
```diff
@@ -704,6 +831,71 @@ int rd_kafka_assignment_clear(rd_kafka_t *rk) {
...
+                /* Materialize the toppar (creating topic + toppar if
+                 * missing) and cache it in rktpar->_private so the next
+                 * loop can reuse the lookup. */
+                rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
+        }
+
+        for (i = 0; i < partitions->cnt; i++) {
+                rd_kafka_topic_partition_t *rktpar = &partitions->elems[i];
+                rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
+        }
```
pranavrth: We don't need this. We are already ensuring above.

src/rdkafka_assignment.c
```diff
@@ -434,6 +476,30 @@ static int rd_kafka_assignment_serve_removals(rd_kafka_t *rk) {
...
+                    rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
+
+                rd_kafka_dbg(rk, CGRP, "SRVPEND",
+                             "Adding share partition %s [%" PRId32
+                             "] to assignment",
+                             rktpar->topic, rktpar->partition);
+
+                rd_kafka_toppar_lock(rktp);
+                rktp->rktp_cgrp = rk->rk_cgrp;
+                rd_kafka_toppar_unlock(rktp);
```
pranavrth: I was thinking if we need this? We can remove this part as well I think. Even if we cant remove this, we should be able to update this without the lock from ShareConsumer's perspective.

src/rdkafka_assignment.c
```diff
@@ -332,6 +333,47 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+
+                /* Remove from pending list if present */
+                rd_kafka_topic_partition_list_del(
+                    rk->rk_consumer.assignment.pending, rktpar->topic,
+                    rktpar->partition);
+
+                /* Detach the partition from the cgrp directly. This sends
+                 * SHARE_SESSION_PARTITION_REMOVE to the partition's broker
+                 * so it queues a forget on the next ShareFetch. */
+                rd_dassert(rktp->rktp_cgrp);
```
pranavrth: Check and remove this handling as well.

src/rdkafka_assignment.c
```diff
@@ -332,6 +333,47 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
         rd_kafka_topic_partition_list_destroy(offsets);
 }
 
+static void rd_kafka_share_assignment_serve_removals(rd_kafka_t *rk) {
```
pranavrth, line 336: Check if we require - `rktpar->metadata`

src/rdkafka_assignment.c
```diff
@@ -332,6 +333,47 @@ static void rd_kafka_assignment_handle_OffsetFetch(rd_kafka_t *rk,
...
+                 * so it queues a forget on the next ShareFetch. */
+                rd_dassert(rktp->rktp_cgrp);
+                rd_kafka_cgrp_partition_del(rktp->rktp_cgrp, rktp);
+
+                /* desired_del requires both topic_wrlock and toppar_lock. */
+                rd_kafka_topic_wrlock(rktp->rktp_rkt);
+                rd_kafka_toppar_lock(rktp);
+
+                /* No longer desired */
+                rd_kafka_toppar_desired_del(rktp);
```
pranavrth, line 360: Understand if we need anything from `rd_kafka_toppar_broker_leave_for_remove`

pranavrth (changes requested):

src/rdkafka_conf.c
PratRanj07: We cant use that as rd_kafka_t is not available here

src/rdkafka_conf.c
PratRanj07: Added a TODO for now

src/rdkafka_assignment.c
PratRanj07: I think we need this. I have removed the lock

src/rdkafka_assignment.c
PratRanj07: rktp_stored_metadata is only written in the offset functions only in the share consumer. So adding it would be a no op as we wouldn't be calling these in the Share Consumer path

src/rdkafka_assignment.c
PratRanj07: No we dont need this. Added a check to stop calling this

src/rdkafka_partition.c
```diff
@@ -683,7 +683,8 @@ void rd_kafka_toppar_desired_del(rd_kafka_toppar_t *rktp) {
                      "Removing (un)desired topic %s [%" PRId32 "]",
                      rktp->rktp_rkt->rkt_topic->str, rktp->rktp_partition);
 
-        if (rktp->rktp_flags & RD_KAFKA_TOPPAR_F_UNKNOWN) {
+        if (rktp->rktp_flags & RD_KAFKA_TOPPAR_F_UNKNOWN &&
+            !RD_KAFKA_IS_SHARE_CONSUMER(rktp->rktp_rkt->rkt_rk)) {
```
pranavrth: We need to keep this as it is. Guard offset management part from `rd_kafka_toppar_broker_leave_for_remove`

src/rdkafka_conf.c
pranavrth: Then this is not the right place to do this check. We will need to remove this property as we don't want user to set this. The below properties should be checked in `rd_kafka_share_consumer_new` function. Just add a TODO for this whole if condition and we will remove it as part of the config task.

src/rdkafka_assignment.c
```diff
@@ -349,7 +349,6 @@ static void rd_kafka_share_assignment_serve_removals(rd_kafka_t *rk) {
                 /* Detach the partition from the cgrp directly. This sends
                  * SHARE_SESSION_PARTITION_REMOVE to the partition's broker
                  * so it queues a forget on the next ShareFetch. */
-                rd_dassert(rktp->rktp_cgrp);
```
pranavrth: you can keep this as we are still using this field. Anyways its `dassert`

pranavrth (changes requested):
Few minor comments.

src/rdkafka_assignment.c
```diff
@@ -612,6 +702,10 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
  * - partition fetcher is stopped
  */
 void rd_kafka_assignment_serve(rd_kafka_t *rk) {
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
+                rd_kafka_share_assignment_serve(rk);
+                return;
+        }
```
pranavrth, line 708: We need to move this after declarations.

pranavrth (approved):
1. Still need to write exhaustive test cases. That would be in a different PR.
2. Do this minor change in that PR as well.

Apart from this, LGTM!. Thanks @PratRanj07 !..

Issue #5195, opened 2025-09-17
runtime error: call to function rd_kafka_assignor_destroy through pointer to incorrect function type 'void (*)(void *)'


olegrok:
On fresh master I see following error:

```
%5|1758098420.245|CONFWARN|rdkafka#consumer-1| [thrd:app]: No `bootstrap.servers` configured: client will not be able to connect to Kafka cluster
rdlist.c:283:33: runtime error: call to function rd_kafka_assignor_destroy through pointer to incorrect function type 'void (*)(void *)'
/home/runner/work/kafka/kafka/librdkafka/src/rdkafka_assignor.c:454: note: rd_kafka_assignor_destroy defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior rdlist.c:283:33 
```

PR #5278, opened 2025-12-16
assignor: fix rd_list destroy callback type mismatch (#5195)

aadish233:
Fix the UBSan warning:
  runtime error: call to function rd_kafka_assignor_destroy through pointer
  to incorrect function type 'void (*)(void *)'
by changing rd_kafka_assignor_destroy to take a void * argument, as expected by rd_list_init() destroy callbacks, and casting internally to rd_kafka_assignor_t *. Removed explicit cast from rd_list_init argument rd_kafka_assignor_destroy as it is exactly the type as expected by rd_list_init.

Test: 
  1. Locally enabled UBSan to reproduce and fix the issue.
  2. make quick on trivup (./interactive_broker_version.py 3.6.0)

### DIFF
--- a/src/rdkafka_assignor.c
+++ b/src/rdkafka_assignor.c
@@ -240,6 +240,13 @@ static void rd_kafka_assignor_topic_destroy(rd_kafka_assignor_topic_t *at) {
         rd_free(at);
 }
 
+/**
+ * @brief Destroy-variant suitable for rd_list free_cb use.
+ */
+static void rd_kafka_assignor_topic_free(void *ptr) {
+        rd_kafka_assignor_topic_destroy((rd_kafka_assignor_topic_t *)ptr);
+}
+
 int rd_kafka_assignor_topic_cmp(const void *_a, const void *_b) {
         const rd_kafka_assignor_topic_t *a =
             *(const rd_kafka_assignor_topic_t *const *)_a;
@@ -267,7 +274,7 @@ rd_kafka_member_subscriptions_map(rd_kafka_cgrp_t *rkcg,
             rd_kafka_metadata_get_internal(metadata);
 
         rd_list_init(eligible_topics, RD_MIN(metadata->topic_cnt, 10),
-                     (void *)rd_kafka_assignor_topic_destroy);
+                     rd_kafka_assignor_topic_free);
 
         /* For each topic in the cluster, scan through the member list
          * to find matching subscriptions. */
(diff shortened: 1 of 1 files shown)


src/rdkafka_assignor.c
```diff
@@ -451,7 +451,8 @@ rd_kafka_assignor_t *rd_kafka_assignor_find(rd_kafka_t *rk,
 /**
  * Destroys an assignor (but does not unlink).
  */
-static void rd_kafka_assignor_destroy(rd_kafka_assignor_t *rkas) {
+static void rd_kafka_assignor_destroy(void *rkasp) {
```
olegrok: That's not enough. `rd_kafka_assignor_topic_destroy` should be fixed in the same way.

aadish233:
Added two wrapper callbacks to avoid changing signature of existing internal functions. 
1. rd_kafka_assignor_free : Calls rd_kafka_assignor_destroy internally after typecasting
2. rd_kafka_assignor_topic_free: Calls rd_kafka_assignor_topic_destroy internally

rd_list_init now uses these new wrapper callbacks to avoid UBSan warning.

Test - Local UBSan and make quick on trivup (./interactive_broker_version.py 3.6.0)

src/rdkafka_assignor.c
```diff
@@ -240,6 +240,13 @@ static void rd_kafka_assignor_topic_destroy(rd_kafka_assignor_topic_t *at) {
         rd_free(at);
 }
 
+/**
+ * @brief Destroy-variant suitable for rd_list free_cb use.
+ */
+static void rd_kafka_assignor_topic_free(void *ptr) {
```
pranavrth, line 246: Name should be `rd_kafka_assignor_topic_destroy_ptr`

aadish233: In existing codebase, the naming convention that is being used has suffix "free". Could you confirm if I should change the name?
Here is one example: 
```
static void rd_kafka_ConfigEntry_destroy(rd_kafka_ConfigEntry_t *entry) {
        rd_strtup_destroy(entry->kv);
        rd_list_destroy(&entry->synonyms);
        rd_free(entry);
}


static void rd_kafka_ConfigEntry_free(void *ptr) {
        rd_kafka_ConfigEntry_destroy((rd_kafka_ConfigEntry_t *)ptr);
}
```

pranavrth (approved):
LGTM!.

PR #5287, opened 2026-01-05
Conditionally include sys/random.h based on HAVE_GETENTROPY

lpsinger:
This fixes Conda build errors on Linux aarch64 and ppc64. See, for example, https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1429288&view=logs&j=60238c3a-ba84-5ca5-39dc-717fbb421dbc&t=9cf66154-4ecc-5714-a3ee-b77d4233a250

### DIFF
--- a/src/rdrand.c
+++ b/src/rdrand.c
@@ -31,7 +31,7 @@
 #include "rdtime.h"
 #include "tinycthread.h"
 #include "rdmurmur2.h"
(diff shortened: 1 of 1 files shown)


emasab:
Thanks for the PR! Sorry there is a problem with CI checks on public PRs so have to include it in #5288.

PR #5290, opened 2026-01-12
FF-25386 - Make JWT OAuth refresh cb configurable for retry and timeout

LGouellec:


src/rdhttp.c
```diff
@@ -432,7 +442,13 @@ rd_http_error_t *rd_http_get(rd_kafka_t *rk,
         if (timeout_s > 0)
                 curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT, timeout_s);
```
azotcsit: This seems to be not precisely the same as [the logic in Java client](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/internals/secured/HttpJwtRetriever.java#L211-L215).

I explored Curl source code, and it looks like [CURLOPT_ACCEPTTIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_ACCEPTTIMEOUT_MS.html) and [CURLOPT_SERVER_RESPONSE_TIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_SERVER_RESPONSE_TIMEOUT_MS.html) options are not suitable for HTTP protocol. However, there are two other options that seem to be applicable:

| Property | Java | librdkafka | Phase Covered |
|--------|--------|--------|--------|
| sasl.login.connect.timeout.ms | [connectTimeout](https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#setConnectTimeout-int-) | [CURLOPT_CONNECTTIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html) | DNS lookup, TCP handshake, and SSL negotiation |
| sasl.login.read.timeout.ms | [readTimeout](https://docs.oracle.com/javase/8/docs/api/java/net/URLConnection.html#setReadTimeout-int-) | - | Waiting for the initial response after a request is sent |
| - | - | [CURLOPT_TIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_TIMEOUT_MS.html) | The entire operation (Connect + Transfer) |

Unfortunately, there is no straightforward way to implement it precisely, however it is possible to achieve a similar (to Java client) behavior using the logic below:
1. Set [CURLOPT_CONNECTTIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html) as `sasl.login.connect.timeout.ms`
2. Set [CURLOPT_TIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_TIMEOUT_MS.html) as `sasl.login.connect.timeout.ms` + `sasl.login.read.timeout.ms`.

I’m not sure introducing an additional property and the associated complexity is justified. What are your thoughts?

azotcsit: I guess you can use [CURLOPT_TIMEOUT_MS](https://curl.se/libcurl/c/CURLOPT_TIMEOUT_MS.html) instead of [CURLOPT_TIMEOUT](https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html), so you do not need to convert milliseconds to seconds.

src/rdhttp.c
```diff
@@ -441,22 +457,32 @@ rd_http_error_t *rd_http_get(rd_kafka_t *rk,
                 herr = rd_http_req_perform_sync(&hreq);
                 len  = rd_buf_len(hreq.hreq_buf);
 
+                herr = rd_http_req_perform_sync(&hreq);
```
azotcsit: These look like duplicated lines. Is this intentional?

src/rdhttp.c
```diff
@@ -432,7 +442,13 @@ rd_http_error_t *rd_http_get(rd_kafka_t *rk,
         if (timeout_s > 0)
                 curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT, timeout_s);
 
-        for (i = 0; i <= retries; i++) {
+        long endMs = current_milliseconds() + retry_backoff_max_ms;
+        long retryBackoffMs = retry_backoff_ms;
```
azotcsit: This variable does not seem to be updated later on. I guess we can remove it and just use `retry_backoff_ms`.

src/rdhttp.c
```diff
@@ -441,22 +457,32 @@ rd_http_error_t *rd_http_get(rd_kafka_t *rk,
...
                                 break; /* Success */
                         /* Empty response */
                         goto done;
                 }
 
+                long waitMs = retryBackoffMs *
+                          (long)pow(2, currAttempt - 1);
+
+                long diff = endMs - current_milliseconds();
+                if (waitMs > diff)
```
azotcsit: While this gives the same result as https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/internals/secured/Retry.java#L88, I like `min` more as it looks more readable.

LGouellec: Done

src/rdhttp.c
LGouellec: Nope, fixed

src/rdhttp.c
LGouellec: Fixed

src/rdhttp.c
```diff
@@ -448,15 +469,24 @@ rd_http_error_t *rd_http_get(rd_kafka_t *rk,
                         goto done;
                 }
 
+                long waitMs = retry_backoff_ms *
+                          (long)pow(2, currAttempt - 1);
+
+                long diff = endMs - current_milliseconds();
+                waitMs = MIN(waitMs, diff);
+
+                if (waitMs > diff)
```
azotcsit: Now, this condition duplicates `waitMs = MIN(waitMs, diff)`, so it can be removed.

src/rdkafka_conf.h
```diff
@@ -196,7 +196,7 @@ typedef enum {
 
 /* Increase in steps of 64 as needed.
  * This must be larger than sizeof(rd_kafka_[topic_]conf_t) */
-#define RD_KAFKA_CONF_PROPS_IDX_MAX (64 * 35)
+#define RD_KAFKA_CONF_PROPS_IDX_MAX (64 * 38)
```
azotcsit: 39?

src/rdkafka_conf.h
```diff
@@ -382,6 +382,13 @@ struct rd_kafka_conf_s {
                                                  const char *oauthbearer_config,
                                                  void *opaque);
                 } oauthbearer;
+
+                struct {
+                        long retry_backoff_max_ms;
```
azotcsit: Does the order matter? Can we double check that this is the right one?

src/rdhttp.c
```diff
@@ -715,7 +764,7 @@ int unittest_http_get(void) {
 
         /* Try the base url first, parse its JSON and extract a key-value. */
         json = NULL;
-        herr = rd_http_get_json(rk, base_url, NULL, 0, 5, 1, 1000, &json);
+        herr = rd_http_get_json(rk, base_url, NULL, 0, 5, 5, 1, 1000, &json);
```
azotcsit, line 767: This used to be 5 retries (5 seconds each) with 1 second delay in between. Now, the parameters changed, we need to reflect that.

src/rdhttp.c
```diff
@@ -546,13 +575,25 @@ rd_http_error_t *rd_http_post_expect_json(rd_kafka_t *rk,
                 return herr;
 
         curl_easy_setopt(hreq.hreq_curl, CURLOPT_HTTPHEADER, headers);
-        curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT, timeout_s);
+
+        if (timeout_ms > 0) {
+                curl_easy_setopt(hreq.hreq_curl, CURLOPT_CONNECTTIMEOUT_MS, timeout_ms);
+
+                if (read_timeout_ms > 0)
```
azotcsit: Is it possible to have `read_timeout_ms` set without  `timeout_ms`? If yes, then this scenario is not handled.

src/rdhttp.c
```diff
@@ -546,13 +575,25 @@ rd_http_error_t *rd_http_post_expect_json(rd_kafka_t *rk,
...
+                        curl_easy_setopt(hreq.hreq_curl, CURLOPT_TIMEOUT_MS, timeout_ms + read_timeout_ms);
+        }
 
         curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDSIZE,
                          post_fields_size);
         curl_easy_setopt(hreq.hreq_curl, CURLOPT_POSTFIELDS, post_fields);
 
-        for (i = 0; i <= retries; i++) {
+        long endMs = current_milliseconds() + retry_backoff_max_ms;
```
azotcsit: What is about code style? Should we use snake_case as the rest of the codebase or is it fine to keep camelCase?

ojasvajain:
@LGouellec Thanks for raising this PR. Will it be possible to add functional test cases to verify if the changes work as expected?

LGouellec:
@ojasvajain I want to but how do you mock the OIDC server ? Even today, there is no unit test which test the retry feature based on the HTTP Error code (408, 425, 429, 500, 502, 503 or 504)

There is this unit test : https://github.com/confluentinc/librdkafka/blob/e0da09cc26241ee123c793f806c27b5df494b094/src/rdhttp.c#L697

but does the server return a retryable HTTP error or just a 400 ? Don't know ..

PR #5299, opened 2026-02-04
ShareGroupHeartBeat API

Ankith-Confluent:
Design :
https://confluentinc.atlassian.net/wiki/x/-QDSMwE

### DIFF
--- a/src/rdkafka_mock_handlers.c
+++ b/src/rdkafka_mock_handlers.c
@@ -3136,6 +3136,8 @@ rd_kafka_mock_handle_ShareGroupHeartbeat(rd_kafka_mock_connection_t *mconn,
                                          * catching up */
                                         member->conn = mconn;
                                         MemberEpoch  = member->member_epoch;
+                                        rd_kafka_mock_sharegroup_member_active(
+                                            mshgrp, member);
                                 }
                         } else {
                                 /* Epoch matches - normal heartbeat */
@@ -3148,6 +3150,8 @@ rd_kafka_mock_handle_ShareGroupHeartbeat(rd_kafka_mock_connection_t *mconn,
                                 }
                                 member->conn = mconn;
                                 MemberEpoch  = member->member_epoch;
+                                rd_kafka_mock_sharegroup_member_active(mshgrp,
+                                                                       member);
                         }
                 }
 
--- a/src/rdkafka_mock_int.h
+++ b/src/rdkafka_mock_int.h
@@ -762,6 +762,10 @@ void rd_kafka_mock_sharegroup_member_destroy(
     rd_kafka_mock_sharegroup_t *mshgrp,
     rd_kafka_mock_sharegroup_member_t *member);
 
+void rd_kafka_mock_sharegroup_member_active(
+    rd_kafka_mock_sharegroup_t *mshgrp,
+    rd_kafka_mock_sharegroup_member_t *member);
+
 rd_kafka_mock_sharegroup_member_t *
 rd_kafka_mock_sharegroup_member_get(rd_kafka_mock_sharegroup_t *mshgrp,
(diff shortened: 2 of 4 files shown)


src/rdkafka_mock_int.h
```diff
@@ -179,6 +179,39 @@ typedef struct rd_kafka_mock_cgrp_consumer_member_s {
...
+        int member_cnt;              /**< Number of share group members */
+        rd_bool_t manual_assignment; /**< Use manual assignment */
+} rd_kafka_mock_sharegroup_t;
+
+/**
+ * @struct Share group member (KIP-932).
+ */
+typedef struct rd_kafka_mock_sharegroup_member_s {
+        TAILQ_ENTRY(rd_kafka_mock_sharegroup_member_s) link;
+        char *id;                          /**< MemberId */
```
k-raina, line 204: Is Member Id server generated, in case client doesn't send memberId in params?

src/rdkafka_mock_sharegrp.c
```diff
@@ -0,0 +1,752 @@
...
+
+
+#include "rdkafka_int.h"
+#include "rdbuf.h"
+#include "rdkafka_mock_int.h"
+
+/**
+ * @brief Share group target assignment (manual)
+ */
+typedef struct rd_kafka_mock_sharegroup_target_assignments_s {
```
k-raina, line 43: For SGHB file, can we reuse CGHB code? 

Eg. these functions seems to be common `_write_TopicPartitions, session timer, find/get/destroy, member_find/active/fenced, set_timeout/interval, target_assignment struct, response builder`

Can we check how much  code can be deduplicated, then plan and commit changes?

src/rdkafka_mock_handlers.c
```diff
@@ -2993,6 +2993,229 @@ rd_kafka_mock_handle_ConsumerGroupHeartbeat(rd_kafka_mock_connection_t *mconn,
...
+                } else {
+                        /* HEARTBEAT: Existing member heartbeat */
+                        member = rd_kafka_mock_sharegroup_member_find(
+                            mshgrp, &MemberId);
+                        if (!member) {
+                                err = RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID;
+                        } else if (MemberEpoch > member->member_epoch) {
+                                /* Client epoch is ahead of server - indicates
+                                 * a bug or stale coordinator. */
+                                err = RD_KAFKA_RESP_ERR_FENCED_MEMBER_EPOCH;
```
k-raina, line 3124: Does this need to be `RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID` ?

k-raina (changes requested):
Thanks for PR!
For first round of review, overall PR looks good. I have suggested few changes. 

Additionally, should we update PR description also? It will help with future reviews

src/rdkafka_mock_handlers.c
Ankith-Confluent: I checked the java implementation, in that they are returning `FencedMemberEpochException` in both the cases if its more or less, so `RD_KAFKA_RESP_ERR_FENCED_MEMBER_EPOCH` is being returned.

src/rdkafka_mock_int.h
Ankith-Confluent: Actually the member Id is generated by the consumer itself, if its not sent with the request an exception (`UnknownMemberIdException`) will be thrown. So we dont need to generate it in the broker side.

src/rdkafka_mock_sharegrp.c
Ankith-Confluent: Sure

k-raina (approved):
Ack, pipeline is broken in base branch. 

PR LGTM! Thanks

PR #5308, opened 2026-02-13
Produce deterministic "client.id" by default

rockdrilla:
It's no secret that many applications don't set `client.id` themselves, making identification on the Kafka broker difficult.

The problem becomes more acute when applications run in container orchestration environments (Kubernetes, Nomad, Docker, etc.) and connect to external Kafka clusters. In such cases, identifying the actual producer/consumer instance becomes nearly impossible without a meaningful `client.id`.

This MR addresses the issue by providing a deterministic default `client.id` in the form `"rdkafka@{hostname}"`, falling back to the previous default `"rdkafka"` if the hostname cannot be retrieved.

*A quick note on intent:*  
This MR isn't meant to call out or blame any specific software or project for not setting `client.id`. Many legitimate use cases rely on defaults, and the goal here is simply to make those defaults more useful for operational visibility - especially in modern cloud-native environments. The change is proposed as a general improvement to librdkafka's out-of-the-box experience.

---

*Nota bene*:

This MR consists of two logical parts:

**Part 1: Non-breaking improvements (safe to merge)**
- `Define RD_NOINLINE as opposite to RD_INLINE`
- `Ensure rd_kafka_transport_init() is called once`
- `Adjust buffer size to avoid hostname truncation` (applies to example application)

**Part 2: Potentially breaking change (requires discussion)**
- `Produce deterministic "client.id" by default`
- `Adjust "client.id" in examples and tests`

---

Regarding `CHANGELOG.md`: I believe these changes should be part of the next release (e.g., 2.15.0). If accepted, I can update the changelog with something like:
  * Ensure Winsock initialization is called only once via `InitOnceExecuteOnce`
  * Add deterministic default `client.id` (`rdkafka@{hostname}`) when none is provided

Thoughts? Feedback welcome!

### DIFF
--- a/examples/kafkatest_verifiable_client.cpp
+++ b/examples/kafkatest_verifiable_client.cpp
@@ -627,12 +627,6 @@ int main(int argc, char **argv) {
     exit(1);
   }
 
-  {
-    char hostname[260] = { 0 };
-    gethostname(hostname, sizeof(hostname) - 1);
-    conf->set("client.id", std::string("rdkafka@") + hostname, errstr);
-  }
-
   conf->set("log.thread.name", "true", errstr);
 
   /* auto commit is explicitly enabled with --enable-autocommit */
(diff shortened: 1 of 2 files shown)


rockdrilla:
@prashah-confluent @emasab friendly ping.

PR #5322, opened 2026-02-20
Add topic and partition deletion functionality to mock cluster

Ankith-Confluent:
- Implement `rd_kafka_mock_topic_delete()` to delete a topic and its partitions.
- Implement `rd_kafka_mock_partition_delete_records()` to delete records before a specified offset.
- Update command handling in the mock cluster to support new delete operations.
- Add tests for topic deletion and partition record deletion to ensure correct behavior.
- Introduce max size limit for share groups with appropriate handling in join requests.


k-raina:
@Ankith-Confluent PR needs rebase

PR #5335, opened 2026-02-28
Fix int_latency calculation for infinite timeout

x0id:
### Before the fix:
```
int_latency_base = now + (message_timeout_ms * 1000);
rd_avg_add(&rkb->rkb_avg_int_latency, int_latency_base - rkm->rkm_ts_timeout);
```
When message_timeout_ms = 0:
int_latency_base = now + 0 = now
rkm_ts_timeout = INT64_MAX (for infinite timeout)
Result: now - INT64_MAX = massive negative number

### After the fix:
```
rd_avg_add(&rkb->rkb_avg_int_latency, now - rkm->rkm_ts_enq);
```
Always works regardless of timeout setting:
now - rkm_ts_enq = current_time - enqueue_time = actual queue time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+# librdkafka v2.15.1
+
+librdkafka v2.15.1 is a maintenance release:
+
+* Fix `int_latency` metric calculation, that was reporting bogus values when
+  `message.timeout.ms` is set to 0 (infinite) (#5335).
+
+
+## Fixes
+
+### Producer fixes
+
+* Issues: #5555.
+  Fix `int_latency` metric calculation. It was derived from the message
+  timeout timestamp (`now + message.timeout.ms - rkm_ts_timeout`), which
+  yields a large negative value when `message.timeout.ms` is 0 (infinite)
+  and `rkm_ts_timeout` is `INT64_MAX`. It's now computed directly as
+  `now - rkm_ts_enq`, the actual time the message spent in the queue,
+  regardless of the timeout setting.
+  Happening since 0.11.0 (#5335).
+
+
+
 # librdkafka v2.15.0
 
 librdkafka v2.15.0 is a feature release:
--- a/src/rdkafka_msgset_writer.c
+++ b/src/rdkafka_msgset_writer.c
@@ -819,16 +819,10 @@ static int rd_kafka_msgset_writer_write_msgq(rd_kafka_msgset_writer_t *msetw,
         size_t max_msg_size =
             RD_MIN((size_t)msetw->msetw_rkb->rkb_rk->rk_conf.max_msg_size,
                    (size_t)msetw->msetw_rkb->rkb_rk->rk_conf.batch_size);
-        rd_ts_t int_latency_base;
         rd_ts_t MaxTimestamp = 0;
         rd_kafka_msg_t *rkm;
         int msgcnt        = 0;
         const rd_ts_t now = rd_clock();
-
-        /* Internal latency calculation base.
-         * Uses rkm_ts_timeout which is enqueue time + timeout */
-        int_latency_base =
-            now + ((rd_ts_t)rktp->rktp_rkt->rkt_conf.message_timeout_ms * 1000);
 
         /* Acquire BaseTimestamp from first message. */
         rkm = TAILQ_FIRST(&rkmq->rkmq_msgs);
@@ -895,8 +889,7 @@ static int rd_kafka_msgset_writer_write_msgq(rd_kafka_msgset_writer_t *msetw,
                 msetw->msetw_messages_kvlen += rkm->rkm_len + rkm->rkm_key_len;
 
                 /* Add internal latency metrics */
-                rd_avg_add(&rkb->rkb_avg_int_latency,
-                           int_latency_base - rkm->rkm_ts_timeout);
+                rd_avg_add(&rkb->rkb_avg_int_latency, now - rkm->rkm_ts_enq);
 
                 /* MessageSet v2's .MaxTimestamp field */
                 if (unlikely(MaxTimestamp < rkm->rkm_timestamp))
--- a/tests/0055-producer_latency.c
+++ b/tests/0055-producer_latency.c
@@ -621,7 +621,7 @@ static void test_int_latency(const char *message_timeout_ms) {
         int64_t abs_timeout;
         int i;
 
-        SUB_TEST("message.timeout.ms=%s", message_timeout_ms);
+        SUB_TEST_QUICK("message.timeout.ms=%s", message_timeout_ms);
 
         mcluster = test_mock_cluster_new(1, &bootstrap_servers);
         rd_kafka_mock_topic_create(mcluster, topic, 1, 1);
--- a/tests/0055-producer_latency.c
+++ b/tests/0055-producer_latency.c
@@ -527,6 +527,145 @@ static void test_producer_latency_first_message(int case_number) {
         SUB_TEST_PASS();
 }
 
+struct int_latency_stats {
+        int64_t minv; /**< Lowest int_latency.min seen */
+        int64_t avgv; /**< Last int_latency.avg seen */
+        int cnt;      /**< Total number of int_latency samples */
+};
+
+
+/**
+ * @brief Crude extraction of the integer JSON field \p field following
+ *        \p start.
+ */
+static int64_t find_json_int(const char *start, const char *field) {
+        char search[64];
+        const char *t;
+
+        rd_snprintf(search, sizeof(search), "\"%s\":", field);
+
+        t = strstr(start, search);
+        TEST_ASSERT(t, "Field %s not found in stats", field);
+
+        t += strlen(search);
+        while (isspace((int)*t))
+                t++;
+
+        return strtoll(t, NULL, 0);
+}
+
+
+/**
+ * @brief Collect the per-broker int_latency gauges from the stats.
+ */
+static int int_latency_stats_cb(rd_kafka_t *rk,
+                                char *json,
+                                size_t json_len,
+                                void *opaque) {
+        struct int_latency_stats *stats = opaque;
+        const char *t                   = json;
+
+        while ((t = strstr(t, "\"int_latency\":"))) {
+                int64_t minv, avgv;
+                int cnt;
+
+                t += strlen("\"int_latency\":");
+
+                /* Brokers that haven't produced anything during this
+                 * window have no samples. */
+                cnt = (int)find_json_int(t, "cnt");
+                if (cnt == 0)
+                        continue;
+
+                minv = find_json_int(t, "min");
+                avgv = find_json_int(t, "avg");
+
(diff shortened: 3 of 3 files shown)


x0id:
Friendly ping on this one — it's been open ~4.5 months with CLA signed and no reviews yet.

To restate the impact: any producer using `message.timeout.ms=0` (infinite timeout, common with idempotent/transactional producers) gets a garbage `int_latency` stat — `now - INT64_MAX` underflows to a massive negative number, which corrupts the `rkb_avg_int_latency` window and anything downstream reading `int_latency` from stats (dashboards, alerting, etc.). The fix just switches the base from `rkm_ts_timeout` (enqueue time + timeout, which is `INT64_MAX` when timeout is disabled) to `rkm_ts_enq` (enqueue time), which is well-defined regardless of the timeout setting and gives the actual queue latency in both cases.

It's a small, self-contained change with no API/ABI impact. Would appreciate a look from a maintainer when you have a moment — @esabellico @PratRanj07 @pranavrth. Happy to add a test case or CHANGELOG entry if that's what's blocking review.

x0id:
Checking in again — the companion issue #5555 was assigned to @ojasvajain a couple of weeks ago but hasn't seen any activity since. The fix here is a one-liner with no API/ABI impact, CLA signed, tests pass. @Ankith-Confluent would you be able to take a quick look or point me to whoever could? Happy to add a CHANGELOG entry or unit test if that helps move it forward. Thanks!

ojasvajain:
@x0id The fix looks correct. Please add a CHANGELOG entry and a unit test for this.

x0id:
Updated this PR:

- **Rebased onto latest `master`** (`0623a303`). The fix applied without conflicts — `src/rdkafka_msgset_writer.c` is byte-identical between v2.13.2 and current master, so no adaptation was needed and the bug is still present on master.
- **CHANGELOG entry added.** Since the top of the file is the already-released v2.15.0, I opened a new `v2.15.1` maintenance-release section; happy to renumber if the next release lands differently. It references #5555.
- **Unit test added.** `test_int_latency()` in `0055_producer_latency_mock` — mock cluster, no broker required. It produces 10 messages and asserts the per-broker `int_latency` stat is non-negative, running once with `message.timeout.ms=0` (infinite) and once with `5000` as a regression guard for the normal path.

The test was checked against the buggy code to make sure it actually catches this. With the fix reverted:

```
int_latency: min -9221585134712460448us, avg 1786902142315360us, cnt 10
### Test "0055_producer_latency_mock (test_int_latency: message.timeout.ms=0)" failed:
    Test assertion failed: "stats.minv >= 0": Expected non-negative int_latency,
    got min -9221585134712460448us
```

With the fix in place, on the rebased tree:

```
[ test_int_latency: message.timeout.ms=0 ]     int_latency: min 1371us, avg 1439us, cnt 10  PASS
[ test_int_latency: message.timeout.ms=5000 ]  int_latency: min 1446us, avg 1473us, cnt 10  PASS
================= Test 0055_producer_latency_mock PASSED =================
```

Note that `avg` comes back as a large *positive* garbage value under the bug (sum overflow), so the `min` assertion is what catches it.

`packaging/tools/style-format.sh` passes clean on the changed test file.

ojasvajain:
@x0id Please look into the build failures on semaphore.

x0id:
@ojasvajain Thanks for looking. I can't see the Semaphore logs (the instance requires Confluent auth), so I went at it from the outside — here's what I found, plus one real fix pushed.

**Pushed `320b3132`: run the new subtests in quick mode.** While investigating I noticed my subtests were not actually running in CI at all. The local jobs run `make -C tests run_local_quick`, and `-Q` sets `test_quick`, which makes `test_sub_start()` return early for anything declared with `SUB_TEST()`:

```c
if (!is_quick && test_quick)
        return 0;
```

The test was therefore completing in 0.001s and "passing" without executing. They are fast (~0.5s) and have no timing assertions — unlike the other subtests in this file, which are deliberately non-quick — so they are now declared with `SUB_TEST_QUICK()`. Re-verified in the exact CI mode (`CI=true ./run-test.sh -l -Q -P`), which also exercises the idempotent producer that quick mode enables:

```
int_latency: min 506156us, avg 506166us, cnt 10   message.timeout.ms=0     PASS
int_latency: min 507238us, avg 507288us, cnt 10   message.timeout.ms=5000  PASS
```

(The ~506ms is initial PID acquisition with idempotence enabled; it also confirms `message.timeout.ms=0` doesn't conflict with `enable.idempotence=true`.)

**On the earlier failure**, what I could check locally all comes back clean:

- The strictest build job's settings — `CFLAGS=-std=gnu90` with `--enable-werror --enable-devel` — compile the changed test file with no warnings. `strtoll`, `rd_snprintf`, `isspace` and `PRId64` are all available via `test.h` (`../src/rd.h` defines `_GNU_SOURCE`, `<ctype.h>` is included), and `strtoll` is already used in `tests/0142-reauthentication.c`.
- `CI=true make -C tests run_local_quick` passes in full locally, with no failures.
- `packaging/tools/style-format.sh` is clean on the changed file.

It may also be worth noting that `master` itself appears red at the commit this branch is rebased onto:

```
$ gh api repos/confluentinc/librdkafka/commits/0623a303/status --jq .state
failure    # ci/semaphoreci/push: build-test-release
```

with `4c3b017c` (the commit before it) green. So at least part of what this PR is seeing may be inherited rather than caused here — but I genuinely can't confirm without the job names, so if you can share which jobs failed (or paste the relevant log lines) I'll dig into anything that's actually mine.

Could you re-run `/sem-approve` for `320b3132` when you get a chance? Happy to rebase onto a green master too if that helps isolate it.

ojasvajain:
Sorry, I should have shared the logs with you. The failure is occurring while building the tests when CI jobs run `make -j -C tests build`. Failures are observed only for Linux runs. Logs:


`
In file included from ../src/rd.h:80,00:10
                 from test.h:32,00:10
                 from 0055-producer_latency.c:30:00:10
0055-producer_latency.c: In function ‘test_int_latency’:00:10
test.h:994:40: error: ‘%s’ directive output truncated writing 501 bytes into a region of size 450 [-Werror=format-truncation=]00:10
  994 |                 TIMING_START(&_timing, "%s", _desc);                           \00:10
      |                                        ^~~~00:10
../src/rdposix.h:87:36: note: in definition of macro ‘rd_snprintf’00:10
   87 | #define rd_snprintf(...)  snprintf(__VA_ARGS__)00:10
      |                                    ^~~~~~~~~~~00:10
test.h:994:17: note: in expansion of macro ‘TIMING_START’00:10
  994 |                 TIMING_START(&_timing, "%s", _desc);                           \00:10
      |                 ^~~~~~~~~~~~00:10
0055-producer_latency.c:642:17: note: in expansion of macro ‘TEST_CALL_ERR__’00:10
  642 |                 TEST_CALL_ERR__(rd_kafka_producev(00:10
      |                 ^~~~~~~~~~~~~~~00:10
test.h:994:41: note: format string is defined here00:10
  994 |                 TIMING_START(&_timing, "%s", _desc);                           \00:10
      |                                         ^~00:10
cp interceptor_test.a interceptor_test-dbg.a00:10
In file included from /usr/include/stdio.h:980,00:10
                 from ../src/rd.h:54:00:10
In function ‘snprintf’,00:10
    inlined from ‘test_int_latency’ at 0055-producer_latency.c:642:17:00:10
/usr/include/x86_64-linux-gnu/bits/stdio2.h:54:10: note: ‘__builtin_snprintf’ output 502 bytes into a destination of size 45000:10
   54 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,00:10
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~00:10
   55 |                                    __glibc_objsize (__s), __fmt,00:10
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~00:10
   56 |                                    __va_arg_pack ());00:10
      |                                    ~~~~~~~~~~~~~~~~~
`

x0id:
Thanks for the logs — that was exactly what I needed, and the failure is mine. Fixed in `fb49e54c`.

**Cause.** `TEST_CALL_ERR__()` stringifies the entire call expression into `test_timing_t.name`, which is `char name[450]`:

```c
const char *_desc = RD_STRINGIFY(FUNC_W_ARGS);
TIMING_START(&_timing, "%s", _desc);
```

My `rd_kafka_producev()` call expands through the `RD_KAFKA_V_*` macros to 501 bytes, so GCC's `-Wformat-truncation` fires and `--enable-werror` turns it into a build error. It only shows up on Linux because clang doesn't implement that warning, which is why it slipped past my macOS build. In hindsight the truncation was visible in my own test output all along — the `duration` log lines were being cut off mid-token at 450 characters.

**Fix.** Capture and check the error explicitly, which is what the other `producev` calls in this file already do:

```c
err = rd_kafka_producev(
    rk, RD_KAFKA_V_TOPIC(topic), RD_KAFKA_V_PARTITION(0),
    RD_KAFKA_V_VALUE("hi", 2),
    RD_KAFKA_V_MSGFLAGS(RD_KAFKA_MSG_F_COPY), RD_KAFKA_V_END);
TEST_ASSERT(!err, "producev #%d failed: %s", i, rd_kafka_err2str(err));
```

The remaining `TEST_CALL_ERR__(rd_kafka_flush(...))` stringifies to 43 bytes, well within the buffer.

**Verified with real GCC this time**, rather than relying on clang. Reproduced the exact error on the previous commit:

```
test.h:994:40: error: '%s' directive output truncated writing 501 bytes
               into a region of size 450 [-Werror=format-truncation=]
```

and confirmed it is gone on `fb49e54c`, using the source-build job's settings (`-std=gnu90 -O2 -Wall -Werror`). The test itself still passes in the CI mode (`CI=true ./run-test.sh -l -Q -P`):

```
int_latency: min 507993us, avg 507997us, cnt 10   message.timeout.ms=0     PASS
int_latency: min 507264us, avg 507265us, cnt 10   message.timeout.ms=5000  PASS
```

`packaging/tools/style-format.sh` is clean. Could you kick off `/sem-approve` for `fb49e54c` when convenient?

x0id:
Not sure why the CLA check flipped to pending. It read "All Contributor License Agreements are signed" on the earlier commits and now says "Not all…". I signed CLA once again, just in case.

ojasvajain:
@x0id There is an internal issue with the CLA check. We are looking into this.

ojasvajain:
@x0id The issue is fixed now. Can you retrigger the check by pushing an empty commit?

x0id:
@ojasvajain Done — pushed empty commit `e4d2c777`, and `cla-status-check` is green again ("All Contributor License Agreements are signed"). Thanks for sorting that out.

The push moved the head SHA, so the Semaphore results from `fb49e54c` (both `build-test-release` and `librdkafka-integration-tests` passed) no longer show against the PR. Could you run `/sem-approve` once more? The new commit is empty, so the tree is byte-identical to the one that already went green — it should just be a formality.

ojasvajain (approved):
LGTM, thanks @x0id

PR #5336, opened 2026-03-02
Add support for different claims for authentication

prashah-confluent:
### Problem
For OIDC / SASL_OAUTHBEARER flows using librdkafka’s built‑in OIDC helper, we currently hard‑require that the JWT payload contain a sub claim. If the IdP omits sub or uses another claim (for example client_id) as the primary subject, librdkafka fails token acquisition with an error such as:

`Failed to acquire SASL OAUTHBEARER token: Expected JSON JWT response with "sub" field`

### Fix
We have introduced a new config `sub_claim_name` using which customers can provide a different claim. The claim value will be honored and used for client side validation. 

***Note***
Requires users to keep broker and client configs aligned for the subject claim name to avoid confusion and consistency.

#### Unit tests
To test the core JWT payload parsing and claim extraction logic inside rd_kafka_oidc_token_try_validate() in isolation (no broker, network, oidc server). 

1. Default Configuration & Backward Compatibility
Implicit Default Behavior: Validate that when sub.claim.name is not provided, the system defaults to extracting the subject from the "sub" claim.
Legacy Parity: Ensure that existing JWTs containing only the "sub" claim continue to be processed correctly, maintaining no regressions for standard OIDC flows.

2. Explicit Claim Configuration (KIP-768 Parity)
Custom Claim Extraction: Verify that setting sub.claim.name (e.g., to client_id) successfully extracts the value from the specified field instead of the hardcoded "sub" key.
Configuration Precedence: Test that when both "sub" and a custom claim (e.g., "client_id") coexist in a JWT, the system honors the user-defined configuration over the OIDC default.

3. Error Handling & Validation Logic
Missing Claim Detection: Verify that providing a sub.claim.name that does not exist in the JWT triggers a specific "Claim not found" error, rather than failing on a hardcoded "sub".
Empty Value Rejection: Ensure the system rejects empty string subjects with a "non-empty value required" error, addressing previous bugs where empty strings passed silently.
Invalid Configuration Recovery: Validate that if sub.claim.name is set to an empty string, the system gracefully falls back to the OIDC default ("sub") to ensure continuous operation.

4. librdkafka & Java Client Alignment
Functional Parity: Ensure all scenarios are cross-validated against the Java KIP-768 implementation to guarantee identical extraction logic and exception triggers (e.g., ValidateException equivalents).

#### Integration tests
 - To validate end-to-end OIDC authentication against a real Kafka cluster (via trivup) with different sasl.oauthbearer.sub.claim.name configurations
 - The trivup setup requires changes for [returning JWT response with multiple claims](https://github.com/emasab/trivup/pull/10)

[Design Doc](https://confluentinc.atlassian.net/wiki/spaces/~712020e58a442b654d4de1924d66f639e7cf1a/pages/5175935729/INIT-9699+-+OIDC+sub-claim+handling+in+librdkafka)

### DIFF
--- a/src/rdkafka_conf.c
+++ b/src/rdkafka_conf.c
@@ -4910,8 +4910,7 @@ int unittest_conf(void) {
         rd_kafka_conf_destroy(conf);
 
         /* Verify rd_kafka_conf_validate_str */
-        RD_UT_ASSERT(!rd_kafka_conf_validate_str(NULL),
-                     "NULL must be invalid");
+        RD_UT_ASSERT(!rd_kafka_conf_validate_str(NULL), "NULL must be invalid");
         RD_UT_ASSERT(!rd_kafka_conf_validate_str(""),
                      "empty string must be invalid");
         RD_UT_ASSERT(!rd_kafka_conf_validate_str("   "),


src/rdkafka_conf.c
```diff
@@ -1121,6 +1121,15 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "OAuth/OIDC issuer token endpoint HTTP(S) URI used to retrieve token. "
      "Only used when `sasl.oauthbearer.method` is set to \"oidc\".",
      _UNSUPPORTED_OIDC},
+    {_RK_GLOBAL, "sasl.oauthbearer.sub.claim.name", _RK_C_STR,
+     _RK(sasl.oauthbearer.sub_claim_name),
+     "JWT claim name to use as the subject (principal) when validating "
+     "OIDC access tokens. Must be present in the JWT payload with a "
+     "non-empty value. Should match the broker's "
+     "sasl.oauthbearer.sub.claim.name configuration for consistent "
```
emasab:
```suggestion
     "`sasl.oauthbearer.sub.claim.name` configuration for consistent "
```

emasab: rebuild `CONFIGURATION.md` as well.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -729,19 +730,28 @@ static char *rd_kafka_oidc_token_try_validate(cJSON *json,
                 goto fail;
         }
 
-        jwt_sub = cJSON_GetObjectItem(payloads, "sub");
+        /* Safety check to default to "sub" if not set */
+        if (!sub_claim_name || !*sub_claim_name)
+                sub_claim_name = "sub";
```
emasab: it shoud be a `rd_dassert` here, not reverting to the default. The configuration should be validated in `rd_kafka_conf_finalize_oauthbearer_oidc` and using `rd_kafka_conf_is_modified` on the field and then check for NULL or a an empty string or whitespace string ("\s+") you can write a function like the Java method `ClaimValidationUtils.validateString`.

tests/0126-oauthbearer_oidc.c
```diff
@@ -494,6 +494,118 @@ void do_test_produce_consumer_with_OIDC_metadata_authentication(
...
+                    "Sub claim variation: %s\n",
+                    oidc_configuration_sub_claim_variation_name(variation));
+
+                if (variation <
+                    OIDC_CONFIGURATION_SUB_CLAIM_VARIATION__FIRST_FAILING) {
+                        /* These variations should succeed */
+                        do_test_produce_consumer_with_OIDC(test_name,
+                                                           sub_claim_conf);
+                } else {
+                        /* These variations should fail due to missing claim */
```
emasab: This function should be agnostic to the failure cause (at the moment it's only this one but in future there can be other test cases).
```suggestion
                        /* These variations should fail */
```

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# librdkafka v2.14.0
+
+librdkafka v2.14.0 is a feature release:
+
+* [KIP-768](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=186877575#KIP768:ExtendSASL/OAUTHBEARERwithSupportforOIDC-ClientConfiguration) Extend SASL/OAUTHBEARER to support OIDC claim mapping beyond the default `sub` claim [(#5336)](https://github.com/confluentinc/librdkafka/pull/5336)
```
```suggestion
* [KIP-768](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=186877575#KIP768:ExtendSASL/OAUTHBEARERwithSupportforOIDC-ClientConfiguration) Extend SASL/OAUTHBEARER to support OIDC claim mapping beyond the default `sub` claim (#5336).
```
The link is already added by GH in the release page, so we don't put the link here.

src/rdkafka_sasl_oauthbearer_oidc.c
```diff
@@ -1312,6 +1323,136 @@ static int ut_sasl_oauthbearer_oidc_post_fields_with_empty_scope(void) {
...
+                const char *test_name;
+                const char *jwt;
+                const char *sub_claim_name;
+                rd_bool_t expect_success;
+                const char *expected_sub;
+        } tests[] = {
+            {"NULL sub_claim_name defaults to 'sub'", UT_JWT_SUB_ONLY, NULL,
+             rd_true, "subject"},
+            {"Empty sub_claim_name defaults to 'sub'", UT_JWT_SUB_ONLY, "",
+             rd_true, "subject"},
```
emasab: We can remove these two cases after putting an assert there as it must not reach this function if those preconditions aren't satisfied.
```suggestion
```

tests/0126-oauthbearer_oidc.c
```diff
@@ -494,6 +494,118 @@ void do_test_produce_consumer_with_OIDC_metadata_authentication(
...
+        case OIDC_CONFIGURATION_SUB_CLAIM_VARIATION_EXPLICIT_SUB:
+                test_conf_set(conf, "sasl.oauthbearer.sub.claim.name", "sub");
+                break;
+        case OIDC_CONFIGURATION_SUB_CLAIM_VARIATION_CUSTOM_CLIENT_ID:
+                test_conf_set(conf, "sasl.oauthbearer.sub.claim.name",
+                              "client_id");
+                break;
+        case OIDC_CONFIGURATION_SUB_CLAIM_VARIATION_EMPTY_STRING:
+                test_conf_set(conf, "sasl.oauthbearer.sub.claim.name", "");
+                break;
```
emasab, line 550: This one should fail as well. Similarly to the Java client, in case it's set it should be valid otherwise the user could want to set something but there was a bug in their code so an empty string was received instead.

prashah-confluent: This should pass due to librdkafka convention which sets the config to default values if empty/NULL string is provided.

src/rdkafka_conf.c
prashah-confluent: Changed to "`sasl.oauthbearer.sub.claim.name`" and rebuilt CONFIGURATION.md

src/rdkafka_sasl_oauthbearer_oidc.c
prashah-confluent: Added rd_dassert and as suggested added helper api which validates string only if config is modified.

src/rdkafka_sasl_oauthbearer_oidc.c
prashah-confluent: Removed the test cases since they can't be valid at this point.

CHANGELOG.md
prashah-confluent: Removed the link

tests/0126-oauthbearer_oidc.c
prashah-confluent: Modified the comment to make the function failure agnostic

PR #5338, opened 2026-03-02
Initial style fix for Share Consumer

pranavrth:
Fixed style issues.
Fixed documentation build issue.
Fixed a few test related issues. Build is still failing which will have separate PR.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -5957,6 +5957,7 @@ const char *rd_kafka_Uuid_base64str(const rd_kafka_Uuid_t *uuid) {
         rd_chariov_t in_base64;
         char *out_base64_str;
         char *uuid_bytes;
+        char *p;
         uint64_t input_uuid[2];
 
         input_uuid[0]  = htobe64(uuid->most_significant_bits);
@@ -5972,7 +5973,7 @@ const char *rd_kafka_Uuid_base64str(const rd_kafka_Uuid_t *uuid) {
                 return NULL;
 
         // Convert to URL-safe Base64
-        for (char *p = out_base64_str; *p; p++) {
+        for (p = out_base64_str; *p; p++) {
                 if (*p == '+')
                         *p = '-';
                 else if (*p == '/')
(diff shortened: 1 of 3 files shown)


src/rdkafka.c
```diff
@@ -2893,7 +2893,8 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
         }
 
         /**
-         * TODO KIP-932: Check if this way of defining share consumer needs to be changed.
+         * TODO KIP-932: Check if this way of defining share consumer needs to
+         * be changed.
```
k-raina, line 2897: I see we are using [script](https://github.com/confluentinc/librdkafka/blob/901af7c79c0ca5877ed187c4452820aecf51430c/packaging/tools/style-format.sh#L4) to run style check in semaphore.   Did we use `make style-fix` for these fixes?

src/rdkafka_mock_handlers.c
```diff
@@ -4343,8 +4330,8 @@ rd_kafka_mock_handle_ShareAcknowledge(rd_kafka_mock_connection_t *mconn,
                 rd_kafka_topic_partition_list_sort_by_topic_id(ack_partitions);
 
                 {
-                        int i         = 0;
-                        int topic_cnt = 0;
+                        int i                         = 0;
```
k-raina, line 4333: Does spacing look excessive here? IG its fine if clang-format updated it

src/rdkafka_mock_sharegrp.c
```diff
@@ -582,7 +582,7 @@ static void rd_kafka_mock_sharegroup_target_assignment_set(
     rd_kafka_mock_sharegroup_t *mshgrp,
     rd_kafka_mock_sharegroup_target_assignment_t *target_assignment) {
         rd_kafka_mock_sharegroup_member_t *member;
-        size_t i;
+        int i;
```
k-raina, line 585: Nice catch! Did clang-format caught it?

src/rdkafka.c
pranavrth: Yes we use that.

src/rdkafka_mock_sharegrp.c
pranavrth: clang-format doesn't catch this but our build process make sure that these kind of issues are caught.

src/rdkafka_mock_handlers.c
pranavrth: So what clang-format tries to do is make all the consecutive "=" align. For this example:

```
                        int i                         = 0;
                        int topic_cnt                 = 0;
                        rd_kafka_Uuid_t current_topic = RD_KAFKA_UUID_ZERO;
```

PR #5316, opened 2026-02-16
[KIP-932]: Segregation of Acquired Records and Acknowledgement mapping

PratRanj07:
This PR intends to:

- Segregate the acquired records based on the records and the AcquiredRecords array
- Implement Acknowledgement Mapping

PR #5339, opened 2026-03-02
Poll flow using acks sent received from the app thread

pranavrth:


pranavrth:
Merged as part of https://github.com/confluentinc/librdkafka/pull/5316

PR #5344, opened 2026-03-05
Fix CI related to mock broker 

k-raina:
**Key changes in PR**
- Remove stale rd_dassert calls in rdkafka_cgrp.c: Removed assertions that fired incorrectly when UNKNOWN_MEMBER_ID is intentionally injected during share group testing, preventing false assertion crashes from blocking CI.
- In quick mode (test_quick=true), SUB_TEST() (non-quick) causes an early return. Any test_ctx_new() called before SUB_TEST() leaked the mock cluster + producer.
- The original 300s timeout was insufficient; 1500s (5×) ensures tests don't time out under heavy load on CI Linux agents.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -6627,13 +6627,6 @@ void rd_kafka_cgrp_consumer_serve(rd_kafka_cgrp_t *rkcg) {
                                                 "member fenced - rejoining");
         }
 
-        /*  Per KIP-932, share consumers have no fencing semantics and
-         * should never enter WAIT_REJOIN / WAIT_REJOIN_TO_COMPLETE.
-         *
-         * TODO: Test current and new behavior with share consumers
-         */
-
-
         switch (rkcg->rkcg_join_state) {
         case RD_KAFKA_CGRP_JOIN_STATE_INIT:
                 rkcg->rkcg_consumer_flags &=


src/rdkafka_cgrp.c
```diff
@@ -6627,14 +6627,11 @@ void rd_kafka_cgrp_consumer_serve(rd_kafka_cgrp_t *rkcg) {
...
         }
 
-        /* There should be no fencing, hence no rejoining - these asserts are to
-         * test only, we don't actually need them. */
-        rd_dassert(!(RD_KAFKA_IS_SHARE_CONSUMER(rkcg->rkcg_rk) &&
-                     (rkcg->rkcg_consumer_flags &
-                      RD_KAFKA_CGRP_CONSUMER_F_WAIT_REJOIN)));
-        rd_dassert(!(RD_KAFKA_IS_SHARE_CONSUMER(rkcg->rkcg_rk) &&
-                     (rkcg->rkcg_consumer_flags &
-                      RD_KAFKA_CGRP_CONSUMER_F_WAIT_REJOIN_TO_COMPLETE)));
```
pranavrth, line 6637: We should remove this TODO.

pranavrth:
> We should run the tests again v4.2.0 kafka broker. Some tests are getting stuck as share consumer is not enable by default in earlier version of kafka. Please update this and then let's check the CI again.

Wait. There are other issues as well with v4.2.0. We are fixing those in our PR. Let's not update the kafka version right now.

tests/0155-share_group_heartbeat_mock.c
```diff
@@ -566,10 +571,14 @@ static void do_test_share_group_session_timeout(void) {
...
-        /* Wait for both to join and rebalance to complete */
-        wait_share_heartbeats(mcluster, 3, 500);
-        rd_kafka_consumer_poll(c1, 2000);
-        rd_kafka_consumer_poll(c2, 2000);
+        /* Wait for both to join and rebalance to complete. Poll both
+         * consumers in short alternating windows so both can heartbeat
+         * and neither session times out while the other is polled. */
+        wait_share_heartbeats(mcluster, 4, 500);
+        for (int i = 0; i < 5; i++) {
+                rd_kafka_consumer_poll(c1, 200);
```
Ankith-Confluent, line 579: wrong poll is used, but that will be fixed in my PR, but there might be merge conflicts

k-raina: Yes, i have kept file structure same for now

PR #5349, opened 2026-03-06
[KIP-932] Mock Broker: Add transaction lifecycle support

Ankith-Confluent:
* Track transaction state in AddPartitionsToTxn and EndTxn handlers — write COMMIT/ABORT control batches, record aborted ranges, and manage LSO
* Add rd_kafka_mock_sharegroup_set_isolation_level() API to configure read_uncommitted vs read_committed isolation for share groups
* Filter aborted transactional data during read_committed ShareFetch using LSO and per-partition aborted transaction ranges
* Fix ShareFetch session partition handling — use session->partitions instead of requested_partitions for acquisition and response writing, as the client sends an empty partition list for epoch>0 continuation requests
* Add integration and unit tests for transactional produce + share consume

### DIFF
--- a/src/rdkafka_mock_int.h
+++ b/src/rdkafka_mock_int.h
@@ -255,7 +255,7 @@ typedef struct rd_kafka_mock_sharegroup_s {
         int record_lock_duration_ms; /**< Per-record lock duration in ms.
                                       *   0 = use session_timeout_ms
                                       *   as fallback (default 0). */
-        int isolation_level;         /**< Share isolation level
+        int isolation_level;         /**< Share isolation level */
         int max_size;                /**< Max members allowed.
                                       *   0 = unlimited (default). */
 } rd_kafka_mock_sharegroup_t;


src/rdkafka_mock_int.h
```diff
@@ -255,6 +255,9 @@ typedef struct rd_kafka_mock_sharegroup_s {
         int record_lock_duration_ms; /**< Per-record lock duration in ms.
                                       *   0 = use session_timeout_ms
                                       *   as fallback (default 0). */
+        int isolation_level;         /**< Share isolation level (KIP-932).
```
k-raina:
```suggestion
        int isolation_level;         /**< Share isolation level
``` 
Nit: AI usually generates these tags, we can configure claude accordingly

src/rdkafka_mock_handlers.c
```diff
@@ -3365,6 +3432,26 @@ rd_kafka_mock_msgset_est_record_size(const rd_kafka_mock_msgset_t *mset) {
...
+        TAILQ_FOREACH(mabort, &mpart->aborted_txns, link) {
+                if (mabort->pid_id == pid_id &&
+                    offset >= mabort->first_offset &&
+                    offset <= mabort->last_offset)
+                        return rd_true;
+        }
+        return rd_false;
+}
+
 static void rd_kafka_mock_sgrp_acquire_available_offsets(
```
k-raina, line 3455: Is there any consensus  on if read_committed and read_uncommitted will deliver/not deliver internal markers to consumers.?

src/rdkafka_mock_handlers.c
```diff
@@ -3408,6 +3504,41 @@ static void rd_kafka_mock_sgrp_acquire_available_offsets(
...
+                                memcpy(&batch_pid,
+                                       (const char *)mset->bytes.data +
+                                           RD_KAFKAP_MSGSET_V2_OF_ProducerId,
+                                       sizeof(batch_pid));
+                                batch_pid = be64toh(batch_pid);
+                                if (!(attrs &
+                                      RD_KAFKA_MSGSET_V2_ATTR_CONTROL) &&
+                                    rd_kafka_mock_offset_is_aborted(
+                                        mpart, offset, batch_pid)) {
+                                        /* Archive aborted data batch */
```
k-raina, line 3527: Is this intentional, broker is also archiving 
  aborted data batches directly rather than letting the client filter them?

Ankith-Confluent: Yes broker is supposed to archive aborted messages and is completely skipped while sending it to the client

src/rdkafka_mock_handlers.c
Ankith-Confluent: yes, both read_committed and read_uncommitted deliver control markers to consumers, the client is responsible for skipping them. This is verified from the implementation in java.

k-raina: >For the read_committed isolation level, the share group only consumes non-transactional records and committed transactional records. The set of records which are eligible to become in-flight records are non-transactional records and committed transactional records only. The SPEO cannot move past the last stable offset, so an open transaction blocks the progress of the share group with read_committed isolation level. The **share-partition leader** itself is responsible for filtering out transactional records which have been aborted. This can be done looking at the headers of the record batches without having to iterate of the individual records.

This is snippet from KIP-932, is this outdated?

src/rdkafka_mock_int.h
```diff
@@ -292,6 +295,56 @@ typedef struct rd_kafka_mock_pid_s {
...
+/**
+ * @brief Transaction state enum.
+ */
+typedef enum {
+        RD_KAFKA_MOCK_TXN_NONE,    /**< No active transaction */
+        RD_KAFKA_MOCK_TXN_ONGOING, /**< Transaction in progress */
+} rd_kafka_mock_txn_state_t;
+
+/**
+ * @brief A partition that is part of an active transaction.
```
k-raina: Does this need to be `@struct`?

src/rdkafka_mock_int.h
```diff
@@ -703,6 +769,22 @@ rd_kafka_resp_err_t rd_kafka_mock_partition_leader_epoch_check(
     const rd_kafka_mock_partition_t *mpart,
     int32_t leader_epoch);
 
+/* Transaction helpers (used across rdkafka_mock.c and
```
k-raina:
```suggestion
/* Transaction helpers 
```

src/rdkafka_mock_handlers.c
Ankith-Confluent: Aborted transactional data batches are filtered by the broker before sending to the client. This is what the mock does also. Control record batches (COMMIT/ABORT markers) are delivered to the client in both read_committed and read_uncommitted. The client is responsible for recognizing and skipping them. This is consistent with how the regular consumer works in Java (verified against the Java implementation).

PR #5353, opened 2026-03-10
[KIP-932] Add commit async API

pranavrth:


tests/0173-share_consumer_commit_async.c
pranavrth, line 1: Self: Check this test file again.

examples/share_consumer_commit_async.c
```diff
@@ -0,0 +1,262 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2019-2022, Magnus Edenhill
+ *               2023, Confluent Inc.
```
PratRanj07: 2026, Confluent Inc.

src/rdkafka.c
PratRanj07: NOT a part of this PR:
But maybe we can remove this redundant check as if we are passing the rkshare, it would already be of type consumer with the share consumer as true
But if we decide to remove this rkshare and keep it as part of rk, maybe we can add this check in rd_kafka_share_commit_async

src/rdkafka.c
```diff
@@ -3667,6 +3698,79 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
...
+ *
+ * In implicit ack mode, all ACQUIRED records are first converted
+ * to ACCEPT before building the ack details.
+ *
+ * @param rkshare Share consumer instance.
+ * @returns NULL on success, or an rd_kafka_error_t* on failure.
+ *
+ * @locality app thread
+ */
+rd_kafka_error_t *rd_kafka_share_commit_async(rd_kafka_share_t *rkshare) {
```
PratRanj07, line 3741: Given that we aren't returning any errors as part of this API, shouldn't the return type be void?

PratRanj07: Or are you planning to return any errors in future versions. If so, it would be great to add a TODO. Because as I know the errors that arise from this API is returned with Acknowledgement Callback

pranavrth: We will add errors in future. I think no need to add a TODO as the functionality itself is not built for which we need to send an error.

PR #5354, opened 2026-03-10
[KIP-932] Skip offset filter for share consumers in msgset parser

pranavrth:
Share consumers may receive re-delivered records at earlier offsets after RELEASE. The three offset checks in the msgset parser that skip messages where offset < fetch_pos.offset are correct for regular consumers (monotonically increasing offsets) but cause share consumers to drop re-delivered records, resulting in spurious GAP acks.

Add RD_KAFKA_IS_SHARE_CONSUMER guard to all three offset checks so share consumers never skip messages based on offset position.

### DIFF
--- a/src/rdkafka_msgset_reader.c
+++ b/src/rdkafka_msgset_reader.c
@@ -653,7 +653,8 @@ rd_kafka_msgset_reader_msg_v0_1(rd_kafka_msgset_reader_t *msetr) {
          *       we cant perform this offset check here
          *       in that case. */
         if (!relative_offsets &&
-            hdr.Offset < rktp->rktp_offsets.fetch_pos.offset)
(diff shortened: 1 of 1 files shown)


src/rdkafka_msgset_reader.c
```diff
@@ -653,7 +653,8 @@ rd_kafka_msgset_reader_msg_v0_1(rd_kafka_msgset_reader_t *msetr) {
          *       we cant perform this offset check here
          *       in that case. */
```
PratRanj07, line 654:
```suggestion
         *       in that case. 
         *       Share consumers may receive re-delivered records at earlier offsets
         *       after RELEASE, so skip this check for them.*/
```

PratRanj07: Or maybe add a TODO to remove this handling after making sure we dont recieve Messages v0 and v1 after broker v4.0

PR #5355, opened 2026-03-13
[KIP-932]: Implement DeliveryCount Public API

PratRanj07:
This PR intends to:

- Implement DeliveryCount Public API
- Add usage of Delivery count in tests to check for the redelivery of RELEASE'd messages
- Add more acknowledge() tests

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3179,9 +3179,10 @@ typedef enum rd_kafka_share_AcknowledgeType_s {
  *
  * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success,
  *          RD_KAFKA_RESP_ERR__INVALID_ARG if there is discrepancy with the
- * input parameters like rkmessage is NULL, RD_KAFKA_RESP_ERR__STATE if there is
- * some internal discrepancy like we are not in explicit acknowledgement mode or
- * offset is a GAP record.
+ *          input parameters like rkmessage is NULL,
+ *          RD_KAFKA_RESP_ERR__STATE if there is some internal discrepancy
+ *          like we are not in explicit acknowledgement mode or
+ *          offset is a GAP record.
  */
 RD_EXPORT
 rd_kafka_resp_err_t
@@ -3197,9 +3198,10 @@ rd_kafka_share_acknowledge(rd_kafka_share_t *rkshare,
  *
  * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success,
  *          RD_KAFKA_RESP_ERR__INVALID_ARG if there is discrepancy with the
- * input parameters like rkmessage is NULL, RD_KAFKA_RESP_ERR__STATE if there is
- * some internal discrepancy like we are not in explicit acknowledgement mode or
- * offset is a GAP record.
+ *          input parameters like rkmessage is NULL,
+ *          RD_KAFKA_RESP_ERR__STATE if there is some internal discrepancy
+ *          like we are not in explicit acknowledgement mode or
+ *          offset is a GAP record.
  */
 RD_EXPORT
 rd_kafka_resp_err_t
@@ -3218,9 +3220,10 @@ rd_kafka_share_acknowledge_type(rd_kafka_share_t *rkshare,
  *
  * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success,
  *          RD_KAFKA_RESP_ERR__INVALID_ARG if there is discrepancy with the
- * input parameters like rkmessage is NULL, RD_KAFKA_RESP_ERR__STATE if there is
- * some internal discrepancy like we are not in explicit acknowledgement mode or
- * offset is a GAP record.
+ *          input parameters like rkmessage is NULL,
+ *          RD_KAFKA_RESP_ERR__STATE if there is some internal discrepancy
+ *          like we are not in explicit acknowledgement mode or
+ *          offset is a GAP record.
  *
  * TODO KIP-932: Check if it should only be allowed for error cases
  *               like Java.
--- a/src/rdkafka_share_acknowledgement.c
+++ b/src/rdkafka_share_acknowledgement.c
@@ -121,14 +121,15 @@ rd_kafka_share_ack_batches_t *
 rd_kafka_share_ack_batches_new(rd_kafka_topic_partition_t *rktpar,
                                int32_t response_leader_id,
                                int32_t response_leader_epoch,
-                               int64_t response_msgs_count) {
+                               int64_t response_acquired_offsets_count) {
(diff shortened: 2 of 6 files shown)


src/rdkafka_msg.c
```diff
@@ -1592,6 +1592,19 @@ int32_t rd_kafka_message_leader_epoch(const rd_kafka_message_t *rkmessage) {
...
+int16_t rd_kafka_message_delivery_count(const rd_kafka_message_t *rkmessage) {
+        rd_kafka_msg_t *rkm;
+
+        if (unlikely(!rkmessage || !rkmessage->rkt ||
+                     rd_kafka_rkt_is_lw(rkmessage->rkt) ||
+                     !rkmessage->rkt->rkt_rk ||
+                     rkmessage->rkt->rkt_rk->rk_type != RD_KAFKA_CONSUMER))
+                return 0;
+
+        rkm = rd_kafka_message2msg((rd_kafka_message_t *)rkmessage);
```
pranavrth, line 1601: Do a NULL check for rkm as well.

src/rdkafka_msg.c
```diff
@@ -1592,6 +1592,19 @@ int32_t rd_kafka_message_leader_epoch(const rd_kafka_message_t *rkmessage) {
         return rkm->rkm_u.consumer.leader_epoch;
 }
 
+int16_t rd_kafka_message_delivery_count(const rd_kafka_message_t *rkmessage) {
+        rd_kafka_msg_t *rkm;
+
+        if (unlikely(!rkmessage || !rkmessage->rkt ||
+                     rd_kafka_rkt_is_lw(rkmessage->rkt) ||
+                     !rkmessage->rkt->rkt_rk ||
+                     rkmessage->rkt->rkt_rk->rk_type != RD_KAFKA_CONSUMER))
```
pranavrth: No need to check topic related information. 

The delivery count should be initialized with 0. We should just do null handling and can directly return the count.

src/rdkafka.h
```diff
@@ -1705,6 +1705,17 @@ rd_kafka_message_status(const rd_kafka_message_t *rkmessage);
...
 rd_kafka_message_leader_epoch(const rd_kafka_message_t *rkmessage);
 
+/**
+ * @brief Get the delivery count for a message.
+ *
+ * @param rkmessage Message object.
+ *
+ * @returns Delivery count (number of times this message has been delivered).
+ *          Returns 0 if delivery count is not available.
+ */
```
pranavrth, line 1717:
```suggestion
/**
 * @brief Get the delivery count for a message associated with the share consumer.
 *
 * @param rkmessage Message object.
 *
 * @returns Delivery count (number of times this message has been delivered).
 *          Returns 0 if delivery count is not available like in case of normal consumer or producer.
 */
```

tests/0171-share_consumer_consume.c
pranavrth: We should set acquisition lock timeout to lower value to make the tests run quicker. Maybe 1s.

tests/0172-share_consumer_acknowledge.c
```diff
@@ -103,6 +112,22 @@ static rd_kafka_share_t *create_explicit_ack_consumer(const char *group_id) {
...
+        int r = rand_r(seed) % 3;
+        switch (r) {
+        case 0:
+                return RD_KAFKA_SHARE_ACKNOWLEDGE_TYPE_ACCEPT;
+        case 1:
+                return RD_KAFKA_SHARE_ACKNOWLEDGE_TYPE_REJECT;
+        default:
+                return RD_KAFKA_SHARE_ACKNOWLEDGE_TYPE_RELEASE;
+        }
+}
```
pranavrth, line 121:
```suggestion
get_random_ack_type(unsigned int *seed) {
        return (rd_kafka_share_AcknowledgeType_t) ((rand_r(seed) % 3) + 1);
}
```

tests/0172-share_consumer_acknowledge.c
```diff
@@ -219,14 +301,68 @@ static void consume_and_acknowledge(ack_test_config_t *config,
...
                                     rd_kafka_topic_name(batch[m]->rkt));
                                 p_idx = batch[m]->partition;
 
-                                if (t_idx >= 0 &&
-                                    p_idx < config->partitions[t_idx]) {
+                                /*
+                                 * In random mode, check if this message was
+                                 * previously RELEASE'd (redelivery).
+                                 */
+                                if (config->use_random_acks) {
```
pranavrth: Reduce cyclomatic complexity of this function.

tests/0172-share_consumer_acknowledge.c
```diff
@@ -219,14 +301,68 @@ static void consume_and_acknowledge(ack_test_config_t *config,
...
+                                            batch[m]->partition,
+                                            batch[m]->offset);
+
+                                        if (released_idx >= 0) {
+                                                /* This is a redelivered msg */
+                                                rd_kafka_topic_partition_t
+                                                    *rktpar;
+
+                                                TEST_ASSERT(
+                                                    delivery_count > 1,
```
pranavrth: Better to check exact count.

src/rdkafka_msg.c
PratRanj07: It will never be NULL if rkmessage is not NULL as it is a simple type casting

tests/0171-share_consumer_consume.c
PratRanj07: Added a TODO for now to check later

src/rdkafka_msg.c
pranavrth: It can be created for testing which won't have msg.

src/rdkafka_share_acknowledgement.c
```diff
@@ -384,13 +385,6 @@ rd_list_t *rd_kafka_share_build_ack_details(rd_kafka_share_t *rkshare) {
                         }
                 }
 
-                /* Propagate sorted flag from source - order is preserved */
-                if (inflight_batches->entries.rl_flags & RD_LIST_F_SORTED) {
```
pranavrth, line 388: Let's add a TODO to check this logic again and maybe optimize.

tests/0172-share_consumer_acknowledge.c
pranavrth, line 1: There is memory leak in this test and 0170 as well. Please check and fix the issue.

PratRanj07: Checking this in different PR.

src/rdkafka_share_acknowledgement.c
PratRanj07: Checking this in different PR.

PR #5363, opened 2026-03-19
[KIP-932]: Use jitter in place of rand_r() for support in windows

PratRanj07:
Use jitter in place of rand_r() for support in windows


pranavrth (approved):
LGTM!.

PR #5366, opened 2026-03-23
[KIP-932]: Added more concurrency tests and multiple groups tests

PratRanj07:
## 0170-share_consumer_subscription.c

- Verify share.auto.offset.reset=earliest receives messages produced before subscription
- Verify share.auto.offset.reset=latest (default) skips messages produced before subscription

## 0174-share_consumer_concurrency.c

**Threading model:** Each producer runs in its own thread (`thrd_create`), producing messages in round-robin across topics/partitions. Consumers run in the main thread, polling in round-robin. Shared state is protected by mutex. Producer threads update `total_produced` atomically, main thread tracks `total_consumed`.

- 1 producer thread, 1 consumer thread, 1 topic, 1 partition (baseline)
- 1 producer, 4 consumers competing for 4 partitions (fan-out)
- 4 producers, 1 consumer aggregating from 4 partitions (fan-in)
- 4 producers, 4 consumers on 4 partitions (symmetric)
- 2 producers, 2 consumers across 3 topics, 2 partitions each
- 2 producers, 2 consumers across 8 topics, 1 partition each
- 1 producer, 8 consumers on single partition (high contention)
- 2 producers, 6 consumers on 2 partitions (3:1 consumer-to-partition ratio)
- 4 producers, 4 consumers with explicit acknowledgement mode
- Consumers start polling before producers start producing (staggered start)
- 4 producers, 4 consumers, 20,000 messages (high volume stress)
- 4 producers, 4 consumers, 8 partitions (high volume many partitions)
- 8 producers, 2 consumers (producer-heavy asymmetric)
- 2 producers, 8 consumers (consumer-heavy asymmetric)
- 4 topics with varied partition counts (1,2,3,4 partitions)
- 10 producers, 10 consumers, 4 topics (maximum concurrency)
- 4 producers (fast), 2 consumers with 50ms delay (slow consumers)
- 2 producers with 50ms delay (slow), 4 consumers (fast consumers)

## 0175-share_consumer_groups.c

- 2 share groups consuming same topic independently, each receives all 100 messages
- 3 share groups with 2 consumers each, all independently consume same topic
- 5 share groups each with 1 consumer, all independently consume same topic
- Group A starts first, Group B joins later, both eventually receive all messages


tests/0170-share_consumer_subscription.c
ojasvajain, line 1: The coverage looks fine. Can you check if we can have a few more test cases related to subscribe / unsubscribe  by adding delay to the ShareGroupHeartbeat? 

For example, consumer is subscribed to topics A and B, add delay to the heartbeat and now call subscribe to topics C and D. 1 case can be for delay < session.timeout.ms and another for delay > session.timeout.ms. This is just one test case. 


Some more test cases I can think of (Ignore if already covered)

1. Subscribe to a topic, consume some msgs and then unsubscribe without acking those msgs. Try to ack after unsubscribe or resubscribe to the same topic and then try to ack.

2. A test case around this: "_If a member is removed from the group because it fails to heartbeat, because there’s intentionally no fencing, at the protocol level, the consumer does not lose the ability to fetch and acknowledge records. A failure to heartbeat is most likely because the consumer has died. If the consumer just failed to heartbeat due to a temporary pause, it could in theory continue to fetch and acknowledge records. When it finally sends a heartbeat and realises it’s been kicked out of the group, it should stop fetching records because its assignment has been revoked, and rejoin the group._" (Source: KIP-932)

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,812 @@
...
+typedef struct {
+        int consumer_cnt;           /**< Number of consumers to create */
+        int producer_cnt;           /**< Number of producer threads */
+        int topic_cnt;              /**< Number of topics */
+        int partitions[MAX_TOPICS]; /**< Partitions per topic */
+        int msgs_per_partition;     /**< Messages to produce per partition */
+        const char *group_name;     /**< Share group name */
+        const char *test_name;      /**< Test description */
+        int max_attempts;           /**< Max poll attempts (0 = default) */
+        int consumer_delay_ms;      /**< Delay between consumer polls */
```
ojasvajain, line 65: This in unused. I think we intended to use it in `test_slow_consumers`.

tests/0170-share_consumer_subscription.c
PratRanj07: Nice suggestions! I will add some more tests regarding this in other PR

tests/test.c
```diff
@@ -582,6 +584,8 @@ struct test tests[] = {
     _TEST(0172_share_consumer_acknowledge, 0, TEST_BRKVER(0, 4, 2, 0)),
     _TEST(0173_share_consumer_commit_async_local, TEST_F_LOCAL),
     _TEST(0173_share_consumer_commit_async, 0, TEST_BRKVER(0, 4, 2, 0)),
+    _TEST(0174_share_consumer_concurrency, 0, TEST_BRKVER(0, 4, 0, 0)),
+    _TEST(0175_share_consumer_groups, 0, TEST_BRKVER(0, 4, 0, 0)),
```
```suggestion
    _TEST(0174_share_consumer_concurrency, 0, TEST_BRKVER(0, 4, 2, 0)),
    _TEST(0175_share_consumer_groups, 0, TEST_BRKVER(0, 4, 2, 0)),
```

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+/**
+ * @brief Shared state for concurrent test threads
+ */
+typedef struct {
+        mtx_t lock;                      /**< Mutex for state access */
+        int total_produced;              /**< Total messages produced */
+        int total_consumed;              /**< Total messages consumed */
+        int expected_total;              /**< Expected total messages */
+        rd_bool_t producers_done;        /**< All producers finished */
+        rd_bool_t consumers_should_stop; /**< Signal consumers to stop */
```
ojasvajain: This is unused.

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+
+/**
+ * @brief Shared state for concurrent test threads
+ */
+typedef struct {
+        mtx_t lock;                      /**< Mutex for state access */
+        int total_produced;              /**< Total messages produced */
+        int total_consumed;              /**< Total messages consumed */
+        int expected_total;              /**< Expected total messages */
+        rd_bool_t producers_done;        /**< All producers finished */
```
ojasvajain: This is never set. It is always false.

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+        /* Create consumers */
+        for (c = 0; c < config->consumer_cnt; c++) {
+                consumers[c] =
+                    test_create_share_consumer(config->group_name, NULL);
+
+                /* Configure group on first consumer */
+                if (c == 0) {
+                        test_share_set_auto_offset_reset(config->group_name,
+                                                         "earliest");
+                }
```
ojasvajain: You can do this after the loop.

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+
+                if (thrd_create(&producer_threads[p], producer_thread_func,
+                                args) != thrd_success) {
+                        TEST_FAIL("Failed to create producer thread %d", p);
+                }
+        }
+
+        /* Poll consumers from main thread while producers run */
+        max_attempts = config->max_attempts > 0 ? config->max_attempts : 100;
+        {
```
ojasvajain: Let's remove these blocks

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+static void test_slow_producers(void) {
+        concurrent_test_config_t config = {
+            .consumer_cnt       = 4,
+            .producer_cnt       = 2,
+            .topic_cnt          = 1,
+            .partitions         = {2},
+            .msgs_per_partition = 10000,
+            .group_name         = "share-conc-slow-producer",
+            .test_name          = "Slow producers, fast consumers",
+            .producer_delay_ms  = 50,
```
ojasvajain: 50ms delay every 100 msgs seems too less for a test case for slow producers. Can we increase it to 10 seconds?

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+        }
+
+        /* Verify results */
+        TEST_SAY("Final: produced=%d, consumed=%d, expected=%d\n",
+                 state.total_produced, state.total_consumed,
+                 state.expected_total);
+
+        TEST_ASSERT(state.total_produced == state.expected_total,
+                    "Expected to produce %d, actually produced %d",
+                    state.expected_total, state.total_produced);
```
ojasvajain: acquire state lock before reading total_produced

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+                                idle_rounds++;
+                        }
+
+                        /* Apply consumer delay if configured */
+                        if (config->consumer_delay_ms > 0)
+                                rd_usleep(config->consumer_delay_ms * 1000,
+                                          NULL);
+
+                        /* Check if producers are done */
+                        mtx_lock(&state.lock);
```
ojasvajain: the locking and unlocking should not be required as we are not setting producers_done

tests/0174-share_consumer_concurrency.c
ojasvajain: You might need to adjust some configs like idle rounds if we are increasing the producer delay.

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+                while (state.total_consumed < state.expected_total &&
+                       idle_rounds < 50) {
+                        int round_consumed = 0;
+
+                        for (c = 0; c < config->consumer_cnt; c++) {
+                                size_t rcvd = 0;
+                                size_t m;
+                                rd_kafka_error_t *err;
+
+                                err = rd_kafka_share_consume_batch(
```
ojasvajain, line 421: Log the error.

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,822 @@
...
+ */
+
+#include "test.h"
+#include "rdkafka.h"
+
+/**
+ * @name Share Consumer Concurrency and Stress Tests
+ *
+ * Tests share consumer behavior under concurrent producer/consumer scenarios.
+ * Each test uses config-based structure with producers and consumers running
```
ojasvajain: Please clarify it to say that consumers are being polled from the same thread. 
"Consumers running in separate threads" is a little misleading.

tests/0175-share_consumer_groups.c
```diff
@@ -0,0 +1,515 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2025, Confluent Inc.
```
ojasvajain: 2026

ojasvajain (commented):
Yet to review 0174.

tests/0175-share_consumer_groups.c
```diff
@@ -0,0 +1,515 @@
...
+        int group_cnt;                       /**< Number of share groups */
+        int consumers_per_group[MAX_GROUPS]; /**< Consumers per group */
+        const char *group_names[MAX_GROUPS]; /**< Group names */
+        rd_bool_t use_earliest[MAX_GROUPS];  /**< Use earliest offset */
+        int partitions;                      /**< Number of partitions */
+        int msgs_per_partition;              /**< Messages per partition */
+        const char *test_name;               /**< Test description */
+        int max_attempts;                    /**< Max poll attempts */
+        rd_bool_t produce_before_subscribe;  /**< Produce before subscribe */
+        int produce_after_subscribe;         /**< Messages after subscribe */
```
ojasvajain: nit: can we rename it to `msgs_produce_after_subscribe`?

tests/0175-share_consumer_groups.c
```diff
@@ -0,0 +1,515 @@
...
+                                continue;
+
+                        all_done = 0;
+
+                        for (c = 0; c < config->consumers_per_group[g]; c++) {
+                                size_t rcvd = 0;
+                                size_t m;
+                                rd_kafka_error_t *err;
+
+                                err = rd_kafka_share_consume_batch(
```
ojasvajain, line 178: can we log the error?

ojasvajain:
Left comments. Please rebase the PR and ensure test cases are consistently passing.

tests/0174-share_consumer_concurrency.c
PratRanj07: We are now setting it
Its just an optimizationj to not wait for faster early exit when producers finish early

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,835 @@
...
+                                args) != thrd_success) {
+                        TEST_FAIL("Failed to create producer thread %d", p);
+                }
+        }
+
+        /* Poll consumers from main thread while producers run */
+        max_attempts = config->max_attempts > 0 ? config->max_attempts : 100;
+        rd_kafka_message_t *batch[BATCH_SIZE];
+        int attempts    = max_attempts;
+        int idle_rounds = 0;
```
ojasvajain: Define these at the top of the function.

tests/0174-share_consumer_concurrency.c
```diff
@@ -0,0 +1,835 @@
...
+                        if (config->consumer_delay_ms > 0)
+                                rd_usleep(config->consumer_delay_ms * 1000,
+                                          NULL);
+                }
+        }
+
+        /* Snapshot results under lock */
+        int final_produced;
+        int final_consumed;
+        rd_bool_t final_failed;
```
ojasvajain: define these at the top

tests/0170-share_consumer_subscription.c
ojasvajain: Hope these are added to the list we are maintaining

ojasvajain (approved):
Couple of nits. Looks like there is some flakiness in 0174 (check build failure logs).

ojasvajain (approved):
lgtm

PR #5374, opened 2026-03-30
[KIP-932] Add rd_kafka_share_commit_sync() API

pranavrth:
1. Core implementation:
   - SHARE_COMMIT_SYNC_FANOUT op type and handler in main thread
   - Segregate acks by partition leader into per-broker pending_commit_sync
   - Dispatch sync ack ops with priority over async_ack_details
   - Timeout timer fills remaining partitions with _TIMED_OUT
   - App thread blocks on temp queue until all broker replies or timeout

2. Data structures:
   - commit_sync_request in rkcg (abs_timeout, results, wait count, replyq, timer)
   - pending_commit_sync in broker (sync_ack_details, abs_timeout, request_id)
   - SHARE_COMMIT_SYNC_FANOUT and FANOUT_REPLY op types with union members

3. Broker thread changes:
   - ShareAcknowledge reply handler populates ack_results in reply op
   - Main thread reply handler copies per-partition errors to sync results
   - Pending sync acks dispatched before async when broker becomes free

4. Tests (0176-share_consumer_commit_sync.c):
   - Basic implicit/explicit ack mode commit_sync
   - No pending acks returns NULL/NULL
   - Commit prevents redelivery (Consumer B gets 0 records)
   - Mixed ack types (ACCEPT/RELEASE/REJECT) with delivery count verification
   - Multiple commit_sync calls with back-to-back no-op commits
   - Multi-topic multi-partition (10 topics x 6 partitions x 10 msgs)
   - Mock: verifies ShareAcknowledge RPC count matches commit_sync calls
   - Mock: timeout handling with 5s RTT / 2s timeout and recovery
   - Mixed commit types (10 async + 1 sync pattern)
   - Mock: broker dispatch priority (sync before async)

5. Example program and build integration

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3937,7 +3937,9 @@ static void rd_kafka_share_commit_sync_timeout_cb(rd_kafka_timers_t *rkts,
         }
         rd_kafka_rdunlock(rk);
 
-        /* Fill partitions still IN_PROGRESS with REQUEST_TIMED_OUT */
+        /* TODO: Verify that __TIMED_OUT (local timeout) is the correct
+         *  error here rather than REQUEST_TIMED_OUT (broker error). */
+        /* Fill partitions still IN_PROGRESS with __TIMED_OUT */
         results = rkcg->rkcg_commit_sync_request.results;
         for (i = 0; i < results->cnt; i++) {
                 rd_kafka_topic_partition_t *rktpar = &results->elems[i];
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3175,6 +3175,12 @@ rd_kafka_share_fetch_fanout_with_backoff(rd_kafka_t *rk,
                 rd_kafka_share_fetch_fanout_enqueue(rk, rko);
 }
 
+/* Forward declarations for commit_sync helpers used in reply handler */
+static void rd_kafka_share_commit_sync_maybe_complete(rd_kafka_t *rk,
+                                                      rd_kafka_cgrp_t *rkcg);
+static void rd_kafka_share_enqueue_sync_ack_op(rd_kafka_t *rk,
+                                               rd_kafka_broker_t *rkb);
+
 /**
  * Handles RD_KAFKA_OP_SHARE_FETCH | RD_KAFKA_OP_REPLY.
  * @locality main thread
@@ -3209,7 +3215,65 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
                 rkcg->rkcg_share.share_fetch_more_records = rd_false;
 
         /*
-         * Step 2: If this was the fetch broker but no records were received,
+         * Step 2: Handle commit_sync reply if this op belongs to
+         * the current commit_sync request.
+         */
+        if (rko_orig->rko_u.share_fetch.commit_sync_request_id != 0 &&
+            rko_orig->rko_u.share_fetch.commit_sync_request_id ==
+                rkcg->rkcg_commit_sync_request.id) {
+                rd_kafka_topic_partition_list_t *ack_results =
+                    rko_orig->rko_u.share_fetch.ack_results;
+
+                if (ack_results) {
+                        int k;
+                        for (k = 0; k < ack_results->cnt; k++) {
+                                rd_kafka_topic_partition_t *src =
+                                    &ack_results->elems[k];
+                                rd_kafka_topic_partition_t *dst =
+                                    rd_kafka_topic_partition_list_find(
+                                        rkcg->rkcg_commit_sync_request.results,
+                                        src->topic, src->partition);
+                                if (dst)
+                                        dst->err = src->err;
+                        }
+                }
+
+                /* TODO KIP-932: When broker reply has a top-level error
+                 * (e.g. SHARE_SESSION_NOT_FOUND, __TRANSPORT, __DESTROY)
+                 * ack_results is NULL but partitions handled by this
+                 * broker remain as _IN_PROGRESS sentinel. Should map
+                 * rko_orig->rko_err to all partitions that were sent
+                 * to this broker so they don't leak _IN_PROGRESS to
+                 * the caller. */
+
+                rkcg->rkcg_commit_sync_request.wait_broker_result_count--;
+
+                rd_kafka_dbg(
+                    rk, CGRP, "SHARE",
+                    "Commit sync reply from broker %s: %s, "
+                    "remaining brokers: %d",
+                    rd_kafka_broker_name(reply_rkb),
+                    rd_kafka_err2str(rko_orig->rko_err),
+                    rkcg->rkcg_commit_sync_request.wait_broker_result_count);
+
+                rd_kafka_share_commit_sync_maybe_complete(rk, rkcg);
+        }
+
+        /*
+         * Step 3: Dispatch pending commit_sync to the replying broker
+         * (highest priority). Acks must always be sent.
+         */
+        if (!reply_rkb->rkb_share_fetch_enqueued &&
+            !rd_kafka_broker_or_instance_terminating(reply_rkb) &&
+            reply_rkb->rkb_pending_commit_sync.sync_ack_details) {
+                rd_kafka_dbg(rk, CGRP, "SHARE",
+                             "Dispatching pending sync acks to broker %s",
+                             rd_kafka_broker_name(reply_rkb));
+                rd_kafka_share_enqueue_sync_ack_op(rk, reply_rkb);
+        }
+
+        /*
+         * Step 4: If this was the fetch broker but no records were received,
          * try to select another broker to fetch from.
          *
          * TODO: KIP-932: Handle the case where all assignments are removed
@@ -3243,7 +3307,7 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
         }
 
         /*
-         * Step 3: If the replying broker has cached ack details,
+         * Step 5: If the replying broker has cached async ack details,
          * send an ack-only SHARE_FETCH op to it.
          */
         if (reply_rkb->rkb_share_async_ack_details &&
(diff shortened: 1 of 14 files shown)


src/rdkafka.c
```diff
@@ -3766,6 +4251,106 @@ rd_kafka_error_t *rd_kafka_share_commit_async(rd_kafka_share_t *rkshare) {
...
+        rd_kafka_q_serve(rk->rk_rep, RD_POLL_NOWAIT, 0, RD_KAFKA_Q_CB_CALLBACK,
+                         rd_kafka_poll_cb, NULL);
+
+        rd_kafka_share_acknowledge_all_if_implicit(rkshare);
+
+        ack_batches = rd_kafka_share_build_ack_details(rk->rk_rkshare);
+
+        if (!ack_batches) {
+                rd_kafka_dbg(rk, CGRP, "SHARE",
+                             "No pending acknowledgements to commit sync");
```
PratRanj07:
```suggestion
                             "No pending acknowledgements to commit");
```

src/rdkafka.c
```diff
@@ -3723,6 +3801,411 @@ rd_kafka_op_res_t rd_kafka_share_commit_async_fanout_op(rd_kafka_t *rk,
...
+
+        /* Initialize commit_sync request state */
+        rkcg->rkcg_share.commit_sync_id_counter++;
+        if (rkcg->rkcg_share.commit_sync_id_counter == INT64_MAX)
+                rkcg->rkcg_share.commit_sync_id_counter = 1;
+        rkcg->rkcg_commit_sync_request.id =
+            rkcg->rkcg_share.commit_sync_id_counter;
+        rkcg->rkcg_commit_sync_request.abs_timeout = abs_timeout;
+        rkcg->rkcg_commit_sync_request.replyq      = rko->rko_replyq.q;
+        rko->rko_replyq.q = NULL; /* Take ownership of replyq */
```
PratRanj07: Keep all the rkcg assignment operations together

src/rdkafka.c
```diff
@@ -3723,6 +3801,411 @@ rd_kafka_op_res_t rd_kafka_share_commit_async_fanout_op(rd_kafka_t *rk,
...
+                rd_kafka_topic_partition_t *result_rktpar;
+
+                rktp = rd_kafka_topic_partition_toppar(rk, batch->rktpar);
+                if (!rktp || !rktp->rktp_leader) {
+                        result_rktpar = rd_kafka_topic_partition_list_find(
+                            rkcg->rkcg_commit_sync_request.results,
+                            batch->rktpar->topic, batch->rktpar->partition);
+                        if (result_rktpar)
+                                result_rktpar->err =
+                                    RD_KAFKA_RESP_ERR_LEADER_NOT_AVAILABLE;
```
PratRanj07, line 4039: Shouldn't it be problematic if result_rktpar is not found? Maybe add a debug log or rd_assert in that case to make sure it should always be there?

src/rdkafka_cgrp.h
```diff
@@ -402,7 +402,43 @@ typedef struct rd_kafka_cgrp_s {
                                                            * reply.
                                                            *   @locality main
                                                            * thread */
+                int64_t commit_sync_id_counter; /**< Global counter for
```
PratRanj07:
```suggestion
                int64_t commit_sync_request_id_counter; /**< Global counter for
```

src/rdkafka_cgrp.h
```diff
@@ -402,7 +402,43 @@ typedef struct rd_kafka_cgrp_s {
...
+                                      *   Set on each new commit_sync call.
+                                      *   Reset to 0 when the reply is sent.
+                                      *   0 means no active request. */
+                rd_ts_t abs_timeout; /**< Absolute timeout for the
+                                      *   commit_sync request. */
+                rd_kafka_topic_partition_list_t
+                    *results;                 /**< Partition-to-error mapping.
+                                               *   Populated as broker replies
+                                               *   arrive. */
+                int wait_broker_result_count; /**< Number of broker
```
PratRanj07:
```suggestion
                int brokers_awaiting_result_cnt; /**< Number of broker
```

or something similar

src/rdkafka_share_acknowledgement.c
```diff
@@ -372,6 +372,13 @@ rd_list_t *rd_kafka_share_build_ack_details(rd_kafka_share_t *rkshare) {
                                                         ->response_leader_epoch,
                                                     0);
                                         }
+                                        rd_kafka_dbg(
+                                            rkshare->rkshare_rk, CGRP,
+                                            "SHAREACK",
+                                            "    Adding ack detail for offsets "
```
PratRanj07, line 378: Why this extra space?

tests/0176-share_consumer_commit_sync.c
```diff
@@ -0,0 +1,1800 @@
...
+
+/**
+ * @brief Share consumer rd_kafka_share_commit_sync() API tests.
+ *
+ * Tests the commit_sync API in both implicit and explicit ack modes.
+ * Verifies that commit_sync synchronously commits acknowledged records
+ * and returns per-partition results.
+ */
+
+#define MAX_MSGS      500
```
PratRanj07: This is not needed

PratRanj07 (changes requested):
Some minor comments

src/rdkafka.c
```diff
@@ -3766,6 +4251,106 @@ rd_kafka_error_t *rd_kafka_share_commit_async(rd_kafka_share_t *rkshare) {
...
+        rd_kafka_dbg(rk, CGRP, "SHARE",
+                     "Committing synchronously with timeout %d ms", timeout_ms);
+
+        /* Drain rk_rep for all pending callbacks (non-blocking) */
+        rd_kafka_q_serve(rk->rk_rep, RD_POLL_NOWAIT, 0, RD_KAFKA_Q_CB_CALLBACK,
+                         rd_kafka_poll_cb, NULL);
+
+        rd_kafka_share_acknowledge_all_if_implicit(rkshare);
+
+        ack_batches = rd_kafka_share_build_ack_details(rk->rk_rkshare);
```
ojasvajain, line 4316: If multiple app threads call commitSync(), there can be race conditions while building the acks.

src/rdkafka.c
```diff
@@ -3723,6 +3801,411 @@ rd_kafka_op_res_t rd_kafka_share_commit_async_fanout_op(rd_kafka_t *rk,
...
+        int batch_cnt = rd_list_cnt(ack_batches);
+
+        while ((batch = rd_list_pop(ack_batches))) {
+                rd_kafka_toppar_t *rktp;
+                rd_kafka_broker_t *leader_rkb;
+                rd_kafka_share_ack_batches_t *existing;
+                rd_kafka_topic_partition_t *result_rktpar;
+
+                rktp = rd_kafka_topic_partition_toppar(rk, batch->rktpar);
+                if (!rktp || !rktp->rktp_leader) {
```
ojasvajain, line 4033: Should we wrap this condition in `unlikely()`?

src/rdkafka_broker.h
```diff
@@ -469,6 +473,25 @@ struct rd_kafka_broker_s { /* rd_kafka_broker_t */
...
+         * @locality main thread
+         */
+        struct {
+                rd_list_t *sync_ack_details;    /**< Ack batches waiting to be
+                                                 *   sent. Type:
+                                                 *   rd_kafka_share_ack_batches_t*.
+                                                 */
+                rd_ts_t abs_timeout;            /**< Absolute timeout from
+                                                 *   the commit_sync request. */
+                int64_t commit_sync_request_id; /**< Request ID this
```
ojasvajain, line 491: Can we get rid of these fields? They are only getting used while creating the rko to be sent from MT to BT. In that situation, we can use values from cgrp struct instead.

src/rdkafka.c
```diff
@@ -3209,7 +3215,65 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
...
+                    rkcg->rkcg_commit_sync_request.wait_broker_result_count);
+
+                rd_kafka_share_commit_sync_maybe_complete(rk, rkcg);
+        }
+
+        /*
+         * Step 3: Dispatch pending commit_sync to the replying broker
+         * (highest priority). Acks must always be sent.
+         */
+        if (!reply_rkb->rkb_share_fetch_enqueued &&
```
ojasvajain, line 3266: Do we need the check on `rkb_share_fetch_enqueued`? We are setting it to false at the beginning of this function.

src/rdkafka.c
```diff
@@ -3723,6 +3801,411 @@ rd_kafka_op_res_t rd_kafka_share_commit_async_fanout_op(rd_kafka_t *rk,
...
+            rko->rko_u.share_commit_sync_fanout.results;
+        int i;
+        int64_t timeout_us;
+
+        /* Take ownership from op */
+        rko->rko_u.share_commit_sync_fanout.ack_batches = NULL;
+        rko->rko_u.share_commit_sync_fanout.results     = NULL;
+
+        /* Initialize commit_sync request state */
+        rkcg->rkcg_share.commit_sync_id_counter++;
```
ojasvajain: Instead of having this global counter, can we generate the id using a random no. generator or use abs_timeout value? If yes, we can get rid of commit_sync_id_counter field.

src/rdkafka.c
```diff
@@ -3209,7 +3215,65 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
...
+         * Step 3: Dispatch pending commit_sync to the replying broker
+         * (highest priority). Acks must always be sent.
+         */
+        if (!reply_rkb->rkb_share_fetch_enqueued &&
+            !rd_kafka_broker_or_instance_terminating(reply_rkb) &&
+            reply_rkb->rkb_pending_commit_sync.sync_ack_details) {
+                rd_kafka_dbg(rk, CGRP, "SHARE",
+                             "Dispatching pending sync acks to broker %s",
+                             rd_kafka_broker_name(reply_rkb));
+                rd_kafka_share_enqueue_sync_ack_op(rk, reply_rkb);
```
ojasvajain, line 3272: Can we simply `return` after this?

ojasvajain (commented):
Did a 2nd pass. Just 1 more comment. Else looks good.

src/rdkafka_share_acknowledgement.c
pranavrth: Kept intentionally to differentiate visually.

src/rdkafka.c
pranavrth: Share consumer is not thread safe. We are going to add a check for it later.

src/rdkafka.c
pranavrth: Its fine as it is not the fast path.

src/rdkafka_broker.h
pranavrth: No. We need them to make sure that we are timing out or updating the correct commit sync requests from the response from the broker.

src/rdkafka.c
pranavrth: I think we should still keep it as it is a future guard.

src/rdkafka.c
pranavrth: As discussed, we will keep it as we don't want to complicate the solution.

ojasvajain (approved):

PratRanj07 (approved):
LGTM! Thanks @pranavrth

PR #5380, opened 2026-04-01
[KIP-932]: Implement acknowledgement callback

PratRanj07:
This PR intends to implement acknowledgement callback and call it for:
- Poll
- Commit Async
- Commit Sync


src/rdkafka_conf.c
```diff
@@ -1516,6 +1516,10 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      _RK(offset_commit_cb),
      "Offset commit result propagation callback. "
      "(set with rd_kafka_conf_set_offset_commit_cb())"},
+    {_RK_GLOBAL | _RK_CONSUMER, "share_acknowledgement_commit_cb", _RK_C_PTR,
```
pranavrth, line 1523: Let's understand how `offset_commit_cb` works and identify gaps in the implementation. Once we know the gaps, we can reason about if those are required or not. One of the gap is handling of this callback in events API.

src/rdkafka_fetcher.c
```diff
@@ -1207,11 +1208,26 @@ static rd_kafka_resp_err_t rd_kafka_share_fetch_reply_handle_partition(
...
+                            !rd_kafkap_str_cmp_str(topic,
+                                                   ack_batch->rktpar->topic) &&
+                            ack_batch->rktpar->partition == PartitionId) {
+                                rd_kafka_share_enqueue_ack_callback(
+                                    rkb->rkb_rk, ack_batch,
+                                    (rd_kafka_resp_err_t)
+                                        AcknowledgementErrorCode);
+                                break;
+                        }
+                }
```
pranavrth: I am planning to simplify this flow in the error handling PR. `ack_details` from both the `ShareAcknowledge` or `ShareFetch` RPC will have `err` field set once the response is received. This `ack_details` is received from the reply handler in the main thread. We can do both acknowledgement callback logic and commit sync reply from there in the main thread.

This simplifies multiple things
* ack_details destruction flow (currently at 6 places - later only 1)
* single place to handle error for commit sync and acknowledgement callback
* responsibility of each thread well defined
* doesn't create unnecessary ack_results for each RPC response

I think we should have this PR on top of error handling PR.

src/rdkafka_op.h
```diff
@@ -211,6 +212,8 @@ typedef enum {
         RD_KAFKA_OP_SHARE_FETCH_RESPONSE, /**< Share fetch response containing
                                            *   all messages and partition acks
                                            *   from a single broker response. */
+        RD_KAFKA_OP_SHARE_ACK_REPLY,      /**< Share acknowledgement callback
```
pranavrth: Better to name it `RD_KAFKA_OP_SHARE_ACK_COMMIT`.

src/rdkafka_op.h
```diff
@@ -211,6 +212,8 @@ typedef enum {
...
+                                           *   reply: broker -> app */
```
pranavrth: Planning to do from main thread to app thread.

src/rdkafka_op.h
```diff
@@ -879,6 +882,24 @@ struct rd_kafka_op_s {
...
+                        rd_kafka_share_partition_offsets_list_t *partitions;
+                        /** Callback function pointer. */
+                        void (*cb)(
+                            rd_kafka_share_t *rkshare,
+                            rd_kafka_share_partition_offsets_list_t *partitions,
+                            rd_kafka_resp_err_t err,
+                            void *opaque);
+                        /** Application opaque. */
+                        void *opaque;
+                } share_ack_reply;
```
pranavrth: `share_ack_commit`

src/rdkafka_share_acknowledgement.c
```diff
@@ -708,3 +708,241 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+ * @param topic Topic name (will be duplicated).
+ * @param partition Partition id.
+ * @param offsets_cnt Number of offsets to allocate space for.
+ */
+void rd_kafka_share_partition_offsets_init(
+    rd_kafka_share_partition_offsets_t *elem,
+    const char *topic,
+    int32_t partition,
+    int offsets_cnt) {
+        elem->partition.topic     = rd_strdup(topic);
```
pranavrth: Check if we need to add `id` as well. Verify with Java.

src/rdkafka_share_acknowledgement.c
```diff
@@ -708,3 +708,241 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+void rd_kafka_share_enqueue_ack_callback(rd_kafka_t *rk,
+                                         rd_kafka_share_ack_batches_t *batches,
+                                         rd_kafka_resp_err_t err) {
+        rd_kafka_op_t *cb_rko;
+        rd_kafka_share_partition_offsets_list_t *partitions;
+
+        if (!rk->rk_conf.share_acknowledgement_commit_cb)
+                return;
+
+        partitions = rd_kafka_share_build_partition_offsets_list(batches);
```
pranavrth, line 893: This is a partition list but Java sends 1 partition per callback. I think you are doing this already but can you just confirm?

src/rdkafka_share_acknowledgement.c
```diff
@@ -708,3 +708,241 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+
+
+int rd_kafka_share_partition_offsets_offsets_cnt(
+    const rd_kafka_share_partition_offsets_t *partition_offsets) {
+        return partition_offsets->cnt;
+}
+
+
+rd_kafka_share_ack_batches_t *
+rd_kafka_share_ack_batch_find(rd_list_t *ack_details,
```
pranavrth: Can we use `rd_kafka_share_find_ack_batch` here?

src/rdkafka_share_acknowledgement.h
```diff
@@ -202,4 +201,132 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
  */
 int rd_kafka_share_ack_entries_sort_cmp_ptr(const void *_a, const void *_b);
 
+
+/**
+ * @struct rd_kafka_share_partition_offsets_s
+ * @brief Partition with set of acknowledged offsets.
+ */
+struct rd_kafka_share_partition_offsets_s {
```
pranavrth, line 232: We should implement this in flexible array way to decrease allocation from 2 to 1. There is a discussion going on in the doc for the same as well. I will updated the doc once it is finalized but its mostly decided as this is a fixed structure which wouldn't be modified.

src/rdkafka_share_acknowledgement.h
```diff
@@ -202,4 +201,132 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
...
+};
+
+/**
+ * @struct rd_kafka_share_partition_offsets_list_s
+ * @brief List of share partition offsets for callback.
+ */
+struct rd_kafka_share_partition_offsets_list_s {
+        int cnt;                                   /**< Number of partitions */
+        int size;                                  /**< Allocated size */
+        rd_kafka_share_partition_offsets_t *elems; /**< Array of partition
```
pranavrth: - We can remove size field. We won't add more elements.
- We should use flexible array approach here as well. Same discussion is going on for this as well. Though we need to have array of pointer to `rd_kafka_share_partition_offsets_t` so that we can calculate the allocation size correctly.

src/rdkafka_share_acknowledgement.c
```diff
@@ -708,3 +708,241 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+ * @param partition Partition id.
+ * @param offsets_cnt Number of offsets to allocate space for.
+ */
+void rd_kafka_share_partition_offsets_init(
+    rd_kafka_share_partition_offsets_t *elem,
+    const char *topic,
+    int32_t partition,
+    int offsets_cnt) {
+        elem->partition.topic     = rd_strdup(topic);
+        elem->partition.partition = partition;
```
pranavrth: Any reason for not using `rd_kafka_topic_partition_*_new()`? If not, we should use that with id as well. We shouldn't have reference to internal `toppar_t` struct though.

tests/0170-share_consumer_subscription.c
```diff
@@ -618,8 +618,20 @@ static void state_cleanup(sub_test_state_t *state) {
         /* Destroy all consumers */
         for (i = 0; i < MAX_CONSUMERS; i++) {
                 if (state->consumers[i]) {
+                        TEST_SAY(
+                            "[state_cleanup/consumer%d]: Closing share "
+                            "consumer\n",
+                            i);
                         rd_kafka_share_consumer_close(state->consumers[i]);
```
pranavrth: Better to create a common function `test_share_consumer_close()` in the `test.c` and use that instead. We can have all the TEST_SAY in there. Check other functions in `test.c`.

I would suggest doing these changes in other PR. We want to review ack callback related tests in this PR.

tests/0170-share_consumer_subscription.c
```diff
@@ -618,8 +618,20 @@ static void state_cleanup(sub_test_state_t *state) {
...
+                        TEST_SAY(
+                            "[state_cleanup/consumer%d]: Destroying share "
+                            "consumer\n",
+                            i);
                         rd_kafka_share_destroy(state->consumers[i]);
```
pranavrth: Better to create a common function `test_share_consumer_destroy()` in the `test.c` and use that instead. We can have all the TEST_SAY in there. Check other functions in `test.c`.

 I would suggest doing these changes in other PR. We want to review ack callback related tests in this PR.

pranavrth (changes requested):
High level first pass implementation comments.

src/rdkafka_share_acknowledgement.c
PratRanj07: This is not needed now

src/rdkafka_share_acknowledgement.c
PratRanj07: Yes Java has topic_id too here. Added that

src/rdkafka_share_acknowledgement.c
PratRanj07: Yes it is called for per partition only

src/rdkafka.c
```diff
@@ -3308,7 +3308,14 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
...
         /*
-         * Step 4: Dispatch pending commit_sync to the replying broker
+         * Step 4: Dispatch acknowledgement callbacks.
+         * Per-partition errors were set by the broker thread.
+         */
+        rd_kafka_share_dispatch_ack_callbacks(
+            rk, rko_orig->rko_u.share_fetch.ack_details);
+
+        /*
+         * Step 5: Dispatch pending commit_sync to the replying broker
```
pranavrth: This should be the first thing that we do in this reply handler.

src/rdkafka_share_acknowledgement.h
```diff
@@ -32,7 +32,6 @@
 #include "rdkafka.h"
 
 /* Forward declarations */
-typedef struct rd_kafka_share_s rd_kafka_share_t;
 typedef struct rd_kafka_op_s rd_kafka_op_t;
```
pranavrth, line 32: Wait. Why did we have these in the first place? Remove `typedef struct rd_kafka_op_s rd_kafka_op_t;` as well.

src/rdkafka_share_acknowledgement.h
```diff
@@ -32,7 +32,6 @@
 #include "rdkafka.h"
```
pranavrth: We shouldn't have this import as well.

src/rdkafka_share_acknowledgement.h
pranavrth: Check if this is needed or not.

src/rdkafka_share_acknowledgement.h
```diff
@@ -228,4 +227,116 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
...
+void rd_kafka_share_enqueue_ack_callback(rd_kafka_t *rk,
+                                         rd_kafka_share_ack_batches_t *batches,
+                                         rd_kafka_resp_err_t err);
+
+
+/**
+ * @brief Dispatch ack callbacks for all partitions in ack_details.
+ *
+ * If err is set (top-level error), all partitions receive the same error.
+ * Otherwise, per-partition results from ack_results are used.
```
pranavrth: Let's not write this.

src/rdkafka_share_acknowledgement.h
```diff
@@ -228,4 +227,116 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
...
+ * @param batches The ack batches for this partition (contains offsets).
+ * @param err Error code to report in callback.
+ */
+void rd_kafka_share_enqueue_ack_callback(rd_kafka_t *rk,
+                                         rd_kafka_share_ack_batches_t *batches,
+                                         rd_kafka_resp_err_t err);
+
+
+/**
+ * @brief Dispatch ack callbacks for all partitions in ack_details.
```
pranavrth: Explain a bit more on what we are doing. We will enqueue op 1 by 1.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,212 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+        if (!rk->rk_conf.share_acknowledgement_commit_cb || !ack_details ||
+            rd_list_cnt(ack_details) == 0)
+                return;
+
+        /* Use per-partition error from each batch */
+        RD_LIST_FOREACH(ack_batch, ack_details, k) {
+                rd_kafka_share_enqueue_ack_callback(
+                    rk, ack_batch,
+                    ack_batch->rktpar ? ack_batch->rktpar->err
+                                      : RD_KAFKA_RESP_ERR_NO_ERROR);
```
pranavrth: We can expect `ack_batch` to always have `rktpar`.

src/rdkafka.c
```diff
@@ -5718,6 +5725,14 @@ rd_kafka_op_res_t rd_kafka_poll_cb(rd_kafka_t *rk,
                                             rko->rko_u.offset_commit.opaque);
                 break;
 
+        case RD_KAFKA_OP_SHARE_ACK_COMMIT:
```
pranavrth: change name to include _CB at the end

src/rdkafka_share_acknowledgement.h
```diff
@@ -228,4 +227,116 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
...
+
+/**
+ * @brief Destroy a partition offsets list.
+ *
+ * Frees all elements and the list itself.
+ *
+ * @param list List to destroy (may be NULL).
+ */
+void rd_kafka_share_partition_offsets_list_destroy(
+    rd_kafka_share_partition_offsets_list_t *list);
```
pranavrth: Don't expose these functions.

src/rdkafka_share_acknowledgement.h
```diff
@@ -228,4 +227,116 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
...
+ *
+ * Creates and enqueues a callback op for the given partition with the
+ * specified error code. Used for both per-partition acknowledgement results
+ * and top-level errors.
+ *
+ * @param rk Kafka handle.
+ * @param batches The ack batches for this partition (contains offsets).
+ * @param err Error code to report in callback.
+ */
+void rd_kafka_share_enqueue_ack_callback(rd_kafka_t *rk,
```
pranavrth:
```suggestion
void rd_kafka_share_enqueue_ack_commit_cb_op(rd_kafka_t *rk,
```

tests/0176-share_consumer_commit_sync.c
```diff
@@ -1931,6 +1931,437 @@ static void do_test_mock_broker_dispatch_priority(void) {
 }
 
 
+/* ===================================================================
```
pranavrth, line 1934: Move common functions to `test.c`

pranavrth (commented):
Please add test for 0182 as well.

src/rdkafka_share_acknowledgement.h
PratRanj07: We need this particular forward declaration as we would have a circular dependency if we include rdkafka_op.h where this particular rd_kafka_op_t is present
rdkafka_share_acknowledgement.h->rdkafka_op.h->rdkafka_msg.h->rdkafka_share_acknowledgement.h

pranavrth:
There are 2 things that we are not doing which are expected:
- The user can update the callback on the fly. Basically user can update the acknowledgement callback which consumer is already running. This is a design gap. I think we should confirm this before implementing though.
- We need to make sure that user is not calling any other function from the callback.

src/rdkafka_op.h
```diff
@@ -40,6 +40,7 @@ typedef struct rd_kafka_q_s rd_kafka_q_t;
 typedef struct rd_kafka_toppar_s rd_kafka_toppar_t;
 typedef struct rd_kafka_op_s rd_kafka_op_t;
 typedef struct rd_kafka_broker_s rd_kafka_broker_t;
+typedef struct rd_kafka_share_ack_result_s rd_kafka_share_ack_result_t;
```
pranavrth: This is not used anywhere.

src/rdkafka_share_acknowledgement.h
pranavrth: Let's try to break this circular dependency chain. Let's move the functions that requires this to other file maybe.

src/rdkafka_conf.c
pranavrth: Let's add a TODO to verify if events API part is needed or not.

src/rdkafka_share_acknowledgement.h
```diff
@@ -228,4 +223,90 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
...
 
+
+/**
+ * @struct rd_kafka_share_partition_offsets_s
+ * @brief Partition with set of acknowledged offsets.
+ */
+struct rd_kafka_share_partition_offsets_s {
+        rd_kafka_topic_partition_t
+            *partition;    /**< Topic partition information */
+        int cnt;           /**< Number of offsets in array */
```
pranavrth: we should use `size_t` itself. I will update in the doc as well.

src/rdkafka_share_acknowledgement.h
```diff
@@ -228,4 +223,90 @@ rd_kafka_share_ensure_all_acknowledged_if_explicit(rd_kafka_share_t *rkshare);
...
+        int64_t offsets[]; /**< Flexible array of acknowledged offsets */
+};
+
+/**
+ * @struct rd_kafka_share_partition_offsets_list_s
+ * @brief List of share partition offsets for callback.
+ */
+struct rd_kafka_share_partition_offsets_list_s {
+        int cnt; /**< Number of partitions */
```
pranavrth: same. `size_t`

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+}
+
+
+const int64_t *rd_kafka_share_partition_offsets_offsets(
+    const rd_kafka_share_partition_offsets_t *partition_offsets) {
+        return partition_offsets->offsets;
+}
+
+
+int rd_kafka_share_partition_offsets_offsets_cnt(
```
pranavrth: return size_t

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+        cb_rko->rko_u.share_ack_commit.opaque = rk->rk_conf.opaque;
+
+        rd_kafka_q_enq(rk->rk_rep, cb_rko);
+}
+
+
+size_t rd_kafka_share_partition_offsets_list_count(
+    const rd_kafka_share_partition_offsets_list_t *list) {
+        if (!list)
+                return 0;
```
pranavrth: I don't think we should do NULL check to be consistent with other places.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+        return (size_t)list->cnt;
+}
+
+
+const rd_kafka_share_partition_offsets_t *
+rd_kafka_share_partition_offsets_list_get(
+    const rd_kafka_share_partition_offsets_list_t *list,
+    size_t index) {
+        if (!list || index >= (size_t)list->cnt)
+                return NULL;
```
pranavrth: Let's not do this check.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+    rd_kafka_share_ack_batches_t *batches) {
+        rd_kafka_share_partition_offsets_list_t *list;
+        rd_kafka_share_partition_offsets_t *elem;
+        rd_kafka_share_ack_batch_entry_t *entry;
+        int total_offsets = 0;
+        int offset_idx    = 0;
+        int j;
+
+        if (!batches || !batches->rktpar || rd_list_cnt(&batches->entries) == 0)
+                return NULL;
```
pranavrth: I think we should remove most of the NULL checks to be consistent with the library. Add only if really necessary.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+
+        /* Count total offsets */
+        RD_LIST_FOREACH(entry, &batches->entries, j) {
+                total_offsets +=
+                    (int)(entry->end_offset - entry->start_offset + 1);
+        }
+
+        if (total_offsets == 0)
+                return NULL;
```
pranavrth: Make sure that batch should always have offsets. If not then there is some issue in the code. We should fix that instead of this check.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+
+        list = rd_kafka_share_partition_offsets_list_new(1);
+        if (!list)
+                return NULL;
```
pranavrth: Same. No need for NULL check.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+ *          Caller must destroy with
+ *          rd_kafka_share_partition_offsets_list_destroy().
+ */
+static rd_kafka_share_partition_offsets_list_t *
+rd_kafka_share_partition_offsets_list_new(int capacity) {
+        rd_kafka_share_partition_offsets_list_t *list;
+        size_t size;
+
+        if (capacity <= 0)
+                return NULL;
```
pranavrth: Remove this null check.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+
+        /* Allocate list + flexible array in one allocation */
+        size = sizeof(*list) + ((size_t)capacity *
+                                sizeof(rd_kafka_share_partition_offsets_t *));
+        list = rd_malloc(size);
```
pranavrth: use calloc instead.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+rd_kafka_share_partition_offsets_new(rd_kafka_Uuid_t topic_id,
+                                     const char *topic,
+                                     int32_t partition,
+                                     int offsets_cnt) {
+        rd_kafka_share_partition_offsets_t *elem;
+        size_t size;
+
+        /* Allocate struct + flexible array in one allocation */
+        size = sizeof(*elem) + ((size_t)offsets_cnt * sizeof(int64_t));
+        elem = rd_malloc(size);
```
pranavrth: calloc instead.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,211 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+    rd_kafka_resp_err_t err) {
+        rd_kafka_op_t *cb_rko;
+        rd_kafka_share_partition_offsets_list_t *partitions;
+
+        if (!rk->rk_conf.share_acknowledgement_commit_cb)
+                return;
+
+        partitions = rd_kafka_share_build_partition_offsets_list(batches);
+        if (!partitions)
+                return;
```
pranavrth: We shouldn't reach this. Confirm and remove this.

tests/0173-share_consumer_commit_async.c
```diff
@@ -1576,6 +1576,310 @@ static void do_test_lock_timeout_redelivery(void) {
 }
 
 
+/* ===================================================================
+ *  Acknowledgement callback helpers.
+ * =================================================================== */
+
+#define MAX_CB_OFFSETS 500
```
pranavrth: Remove unused `MAX_CB_OFFSETS`

pranavrth (changes requested):
Some nits. Not going through the tests as we might need to change the way we have implemented ack callback.

src/rdkafka.h
```diff
@@ -2167,6 +2167,115 @@ void rd_kafka_conf_set_offset_commit_cb(
...
+ */
+RD_EXPORT size_t rd_kafka_share_partition_offsets_offsets_cnt(
+    const rd_kafka_share_partition_offsets_t *partition_offsets);
+
+/**
+ * @brief Set share acknowledgement commit callback.
+ *
+ * The share acknowledgement commit callback is called when share consumer
+ * acknowledgements (from rd_kafka_share_acknowledge* calls followed by
+ * rd_kafka_share_commit_async() or rd_kafka_share_consume_batch()) complete.
```
pranavrth: We can have it with _sync() as well.

src/rdkafka.h
pranavrth, line 1: I think we will need to update this information for all the APIs before releasing so not checking the doc thoroughly.

src/rdkafka.h
```diff
@@ -2167,6 +2167,115 @@ void rd_kafka_conf_set_offset_commit_cb(
...
+RD_EXPORT const rd_kafka_topic_partition_t *
+rd_kafka_share_partition_offsets_partition(
+    const rd_kafka_share_partition_offsets_t *partition_offsets);
+
+/**
+ * @brief Get the offsets array from a partition offsets entry.
+ * @param partition_offsets The partition offsets entry.
+ * @returns Array of acknowledged offsets (valid for the lifetime of the list).
+ */
+RD_EXPORT const int64_t *rd_kafka_share_partition_offsets_offsets(
```
pranavrth, line 2235: This feels weird name. Can you add a TODO to recheck this name once the public interfaces are finalized.

src/rdkafka.h
```diff
@@ -2167,6 +2167,115 @@ void rd_kafka_conf_set_offset_commit_cb(
...
+
+/**
+ * @brief Set share acknowledgement commit callback.
+ *
+ * The share acknowledgement commit callback is called when share consumer
+ * acknowledgements (from rd_kafka_share_acknowledge* calls followed by
+ * rd_kafka_share_commit_async() or rd_kafka_share_consume_batch()) complete.
+ *
+ * The callback is called from rd_kafka_share_consume_batch() or
+ * rd_kafka_poll() context.
```
pranavrth: This is wrong. Shouldn't mention this.

src/rdkafka.h
```diff
@@ -2167,6 +2167,115 @@ void rd_kafka_conf_set_offset_commit_cb(
...
+ *
+ * @param conf Configuration object.
+ * @param share_acknowledgement_commit_cb Callback to set.
+ *
+ * The callback parameters are:
+ *   - \p rkshare: Share consumer handle.
+ *   - \p partitions: List of partitions with their acknowledged offsets.
+ *   - \p err: Overall error code (RD_KAFKA_RESP_ERR_NO_ERROR on success).
```
pranavrth: Explain that this `err` applies to all the partitions.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,194 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+rd_kafka_share_partition_offsets_list_new(size_t capacity) {
+        rd_kafka_share_partition_offsets_list_t *list;
+        size_t size;
+
+        /* Allocate list + flexible array in one allocation */
+        size = sizeof(*list) +
+               (capacity * sizeof(rd_kafka_share_partition_offsets_t *));
+        list = rd_calloc(1, size);
+
+        list->cnt = 0;
```
pranavrth: `calloc` will ensure this.

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,194 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+                                     size_t offsets_cnt) {
+        rd_kafka_share_partition_offsets_t *elem;
+        size_t size;
+
+        /* Allocate struct + flexible array in one allocation */
+        size = sizeof(*elem) + (offsets_cnt * sizeof(int64_t));
+        elem = rd_calloc(1, size);
+
+        elem->partition = rd_kafka_topic_partition_new_with_id_and_name(
+            topic_id, topic, partition);
```
pranavrth: Why don't we use `rd_kafka_topic_partition_copy` directly?

src/rdkafka_share_acknowledgement.c
```diff
@@ -781,3 +781,194 @@ rd_kafka_share_acknowledge_offset(rd_kafka_share_t *rkshare,
...
+ * @brief Initialize a partition offsets element with topicid,topic, partition,
+ * and offsets.
+ *
+ * @param topic_id Topic UUID.
+ * @param topic Topic name (will be duplicated).
+ * @param partition Partition id.
+ * @param offsets_cnt Number of offsets to allocate space for.
+ */
+static rd_kafka_share_partition_offsets_t *
+rd_kafka_share_partition_offsets_new(rd_kafka_Uuid_t topic_id,
```
pranavrth: Let's take `topic_partition_t` as input.

src/rdkafka_op.h
```diff
@@ -979,6 +999,12 @@ rd_kafka_op_process_share_fetch_response(rd_kafka_op_t *rko,
                                          rd_kafka_message_t **rkmessages,
                                          unsigned int cnt);
 
+void rd_kafka_share_build_inflight_acks_map(rd_kafka_share_t *rkshare,
+                                            rd_kafka_op_t *response_rko);
+
+void rd_kafka_share_fetch_op_reply_with_err(rd_kafka_op_t *rko,
+                                            rd_kafka_resp_err_t err);
```
pranavrth: We will have to move the implementation as well. Let's discuss this part first and then do the change.

pranavrth (commented):
Few minor comments. Still going through the tests.

src/rdkafka.h
```diff
@@ -2167,6 +2167,121 @@ void rd_kafka_conf_set_offset_commit_cb(
...
+ *
+ * @param conf Configuration object.
+ * @param share_acknowledgement_commit_cb Callback to set.
+ *
+ * The callback parameters are:
+ *   - \p rkshare: Share consumer handle.
+ *   - \p partitions: List containing exactly one partition with its
+ *                    acknowledged offsets.
+ *   - \p err: Error code for this partition (RD_KAFKA_RESP_ERR_NO_ERROR on
+ *             success). This error applies to all offsets in the partition.
```
pranavrth, line 2265:
```suggestion
 *   - \p err: Error code for these partitions (RD_KAFKA_RESP_ERR_NO_ERROR on
 *             success). This error applies to all offsets in the partition.
```

pranavrth (approved):
Provided a nit that we can fix in the next PR. LGTM otherwise!

Thanks @PratRanj07 !.

PR #5381, opened 2026-04-01
Wrote tests for checking transactions and control records filtering h…

PratRanj07:
Share Consumer Transaction Tests

Tests share consumer behavior with transactional messages using the broker-side share.isolation.level configuration (KIP-932). Unlike regular consumers where isolation level is a client config, share consumers rely on broker-side filtering configured via share group settings.

**Isolation Level Behavior**
Each test runs twice - once with read_committed and once with read_uncommitted

1. Committed transactions - Produces 100 messages in a committed transaction; verifies both isolation levels receive all messages (baseline sanity check)

2. Aborted transactions - Produces 100 messages then aborts the transaction; read_committed should receive 0 messages while read_uncommitted receives all 100
 
3. Mixed commit/abort transactions - Executes 5 transactions with varying message counts (50 commit, 50 abort, 30 commit, 20 abort, 40 commit); validates read_committed sees only 120 committed messages while read_uncommitted sees all 190

4. Control record filtering - Produces 1 aborted + 5 committed messages, creating control records at specific offsets; verifies commit/abort markers (offsets 1 and 7) are never delivered to applications regardless of isolation level; validates exact offset sequences

5. Multi-partition transactions - Creates topic with 3 partitions; runs 3 transactions (commit, abort, commit) each writing to all partitions; verifies correct aggregate counts across partitions

6. Interleaved producers - Two transactional producers begin transactions simultaneously, interleave their writes, then one commits while the other aborts; tests broker's ability to track independent transaction states

7. Mixed transactional and non-transactional - Combines 50 non-transactional messages with 30 committed + 20 aborted transactional messages; verifies non-transactional messages are always visible and transaction filtering only applies to transactional batches
 
8. Interval-based abort pattern - Produces 20 transactions (5 messages each) where every 5th transaction is aborted; read_committed should see exactly 80 messages (16 committed × 5), read_uncommitted sees all 100

**Dynamic Isolation Level Change**

Runtime isolation level switch (read_uncommitted → read_committed) - Starts consumer with read_uncommitted, consumes committed and aborted transactions, then changes group config to read_committed via IncrementalAlterConfigs; produces new aborted transaction and verifies it's now filtered; produces committed transaction and verifies it's still visible. Demonstrates isolation level changes take effect on next poll without requiring consumer restart.

### DIFF
--- a/tests/0177-share_consumer_transactions.c
+++ b/tests/0177-share_consumer_transactions.c
@@ -1,0 +1,1139 @@
+/*
+ * librdkafka - Apache Kafka C library
+ *
+ * Copyright (c) 2025, Confluent Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test.h"
+
+/**
+ * @brief Share consumer transaction tests.
+ *
+ * Tests that share consumers correctly handle transactional messages
+ * based on the isolation.level configuration:
+ *
+ * - read_committed: Only see committed messages, aborted filtered out
+ * - read_uncommitted: See ALL messages including aborted ones
+ *
+ * Both modes are tested to verify correct behavior.
+ */
+
+#define BATCH_SIZE 1000
+
+
+/**
+ * @brief Create a transactional producer.
+ */
+static rd_kafka_t *create_txn_producer(const char *txn_id) {
+        rd_kafka_t *rk;
+        rd_kafka_conf_t *conf;
+
+        test_conf_init(&conf, NULL, 60);
+        test_conf_set(conf, "transactional.id", txn_id);
+        rd_kafka_conf_set_dr_msg_cb(conf, test_dr_msg_cb);
+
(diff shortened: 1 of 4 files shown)


tests/0177-share_consumer_transactions.c
```diff
@@ -0,0 +1,1194 @@
...
+        return consumed;
+}
+
+
+/* ===================================================================
+ *  Test: Committed transaction visibility.
+ *
+ *  Both read_committed and read_uncommitted should see committed
+ *  messages.
+ * =================================================================== */
```
Ankith-Confluent: Are there plans to remove these markers later(`====`) ?

tests/0177-share_consumer_transactions.c
```diff
@@ -0,0 +1,1194 @@
...
+        rd_kafka_share_destroy(consumer);
+        rd_kafka_destroy(producer);
+        rd_kafka_destroy(admin_client);
+
+        SUB_TEST_PASS();
+}
+
+
+int main_0177_share_consumer_transactions(int argc, char **argv) {
+
```
Ankith-Confluent: Maybe we need a `test_timeout_set()` here?

tests/0177-share_consumer_transactions.c
```diff
@@ -0,0 +1,1194 @@
...
+
+/**
+ * @brief Create a share consumer with specified group.
+ *
+ * Note: For share consumers, isolation.level is configured on the
+ * BROKER-SIDE via share group configuration (share.isolation.level),
+ * not on the client. See configure_share_group().
+ */
+static rd_kafka_share_t *create_share_consumer(const char *group) {
```
Ankith-Confluent: why not use the existing `test_create_share_consumer` function to create a share consumer?

tests/0177-share_consumer_transactions.c
```diff
@@ -0,0 +1,1194 @@
...
+ *
+ * Note: share.isolation.level is a BROKER-SIDE share group configuration.
+ * Transaction filtering happens on the broker, not the client.
+ *
+ * @param admin_rk Admin client handle (producer) for admin API calls.
+ */
+static void configure_share_group(rd_kafka_t *admin_rk,
+                                  const char *group,
+                                  const char *isolation_level) {
+        const char *offset_cfg[]    = {"share.auto.offset.reset", "SET",
```
Ankith-Confluent: This can be combined into a single call
`const char *cfg[] = {"share.auto.offset.reset", "SET", "earliest",
                     "share.isolation.level", "SET", isolation_level};
test_IncrementalAlterConfigs_simple(rk, RD_KAFKA_RESOURCE_GROUP, group, cfg, 2);`

tests/0177-share_consumer_transactions.c
```diff
@@ -0,0 +1,1194 @@
...
+         * - Offset 7: commit control record
+         *
+         * read_committed: should see offsets 2,3,4,5,6 (committed only)
+         * read_uncommitted: should see offsets 0,2,3,4,5,6 (aborted +
+         * committed) Control records (offsets 1 and 7) should NEVER be visible.
+         */
+        expected_msgs = is_read_committed ? 5 : 6;
+
+        /* Consume and verify exact offsets */
+        {
```
Ankith-Confluent: Can it be written without this block? (without creating a new scope)

tests/0177-share_consumer_transactions.c
```diff
@@ -0,0 +1,1194 @@
...
+
+        /* Test both isolation levels */
+        const char *isolation_levels[] = {"read_committed", "read_uncommitted"};
+        int i;
+
+        for (i = 0; i < 2; i++) {
+                const char *level = isolation_levels[i];
+
+                TEST_SAY("\n");
+                TEST_SAY("========================================\n");
```
Ankith-Confluent, line 1099: Will these be removed later?

Ankith-Confluent (commented):
I have provided the first round of review

tests/0177-share_consumer_transactions.c
PratRanj07: These aren't very long running tests so ideally they shouldn't need a test timeout set, the default timeout should work

tests/0177-share_consumer_transactions.c
PratRanj07: These are just seperators and I think we should keep them for reading easibility on the test logs

tests/0177-share_consumer_transactions.c
```diff
@@ -0,0 +1,1135 @@
...
+
+
+/**
+ * @brief Create a transactional producer.
+ */
+static rd_kafka_t *create_txn_producer(const char *txn_id) {
+        rd_kafka_t *rk;
+        rd_kafka_conf_t *conf;
+
+        test_conf_init(&conf, NULL, 60);
```
Ankith-Confluent, line 57: should this be set to 0?

Ankith-Confluent (commented):
One small comment

Ankith-Confluent (approved):
LGTM

PR #5391, opened 2026-04-07
Add s390x architecture support for CI, packaging, and binary validation

Devarsh010:


packaging/nuget/nugetpackage.py
```diff
@@ -109,6 +109,15 @@ class NugetPackage (Package):
                 './usr/local/lib/librdkafka.so.1',
                 'runtimes/linux-arm64/native/alpine-librdkafka.so'),
 
+        # Linux glibc centos8 s390x without GSSAPI (no external deps)
```
prashah-confluent, line 112: Check with @emasab on whether we need similar one for `with GSSAPI`

packaging/nuget/packaging.py
```diff
@@ -46,6 +46,7 @@ def unquote(path):
     ('win', 'x86', '.lib'): re.compile('current ar archive'),
     ('linux', 'x64', '.so'): re.compile('ELF 64.* x86-64'),
     ('linux', 'arm64', '.so'): re.compile('ELF 64.* ARM aarch64'),
+    ('linux', 's390x', '.so'): re.compile('ELF 64.* IBM S/390'),
```
prashah-confluent, line 49: If we run `file librdkafka.so` then it does print the below string
`artifacts-s390x-qemu/usr/local/lib/librdkafka.so.1: ELF 64-bit MSB shared object, IBM S/390, version 1 (SYSV), dynamically linked, BuildID[sha1]=`

The format looks correct but lets reconfirm once if we need this change.

prashah-confluent (commented):
Lets do couple of checks as commented below

PR #5396, opened 2026-04-08
[KIP-932] Add Share Consumer mode to rdkafka_performance example

Ankith-Confluent:
Added -S <groupid> option to the performance example for benchmarking
share consumer throughput and latency using implicit acknowledgement
mode with rd_kafka_share_consume_batch().
Change sequence to use -F instead.

### DIFF
--- a/examples/rdkafka_performance.c
+++ b/examples/rdkafka_performance.c
@@ -1785,14 +1785,12 @@ int main(int argc, char **argv) {
                 rd_kafka_message_t **rkmessages = NULL;
                 int share_batch_size;
 
-                share_batch_size =
-                    batch_size ? batch_size : 10001;
-                rkmessages =
-                    malloc(sizeof(*rkmessages) * share_batch_size);
+                share_batch_size = batch_size ? batch_size : 10001;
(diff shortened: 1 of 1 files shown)


examples/rdkafka_performance.c
```diff
@@ -1748,6 +1778,73 @@ int main(int argc, char **argv) {
...
                 rd_kafka_queue_destroy(rkqu);
                 rd_kafka_destroy(rk);
+
+        } else if (mode == 'S') {
+                /* Share Consumer */
+                rd_kafka_share_t *rkshare;
+                rd_kafka_message_t **rkmessages = NULL;
+                int share_batch_size;
+
+                share_batch_size = batch_size ? batch_size : 10001;
```
PratRanj07: I think we should just keep it as default 10001. We would have to think about the batch_size that we are taking as an input. How do we want to process that with max.poll.records or like something else

PratRanj07 (changes requested):
Thanks @Ankith-Confluent for the PR!
One change requested

PratRanj07 (approved):
LGTM!

PR #5398, opened 2026-04-10
admin: Fix duplicate groups in ListConsumerGroups response merge

adizaimi:
ListConsumerGroups fans out ListGroupsRequest to all brokers and merges the per-broker responses.  The merge step used rd_list_copy_to() to unconditionally append all valid group listings from each broker, producing duplicate entries when multiple brokers returned the same group id.

Replace the bulk copy with a deduplicating helper, rd_kafka_ConsumerGroupListing_list_append_dedup(), that uses an ephemeral rd_map_t keyed by group_id for O(1) lookups.  When the same group is reported by multiple brokers the first response wins; this is acceptable because group state is point-in-time and any single broker s view is equally valid.

Note: the helper assumes dst contains no duplicates on entry.  The current code path upholds this (dst starts empty and is only modified through this function).  If future changes break that assumption the function will still not introduce new duplicates, but will not clean up pre-existing ones either.

Add 4 unit tests covering: identical groups, disjoint groups, three-broker complex overlap, and duplicates with interleaved errors.

### DIFF
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -7608,14 +7608,13 @@ rd_kafka_ConsumerGroupListing_list_append_dedup(rd_list_t *dst,
                 return;
 
         /* Ensure dst is sorted for bsearch lookups. */
-        if (rd_list_cnt(dst) && !(dst->rl_flags & RD_LIST_F_SORTED))
+        if (!(dst->rl_flags & RD_LIST_F_SORTED))
                 rd_list_sort(dst, rd_kafka_ConsumerGroupListing_cmp_group_id);
 
         /* Collect pointers to src groups not already in dst.
          * dst is not modified here so the sorted flag stays valid. */
         RD_LIST_FOREACH(grp, src, j) {
-                if (rd_list_cnt(dst) &&
-                    rd_list_find(dst, grp,
+                if (rd_list_find(dst, grp,
                                  rd_kafka_ConsumerGroupListing_cmp_group_id))
                         continue;
                 if (!to_add)
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -7618,10 +7618,8 @@ rd_kafka_ConsumerGroupListing_list_append_dedup(rd_list_t *dst,
                     rd_list_find(dst, grp,
                                 rd_kafka_ConsumerGroupListing_cmp_group_id))
                         continue;
-                if (!to_add) {
-                        to_add = rd_alloca(sizeof(*to_add));
-                        rd_list_init(to_add, rd_list_cnt(src), NULL);
-                }
+                if (!to_add)
+                        to_add = rd_list_new(rd_list_cnt(src), NULL);
                 rd_list_add(to_add, (void *)grp);
         }
 
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -10138,6 +10138,44 @@ static int ut_ListConsumerGroups_response_merge(void) {
                 rd_kafka_op_destroy(rko_fanout);
         }
 
+       /*
+         * Test 5: Empty broker response interleaved.
+         * Broker1: {A, B}, Broker2: {}, Broker3: {B, C}
+         * → expect {A, B, C}
+         */
+        {
+                const char *groups1[]  = {"groupB", "groupA"};
+                const char *groups3[]  = {"groupC", "groupB"};
+                const char *expected[] = {"groupA", "groupB", "groupC"};
+
+                rko_fanout = ut_make_ListConsumerGroups_fanout();
+
+                rko_partial =
+                    ut_make_ListConsumerGroups_partial(groups1, 2);
+                rd_kafka_ListConsumerGroups_response_merge(rko_fanout,
+                                                          rko_partial);
+                rd_kafka_op_destroy(rko_partial);
+
+                rko_partial =
+                    ut_make_ListConsumerGroups_partial(NULL, 0);
+                rd_kafka_ListConsumerGroups_response_merge(rko_fanout,
+                                                          rko_partial);
+                rd_kafka_op_destroy(rko_partial);
+
+                rko_partial =
+                    ut_make_ListConsumerGroups_partial(groups3, 2);
+                rd_kafka_ListConsumerGroups_response_merge(rko_fanout,
+                                                          rko_partial);
+                rd_kafka_op_destroy(rko_partial);
+
+                if (ut_verify_ListConsumerGroups_result(
+                        rko_fanout, expected, 3, 0,
+                        "empty broker response interleaved"))
+                        return 1;
+
+                rd_kafka_op_destroy(rko_fanout);
+        }
+
         RD_UT_PASS();
 }
 
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -10118,7 +10118,7 @@ static int ut_ListConsumerGroups_response_merge(void) {
                 rd_kafka_op_destroy(rko_fanout);
         }
 
-       /*
+        /*
          * Test 5: Empty broker response interleaved.
          * Broker1: {A, B}, Broker2: {}, Broker3: {B, C}
          * → expect {A, B, C}
@@ -10130,22 +10130,19 @@ static int ut_ListConsumerGroups_response_merge(void) {
 
                 rko_fanout = ut_make_ListConsumerGroups_fanout();
 
-                rko_partial =
-                    ut_make_ListConsumerGroups_partial(groups1, 2);
+                rko_partial = ut_make_ListConsumerGroups_partial(groups1, 2);
                 rd_kafka_ListConsumerGroups_response_merge(rko_fanout,
-                                                          rko_partial);
+                                                           rko_partial);
                 rd_kafka_op_destroy(rko_partial);
 
-                rko_partial =
-                    ut_make_ListConsumerGroups_partial(NULL, 0);
+                rko_partial = ut_make_ListConsumerGroups_partial(NULL, 0);
                 rd_kafka_ListConsumerGroups_response_merge(rko_fanout,
-                                                          rko_partial);
+                                                           rko_partial);
                 rd_kafka_op_destroy(rko_partial);
 
-                rko_partial =
-                    ut_make_ListConsumerGroups_partial(groups3, 2);
+                rko_partial = ut_make_ListConsumerGroups_partial(groups3, 2);
                 rd_kafka_ListConsumerGroups_response_merge(rko_fanout,
-                                                          rko_partial);
+                                                           rko_partial);
                 rd_kafka_op_destroy(rko_partial);
 
                 if (ut_verify_ListConsumerGroups_result(
--- a/src/rdkafka_admin.c
+++ b/src/rdkafka_admin.c
@@ -31,6 +31,7 @@
 #include "rdkafka_admin.h"
 #include "rdkafka_request.h"
 #include "rdkafka_aux.h"
+#include "rdunittest.h"
 
 #include <stdarg.h>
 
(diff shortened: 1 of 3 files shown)


src/rdkafka_admin.c
```diff
@@ -7581,6 +7582,41 @@ rd_kafka_ListConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
...
 }
 
+/**
+ * @brief Append groups from \p src to \p dst, skipping duplicates.
+ *
+ * Uses a hash map for O(1) lookups.  When the same group is reported
+ * by multiple brokers the first response wins; this is acceptable
+ * because group state is point-in-time and any single broker's view
+ * is equally valid.
+ */
```
pranavrth, line 7599: Consider replacing the ephemeral `rd_map_t` with a sorted list approach using `rd_list_find` (bsearch on sorted lists) and `rd_list_sort`. We keep the result list (`res->valid`) sorted after each broker response is processed and updates `res->valid`, so `rd_list_find` uses bsearch for O(log n) dedup lookups with zero extra allocations.

The map is rebuilt from scratch on every broker response — seeding all accumulated entries into a new map, checking new entries, then destroying the map. This means O(n) hash inserts + n `rd_map_elem_t` mallocs per broker just to reconstruct state that already exists in `dst`. With a sorted list, the sorted state persists naturally across broker responses. In the WarpStream use case where all agents return the same groups, after the first broker populates and sorts the list, every subsequent broker is pure bsearch lookups — all groups found, zero inserts, zero sorts, zero allocations.

This assumes that a single broker response won't contain duplicate group IDs, which is valid with Apache Kafka today — each group has exactly one coordinator, and `ListGroupsResponse` enumerates groups the broker coordinates. This assumption holds in both the map and sorted list approach — deduplication is only needed across broker responses, not within a single response.

src/rdkafka_admin.c
```diff
@@ -7581,6 +7582,41 @@ rd_kafka_ListConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
         return RD_KAFKA_RESP_ERR_NO_ERROR;
 }
 
+/**
+ * @brief Append groups from \p src to \p dst, skipping duplicates.
+ *
+ * Uses a hash map for O(1) lookups.  When the same group is reported
+ * by multiple brokers the first response wins; this is acceptable
```
pranavrth: "First response wins" is not useful here — broker responses arrive in arbitrary order depending on network latency. The Java client does the opposite (last response wins) and the outcome is identical.

src/rdkafka_admin.c
```diff
@@ -7620,11 +7656,8 @@ rd_kafka_ListConsumerGroups_response_merge(rd_kafka_op_t *rko_fanout,
                 new_valid_count  = rd_list_cnt(&newres->valid);
                 new_errors_count = rd_list_cnt(&newres->errors);
                 if (new_valid_count) {
-                        new_valid_list = &newres->valid;
-                        rd_list_grow(&res->valid, new_valid_count);
-                        rd_list_copy_to(
-                            &res->valid, new_valid_list,
-                            rd_kafka_ConsumerGroupListing_copy_opaque, NULL);
+                        rd_kafka_ConsumerGroupListing_list_append_dedup(
```
pranavrth, line 7677: `new_valid_list` is declared but no longer read after the `rd_list_copy_to` call was replaced. Will trigger a compiler warning.

src/rdkafka_admin.c
```diff
@@ -9809,3 +9842,291 @@ void rd_kafka_ElectLeaders(rd_kafka_t *rk,
...
+
+                if (ut_verify_ListConsumerGroups_result(
+                        rko_fanout, expected, 3, 1,
+                        "duplicates with interleaved error"))
+                        return 1;
+
+                rd_kafka_op_destroy(rko_fanout);
+        }
+
+        RD_UT_PASS();
```
pranavrth, line 10179: Add an empty broker response case (broker returns 0 groups) to cover the `new_valid_count == 0` branch.

pranavrth (changes requested):

src/rdkafka_admin.c
adizaimi: Good points about the map costs. Having dst a sorted list was an option I did consider. Just to run this by you, the idea you are proposing is to  build a separate list of unique new items first, then bulk-append to dst and re-sort dst:

static void
rd_kafka_ConsumerGroupListing_list_append_dedup(rd_list_t *dst,
                                                const rd_list_t *src) {
        const rd_kafka_ConsumerGroupListing_t *grp;
        int j;
        rd_list_t to_add;

        if (!rd_list_cnt(src))
                return;

        /* Ensure dst is sorted for bsearch lookups */
        if (rd_list_cnt(dst) && !(dst->rl_flags & RD_LIST_F_SORTED))
                rd_list_sort(dst, rd_kafka_ConsumerGroupListing_cmp);

        /* Collect new unique items without mutating dst */
        rd_list_init(&to_add, rd_list_cnt(src), NULL);
        RD_LIST_FOREACH(grp, src, j) {
                if (rd_list_cnt(dst) &&
                    rd_list_find(dst, grp,
                                rd_kafka_ConsumerGroupListing_cmp))
                        continue;
                rd_list_add(&to_add,
                            rd_kafka_ConsumerGroupListing_copy(grp));
        }

        /* Bulk-append and re-sort */
        if (rd_list_cnt(&to_add)) {
                rd_list_grow(dst, rd_list_cnt(&to_add));
                rd_list_copy_to(
                            &res->valid, to_add,
                            rd_kafka_ConsumerGroupListing_copy_opaque, NULL);
                rd_list_sort(dst, rd_kafka_ConsumerGroupListing_cmp);
        }

        rd_list_destroy(&to_add);
}
This keeps dst immutable during the lookup phase — all bsearches hit the sorted array. Then one bulk append + one re-sort at the end. The to_add list is just a pointer array (no deep copies to free on destroy since ownership transfers to dst), so the only allocation is the rd_list_init backing array.

pranavrth: The approach is what I was suggesting. There are few things to improve here though:

1. `to_add` should be a pointer with default value as `NULL`. `rd_list_t *to_add = NULL;`
2. Initialize this list lazily when the first element is not found in the `dst` list. In the common case for Warpstream, it will never be initialized so later operations are no-op
3. Its better to compare group id instead of whole group so that the solution is still valid in the rare duplicate scenario in AK as well. string comparator instead of `rd_kafka_ConsumerGroupListing_cmp`

adizaimi: Sounds good. For 3. we need a comparator that knows how to compare groups by their id:
static int rd_kafka_ConsumerGroupListing_cmp_group_id(const void *a,
                                                      const void *b) {
        const rd_kafka_ConsumerGroupListing_t *ga = a, *gb = b;
        return strcmp(ga->group_id, gb->group_id);
}

We **can** use strcmp since today group_id is the first member of the struct:
struct rd_kafka_ConsumerGroupListing_s {
        char *group_id; /**< Group id */
        /** Is it a simple consumer group? That means empty protocol_type. */
        rd_bool_t is_simple_consumer_group;
        rd_kafka_consumer_group_state_t state; /**< Consumer group state. */
        rd_kafka_consumer_group_type_t type;   /**< Consumer group type. */
};
but nothing stops future code changes to change the ordering so it's only safe to have rd_kafka_ConsumerGroupListing_cmp_group_id.

Maybe you were just saying that we need to compare just the group_id field, which the previously named rd_kafka_ConsumerGroupListing_cmp i had above does not clearly indicate.

src/rdkafka_admin.c
adizaimi: Removed, thanks.

src/rdkafka_admin.c
adizaimi: Removed that part, thanks.

src/rdkafka_admin.c
adizaimi: Done, added Test 5 for that.

src/rdkafka_admin.c
```diff
@@ -7581,6 +7582,59 @@ rd_kafka_ListConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
...
+        /* Collect pointers to src groups not already in dst.
+         * dst is not modified here so the sorted flag stays valid. */
+        RD_LIST_FOREACH(grp, src, j) {
+                if (rd_list_cnt(dst) &&
+                    rd_list_find(dst, grp,
+                                rd_kafka_ConsumerGroupListing_cmp_group_id))
+                        continue;
+                if (!to_add) {
+                        to_add = rd_alloca(sizeof(*to_add));
+                        rd_list_init(to_add, rd_list_cnt(src), NULL);
```
pranavrth: Better to call `rd_list_new` instead as used in different places in the code base.

pranavrth (changes requested):

src/rdkafka_admin.c
adizaimi: Done.

src/rdkafka_admin.c
pranavrth: Yes thats what I meant that.

src/rdkafka_admin.c
```diff
@@ -7581,6 +7582,57 @@ rd_kafka_ListConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
...
+                return;
+
+        /* Ensure dst is sorted for bsearch lookups. */
+        if (rd_list_cnt(dst) && !(dst->rl_flags & RD_LIST_F_SORTED))
+                rd_list_sort(dst, rd_kafka_ConsumerGroupListing_cmp_group_id);
+
+        /* Collect pointers to src groups not already in dst.
+         * dst is not modified here so the sorted flag stays valid. */
+        RD_LIST_FOREACH(grp, src, j) {
+                if (rd_list_cnt(dst) &&
```
pranavrth: Run the loop only if this condition is met

```
if (rd_list_cnt(dst))
```

src/rdkafka_admin.c
```diff
@@ -7581,6 +7582,57 @@ rd_kafka_ListConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
...
+                                                const rd_list_t *src) {
+        const rd_kafka_ConsumerGroupListing_t *grp;
+        int j;
+        rd_list_t *to_add = NULL;
+
+        if (!rd_list_cnt(src))
+                return;
+
+        /* Ensure dst is sorted for bsearch lookups. */
+        if (rd_list_cnt(dst) && !(dst->rl_flags & RD_LIST_F_SORTED))
```
pranavrth: I think we don't need to check the count. All the APIs already handles empty list part.

src/rdkafka_admin.c
pranavrth: Same count comment.

src/rdkafka_admin.c
```diff
@@ -7581,6 +7582,57 @@ rd_kafka_ListConsumerGroupsResponse_parse(rd_kafka_op_t *rko_req,
...
+        /* Collect pointers to src groups not already in dst.
+         * dst is not modified here so the sorted flag stays valid. */
+        RD_LIST_FOREACH(grp, src, j) {
+                if (rd_list_cnt(dst) &&
+                    rd_list_find(dst, grp,
+                                rd_kafka_ConsumerGroupListing_cmp_group_id))
+                        continue;
+                if (!to_add)
+                        to_add = rd_list_new(rd_list_cnt(src), NULL);
+                rd_list_add(to_add, (void *)grp);
```
pranavrth, line 7622: No need for (void *)

pranavrth (changes requested):
Style check is not passing. Can you please check and fix? Use `make style-check` or `make style-check-changed` to check what is the issue and `make style-fix` and `make style-fix-changed` to fix the issue. `changed` uses `git diff` to identify the changed part.

Once this is fixed, we can run integration tests.

src/rdkafka_admin.c
adizaimi: I think we should leave it because compiler gave me a warning without the cast:
```
warning: passing 'const rd_kafka_ConsumerGroupListing_t *' (aka 'const struct rd_kafka_ConsumerGroupListing_s *') to parameter of type 'void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
 7624 |                 rd_list_add(to_add, grp);
      |                                     ^~~
./rdlist.h:138:40: note: passing argument to parameter 'elem' here
```

src/rdkafka_admin.c
adizaimi: Sure, the rd_list_cnt(dst) checks were defensive guards, but both downstream functions handle empty lists safely so I'll remove them.

adizaimi:
will continue the review/merging process in internal PR https://github.com/confluentinc/librdkafka/pull/5417.

PR #5400, opened 2026-04-14
[KIP-932]: Test 0177 single broker fix

PratRanj07:
Fix transaction init timeout with single broker

- Add transaction.state.log.replication.factor=1 and min.isr=1 to Single Broker Tests in CI
- Change default broker count to 1 in interactive_broker_version.py and broker_version_tests.py

Transaction initialization was timing out after 120s because the default
replication factor (3) and min ISR (2) cannot be satisfied with a single broker.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -221,7 +221,7 @@ blocks:
             - (cd tests && python3 -m trivup.clusters.KafkaCluster --kraft
               --version ${KAFKA_VERSION} --cpversion ${CP_VERSION}
               --brokers 1
-              --conf '["share.coordinator.state.topic.replication.factor=1", "share.coordinator.state.topic.min.isr=1", "group.share.min.record.lock.duration.ms=1000"]'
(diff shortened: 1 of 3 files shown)


.semaphore/semaphore.yml
```diff
@@ -221,7 +221,7 @@ blocks:
             - (cd tests && python3 -m trivup.clusters.KafkaCluster --kraft
               --version ${KAFKA_VERSION} --cpversion ${CP_VERSION}
               --brokers 1
-              --conf '["share.coordinator.state.topic.replication.factor=1", "share.coordinator.state.topic.min.isr=1", "group.share.min.record.lock.duration.ms=1000"]'
+              --conf '["share.coordinator.state.topic.replication.factor=1", "share.coordinator.state.topic.min.isr=1", "group.share.min.record.lock.duration.ms=1000", "transaction.state.log.replication.factor=1", "transaction.state.log.min.isr=1"]'
```
pranavrth, line 224: This is needed to be done in other places as well. Please do those changes as well.

pranavrth (changes requested):

pranavrth (approved):
LGTM!.

PR #5384, opened 2026-04-02
[KIP-932] Implement close APIs

ojasvajain:
The PR implements the following Share Consumer APIs:
1. `rd_kafka_error_t *rd_kafka_share_consumer_close(rd_kafka_share_t *rkshare)`
2. `int rd_kafka_share_consumer_closed(rd_kafka_share_t *rkshare)`
3. `rd_kafka_error_t *rd_kafka_share_consumer_close_queue(rd_kafka_share_t *rkshare,
                                                      rd_kafka_queue_t *rkqu)`
                                                      
                                                    

Pending:

- [x]  Refactor test cases
- [x]  Base the PR against the commit_sync branch.
- [x] Address the TODO regarding finding brokers with active sessions based on discussion.
- [x] ~~Merge pending sync acks to pending async acks before sending the leave request~~
- [x] Add more test cases and check for memory leaks and flakiness
- [x] Verify if queue forwarding in close() is safe



Test Cases

  ```
  Real-broker tests (main_0178_share_consumer_close)                                                                                                                          
                                                                                                                                                                              
  ┌─────┬────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐  
  │  #  │              Test              │                                                           Description                                                           │  
  ├─────┼────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤  
  │ 1   │ test_close_with_acknowledge    │ Matrix of 12 sub-tests (3 commit modes × 4 topologies). C1 consumes + ACCEPT-acks N messages, optionally calls                  │
  │     │                                │ commitAsync/commitSync, then closes. C2 verifies acknowledged messages are not redelivered.                                     │
  ├─────┼────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤  
  │ 2   │ test_close_without_acknowledge │ Matrix of 4 topology sub-tests. C1 consumes without acking, then closes — close() must release unacked records back to the      │
  │     │                                │ share group. C2 verifies it receives every released message.                                                                    │  
  └─────┴────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
                                                                                                                                                                              
  Mock-broker tests (main_0178_share_consumer_close_local)

  ┌─────┬───────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
[...]

PR #5406, opened 2026-04-16
KIP-932: Implement destroy APIs

ojasvajain:
This PR implements the following APIs related to share consumers:

1. `void rd_kafka_share_destroy(rd_kafka_share_t *rkshare);`
2. `void rd_kafka_share_destroy_flags(rd_kafka_share_t *rkshare, int flags);`

Note, handling of _DESTROY* in acknowledgement callback will be covered in a separate PR

Pending:

- [x] Handing of _DESTROY* in commit_sync
- [x] Broker decomission test cases
- [x] More test cases related to destroy API
- [x] Fix "Build configuration check" job getting stuck in destroy
- [ ] Add missing test cases from [this](https://github.com/confluentinc/librdkafka/pull/5384#issuecomment-4368872695) list
- [ ] Add fatal error handling in close API
- [x] Evaluate `RD_KAFKA_RESP_ERR__DESTROY` check in `rd_kafka_share_fetch_reply_op`

Test Cases Added (so far):
```
  ┌─────┬──────────────────────────────────────────────────┬──────┬────────────────────────────────────────────────────────────────────────────────────────────────────────┐  
  │  #  │             Test (subtest function)              │ Type │                                              Description                                               │
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤  
  │     │                                                  │      │ Acks across commit_async + commit_sync with 5s ShareAck RTT delays, then destroys. Verifies destroy    │
  │ 1   │ test_destroy_with_cached_acks_and_delayed_broker │ mock │ completes within bounded time and that commit_sync returns REQUEST_TIMED_OUT for the still-pending     │  
  │     │                                                  │      │ partition.                                                                                             │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 2   │ test_broker_decommission_with_commit_sync        │ mock │ Decommissions a broker mid-commit_sync (background thread moves leader + removes broker from           │  
  │     │                                                  │      │ metadata). Asserts per-partition err = __DESTROY_BROKER for the orphaned partition.                    │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 3   │ test_broker_decommission_with_consume_batch      │ mock │ Decommissions a broker mid-consume_batch. Verifies the in-flight ShareFetch fails locally with         │  
  │     │                                                  │      │ __DESTROY_BROKER and the consumer keeps polling the surviving broker.                                  │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 4   │ test_broker_decommission_during_close            │ mock │ Decommissions a broker mid-close. Verifies close completes despite the in-flight leave-ShareAck being  │  
  │     │                                                  │      │ killed.                                                                                                │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 5   │ test_broker_decommission_with_commit_async       │ mock │ Decommissions a broker mid-commit_async. Verifies the async commit returns promptly with NULL error    │  
  │     │                                                  │      │ and the consumer survives.                                                                             │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 6   │ test_destroy_during_rebalance                    │ mock │ Drives the cgrp into wait-incr-unassign-to-complete (heartbeat blocked by injected 10s RTT), then      │  
  │     │                                                  │      │ destroys. NO_CONSUMER_CLOSE must return in <2s; full close must complete in <15s.                      │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 7   │ test_destroy_with_fatal_error                    │ mock │ Acks half, injects a 30s ShareAck delay, triggers fatal error, then destroys. Verifies the fatal-error │  
  │     │                                                  │      │  path (auto-promoted to NO_CONSUMER_CLOSE) returns in <2s regardless of caller flags.                  │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 8   │ do_test_destroy_with_explicit_ack                │ real │ Consumes a batch with explicit-ack mode, acks all or half (parameterised), then destroys without       │  
  │     │                                                  │      │ commit. Verifies destroy handles pending acks correctly.                                               │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 9   │ do_test_destroy_with_implicit_ack                │ real │ Consumes a batch with implicit-ack mode, never makes a follow-up poll (so implicit ack never fires),   │  
  │     │                                                  │      │ then destroys. Verifies destroy handles un-acked records gracefully.                                   │  
  ├─────┼──────────────────────────────────────────────────┼──────┼────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 10  │ do_test_destroy_with_subscribe_unsubscribe       │ real │ Creates a share consumer with various subscribe/unsubscribe combinations and destroys. Sanity check    │  
  │     │                                                  │      │ for state transitions through subscribe/unsubscribe.                                                   │  
  └─────┴──────────────────────────────────────────────────┴──────┴────────────────────────────────────────────────────────────────────────────────────────────────────────┘
                                                                                                                                                                              
  Notes:          
  - Each test runs with both destroy_flags=0 and RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE variants.
  - Tests 2–5 each also run with both explicit_ack=true/false variants (except _with_consume_batch which is implicit only).                                                   
  - Test 8 also runs with both ack_half=true/false variants.                                                               
  - Mock tests live in main_0179_share_consumer_destroy_local; real-broker tests live in main_0179_share_consumer_destroy.                                                    
```


.semaphore/semaphore.yml
```diff
@@ -222,7 +222,7 @@ blocks:
               --version ${KAFKA_VERSION} --cpversion ${CP_VERSION}
               --brokers 1
               --conf '["share.coordinator.state.topic.replication.factor=1", "share.coordinator.state.topic.min.isr=1", "group.share.min.record.lock.duration.ms=1000", "transaction.state.log.replication.factor=1", "transaction.state.log.min.isr=1"]'
-              --cmd 'TESTS_SKIP_BEFORE=0170 python3 run-test-batches.py')
+              --cmd 'TESTS_SKIP_BEFORE=0157 python3 run-test-batches.py')
```
pranavrth: Add TODO to start running 155 and 156 as well after leader migration.

packaging/tools/run-share-consumer-tests.sh
```diff
@@ -22,4 +22,4 @@ echo "arguments: $TEST_ARGS"
  --conf '["group.share.min.record.lock.duration.ms=1000", "transaction.state.log.replication.factor=1", "transaction.state.log.min.isr=1"]' \
  --version "$TEST_KAFKA_GIT_REF" \
  --cpversion "$TEST_CP_VERSION" \
- --cmd "TESTS_SKIP_BEFORE=0170 python run-test-batches.py $TEST_ARGS")
+ --cmd "TESTS_SKIP_BEFORE=0157 python run-test-batches.py $TEST_ARGS")
```
pranavrth, line 26: same

src/rdkafka.c
```diff
@@ -1201,6 +1202,18 @@ static void rd_kafka_destroy_app(rd_kafka_t *rk, int flags) {
...
 
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
+                /* Destroy inflight acks map to release
+                 * toppar references held by topic_partition objects in the map.
+                 * Otherwise rd_kafka_destroy_app() deadlocks: it joins broker
+                 * threads, which wait for refcnt <= 1, but the toppar holds a
+                 * broker ref via rktp_leader that is only released when the
+                 * toppar is destroyed, which requires refcnt 0, which requires
+                 * releasing the rktp ref held by the inflight_acks map entry.
+                 */
```
pranavrth, line 1214: Correct wordings.

src/rdkafka.c
```diff
@@ -3215,7 +3234,10 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
         rd_bool_t records_fetched = rko_orig->rko_u.share_fetch.records_fetched;
         rd_bool_t should_leave    = rko_orig->rko_u.share_fetch.should_leave;
 
-        rd_kafka_assert(rk, thrd_is_current(rk->rk_thread));
+        /* TODO KIP-932: This assertion has been commented out as the function
+         * can also be called in broker thread during when rko has __DESTROY set
+         * (client termination). Check if it is safe to keep it this way */
+        // rd_kafka_assert(rk, thrd_is_current(rk->rk_thread));
```
pranavrth: 1. Let's change this to `rd_kafka_dassert`
2. Only condition in which this can be called from different thread when main thread has come out of the loop. Let's add that || the thread is main check here.

src/rdkafka.c
pranavrth, line 3282: Just to remember - Let's add ack callback tests as well.

pranavrth, line 3288: Change these to be atomic.

src/rdkafka_cgrp.c
```diff
@@ -6196,6 +6198,14 @@ void rd_kafka_cgrp_terminate0(rd_kafka_cgrp_t *rkcg, rd_kafka_op_t *rko) {
                                                         rd_false, rd_true);
                 }
                 rd_kafka_rdunlock(rkcg->rkcg_rk);
+        } else if (RD_KAFKA_IS_SHARE_CONSUMER(rkcg->rkcg_rk) &&
+                   rko->rko_u.share_fetch_fanout.ack_batches) {
+                /* NO_CONSUMER_CLOSE: we skip dispatching the final ack
+                 * batches to brokers, but still take ownership of the
+                 * ack_batches list so the toppar refs held inside it
+                 * are released. */
+                rd_list_destroy(rko->rko_u.share_fetch_fanout.ack_batches);
```
pranavrth, line 6210: We might need to call ack callback here. This flow is not the trivial flow and might be needed only for .NET client and hence we can take that up later. For now this should be fine. Let's discuss internally as well.

src/rdkafka.c
```diff
@@ -3253,8 +3275,9 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
          * enqueue a session leave op on the replying broker thread and return
          */
         if (rkcg->rkcg_flags & RD_KAFKA_CGRP_F_TERMINATE) {
```
pranavrth, line 3307: Check in no_consumer_close flow if `rkcg` is available when this handler is being called from the broker thead.

src/rdkafka_broker.c
```diff
@@ -3761,7 +3768,29 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
...
+                if (rkb->rkb_pending_commit_sync.sync_ack_details) {
+                        rd_rkb_dbg(
+                            rkb, BROKER, "TERM",
+                            "Clearing %d pending commit sync ack batch(es)",
+                            rd_list_cnt(
+                                rkb->rkb_pending_commit_sync.sync_ack_details));
+                        rd_list_destroy(
+                            rkb->rkb_pending_commit_sync.sync_ack_details);
+                        rkb->rkb_pending_commit_sync.sync_ack_details = NULL;
+                }
```
pranavrth: We need to:

1. clear these in the main thread itself after we set termination in progress flag
2. call acknowledgement callbacks for these acks with appropriate errors.
3. decrement brokers_awaiting_result_cnt counter for commit sync part
4. call maybe complete

Check step 4 in `rd_kafka_share_fetch_reply_op`.

src/rdkafka_broker.c
```diff
@@ -3534,6 +3534,13 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                 TAILQ_REMOVE(&rkb->rkb_toppars, rktp, rktp_rkblink);
                 rkb->rkb_toppar_cnt--;
                 rd_kafka_broker_share_session_toppar_remove(rkb, rktp);
+                /* TODO KIP-932: Verify this while working on leader migration.
+                 * Free rktp_rkb_session_link before the new leader broker
+                 * reuses it via TAILQ_INSERT_TAIL on its own
+                 * toppars_in_session; otherwise old broker's list head is
+                 * left dangling and destroy_final asserts. */
+                rd_kafka_broker_session_remove_partition_from_toppars_in_session(
```
pranavrth, line 3542: This should be done from the `session_clear` functions and not from here.

src/rdkafka_broker.c
```diff
@@ -5132,11 +5109,15 @@ void rd_kafka_broker_destroy_final(rd_kafka_broker_t *rkb) {
         rd_assert(TAILQ_EMPTY(&rkb->rkb_waitresps.rkbq_bufs));
         rd_assert(TAILQ_EMPTY(&rkb->rkb_retrybufs.rkbq_bufs));
         rd_assert(TAILQ_EMPTY(&rkb->rkb_toppars));
+
```
pranavrth: remove extra line

src/rdkafka_cgrp.c
pranavrth: remove this variable.

src/rdkafka_fetcher.c
```diff
@@ -1694,7 +1694,7 @@ static void rd_kafka_broker_session_add_partition_to_toppars_in_session(
         rkb->rkb_share_fetch_session.toppars_in_session_cnt++;
 }
 
-static void rd_kafka_broker_session_remove_partition_from_toppars_in_session(
+void rd_kafka_broker_session_remove_partition_from_toppars_in_session(
```
pranavrth, line 1704: Might need to revert this.

src/rdkafka_fetcher.h
```diff
@@ -44,6 +44,9 @@ void rd_kafka_broker_share_fetch_session_leave(rd_kafka_broker_t *rkb,
                                                rd_kafka_op_t *rko_orig,
                                                rd_ts_t now);
 void rd_kafka_broker_share_fetch_session_clear(rd_kafka_broker_t *rkb);
+void rd_kafka_broker_session_remove_partition_from_toppars_in_session(
+    rd_kafka_broker_t *rkb,
+    rd_kafka_toppar_t *rktp);
```
pranavrth, line 49: We might remove this.

src/rdkafka_int.h
```diff
@@ -1331,7 +1330,6 @@ rd_kafka_share_consumer_closed_error(rd_kafka_share_t *rkshare);
  */
 rd_kafka_resp_err_t
 rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare);
-
```
pranavrth, line 1334: Revert.

tests/test.c
```diff
@@ -584,6 +586,8 @@ struct test tests[] = {
     _TEST(0177_share_consumer_transactions, 0, TEST_BRKVER(0, 4, 0, 0)),
     _TEST(0178_share_consumer_close, 0, TEST_BRKVER(0, 4, 0, 0)),
     _TEST(0178_share_consumer_close_local, TEST_F_LOCAL),
+    _TEST(0179_share_consumer_destroy, 0, TEST_BRKVER(0, 4, 0, 0)),
```
pranavrth: Try changing this to 4.2.0 for other share consumer tests as well.

tests/logs.ci
pranavrth, line 1: Remove `logs.ci`

pranavrth (changes requested):
Added comments for the implementation part. I will go through the tests in another pass.

src/rdkafka.c
ojasvajain: Will address this in a separate PR. We can keep this comment open.

src/rdkafka_cgrp.c
ojasvajain: Sure. Adding a TODO for now.

src/rdkafka.c
ojasvajain: I checked this. `rkcg` is destroyed in `rd_kafka_cgrp_destroy_final` which is called by `rd_kafka_destroy_final` (much later). In case of NO_CONSUMER_CLOSE, it is possible that `rd_kafka_cgrp_terminated` would have been called before we reach here but `rd_kafka_cgrp_terminated` does not clear the rkcg_flags.

src/rdkafka_broker.c
ojasvajain: I checked this, I don't think we would be able to move it to session_clear as leader change might not always trigger broker decommission. There is another reason related to links in toppars_in_session list. Will discuss it with you.

tests/0173-share_consumer_commit_async.c
```diff
@@ -1590,9 +1590,9 @@ static void do_test_commit_async_callback(void) {
         rd_kafka_message_t *rkmessages[CONSUME_ARRAY];
         size_t rcvd;
         size_t j;
-        int consumed = 0;
-        int attempts = 0;
-        test_ack_cb_state_t state;
+        int consumed              = 0;
+        int attempts              = 0;
+        test_ack_cb_state_t state = {0};
```
ojasvajain, line 1595: This issue was causing the tests fail on the feature branch. It should be fixed now.

tests/0176-share_consumer_commit_sync.c
```diff
@@ -1991,9 +1991,9 @@ static void do_test_commit_sync_callback(void) {
         rd_kafka_message_t *rkmessages[CONSUME_ARRAY];
         size_t rcvd;
         size_t j;
-        int consumed = 0;
-        int attempts = 0;
-        ack_cb_state_t state;
+        int consumed         = 0;
+        int attempts         = 0;
+        ack_cb_state_t state = {0};
```
ojasvajain, line 1996: Same as above

tests/0172-share_consumer_acknowledge.c
```diff
@@ -943,7 +943,7 @@ static void test_change_ack_type_before_commit(void) {
         attempts = 30;
         while ((rcvd < 4 || state.callback_cnt < 1) && attempts-- > 0) {
                 size_t batch_rcvd = 0;
-                err = rd_kafka_share_consume_batch(rkshare, 2000, batch,
+                err = rd_kafka_share_consume_batch(rkshare, 2000, batch + rcvd,
```
ojasvajain, line 946: If msgs don't get redelivered msgs in one consume_batch call, some msgs might leak. Observed this issue in one of the runs.

src/rdkafka_share_acknowledgement.c
```diff
@@ -983,3 +983,59 @@ void rd_kafka_share_dispatch_ack_callbacks(rd_kafka_t *rk,
                                                         ack_batch->rktpar->err);
         }
 }
+
+void rd_kafka_share_acks_clear_during_broker_decommission(
```
pranavrth, line 1082: Add test case for this flow.

src/rdkafka_share_acknowledgement.c
```diff
@@ -983,3 +983,59 @@ void rd_kafka_share_dispatch_ack_callbacks(rd_kafka_t *rk,
...
+
+                rd_list_destroy(rkb->rkb_share_async_ack_details);
+                rkb->rkb_share_async_ack_details = NULL;
+        }
+
+        /* Sync acks: stamp SHARE_SESSION_NOT_FOUND on each batch and
+         * apply the result to the in-flight commit_sync request (copies
+         * err into the cgrp's result list, decrements awaiting count,
+         * completes the sync if this was the last broker outstanding). */
+        if (rkb->rkb_pending_commit_sync.sync_ack_details) {
```
pranavrth, line 1114: We need to dispatch callbacks as well.

src/rdkafka_share_acknowledgement.c
pranavrth, line 1082: add broker id as well in the logs.

src/rdkafka.c
```diff
@@ -3284,35 +3312,24 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
                  * normal helper path overwrites _IN_PROGRESS first
                  * (buf callback init -> INVALID_RECORD_STATE -> per-
                  * partition err or top-level err). */
-                if (ack_details) {
+                if (ack_details && rko_orig->rko_err) {
```
pranavrth: Move it to before dispatch ack_details (step 1) part.

src/rdkafka.c
```diff
@@ -3284,35 +3312,24 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
...
-                                        dst->err = batch->rktpar->err;
                         }
                 }
 
-                rkcg->rkcg_commit_sync_request.brokers_awaiting_result_cnt--;
-
+                rd_kafka_share_commit_sync_apply_result(rk, rkcg, ack_details);
                 rd_kafka_dbg(
                     rk, CGRP, "SHARE",
                     "Commit sync reply from broker %s: %s, "
```
pranavrth, line 3323: We can change wording for this.

src/rdkafka.c
```diff
@@ -3887,6 +3905,40 @@ static void rd_kafka_share_commit_sync_maybe_complete(rd_kafka_t *rk,
...
+                            rd_kafka_topic_partition_list_find(
+                                rkcg->rkcg_commit_sync_request.results,
+                                batch->rktpar->topic, batch->rktpar->partition);
+                        if (dst)
+                                dst->err = batch->rktpar->err;
+                }
+        }
+
+        rkcg->rkcg_commit_sync_request.brokers_awaiting_result_cnt--;
+        rd_kafka_share_commit_sync_maybe_complete(rk, rkcg);
```
pranavrth: Add log inside maybe_complete

src/rdkafka_broker.c
```diff
@@ -6541,26 +6503,16 @@ void rd_kafka_broker_decommission(rd_kafka_t *rk,
...
-                 *               might be called before reply for the op
-                 *               is processed. Check properly.
-                 * TODO KIP-932: Surface SHARE_SESSION_NOT_FOUND on the
-                 *               pending acks via the acknowledgement
-                 *               callback before discarding them. */
-                if (rkb->rkb_share_async_ack_details) {
-                        rd_list_destroy(rkb->rkb_share_async_ack_details);
-                        rkb->rkb_share_async_ack_details = NULL;
-                }
+                rd_kafka_share_acks_clear_during_broker_decommission(rk, rkb);
```
pranavrth: We are clearing states here. Make sure that we are not filling these fields after this step.
Add termination check in segregate_by_leader flow.

src/rdkafka.c
```diff
@@ -3253,8 +3280,9 @@ rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
          * enqueue a session leave op on the replying broker thread and return
          */
         if (rkcg->rkcg_flags & RD_KAFKA_CGRP_F_TERMINATE) {
```
pranavrth, line 3307: Check if we need to check broker termination here as well.

src/rdkafka.c
```diff
@@ -1393,6 +1401,18 @@ static void rd_kafka_destroy_internal(rd_kafka_t *rk) {
...
+                 * have been populated on the internal broker during a
+                 * leader migration (PARTITION_JOIN runs on whichever
+                 * broker the toppar is transiently delegated to,
+                 * including :0/internal). Once the leader-migration
+                 * path is fixed to skip non-LEARNED brokers, this
+                 * SESSION_CLEAR enqueue can be removed. */
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rk))
+                        rd_kafka_q_enq(
+                            rk->rk_internal_rkb->rkb_ops,
+                            rd_kafka_op_new(RD_KAFKA_OP_SHARE_SESSION_CLEAR));
```
pranavrth, line 1414: Check if we are not adding ack sync and async details for internal broker.

src/rdkafka_broker.c
```diff
@@ -6541,26 +6503,16 @@ void rd_kafka_broker_decommission(rd_kafka_t *rk,
         if (rd_atomic32_get(&rkb->termination_in_progress) > 0)
                 return;
 
+        rd_atomic32_add(&rkb->termination_in_progress, 1);
+
         if (RD_KAFKA_IS_SHARE_CONSUMER(rk) &&
             rkb->rkb_source == RD_KAFKA_LEARNED) {
```
pranavrth: Let's verify this is correct.

src/rdkafka_int.h
```diff
@@ -1337,4 +1337,10 @@ void rd_kafka_share_enqueue_fetch_op(rd_kafka_t *rk,
                                      rd_bool_t should_fetch,
                                      rd_bool_t should_leave);
 
+void rd_kafka_share_commit_sync_apply_result(rd_kafka_t *rk,
```
pranavrth: Check if this should be moved to any other file?

tests/0179-share_consumer_destroy.c
```diff
@@ -0,0 +1,2203 @@
...
+        int offset_cnt;
+        rd_kafka_resp_err_t err;
+} ack_receipt_t;
+
+typedef struct {
+        ack_receipt_t *receipts;
+        int receipt_cnt;
+        int receipt_capacity;
+        int callback_invocations;
+        mtx_t lock;
```
pranavrth: Check if it is needed.

tests/0179-share_consumer_destroy.c
```diff
@@ -0,0 +1,2203 @@
...
+                        if (explicit_ack) {
+                                size_t k;
+                                for (k = rcvd - batch_rcvd; k < rcvd; k++)
+                                        rd_kafka_share_acknowledge(
+                                            rkshare, rkmessages[k]);
+                        }
+                }
+                attempts++;
+        }
+        TEST_ASSERT(rcvd >= TEST_MSGS,
```
pranavrth: ==

tests/0179-share_consumer_destroy.c
```diff
@@ -0,0 +1,2203 @@
...
+         *   broker before we stage the decommission).
+         *
+         *   Second consume_batch: the piggybacked ShareAck for the
+         *   target partition's ACQUIRED batch parks on the target
+         *   broker and gets stamped __DESTROY_BROKER when the
+         *   decommission fires.
+         *
+         *   surviving_partition: TEST_MSGS/2 offsets, NO_ERROR
+         *   target_partition:    TEST_MSGS/2 offsets, __DESTROY_BROKER */
+        {
```
pranavrth, line 1018: Fix this.

pranavrth (changes requested):

src/rdkafka.c
pranavrth: Add test as well for this.

src/rdkafka_broker.c
pranavrth: Let's also verify

- rebalance flow
- session management flow

src/rdkafka_share_acknowledgement.c
ojasvajain: I am using `rd_rkb_dbg` which already prefixes the broker name (which includes the broker id) automatically

src/rdkafka.c
ojasvajain: Moved the log to above `apply_result` function.

src/rdkafka.c
ojasvajain: Yes, we should add it here. In addition to this, I think we also need a way to decrement the leave count during broker_decommission. Will discuss this with you.

src/rdkafka.c
ojasvajain: Checked this. Don't think we are.

` rktp->rktp_leader` is always upated to a non-zero broker ID (internal broker's ID is -1). Also while segregating acks, we use rktp_leader, so async and sync details should not cached in internal broker.

src/rdkafka.c
ojasvajain: Added in 0179. Will be adding in 0178 in a separate PR.

src/rdkafka.c
ojasvajain: Discussed this. Should not be required as broker thread will reply immediately with error if it is terminating which should decrement the count.

src/rdkafka_broker.c
ojasvajain: Checked this. The other broker types can not have cached acks as they can never be leaders to a rktp.

src/rdkafka_broker.c
ojasvajain: Checked all places where async cache in broker is populated. Have added the terminating check in segregate_by_leader function. Acks also get moved in `rd_kafka_share_commit_sync_timeout_cb` but I checked, it is getting handled implicitly.

src/rdkafka.c
ojasvajain: Will add in a separate PR

src/rdkafka_share_acknowledgement.c
ojasvajain: Will add in a separate PR

src/rdkafka_int.h
ojasvajain: Moved all commit sync related functions to share_acknowledgement.c and .h files.

src/rdkafka_share_acknowledgement.c
```diff
@@ -1035,6 +1131,9 @@ void rd_kafka_share_acks_clear_during_broker_decommission(
                 rd_kafka_share_commit_sync_apply_result(
                     rk, rkcg, rkb->rkb_pending_commit_sync.sync_ack_details);
 
+                rd_kafka_share_dispatch_ack_callbacks(
```
pranavrth, line 1134: Add tests for this.

tests/0179-share_consumer_destroy.c
```diff
@@ -1917,29 +1904,27 @@ static void do_test_destroy_with_explicit_ack(int destroy_flags,
          *   destroy_flags=0x8 (NO_CONSUMER_CLOSE): destroy short-
          *   circuits, so second-batch acks are dropped. Only
          *   first_batch_part appears with TEST_MSGS/2. */
-        {
-                rd_bool_t no_close =
-                    (destroy_flags & RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE) != 0;
-                const char *label = no_close ? "explicit-ack NO_CONSUMER_CLOSE"
-                                             : "explicit-ack full close";
+        rd_bool_t no_close =
```
pranavrth, line 1907: Move this to start of the function.

tests/test.c
```diff
@@ -571,19 +573,21 @@ struct test tests[] = {
     _TEST(0153_memberid, TEST_F_LOCAL),
     _TEST(0155_share_group_heartbeat_mock, TEST_F_LOCAL),
     _TEST(0156_share_consumer_fetch_mock, TEST_F_MANUAL),
-    _TEST(0157_share_consumer_ack_mock, TEST_F_MANUAL),
```
pranavrth, line 574: Update timeout for mock tests.

pranavrth (approved):
Provided minor nits: please address them in the next PR.

LGTM! Thanks @ojasvajain

PR #5410, opened 2026-04-16
 [KIP-932]: Add comprehensive MessageSet error handling

PratRanj07:
### Add comprehensive MessageSet error handling and added unit tests
Add unit tests for share consumer MessageSet error handling covering:
- CRC errors (REJECT ack_type)
- Decompression errors (RELEASE ack_type)
- Unsupported MagicByte errors (REJECT ack_type)
- Mixed scenarios: success/error state transitions
- Multiple consecutive errors and successes

### DIFF
--- a/src/rdkafka_fetcher.c
+++ b/src/rdkafka_fetcher.c
@@ -932,8 +932,16 @@ void rd_kafka_share_filter_acquired_records_and_update_ack_type(
                                     delivery_count;
                         } else if (rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rkm = &rko->rko_u.err.rkm;
-                                rkm->rkm_u.consumer.ack_type =
-                                    RD_KAFKA_SHARE_INTERNAL_ACK_REJECT;
+                                /* Set ack_type to REJECT only if not already set
+                                 * by rd_kafka_share_consumer_err_range() (which
+                                 * sets REJECT for CRC/unsupported errors, RELEASE
+                                 * for decompression errors). */
+                                if (rkm->rkm_u.consumer.ack_type ==
+                                        RD_KAFKA_SHARE_INTERNAL_ACK_GAP ||
+                                    rkm->rkm_u.consumer.ack_type ==
+                                        RD_KAFKA_SHARE_INTERNAL_ACK_ACQUIRED)
+                                        rkm->rkm_u.consumer.ack_type =
+                                            RD_KAFKA_SHARE_INTERNAL_ACK_REJECT;
                         }
 
                         /* Add to filtered messages list */
@@ -1100,11 +1108,17 @@ rd_kafka_share_build_response_rko(rd_kafka_broker_t *rkb,
                                         .message_rkos,
                                     msg_rko);
                                 msg_cnt++;
-                        } else {
-                                /* RD_KAFKA_OP_CONSUMER_ERR: forward to
-                                 * consumer group queue */
-                                rd_kafka_q_enq(rkb->rkb_rk->rk_cgrp->rkcg_q,
-                                               msg_rko);
+                        } else if (msg_rko->rko_type ==
+                                   RD_KAFKA_OP_CONSUMER_ERR) {
+                                /* Add error ops to message list along with
+                                 * fetch ops. Error ops have ack_type already set
+                                 * and will be delivered to application in the
+                                 * same message stream. */
+                                rd_list_add(
+                                    response_rko->rko_u.share_fetch_response
+                                        .message_rkos,
+                                    msg_rko);
+                                msg_cnt++;
                         }
                 }
 
--- a/src/rdkafka_fetcher.c
+++ b/src/rdkafka_fetcher.c
@@ -1102,18 +1102,8 @@ rd_kafka_share_build_response_rko(rd_kafka_broker_t *rkb,
                         entry->types[offset - entry->start_offset] =
                             rd_kafka_share_ack_type_from_msg_op(msg_rko);
 
-                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH) {
-                                rd_list_add(
-                                    response_rko->rko_u.share_fetch_response
-                                        .message_rkos,
-                                    msg_rko);
-                                msg_cnt++;
-                        } else if (msg_rko->rko_type ==
-                                   RD_KAFKA_OP_CONSUMER_ERR) {
-                                /* Add error ops to message list along with
-                                 * fetch ops. Error ops have ack_type already
-                                 * set and will be delivered to application in
-                                 * the same message stream. */
+                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH ||
+                            msg_rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rd_list_add(
                                     response_rko->rko_u.share_fetch_response
                                         .message_rkos,
--- a/src/rdkafka_fetcher.c
+++ b/src/rdkafka_fetcher.c
@@ -932,10 +932,10 @@ void rd_kafka_share_filter_acquired_records_and_update_ack_type(
                                     delivery_count;
                         } else if (rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rkm = &rko->rko_u.err.rkm;
-                                /* Set ack_type to REJECT only if not already set
-                                 * by rd_kafka_share_consumer_err_range() (which
-                                 * sets REJECT for CRC/unsupported errors, RELEASE
-                                 * for decompression errors). */
+                                /* Set ack_type to REJECT only if not already
+                                 * set by rd_kafka_share_consumer_err_range()
+                                 * (which sets REJECT for CRC/unsupported
+                                 * errors, RELEASE for decompression errors). */
                                 if (rkm->rkm_u.consumer.ack_type ==
                                         RD_KAFKA_SHARE_INTERNAL_ACK_GAP ||
                                     rkm->rkm_u.consumer.ack_type ==
@@ -1111,9 +1111,9 @@ rd_kafka_share_build_response_rko(rd_kafka_broker_t *rkb,
                         } else if (msg_rko->rko_type ==
                                    RD_KAFKA_OP_CONSUMER_ERR) {
                                 /* Add error ops to message list along with
-                                 * fetch ops. Error ops have ack_type already set
-                                 * and will be delivered to application in the
-                                 * same message stream. */
+                                 * fetch ops. Error ops have ack_type already
+                                 * set and will be delivered to application in
+                                 * the same message stream. */
                                 rd_list_add(
                                     response_rko->rko_u.share_fetch_response
                                         .message_rkos,
--- a/src/rdkafka_fetcher.c
+++ b/src/rdkafka_fetcher.c
@@ -933,7 +933,7 @@ void rd_kafka_share_filter_acquired_records_and_update_ack_type(
                         } else if (rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rkm = &rko->rko_u.err.rkm;
                                 /* Set ack_type to REJECT only if not already
-                                 * set by rd_kafka_share_consumer_err_range()
+                                 * set by rd_kafka_share_message_set_err_ops()
                                  * (which sets REJECT for CRC/unsupported
                                  * errors, RELEASE for decompression errors). */
                                 if (rkm->rkm_u.consumer.ack_type ==
@@ -941,7 +941,7 @@ void rd_kafka_share_filter_acquired_records_and_update_ack_type(
                                     rkm->rkm_u.consumer.ack_type ==
                                         RD_KAFKA_SHARE_INTERNAL_ACK_ACQUIRED)
                                         rkm->rkm_u.consumer.ack_type =
-                                            RD_KAFKA_SHARE_INTERNAL_ACK_REJECT;
+                                            RD_KAFKA_SHARE_INTERNAL_ACK_RELEASE;
                         }
 
                         /* Add to filtered messages list */
@@ -1102,6 +1102,10 @@ rd_kafka_share_build_response_rko(rd_kafka_broker_t *rkb,
                         entry->types[offset - entry->start_offset] =
                             rd_kafka_share_ack_type_from_msg_op(msg_rko);
 
+                        /* Forward the rko to the application only for
+                         * messages and per-message errors; GAP entries
+                         * are tracked in the per-offset ack-type table
+                         * above but not surfaced as rkmessages. */
                         if (msg_rko->rko_type == RD_KAFKA_OP_FETCH ||
                             msg_rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rd_list_add(
--- a/src/rdkafka_fetcher.c
+++ b/src/rdkafka_fetcher.c
@@ -1465,6 +1465,19 @@ static rd_kafka_resp_err_t rd_kafka_share_fetch_reply_handle_partition(
                  * parse errors (which are partition-specific) */
         }
 
+        /**
+         * Only inner-record parsing errors (RD_KAFKA_RESP_ERR__BAD_MSG /
+         * RD_KAFKA_RESP_ERR__UNDERFLOW) propagate out of
+         * rd_kafka_share_msgset_parse for share consumers — per-batch
+         * codec / CRC / unsupported-MagicByte failures are handled
+         * inline by emitting per-offset RELEASE/REJECT ops and are
+         * swallowed by the v2 reader (msgset_reader.c). When a true
+         * parse error bubbles up the wire data is corrupt and we can no
+         * longer trust the buffer position or any per-batch metadata.
+         * Stop processing the rest of this ShareFetch response and
+         * propagate the error to the caller, which aborts the whole
+         * response handling.
+         */
         if (err)
                 goto done;
 
--- a/src/rdkafka_fetcher.c
+++ b/src/rdkafka_fetcher.c
@@ -932,8 +932,16 @@ void rd_kafka_share_filter_acquired_records_and_update_ack_type(
                                     delivery_count;
                         } else if (rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rkm = &rko->rko_u.err.rkm;
-                                rkm->rkm_u.consumer.ack_type =
-                                    RD_KAFKA_SHARE_INTERNAL_ACK_REJECT;
+                                /* Set ack_type to REJECT only if not already
+                                 * set by rd_kafka_share_consumer_err_range()
+                                 * (which sets REJECT for CRC/unsupported
+                                 * errors, RELEASE for decompression errors). */
+                                if (rkm->rkm_u.consumer.ack_type ==
+                                        RD_KAFKA_SHARE_INTERNAL_ACK_GAP ||
+                                    rkm->rkm_u.consumer.ack_type ==
+                                        RD_KAFKA_SHARE_INTERNAL_ACK_ACQUIRED)
+                                        rkm->rkm_u.consumer.ack_type =
+                                            RD_KAFKA_SHARE_INTERNAL_ACK_REJECT;
                         }
 
                         /* Add to filtered messages list */
@@ -1094,17 +1102,13 @@ rd_kafka_share_build_response_rko(rd_kafka_broker_t *rkb,
                         entry->types[offset - entry->start_offset] =
                             rd_kafka_share_ack_type_from_msg_op(msg_rko);
 
-                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH) {
+                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH ||
+                            msg_rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rd_list_add(
                                     response_rko->rko_u.share_fetch_response
                                         .message_rkos,
                                     msg_rko);
                                 msg_cnt++;
-                        } else {
-                                /* RD_KAFKA_OP_CONSUMER_ERR: forward to
-                                 * consumer group queue */
-                                rd_kafka_q_enq(rkb->rkb_rk->rk_cgrp->rkcg_q,
-                                               msg_rko);
                         }
                 }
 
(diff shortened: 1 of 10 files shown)


src/rdkafka_fetcher.c
```diff
@@ -1094,17 +1102,13 @@ rd_kafka_share_build_response_rko(rd_kafka_broker_t *rkb,
                         entry->types[offset - entry->start_offset] =
                             rd_kafka_share_ack_type_from_msg_op(msg_rko);
 
-                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH) {
+                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH ||
+                            msg_rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
```
pranavrth, line 1113: Let's write a comment explaining what are we doing here.

src/rdkafka_fetcher.c
```diff
@@ -932,8 +932,16 @@ void rd_kafka_share_filter_acquired_records_and_update_ack_type(
...
+                                /* Set ack_type to REJECT only if not already
+                                 * set by rd_kafka_share_consumer_err_range()
+                                 * (which sets REJECT for CRC/unsupported
+                                 * errors, RELEASE for decompression errors). */
+                                if (rkm->rkm_u.consumer.ack_type ==
+                                        RD_KAFKA_SHARE_INTERNAL_ACK_GAP ||
+                                    rkm->rkm_u.consumer.ack_type ==
+                                        RD_KAFKA_SHARE_INTERNAL_ACK_ACQUIRED)
+                                        rkm->rkm_u.consumer.ack_type =
+                                            RD_KAFKA_SHARE_INTERNAL_ACK_REJECT;
```
pranavrth: Check if Java sets release or reject in this case. I feel it should be release as any other consumer might be able to get these records but lets check Java part.

src/rdkafka_msgset_reader.c
```diff
@@ -517,12 +517,35 @@ rd_kafka_msgset_reader_decompress(rd_kafka_msgset_reader_t *msetr,
 err:
         /* Enqueue error messsage:
          * Create op and push on temporary queue. */
-        rd_kafka_consumer_err(
-            &msetr->msetr_rkq, msetr->msetr_broker_id, err,
-            msetr->msetr_tver->version, NULL, rktp, Offset,
-            "Decompression (codec 0x%x) of message at %" PRIu64 " of %" PRIusz
-            " bytes failed: %s",
-            codec, Offset, compressed_size, rd_kafka_err2str(err));
+        if (RD_KAFKA_IS_SHARE_CONSUMER(msetr->msetr_rkb->rkb_rk) &&
```
pranavrth: First condition should be regarding normal consumer. We don't want to affect normal consumer flow.

src/rdkafka_msgset_reader.c
```diff
@@ -1088,19 +1111,49 @@ rd_kafka_msgset_reader_v2(rd_kafka_msgset_reader_t *msetr) {
...
-                            msetr->msetr_tver->version, NULL, rktp,
-                            hdr.BaseOffset,
-                            "MessageSet at offset %" PRId64 " (%" PRId32
-                            " bytes) "
-                            "failed CRC32C check "
-                            "(original 0x%" PRIx32
-                            " != "
-                            "calculated 0x%" PRIx32 ")",
-                            hdr.BaseOffset, hdr.Length, hdr.Crc, calc_crc);
-                        rd_kafka_buf_skip_to(rkbuf, crc_len);
```
pranavrth, line 1103: Keep this condition as it is for the normal consumer. including `crc_len` part.

src/rdkafka_msgset_reader.c
```diff
@@ -1088,19 +1111,49 @@ rd_kafka_msgset_reader_v2(rd_kafka_msgset_reader_t *msetr) {
...
+                                    " bytes) "
+                                    "failed CRC32C check "
+                                    "(original 0x%" PRIx32
+                                    " != "
+                                    "calculated 0x%" PRIx32 ")",
+                                    hdr.BaseOffset, hdr.Length, hdr.Crc,
+                                    calc_crc);
+                        }
+                        /* Skip to end of MessageSet */
+                        rd_kafka_buf_skip_to(rkbuf, len_start + hdr.Length);
```
pranavrth: Check if this is correct as we are reading 2 fields in share consumer path. I think it is correct but verify once.

src/rdkafka_msgset_reader.c
```diff
@@ -1155,8 +1208,20 @@ rd_kafka_msgset_reader_v2(rd_kafka_msgset_reader_t *msetr) {
...
-                        goto err;
+                if (err) {
+                        /* For share consumer: decompression failed, error ops
+                         * already created. Buffer position already advanced
+                         * past compressed payload. Continue parsing next
+                         * MessageSet. For regular consumer: stop parsing. */
+                        if (RD_KAFKA_IS_SHARE_CONSUMER(
+                                msetr->msetr_rkb->rkb_rk)) {
+                                rd_atomic64_add(&msetr->msetr_rkb->rkb_c.rx_err,
+                                                1);
```
pranavrth: What is the significance of this field? Is it correctly incremented by 1 as we are sending individual record level errors in share consumer.

src/rdkafka_msgset_reader.c
```diff
@@ -1243,26 +1308,66 @@ rd_kafka_msgset_reader_peek_msg_version(rd_kafka_msgset_reader_t *msetr,
                            (int)*MagicBytep, Offset, read_offset,
                            rd_slice_size(&rkbuf->rkbuf_reader));
 
-                if (Offset >=
-                        msetr->msetr_rktp->rktp_offsets.fetch_pos.offset &&
-                    !RD_KAFKA_IS_SHARE_CONSUMER(msetr->msetr_rkb->rkb_rk)) {
-                        rd_kafka_consumer_err(
+                rd_kafka_buf_read_i32(rkbuf, &Length);
+
+                if (RD_KAFKA_IS_SHARE_CONSUMER(msetr->msetr_rkb->rkb_rk)) {
```
pranavrth: Let's keep normal consumer path as it is and it should be first condition.

src/rdkafka_op.c
```diff
@@ -684,6 +684,72 @@ void rd_kafka_consumer_err(rd_kafka_q_t *rkq,
...
+ * @param end_offset Last offset in the error range (inclusive).
+ * @param ack_type Acknowledgement type for share consumer
+ *                 (REJECT for permanent errors, RELEASE for retryable).
+ *
+ * Creates one error op per offset in the range [start_offset, end_offset].
+ * Each op has ack_type pre-set for share consumer acknowledgement tracking.
+ *
+ * @sa rd_kafka_consumer_err()
+ */
+void rd_kafka_share_consumer_err_range(
```
pranavrth:
```suggestion
void rd_kafka_share_message_set_err_ops(
```

pranavrth (changes requested):
High level comments. Still going through the flow.

src/rdkafka_fetcher.c
PratRanj07: There isn't any such case in Java as they handle it in a different way. We have taken care of the CRC and unsupported magic byte errors and decompression errors, this particular condition is just for gatekeeping purpose but we shouldn't hit this. Changed it to release as a gatekeeping purpose as for the decompression errors, in case it is a v0/v1 message we have to have a single op and that op wont have the ack type set.

src/rdkafka_msgset_reader.c
PratRanj07: Yes, it's correct. The skip is absolute, not relative, so the extra reads in the share-consumer path don't affect it.

src/rdkafka_msgset_reader.c
PratRanj07: I think it is correct because the error that we recieved is batch level, we recieved one bad batch, I dont think we should care about how many records are inside there

src/rdkafka_msgset_reader.c
```diff
@@ -517,12 +517,33 @@ rd_kafka_msgset_reader_decompress(rd_kafka_msgset_reader_t *msetr,
...
         /* Enqueue error messsage:
          * Create op and push on temporary queue. */
-        rd_kafka_consumer_err(
-            &msetr->msetr_rkq, msetr->msetr_broker_id, err,
-            msetr->msetr_tver->version, NULL, rktp, Offset,
-            "Decompression (codec 0x%x) of message at %" PRIu64 " of %" PRIusz
-            " bytes failed: %s",
-            codec, Offset, compressed_size, rd_kafka_err2str(err));
+        if (!RD_KAFKA_IS_SHARE_CONSUMER(msetr->msetr_rkb->rkb_rk) ||
+            msetr->msetr_v2_hdr == NULL) {
```
pranavrth: Why do we need this condition? When can this case happen?

src/rdkafka_msgset_reader.c
```diff
@@ -1305,8 +1422,15 @@ rd_kafka_msgset_reader(rd_kafka_msgset_reader_t *msetr) {
...
-                         * MessageSet will be skipped. */
+                        /* For share consumer: continue on unsupported
+                         * MsgVersions, the MessageSet will be skipped. Reset
+                         * error so loop continues parsing remaining
+                         * MessageSets. For regular consumer: stop parsing (old
+                         * behavior). */
+                        if (RD_KAFKA_IS_SHARE_CONSUMER(
+                                msetr->msetr_rkb->rkb_rk)) {
+                                err = RD_KAFKA_RESP_ERR_NO_ERROR;
+                        }
```
pranavrth, line 1475: Remove this handling.

src/rdkafka_msgset_reader.c
```diff
@@ -517,12 +517,33 @@ rd_kafka_msgset_reader_decompress(rd_kafka_msgset_reader_t *msetr,
 err:
```
pranavrth, line 517: Check the followings as well:

- Message size too large path
- fetch_pos.offset handling
- buffer underflow error / err_parse path
- Check this error as well - "err = rd_kafka_share_msgset_parse"

src/rdkafka_msgset_reader.c
```diff
@@ -1243,27 +1311,76 @@ rd_kafka_msgset_reader_peek_msg_version(rd_kafka_msgset_reader_t *msetr,
...
+                        rd_kafka_buf_read_i8(rkbuf, &MagicByteActual);
+                        rd_kafka_buf_read_i32(rkbuf, &Crc);
+                        rd_kafka_buf_read_i16(rkbuf, &Attributes);
+                        rd_kafka_buf_read_i32(rkbuf, &LastOffsetDelta);
+
+                        LastOffset = Offset + LastOffsetDelta;
+
+                        rd_kafka_share_message_set_err_ops(
                             &msetr->msetr_rkq, msetr->msetr_broker_id,
                             RD_KAFKA_RESP_ERR__NOT_IMPLEMENTED,
```
pranavrth, line 1417: Return corrupt record error.

src/rdkafka_msgset_reader.c
```diff
@@ -1243,27 +1311,76 @@ rd_kafka_msgset_reader_peek_msg_version(rd_kafka_msgset_reader_t *msetr,
...
                            rd_slice_size(&rkbuf->rkbuf_reader));
 
-                if (Offset >=
-                        msetr->msetr_rktp->rktp_offsets.fetch_pos.offset &&
-                    !RD_KAFKA_IS_SHARE_CONSUMER(msetr->msetr_rkb->rkb_rk)) {
-                        rd_kafka_consumer_err(
+                if (!RD_KAFKA_IS_SHARE_CONSUMER(msetr->msetr_rkb->rkb_rk)) {
+                        if (Offset >= msetr->msetr_rktp->rktp_offsets.fetch_pos
+                                          .offset &&
+                            !RD_KAFKA_IS_SHARE_CONSUMER(
```
pranavrth: Remove this check.

src/rdkafka_msgset_reader.c
```diff
@@ -517,12 +517,33 @@ rd_kafka_msgset_reader_decompress(rd_kafka_msgset_reader_t *msetr,
...
-        rd_kafka_consumer_err(
-            &msetr->msetr_rkq, msetr->msetr_broker_id, err,
-            msetr->msetr_tver->version, NULL, rktp, Offset,
-            "Decompression (codec 0x%x) of message at %" PRIu64 " of %" PRIusz
-            " bytes failed: %s",
-            codec, Offset, compressed_size, rd_kafka_err2str(err));
+        if (!RD_KAFKA_IS_SHARE_CONSUMER(msetr->msetr_rkb->rkb_rk) ||
+            msetr->msetr_v2_hdr == NULL) {
+                /* Regular consumer, or v0/v1 MessageSet: single error. */
+                rd_kafka_consumer_err(
```
pranavrth, line 522: Move v0/v1 handling for share consumer also in the else condition below.

src/rdkafka_msgset_reader.c
```diff
@@ -1155,8 +1211,20 @@ rd_kafka_msgset_reader_v2(rd_kafka_msgset_reader_t *msetr) {
...
                     hdr.BaseTimestamp, hdr.BaseOffset, compressed,
                     payload_size);
-                if (err)
-                        goto err;
+                if (err) {
+                        /* For share consumer: decompression failed, error ops
+                         * already created. Buffer position already advanced
+                         * past compressed payload. Continue parsing next
+                         * MessageSet. For regular consumer: stop parsing. */
+                        if (RD_KAFKA_IS_SHARE_CONSUMER(
```
pranavrth: Check this again. We can get parse error from this path as well.

pranavrth (changes requested):

src/rdkafka_msgset_reader.c
PratRanj07: This handling is required as in case of an Unsupported magic bytes, we are throwing away the whole parsing for the share consumers, but we have to continue processing for other message sets and other partitions. That why we need this particular handling. Changed it to also check that the error is RD_KAFKA_RESP_ERR__NOT_IMPLEMENTED which is being returned in case of unsupported magic byte.

src/rdkafka_msgset_reader.c
PratRanj07: As discussed will be taken up in another task

src/rdkafka_msgset_reader.c
PratRanj07: Refactored this a bit in case of decompression errors, now we will just put an error op and in case of parsing errors, it will take the usual route

src/rdkafka_fetcher.c
pranavrth: Verify and remove this TODO.

src/rdkafka_fetcher.c
```diff
@@ -1457,6 +1465,9 @@ static rd_kafka_resp_err_t rd_kafka_share_fetch_reply_handle_partition(
                  * parse errors (which are partition-specific) */
         }
 
+        if (err)
```
pranavrth: Add comment to explain what are we doing here and why.

src/rdkafka_fetcher.c
pranavrth, line 1443: I had done this change recently. Please verify that this change is not causing any issue.

src/rdkafka_msgset_reader.c
pranavrth: We might hit this case as well. Understand what is needed here and let's fix this.

src/rdkafka_msgset_reader.c
pranavrth: Let's verify same happens in Java client.

src/rdkafka_msgset_reader.c
```diff
@@ -1243,27 +1325,74 @@ rd_kafka_msgset_reader_peek_msg_version(rd_kafka_msgset_reader_t *msetr,
...
+                        int32_t PartitionLeaderEpoch;
+                        int8_t MagicByteActual;
+                        int32_t Crc;
+                        int16_t Attributes;
+                        int32_t LastOffsetDelta;
+                        int64_t LastOffset;
+
+                        rd_kafka_buf_read_i32(rkbuf, &Length);
+
+                        len_start = rd_slice_offset(&rkbuf->rkbuf_reader);
```
pranavrth: Let's calculate the size of the read information and subtract that from Length. It should be Length - 15 for below buf reads

src/rdkafka_fetcher.c
```diff
@@ -1457,6 +1465,22 @@ static rd_kafka_resp_err_t rd_kafka_share_fetch_reply_handle_partition(
...
+         * inline by emitting per-offset RELEASE/REJECT ops and are
+         * swallowed by the v2 reader (msgset_reader.c). When a true
+         * parse error bubbles up the wire data is corrupt and we can no
+         * longer trust the buffer position or any per-batch metadata.
+         * Stop processing the rest of this ShareFetch response and
+         * propagate the error to the caller, which aborts the whole
+         * response handling.
+         */
+        if (err)
+                goto done;
```
pranavrth: Move this error handling to inside message set parsing block as we are adding this for those errors mainly.

pranavrth (changes requested):

src/rdkafka_msgset_reader.c
```diff
@@ -517,14 +517,59 @@ rd_kafka_msgset_reader_decompress(rd_kafka_msgset_reader_t *msetr,
...
+                } else {
+                        /* Share consumer with MessageSet v2: report error for
+                         * the entire batch with RELEASE acknowledgement —
+                         * decompression capability varies by consumer, so
+                         * another consumer might succeed on the same batch. */
+                        int64_t LastOffset =
+                            msetr->msetr_v2_hdr->BaseOffset +
+                            msetr->msetr_v2_hdr->LastOffsetDelta;
+
+                        rd_kafka_share_message_set_err_ops(
```
pranavrth: Add a TODO to translate errors inside this function. __BAD_COMPRESSION and __BAD_MSG should be INVALID_MSG error. __CRIT_SYS_RESOURCES should be sent as it is IMO.

src/rdkafka_msgset_reader.c
```diff
@@ -1243,27 +1325,70 @@ rd_kafka_msgset_reader_peek_msg_version(rd_kafka_msgset_reader_t *msetr,
...
+                                    .offset = Offset + 1;
+                        }
+
+                        /* Skip this Message(Set).
+                         * If the message is malformed, the skip may trigger
+                         * err_parse and return ERR__BAD_MSG. */
+                        rd_kafka_buf_read_i32(rkbuf, &Length);
+                        rd_kafka_buf_skip(rkbuf, Length);
+                } else {
+                        /* Share consumer: peek further into the MessageSet v2
```
pranavrth, line 1380: This is for version <0 or >2. We are parsing this assuming that it is in v2 format. Let's confirm it if this is correct or not.

src/rdkafka_fetcher.c
```diff
@@ -1729,7 +1759,7 @@ rd_kafka_share_fetch_reply_handle(rd_kafka_broker_t *rkb,
                    "Bad message (Fetch v%d): "
```
pranavrth: Let's correct this message.

src/rdkafka_fetcher.c
```diff
@@ -1094,17 +1102,17 @@ rd_kafka_share_build_response_rko(rd_kafka_broker_t *rkb,
...
                             rd_kafka_share_ack_type_from_msg_op(msg_rko);
 
-                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH) {
+                        /* Forward the rko to the application only for
+                         * messages and per-message errors; GAP entries
+                         * are tracked in the per-offset ack-type table
+                         * above but not surfaced as rkmessages. */
+                        if (msg_rko->rko_type == RD_KAFKA_OP_FETCH ||
+                            msg_rko->rko_type == RD_KAFKA_OP_CONSUMER_ERR) {
                                 rd_list_add(
```
pranavrth, line 1114: Add a TODO to change the op name to `RD_KAFKA_OP_CONSUMER_MESSAGE_ERR` instead.

src/rdkafka_fetcher.c
pranavrth: Change wording to explain that we are creating errored message op for crc, decompression and magic byte errors.

src/rdkafka_fetcher.c
```diff
@@ -1605,6 +1629,7 @@ rd_kafka_share_fetch_reply_handle(rd_kafka_broker_t *rkb,
         rd_list_t *inflight_acks      = NULL;
         rd_kafka_op_t *rko_orig       = request->rkbuf_opaque;
         rd_kafka_op_t *response_rko   = NULL;
+        rd_kafka_resp_err_t err;
```
pranavrth: Add NO_ERROR

pranavrth (changes requested):

src/rdkafka_msgset_reader.c
pranavrth:
```suggestion
                        rd_kafka_share_msgset_err_ops(
```

src/rdkafka_op.c
```diff
@@ -684,6 +688,72 @@ void rd_kafka_consumer_err(rd_kafka_q_t *rkq,
...
+        char buf[2048];
+        char *errstr;
+        int64_t offset;
+
+        /* Format error message once for all offsets */
+        va_start(ap, fmt);
+        rd_vsnprintf(buf, sizeof(buf), fmt, ap);
+        va_end(ap);
+
+        errstr = rd_strdup(buf);
```
pranavrth: No need to rd_strdup here as we are already doing in the loop.

src/rdkafka_msgset_reader.c
```diff
@@ -1088,19 +1135,54 @@ rd_kafka_msgset_reader_v2(rd_kafka_msgset_reader_t *msetr) {
...
+                                    "MessageSet at offset %" PRId64 " (%" PRId32
+                                    " bytes) "
+                                    "failed CRC32C check "
+                                    "(original 0x%" PRIx32
+                                    " != "
+                                    "calculated 0x%" PRIx32 ")",
+                                    hdr.BaseOffset, hdr.Length, hdr.Crc,
+                                    calc_crc);
+                                /* Skip to end of MessageSet */
+                                rd_kafka_buf_skip_to(rkbuf, crc_len);
```
pranavrth, line 1183: Add a FIXME: to fix the incorrect skip_to that the normal consumer is using that you have found.

src/rdkafka_op.c
```diff
@@ -648,6 +648,10 @@ void rd_kafka_q_op_err(rd_kafka_q_t *rkq,
  *
  * @sa rd_kafka_q_op_err()
  */
+/**
+ * TODO KIP-932 GA: Improve handling of this particular function
```
pranavrth:
```suggestion
 * TODO KIP-932: GA: Improve handling of this particular function
```

pranavrth (changes requested):

pranavrth (approved):
LGTM!. Nice work @PratRanj07 ! Thanks.

PR #5388, opened 2026-04-06
[KIP-932]: Refactor test files to make the build pass

PratRanj07:
Refactors share consumer tests to use common infrastructure pattern, fixes MSVC build error, and adds new utility functions.

Key Changes
1. Common Producer/Admin Pattern (tests 0170, 0171, 0172, 0173, 0177)

- Share producer and admin client handles across tests
- Reduces test execution time and code duplication

2. Enhanced test_create_share_consumer() API

// New signature supports acknowledgement mode configuration
rd_kafka_share_t *test_create_share_consumer(const char *group_name, 
                                               const char *ack_mode);

3. New API: rd_kafka_share_consumer_get_rk()

// Get the underlying rd_kafka_t handle from a share consumer
rd_kafka_t *rd_kafka_share_consumer_get_rk(rd_kafka_share_t *rkshare);

- Enables access to low-level Kafka client for advanced operations
- Useful for testing and debugging scenarios

4. MSVC Build Fix

- Replaced POSIX-specific strtok_r() with C89-standard string functions
- Fixes Windows build compatibility

5. Debug Test Selection

- Added TESTS_TO_DEBUG environment variable for selective test execution during development
- Example: TESTS_TO_DEBUG="0170,0171,0172" ./test-runner
- Speeds up development cycle by running only specific tests

PR #5411, opened 2026-04-16
[KIP-932] Add rd_kafka_share_sasl_set_credentials API and SASL PLAIN share consumer tests

Ankith-Confluent:
Add rd_kafka_share_sasl_set_credentials() public API as a share consumer
wrapper around rd_kafka_sasl_set_credentials() for SASL PLAIN/SCRAM.

Tests added:
- 0135: Share consumer credential update test - verifies share consumer
  can recover from wrong credentials by calling set_credentials, with
  both proactive (set_after_auth_failure=false) and reactive
  (set_after_auth_failure=true) scenarios.
- 0142: Share consumer reauth failure test - produces messages, corrupts
  credentials, then consumes while broker demands reauthentication,
  verifying auth error is correctly reported.

CI pipeline updated to support PLAIN mechanism and include test 0135.


tests/0135-sasl_credentials.cpp
```diff
@@ -126,6 +132,139 @@ static void do_test(bool set_after_auth_failure) {
...
 }
 
+static rd_bool_t share_error_seen = rd_false;
+
+static void share_auth_error_cb(rd_kafka_t *rk,
+                                int err,
+                                const char *reason,
+                                void *opaque) {
+  if (err == RD_KAFKA_RESP_ERR__AUTHENTICATION ||
+      err == RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN) {
```
ojasvajain: I dont think we should include this `RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN`.

tests/0135-sasl_credentials.cpp
```diff
@@ -126,6 +132,139 @@ static void do_test(bool set_after_auth_failure) {
...
+    Test::Say(tostr() << "Share consumer auth error: "
+                      << rd_kafka_err2str((rd_kafka_resp_err_t)err) << ": "
+                      << reason << "\n");
+    share_error_seen = rd_true;
+  }
```
ojasvajain, line 146: We should try to reuse the event_cb being used by existing test cases.

tests/0135-sasl_credentials.cpp
```diff
@@ -126,6 +132,139 @@ static void do_test(bool set_after_auth_failure) {
...
+ * @brief Test setting SASL credentials on a share consumer.
+ *
+ * 1. Produce a message to a topic.
+ * 2. Create share consumer with wrong credentials (or correct if
+ *    set_after_auth_failure is false).
+ * 3. Optionally wait for auth failure.
+ * 4. Set correct credentials via rd_kafka_share_sasl_set_credentials().
+ * 5. Verify share consumer can consume the message.
+ */
+static void do_test_share_consumer(bool set_after_auth_failure) {
```
ojasvajain, line 159: can we write it in cpp just to be consistent?

ojasvajain (commented):
Left some comments

tests/0135-sasl_credentials.cpp
Ankith-Confluent: So the share consumer API's like `rd_kafka_share_consumer_new`, `rd_kafka_share_consume_batch` is C only API, and they dont have  C++ wrappers. We will have to make C++ wrapper class for share consumers to write it in CPP.

tests/0135-sasl_credentials.cpp
Ankith-Confluent: `authErrorEventCb` uses `RdKafka::Conf::set("event_cb", ...)` which is a C++ API, should I create a wrapper for this?

tests/0142-reauthentication.c
```diff
@@ -752,9 +854,9 @@ int main_0142_reauthentication(int argc, char **argv) {
                 return 0;
         }
 
-        /* Each test (12 of them) will take slightly more than 1 reauth_time
+        /* Each test (13 of them) will take slightly more than 1 reauth_time
```
ojasvajain: why not 15? I see 15 test cases.

tests/0142-reauthentication.c
```diff
@@ -683,6 +683,108 @@ void do_test_reauth_failure(int64_t reauth_time, const char *topic) {
 }
 
 
+/* Test share consumer reauth failure with wrong PLAIN/SCRAM credentials. */
+void do_test_share_reauth_failure(int64_t reauth_time) {
```
ojasvajain, line 687: There is a [PR](https://github.com/confluentinc/librdkafka/pull/5388) out which adds some common helpers in test.c. It should be possible to use some of those helpers (like for producing msgs or setting group offset). That PR is already approved, so please try to use those helpers.

tests/0135-sasl_credentials.cpp
ojasvajain, line 159: [Same](https://github.com/confluentinc/librdkafka/pull/5411/changes#r3122949244)

ojasvajain: It's fine then.

tests/0135-sasl_credentials.cpp
```diff
@@ -126,6 +132,139 @@ static void do_test(bool set_after_auth_failure) {
...
+
+  /* Get correct credentials from test config */
+  test_conf_init(&conf, NULL, 30);
+  username = rd_strdup(test_conf_get(conf, "sasl.username"));
+  password = rd_strdup(test_conf_get(conf, "sasl.password"));
+  if (!username || !password) {
+    rd_kafka_conf_destroy(conf);
+    SUB_TEST_SKIP("sasl.username and/or sasl.password not configured\n");
+    return;
+  }
```
ojasvajain: username or password will never be NULL

tests/0135-sasl_credentials.cpp
ojasvajain: No, ignore

tests/0142-reauthentication.c
Ankith-Confluent: missed this, will update

ojasvajain (approved):
lgtm

PR #5413, opened 2026-04-17
[KIP-932] Add Windows share consumer tests pipeline

Ankith-Confluent:
Add a new Semaphore CI block to run share consumer tests (0170+) on
Windows (s1-prod-windows) against a native Kafka broker.

- Install OpenJDK 21 and download Kafka binary on Windows Server 2019
- Start KRaft broker natively using kafka-server-start.bat
- Run share consumer tests (x64 and x86) with TESTS_SKIP_BEFORE=0170
- Skip topic deletion in tests 0170, 0171, 0172 on Windows to avoid
  broker crash from NTFS file locking (KAFKA-1194)


pranavrth:
Have you checked all other test files?

Ankith-Confluent:
Yes, only `170`, `171`, `172` has `test_delete_topic` calls, other tests does not have
So the tests have passed in the CI

tests/0170-share_consumer_subscription.c
```diff
@@ -600,11 +600,13 @@ static void state_cleanup(sub_test_state_t *state) {
         /* Delete all created topics (skip already deleted ones) */
         for (i = 0; i < state->all_topic_cnt; i++) {
                 if (state->all_topics[i]) {
+#ifndef _WIN32
                         if (!state->topic_deleted[i]) {
                                 test_delete_topic(test_share_consumer_get_rk(
                                                       state->consumers[0]),
                                                   state->all_topics[i]);
                         }
+#endif
```
pranavrth: 1. We are checking if the client is in windows but we need to check if the broker is in windows. I would suggest that the tests should automatically do this figure this out and skip this topic deletion part.
2. Even if we want to have some check then it should be inside `test_delete_topic`.

pranavrth (changes requested):

pranavrth (approved):
LGTM!.

PR #5414, opened 2026-04-17
[KIP-932] Mock Broker : Minor fixes for ack workflow of Share Consumer

k-raina:
**Summary**
1. `epoch=-1` ShareFetch no longer acquires new records. It only processes piggybacked acks, releases remaining locks, and closes the session.
2. When log retention advances SPSO past acquired records, the consumer's ack for those offsets succeeds instead of returning INVALID_RECORD_STATE.
3. Ack validation uses a two-pass approach (validate all, then apply all). If any record in the batch fails, the entire batch is rejected with no partial application. 
4. ShareFetch and ShareAcknowledge now return NOT_LEADER_OR_FOLLOWER per-partition when the receiving broker isn't the partition leader, with actual leader ID and epoch in CurrentLeader.

### DIFF
--- a/src/rdkafka_mock_handlers.c
+++ b/src/rdkafka_mock_handlers.c
@@ -3760,9 +3760,21 @@ rd_kafka_mock_sgrp_apply_ack(rd_kafka_mock_sharegroup_t *sgrp,
         if (!pmeta)
                 return RD_KAFKA_RESP_ERR_NO_ERROR;
 
+        /* Pass 1: Validate all offsets in the range.
+         * Acks for a partition in a single request are atomic —
+         * all succeed or all fail together. */
         for (offset = first_offset; offset <= last_offset; offset++) {
-                rd_kafka_mock_sgrp_record_state_t *state =
-                    rd_kafka_mock_sgrp_record_state_find(pmeta, offset);
+                rd_kafka_mock_sgrp_record_state_t *state;
+
+                /* Offsets below SPSO are already logically archived
+                 * (e.g. due to log retention).  Acks for such
+                 * records are silently accepted — no error is
+                 * thrown for a record that is about to become
+                 * Archived. */
+                if (offset < pmeta->spso)
+                        continue;
+
+                state = rd_kafka_mock_sgrp_record_state_find(pmeta, offset);
                 if (!state)
                         continue;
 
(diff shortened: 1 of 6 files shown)


tests/0157-share_consumer_ack_mock.c
```diff
@@ -1525,5 +1791,11 @@ int main_0157_share_consumer_ack_mock(int argc, char **argv) {
         do_test_ack_after_lock_expiry_redelivers();
         do_test_ack_success_advances_spso();
 
+        /* Mock broker fix tests */
```
Ankith-Confluent, line 1794: we can remove this comment

src/rdkafka_mock_handlers.c
```diff
@@ -3756,30 +3768,48 @@ rd_kafka_mock_sgrp_apply_ack(rd_kafka_mock_sharegroup_t *sgrp,
...
-                        rd_kafka_mock_sgrp_record_release(sgrp, pmeta, state);
-                        break;
-                default:
-                        break;
+        /* Pass 2: Apply state transitions only if validation passed. */
+        if (!err) {
+                for (offset = first_offset; offset <= last_offset; offset++) {
+                        rd_kafka_mock_sgrp_record_state_t *state;
+
+                        if (offset < pmeta->spso)
```
Ankith-Confluent, line 3780: This is already validated in line 3755 right? This would not be required

src/rdkafka_mock_handlers.c
```diff
@@ -3756,30 +3768,48 @@ rd_kafka_mock_sgrp_apply_ack(rd_kafka_mock_sharegroup_t *sgrp,
...
+                                continue;
+
+                        state = rd_kafka_mock_sgrp_record_state_find(pmeta,
+                                                                     offset);
+                        if (!state)
```
Ankith-Confluent, line 3785: This check is not required

src/rdkafka_mock_handlers.c
```diff
@@ -4845,10 +4921,20 @@ rd_kafka_mock_handle_ShareAcknowledge(rd_kafka_mock_connection_t *mconn,
...
-                                        rd_kafka_buf_write_i32(
-                                            resp, -1); /* LeaderEpoch */
+                                        if (mpart && mpart->leader)
+                                                rd_kafka_buf_write_i32(
+                                                    resp,
+                                                    mpart->leader->id);
+                                        else
+                                                rd_kafka_buf_write_i32(
+                                                    resp, -1); /* LeaderId */
+                                        if (mpart)
```
Ankith-Confluent, line 4931: We need to check `mpart->leader` here also

src/rdkafka_mock_handlers.c
```diff
@@ -4456,8 +4504,18 @@ static int rd_kafka_mock_handle_ShareFetch(rd_kafka_mock_connection_t *mconn,
...
                                         /* Response: CurrentLeader */
-                                        rd_kafka_buf_write_i32(resp, -1);
-                                        rd_kafka_buf_write_i32(resp, -1);
+                                        if (mpart && mpart->leader)
+                                                rd_kafka_buf_write_i32(
+                                                    resp,
+                                                    mpart->leader->id);
+                                        else
+                                                rd_kafka_buf_write_i32(resp, -1);
+                                        if (mpart)
```
Ankith-Confluent, line 4513: nit: add a check for `mpart->leader` here too

tests/0157-share_consumer_ack_mock.c
```diff
@@ -1486,6 +1486,272 @@ static void do_test_ack_success_advances_spso(void) {
...
+        consumed_b = consume_n(consumer, 3, 40);
+        TEST_SAY("final_fetch_no_acq: B consumed %d/3\n", consumed_b);
+
+        extra = consume_n(consumer, 1, 5); /* ack B's records */
+
+        rd_kafka_share_consumer_close(consumer);
+        rd_kafka_share_destroy(consumer);
+
+        /* C should see nothing — everything acked. */
+        {
```
Ankith-Confluent, line 1545: can we remove these types of blocks

Ankith-Confluent (commented):
First pass of reviews

src/rdkafka_mock_handlers.c
```diff
@@ -3741,9 +3741,21 @@ rd_kafka_mock_sgrp_apply_ack(rd_kafka_mock_sharegroup_t *sgrp,
...
-                    rd_kafka_mock_sgrp_record_state_find(pmeta, offset);
+                rd_kafka_mock_sgrp_record_state_t *state;
+
+                /* Offsets below SPSO are already logically archived
+                 * (e.g. due to log retention).  Acks for such
+                 * records are silently accepted — no error is
+                 * thrown for a record that is about to become
+                 * Archived. */
+                if (offset < pmeta->spso)
+                        continue;
```
PratRanj07, line 3775: Have you checked the broker's behaviour in this case? It is the case where the acquisition lock would have expired or it has been moved out of cache right? Shouldn't it throw InvalidRecord in that case? Please check once the behaviour of the broker in Java client

tests/0157-share_consumer_ack_mock.c
```diff
@@ -1486,6 +1486,272 @@ static void do_test_ack_success_advances_spso(void) {
...
+        rd_kafka_mock_sharegroup_set_max_record_locks(ctx.mcluster, 0);
+
+        /* B picks up the released batch2 records. */
+        consumer = new_share_consumer(ctx.bootstraps, "sg-final-no-acq");
+        subscribe_topics(consumer, &topic, 1);
+
+        consumed_b = consume_n(consumer, 3, 40);
+        TEST_SAY("final_fetch_no_acq: B consumed %d/3\n", consumed_b);
+
+        extra = consume_n(consumer, 1, 5); /* ack B's records */
```
PratRanj07: I see this function has been created in other test files too. Can we create a similar function in test.c or use any of the existing ones

tests/0157-share_consumer_ack_mock.c
```diff
@@ -1486,6 +1486,272 @@ static void do_test_ack_success_advances_spso(void) {
...
+        SUB_TEST();
+        ctx = test_ctx_new();
+
+        /* Limit acquisition to 3 records at a time. */
+        rd_kafka_mock_sharegroup_set_max_record_locks(ctx.mcluster, 3);
+
+        TEST_ASSERT(rd_kafka_mock_topic_create(ctx.mcluster, topic, 1, 1) ==
+                        RD_KAFKA_RESP_ERR_NO_ERROR,
+                    "Failed to create mock topic");
+        produce_messages(ctx.producer, topic, msgcnt);
```
PratRanj07: Same for these. Can we use any of the existing funcions

PratRanj07 (changes requested):
Thanks @k-raina !
Here are some comments

src/rdkafka_mock_handlers.c
k-raina: The two-pass pattern guarantees they iterate the same data in the same order with no future in-between  mutations  affecting results.

src/rdkafka_mock_handlers.c
k-raina: Same as above, The two-pass pattern guarantees they iterate the same data in the same order with no future in-between mutations affecting results.

src/rdkafka_mock_handlers.c
k-raina: @PratRanj07 
Java test `testAcknowledgeBatchPostLsoMovement (line 7255)` indicates this behaviour, it acks offsets 2–14 with SPSO at 14, expects success, and only offset 14 gets acknowledged.

src/rdkafka_mock_handlers.c
Ankith-Confluent: but since both passes are right next to each other in the same function, is it really needed here?

src/rdkafka_mock_handlers.c
Ankith-Confluent: same as above
https://github.com/confluentinc/librdkafka/pull/5414#discussion_r3144874762

PratRanj07 (approved):
LGTM! Thanks @k-raina

Ankith-Confluent (approved):
LGTM!
Thanks @k-raina

PR #5417, opened 2026-04-22
admin: Fix duplicate groups in ListConsumerGroups response merge

adizaimi:
ListConsumerGroups fans out ListGroupsRequest to all brokers and 
merges the per-broker responses. The previous merge used rd_list_copy_to()
to unconditionally append every valid group listing from each broker, 
producing duplicate entries whenever multiple brokers reported the same group id.

Changes:
Replaced the bulk copy in the merge path with a deduplicating helper, rd_kafka_ConsumerGroupListing_list_append_dedup(), that keeps the
 destination list sorted by group_id and uses rd_list_find() (bsearch) for lookups. 
The sorted state persists across broker merges, so the common case 
(full overlap across brokers) is pure bsearch lookups with zero allocations.
Added unit tests covering: identical groups, disjoint groups, three-broker 
complex overlap, and duplicates with interleaved errors.
CHANGELOG: entry added under Unreleased → Admin client fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# Unreleased
+
+## Fixes
+
+### Admin client fixes
+
+* Fix duplicate groups in `ListConsumerGroups` when multiple brokers return the same group.
+
+
 # librdkafka v2.14.1
 
 librdkafka v2.14.1 is a maintenance release:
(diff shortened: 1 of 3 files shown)


adizaimi:
Note that this was created as an internal PR after initial review in https://github.com/confluentinc/librdkafka/pull/5398.

pranavrth (approved):
LGTM!. Thanks for the contribution @adizaimi!.

adizaimi:
Thank you @pranavrth for your review and guidance with this PR.

PR #5421, opened 2026-04-27
[KIP-932]: Remove use of rd_kafka_share_consumer_get_rk

PratRanj07:


pranavrth (approved):
LGTM!.

PratRanj07 (commented):
There is an intermittent failure in the tests. It seems like it could be due to an issue with close or destoy. It will be taken care of in another PR

PR #5422, opened 2026-04-28
[KIP-932]: Create wrappers around close and destroy

PratRanj07:
This PR creates wrapper functions around rd_kafka-share_consumer_close and rd_kafka_share_destroy.
Add loggings around the share and close to test on which the tests are hanging
Use the wrappers in all the tests

### DIFF
--- a/tests/test.c
+++ b/tests/test.c
@@ -8203,7 +8203,7 @@ void test_share_set_isolation_level(const char *group_name,
  * @param rkshare The share consumer to close.
  */
 void test_share_consumer_close(rd_kafka_share_t *rkshare) {
-        rd_kafka_error_t *error;
+        rd_kafka_resp_err_t error;
         test_timing_t timing;
 
         TEST_SAY("Calling rd_kafka_share_consumer_close\n");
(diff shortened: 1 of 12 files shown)


tests/test.c
```diff
@@ -8196,3 +8196,37 @@ void test_share_set_isolation_level(const char *group_name,
         const char *cfg[] = {"share.isolation.level", "SET", isolation_level};
         test_alter_group_configurations(group_name, cfg, 1);
 }
+
+/**
+ * @brief Close a share consumer and verify it succeeds.
```
```suggestion
 * @brief Close a share consumer.
```

We are not verifying if the close completed successfully. The error will be returned very rarely. And it is possible that close() does not work as expected (for eg, request times out) and still return NULL. So let's not say we are "verifying it succeeds". No harm in keeping the error check though.

tests/test.c
```diff
@@ -8196,3 +8196,37 @@ void test_share_set_isolation_level(const char *group_name,
...
+ *
+ * @param rkshare The share consumer to close.
+ */
+void test_share_consumer_close(rd_kafka_share_t *rkshare) {
+        rd_kafka_resp_err_t error;
+        test_timing_t timing;
+
+        TEST_SAY("Calling rd_kafka_share_consumer_close\n");
+
+        TIMING_START(&timing, "SHARE_CONSUMER.CLOSE");
```
ojasvajain: Why do we need to time it? close() can take longer based on `socket.timeout.ms`. Same for `destroy`.

ojasvajain (changes requested):

ojasvajain (approved):

PR #5424, opened 2026-04-29
[KIP-932]: Fix 0155 fatal-state test cleanup and 0177 replication factor

Ankith-Confluent:
- 0155: switch fatal-state subtests back to calling
  `rd_kafka_share_consumer_close` directly. The `test_share_consumer_close`
  wrapper TEST_FAILs on any non-zero return, but a consumer that has
  transitioned to fatal state legitimately returns an error from close,
  so the wrapper was failing the test at cleanup time.
- 0155: raise the `wait_fatal_error` timeout in
  `do_test_share_group_error_injection` from 5000 ms to 7000 ms. The mock's
  default ShareGroupHeartbeat interval is 5000 ms, so the next heartbeat
  (carrying the injected INVALID_REQUEST) fires almost exactly when the
  poll deadline elapses — the fatal flag was being set ~ms after the
  poll loop returned NO_ERROR.
- 0177: pass `-1` instead of a hardcoded `1` for the replication factor
  in all `test_create_topic` calls so the tests honor the broker's
  configured default rather than assuming RF=1.

### DIFF
--- a/tests/0155-share_group_heartbeat_mock.c
+++ b/tests/0155-share_group_heartbeat_mock.c
@@ -546,7 +546,7 @@ static void do_test_share_group_error_injection(void) {
             RD_KAFKA_RESP_ERR_INVALID_REQUEST, 0);
 
         /* Wait for the fatal error to propagate. */
-        fatal_err = wait_fatal_error(share_c, 5000, errstr, sizeof(errstr));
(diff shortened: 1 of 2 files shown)


PratRanj07 (approved):
LGTM

PR #5428, opened 2026-05-04
Add enum and defensive stub for AWS IAM

prashah-confluent:
## Summary

Adds `aws_iam` as a valid value for `sasl.oauthbearer.metadata.authentication.type`.

librdkafka does not implement the AWS IAM token flow natively as of now and high-level
clients are expected to register their own token-refresh callback via
`rd_kafka_conf_set_oauthbearer_token_refresh_cb` before `rd_kafka_new`.

A defensive stub callback handles the misconfiguration case where no
callback is registered, failing the token refresh with a clear message
that points the user at the high-level client.

## When the stub fires

Only when ALL three are true:
1. `sasl.oauthbearer.method=oidc`
2. `sasl.oauthbearer.metadata.authentication.type=aws_iam`
3. No `oauthbearer_token_refresh_cb` registered on the conf


tests/0126-oauthbearer_oidc.c
```diff
@@ -996,13 +1009,122 @@ void do_test_produce_share_consumer_with_OIDC_sub_claim(rd_kafka_conf_t *conf) {
...
+}
+
 int main_0126_oauthbearer_oidc(int argc, char **argv) {
         rd_kafka_conf_t *conf;
         const char *sec;
         const char *oidc;
 
         test_conf_init(&conf, NULL, 60);
 
+        do_test_OIDC_aws_iam_conf_roundtrip();  // no broker connection needed
```
emasab: Please move this test below together with the other two

prashah-confluent: Moved the test down with the other 2 tests.

emasab (approved):
LGTM, Thanks @prashah-confluent !

PR #5429, opened 2026-05-05
[KIP-932] Fix race: SHARE_FETCH __DESTROY reply silently dropped

pranavrth:
When a broker is decommissioned by the main thread (metadata update) between the time terminate0/commit_sync enqueues a SHARE_FETCH op on B.rkb_ops and the time B's broker thread processes it, the broker thread sees or_instance_terminating(B) == true and replies with __DESTROY via rd_kafka_share_fetch_op_reply_with_err.

Without this fix, the reply carried RD_KAFKA_OP_REPLY which makes rd_kafka_op_handle_std take the __DESTROY shortcut at rdkafka_op.c:987-991 and silently destroy the op, so:
  - share_session_leave_remaining_cnt never decremented -> close hangs
  - brokers_awaiting_result_cnt never decremented -> commit_sync hangs until its own timeout, then surfaces __TIMED_OUT instead of the real __DESTROY
  - per-batch errors set on rktpar->err by the helper are lost

Set rko_op_cb on SHARE_FETCH ops at both enqueue sites (rd_kafka_share_enqueue_fetch_op and rd_kafka_share_enqueue_sync_ack_op). rd_kafka_op_reply then stamps RD_KAFKA_OP_CB instead of RD_KAFKA_OP_REPLY, which the handle_std __DESTROY shortcut does not match -> the reply handler runs and the counters/errors propagate.

Drop the now-unreachable RD_KAFKA_OP_SHARE_FETCH | RD_KAFKA_OP_REPLY case from rd_kafka_poll_cb.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3184,6 +3184,23 @@ static void rd_kafka_share_commit_sync_maybe_complete(rd_kafka_t *rk,
                                                       rd_kafka_cgrp_t *rkcg);
 static void rd_kafka_share_enqueue_sync_ack_op(rd_kafka_t *rk,
                                                rd_kafka_broker_t *rkb);
+rd_kafka_op_res_t rd_kafka_share_fetch_reply_op(rd_kafka_t *rk,
+                                                rd_kafka_op_t *rko_orig);
+
+/**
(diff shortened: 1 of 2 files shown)


ojasvajain (approved):
LGTM

ojasvajain (approved):

PR #5432, opened 2026-05-06
KIP-932: Verifiable Clients Implementation to Run KIP-932 System Tests

ojasvajain:
This PR implements a verifiable producer and a verifiable share consumer client that are compatible with Ducktape framework used by AK repo to run system test cases. More details in the README.

### DIFF
--- a/kafkatest/common/common.c
+++ b/kafkatest/common/common.c
@@ -1,0 +1,188 @@
+#define _POSIX_C_SOURCE 200809L
+
+#include <ctype.h>
+#include <errno.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/time.h>
+#include <time.h>
+
+#include "common.h"
+
+volatile sig_atomic_t run = 1;
+
+static pthread_mutex_t stdout_mtx = PTHREAD_MUTEX_INITIALIZER;
+
+void stdout_lock(void) {
+        pthread_mutex_lock(&stdout_mtx);
+}
+
+void stdout_unlock(void) {
+        pthread_mutex_unlock(&stdout_mtx);
+}
+
+static void sig_handler(int sig) {
+        (void)sig;
+        run = 0;
+}
+
+void install_signals(void) {
+        struct sigaction sa;
+        memset(&sa, 0, sizeof(sa));
+        sa.sa_handler = sig_handler;
+        sigaction(SIGINT, &sa, NULL);
+        sigaction(SIGTERM, &sa, NULL);
+}
+
+int64_t now_ms(void) {
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+        return (int64_t)ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
+}
+
+void json_write_string(FILE *out, const char *s) {
+        fputc('"', out);
+        if (s) {
+                for (const unsigned char *p = (const unsigned char *)s; *p;
+                     p++) {
+                        switch (*p) {
+                        case '"':
+                                fputs("\\\"", out);
+                                break;
+                        case '\\':
+                                fputs("\\\\", out);
+                                break;
+                        case '\b':
+                                fputs("\\b", out);
+                                break;
+                        case '\f':
+                                fputs("\\f", out);
+                                break;
+                        case '\n':
+                                fputs("\\n", out);
+                                break;
+                        case '\r':
+                                fputs("\\r", out);
+                                break;
+                        case '\t':
+                                fputs("\\t", out);
+                                break;
+                        default:
+                                if (*p < 0x20)
+                                        fprintf(out, "\\u%04x", *p);
+                                else
+                                        fputc(*p, out);
+                        }
+                }
+        }
+        fputc('"', out);
+}
+
+void emit_event(const char *name) {
+        stdout_lock();
+        fputs("{\"name\":", stdout);
+        json_write_string(stdout, name);
+        fprintf(stdout, ",\"timestamp\":%" PRId64 "}\n", now_ms());
+        fflush(stdout);
+        stdout_unlock();
+}
+
+static char *trim(char *s) {
+        while (*s && isspace((unsigned char)*s))
+                s++;
+        char *end = s + strlen(s);
+        while (end > s && isspace((unsigned char)end[-1]))
+                end--;
+        *end = '\0';
+        return s;
+}
+
+static int apply_kv(rd_kafka_conf_t *conf,
+                    const char *key,
+                    const char *val,
+                    char *errstr,
+                    size_t errstr_size) {
+        if (rd_kafka_conf_set(conf, key, val, errstr, errstr_size) !=
+            RD_KAFKA_CONF_OK)
+                return -1;
+        return 0;
+}
+
+int load_properties_file(const char *path,
+                         rd_kafka_conf_t *conf,
+                         char *errstr,
+                         size_t errstr_size) {
+        FILE *fp = fopen(path, "r");
+        if (!fp) {
+                snprintf(errstr, errstr_size,
+                         "Failed to open properties file %s: %s", path,
+                         strerror(errno));
+                return -1;
+        }
+
+        char buf[4096];
+        int rc = 0;
+        while (fgets(buf, sizeof(buf), fp)) {
+                char *line = trim(buf);
+                if (!*line || *line == '#' || *line == '!')
+                        continue;
+
+                char *sep = strpbrk(line, "=:");
+                if (!sep || sep == line) {
+                        /* Malformed line; skip silently (matches Java
+                         * Properties behavior of ignoring lines without a
+                         * separator). */
+                        continue;
+                }
+
+                *sep = '\0';
+                char *key = trim(line);
+                char *val = trim(sep + 1);
+
+                if (apply_kv(conf, key, val, errstr, errstr_size) == -1) {
+                        rc = -1;
+                        break;
+                }
+        }
+
+        fclose(fp);
+        return rc;
+}
+
+int apply_x_properties(const char *csv,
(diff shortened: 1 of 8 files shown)


pranavrth:
There is already an existing verifyiable client - https://github.com/confluentinc/librdkafka/blob/b889b1cf6062c2ce626740de8a00169ceed5fd45/examples/kafkatest_verifiable_client.cpp#L6

Have you checked this?

ojasvajain:
@pranavrth Yes. As discussed, will deprecate this when we add the verifiable consumer client for regular consumer as the example client is outdated and is not compatible with the current set of test cases in apache/kafka.

kafkatest/verifiable_share_consumer/verifiable_share_consumer.c
```diff
@@ -0,0 +1,652 @@
...
+                if (pb->items[i].partition == partition &&
+                    strcmp(pb->items[i].topic, topic) == 0)
+                        return &pb->items[i];
+        }
+        if (pb->cnt >= MAX_PARTITIONS) {
+                fprintf(stderr,
+                        "partition bucket full (%d); dropping %s[%" PRId32
+                        "]\n",
+                        MAX_PARTITIONS, topic, partition);
+                return &pb->items[pb->cnt - 1];
```
k-raina: Why are we returning last partition is case of `pb->cnt >= MAX_PARTITIONS`?

kafkatest/common/common.c
```diff
@@ -0,0 +1,227 @@
...
+        fflush(stdout);
+        stdout_unlock();
+}
+
+int conf_set(rd_kafka_conf_t *conf, const char *key, const char *value) {
+        char errstr[512];
+        if (rd_kafka_conf_set(conf, key, value, errstr, sizeof(errstr)) !=
+            RD_KAFKA_CONF_OK) {
+                fprintf(stderr, "%s: %s\n", key, errstr);
+                rd_kafka_conf_destroy(conf);
```
k-raina, line 149: When value is UNKNOWN, should we log a warning and skip it?

kafkatest/verifiable_share_consumer/verifiable_share_consumer.c
```diff
@@ -0,0 +1,652 @@
...
+
+        while (run && (max_messages < 0 || total_acknowledged < max_messages)) {
+                size_t rcvd             = 0;
+                rd_kafka_error_t *error = rd_kafka_share_consume_batch(
+                    rkshare, POLL_TIMEOUT_MS, batch, &rcvd);
+                if (error) {
+                        fprintf(stderr, "consume_batch: %s\n",
+                                rd_kafka_error_string(error));
+                        rd_kafka_error_destroy(error);
+                        continue;
```
k-raina, line 626: This code treats every error the same way: print it and try again. Do we need to consider breaking it for fatal error?

kafkatest/verifiable_producer/verifiable_producer.c
```diff
@@ -0,0 +1,408 @@
...
+            {"command-config", required_argument, 0, 'C'},
+            {"debug", required_argument, 0, 'd'},
+            {"property", required_argument, 0, 'X'},
+            {"help", no_argument, 0, 'h'},
+            {0, 0, 0, 0}};
+
+        int opt;
+        while ((opt = getopt_long(argc, argv, "X:h", long_opts, NULL)) != -1) {
+                switch (opt) {
+                case 't':
```
k-raina, line 241: Should we accept for more than 1 topics?

kafkatest/verifiable_producer/verifiable_producer.c
```diff
@@ -0,0 +1,408 @@
...
+                if (err == RD_KAFKA_RESP_ERR__QUEUE_FULL) {
+                        rd_kafka_poll(rk, 100);
+                        continue;
+                }
+                if (err) {
+                        state.num_errors++;
+                        emit_send_error(topic, key_ptr, key_len, value_buf,
+                                        (size_t)value_len,
+                                        rd_kafka_err2str(err),
+                                        rd_kafka_err2name(err));
```
k-raina, line 402: Should counter be incremented here?

kafkatest/verifiable_producer/verifiable_producer.c
```diff
@@ -0,0 +1,408 @@
...
+                }
+
+                rate_limit(rk, rate_sleep_us);
+        }
+
+        rd_kafka_flush(rk, 30 * 1000);
+        rd_kafka_destroy(rk);
+
+        emit_event("shutdown_complete");
+        emit_tool_data();
```
k-raina, line 415: Code emits `tool_data` once, at shutdown. So a long-running test sees nothing from us until it ends. If the harness has a "no progress detected" timeout, it'll falsely flag us as hung. Can we check for liveliness?

kafkatest/verifiable_share_consumer/verifiable_share_consumer.c
```diff
@@ -0,0 +1,652 @@
...
+        rd_kafka_ConfigResource_destroy(res);
+
+        rd_kafka_event_t *ev = rd_kafka_queue_poll(q, 30000);
+        int rc               = 0;
+        if (!ev) {
+                snprintf(errbuf, errbuf_size,
+                         "IncrementalAlterConfigs: timeout");
+                rc = -1;
+        } else {
+                rd_kafka_resp_err_t err = rd_kafka_event_error(ev);
```
k-raina: Are we checking resource level error?

kafkatest/verifiable_producer/verifiable_producer.c
```diff
@@ -0,0 +1,408 @@
...
+
+static void usage(const char *prog) {
+        fprintf(stderr,
+                "Usage: %s --topic <topic> --bootstrap-server <servers> "
+                "[options]\n"
+                "\nOptions:\n"
+                "  --topic <t>               Topic to produce to (required)\n"
+                "  --bootstrap-server <s>    Broker list (required; also "
+                "--broker-list)\n"
+                "  --max-messages <n>        Messages to send; -1=infinite "
```
k-raina, line 190: What does infinite means for system tests?

kafkatest/verifiable_share_consumer/verifiable_share_consumer.c
```diff
@@ -0,0 +1,652 @@
...
+static void usage(const char *prog) {
+        fprintf(stderr,
+                "Usage: %s --topic <t> --group-id <g> --bootstrap-server <s> "
+                "[options]\n"
+                "\nOptions:\n"
+                "  --topic <t>                    Topic (required)\n"
+                "  --group-id <g>                 Share group id (required)\n"
+                "  --bootstrap-server <s>         Broker list (required; "
+                "also --broker-list)\n"
+                "  --max-messages <n>             Default: -1 (infinite)\n"
```
k-raina, line 420: **[Same](https://github.com/confluentinc/librdkafka/pull/5432/changes#r3246109196)**

k-raina (changes requested):
Thanks @ojasvajain for PR

Couple of questions: 
1. How are we planning to run these tests, i don't see any trigger for these tests?
2. KIP-932 has three ways to acknowledge a record : ACCEPT, RELEASE, REJECT. Should we add Redelivery tests & Rejected Tests? Maybe add a CLI flag like --release-rate 0.1 to randomly RELEASE 10% of records, plus --reject-rate for REJECT

kafkatest/verifiable_producer/verifiable_producer.c
ojasvajain: The implementation mirrors AK's VerifiableShareConsumer impl present [here](https://github.com/apache/kafka/blob/trunk/tools/src/main/java/org/apache/kafka/tools/VerifiableShareConsumer.java). If Java's impl doesn't have it, then it is not used in test cases, so we shouldn't need it here.

kafkatest/verifiable_producer/verifiable_producer.c
ojasvajain: It means the share consumer will consume until the process is killed externally

ojasvajain:
@k-raina 
> 1. How are we planning to run these tests, i don't see any trigger for these tests?

AK contains a `ducker-ak` script that will take care of wiring these verifiable clients and running it against the system test cases in the AK repo. You can go through the README.md added in this PR. 


> 2. KIP-932 has three ways to acknowledge a record : ACCEPT, RELEASE, REJECT. Should we add Redelivery tests & Rejected Tests? Maybe add a CLI flag like --release-rate 0.1 to randomly RELEASE 10% of records, plus --reject-rate for REJECT

This is not present in AK's implementation so decide to skip it.

kafkatest/verifiable_producer/verifiable_producer.c
ojasvajain: [Same](https://github.com/confluentinc/librdkafka/pull/5432#discussion_r3257296240)

kafkatest/verifiable_share_consumer/verifiable_share_consumer.c
ojasvajain: Fixed it. I am now terminating the client if we go inside the if condition. This is a defensive check, we should never reach here.

kafkatest/verifiable_share_consumer/verifiable_share_consumer.c
ojasvajain: Fixed it. I am now terminating the client if it is a fatal error.

kafkatest/common/common.c
ojasvajain: [Same](https://github.com/confluentinc/librdkafka/pull/5432#discussion_r3257296240)

kafkatest/verifiable_producer/verifiable_producer.c
ojasvajain: We are already incrementing the error counter. Are we talking about something else?

PR #5435, opened 2026-05-11
[KIP-932] Implement client-side poll metrics and max.poll.interval.ms enforcement for the share consumer

Ankith-Confluent:
Implements three KIP-932 share consumer client telemetry metrics:

- `org.apache.kafka.consumer.share.poll.idle.ratio.avg` — average fraction of poll cycle time the consumer is idle inside `share_consume_batch`
- `org.apache.kafka.consumer.share.time.between.poll.avg` — average delay (ms) between successive `share_consume_batch` invocations
- `org.apache.kafka.consumer.share.time.between.poll.max` — maximum delay (ms) between successive `share_consume_batch` invocations

Formula used for poll idle ratio 
```
pollIdleRatio = (cur_poll_end - cur_poll_start) / ((cur_poll_end - cur_poll_start) + start_to_start_interval)
= current_idle / (current_idle + start_to_start_interval) 
```
If we reused the regular formula, our `consumer.share.poll.idle.ratio.avg`
would report a different number than Java's share consumer for the same
workload, breaking cross-client dashboards. So we mirror Java's formula
in a new helper instead.

The gate added in `rd_kafka_app_poll_start` only wraps the **poll-idle-ratio
sampling block**, not the `rk_ts_last_poll` write that `max.poll.interval.ms`
relies on. That write happens earlier in the function and runs for every
consumer (regular and share) as before. So the regular consumer's session
timeout enforcement keeps working exactly the same.

### max.poll.interval.ms Enforcement

Moves max.poll.interval.ms enforcement from the generic queue path to dedicated
share consumer poll start/end functions. The `INT64_MAX` blocked-window sentinel
and heartbeat expedite logic are now in `rd_kafka_share_record_poll_start()` and
`rd_kafka_share_record_poll_end()`, ensuring only `rd_kafka_share_consume_batch()`
resets the timer (not `rd_kafka_poll()` or other queue operations).

The regular consumer's max.poll.interval.ms enforcement is unaffected — the
`rk_ts_last_poll` write in `rd_kafka_app_poll_start()` runs for all consumers
before any share-specific gates.

### Tests

Adds test suite 0185-share_consumer_max_poll_interval with 7 real-broker subtests
and 2 mock regression checks, covering eviction, rejoin, steady consumption,
multi-consumer, and blocking-poll scenarios.


src/rdkafka.c
k-raina, line 1062: Can these checks be simplified further?

src/rdkafka.c
```diff
@@ -1058,16 +1058,30 @@ void rd_kafka_destroy_final(rd_kafka_t *rk) {
                         rd_kafka_q_destroy(rk->rk_consumer.q);
                 rd_avg_destroy(
```
k-raina: Does this needs to be moved into `if (!RD_KAFKA_IS_SHARE_CONSUMER(rk)) { `?

src/rdkafka.c
```diff
@@ -3698,6 +3734,45 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
...
+ */
+static void rd_kafka_share_record_poll_start(rd_kafka_t *rk) {
+        rd_ts_t now                          = rd_clock();
+        rk->rk_telemetry.ts_share_poll_start = now;
+
+        if (rk->rk_telemetry.ts_last_share_poll_start)
+                rk->rk_telemetry.time_since_last_share_poll =
+                    now - rk->rk_telemetry.ts_last_share_poll_start;
+        else
+                rk->rk_telemetry.time_since_last_share_poll = 0;
```
k-raina: Is this logic intentional? 
Eg. for 3 polls -- 1st poll=0ms, 2nd poll = 50ms, 3rd poll = 50ms.
Avg is calculated as (0 + 50 + 50) / 3 = 33 ms

src/rdkafka.c
```diff
@@ -3698,6 +3734,45 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
...
+
+        rd_avg_add(
+            &rk->rk_telemetry.rd_avg_current.rk_avg_share_time_between_poll,
+            rk->rk_telemetry.time_since_last_share_poll / 1000); /* μs → ms */
```
k-raina: Is there rounding problem in this calculation?  i.e `999 μs  ÷ 1000 = 0`

So. for N polls : 
1. 800 μs ÷ 1000 -> 0 ms
2. 750 μs ÷ 1000 -> 0 ms
3. 900 μs ÷ 1000-> 0 ms
... N

The metric reports average = 0 ms, even though polls are happening 1200 times a second.

src/rdkafka.c
```diff
@@ -3698,6 +3734,45 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
...
+ * @brief Record the end of a share consume batch for telemetry.
+ *        Computes pollTime = now − pollStart, then idle ratio =
+ *        pollTime / (pollTime + timeSinceLastPoll), and records the sample.
+ */
+static void rd_kafka_share_record_poll_end(rd_kafka_t *rk) {
+        rd_ts_t end       = rd_clock();
+        rd_ts_t poll_time = end - rk->rk_telemetry.ts_share_poll_start;
+        rd_ts_t total = poll_time + rk->rk_telemetry.time_since_last_share_poll;
+        if (total > 0) {
+                int64_t poll_idle_ratio = poll_time * 1000000 / total;
```
k-raina, line 3961: `poll_time * 1000000` is the most overflow-prone. Although for realistic poll durations (ms to seconds) this multiplication is nowhere close to overflow(100d+), but we could be defensive?

src/rdkafka_telemetry_decode.c
```diff
@@ -561,13 +568,25 @@ int unit_test_telemetry(rd_kafka_type_t rk_type,
 
         rd_avg_init(&rk->rk_telemetry.rd_avg_current.rk_avg_poll_idle_ratio,
                     RD_AVG_GAUGE, 0, 500 * 1000, 2, rd_true);
+        rd_avg_init(
+            &rk->rk_telemetry.rd_avg_current.rk_avg_share_poll_idle_ratio,
+            RD_AVG_GAUGE, 0, 500 * 1000, 2, rd_true);
```
k-raina: Why is histogram range higher in Test when its set as (0,1) in implementation here: https://github.com/confluentinc/librdkafka/pull/5435/changes#diff-eef17694b5807cd63a95c3d86c39752b62f8bdb46e46c59231bf02353a5daef5R2765-R2768

src/rdkafka_telemetry_decode.c
k-raina, line 1: Are we unit testing these newly added functions rd_kafka_share_record_poll_start & record_poll_end ?

src/rdkafka_telemetry_decode.c
```diff
@@ -546,7 +546,14 @@ int unit_test_telemetry(rd_kafka_type_t rk_type,
         rk->rk_conf = *conf;
         rd_free(conf);
 
-        rk->rk_type           = rk_type;
+        rk->rk_type = rk_type;
+        /* Auto-detect share consumer from metric name prefix. */
+        if (strncmp(expected_name,
```
k-raina, line 551: nit: test code looks at the metric name string and guesses type of consumer. Is there more relable way to check for share consumer in test setup?

src/rdkafka.c
```diff
@@ -3698,6 +3734,45 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
         return err;
 }
 
+/**
+ * @brief Record the start of a share consume batch for telemetry.
+ *        Captures the start timestamp, computes the interval since the
+ *        previous poll, records the time-between-poll sample, and updates
+ *        the previous-poll-start timestamp.
+ */
+static void rd_kafka_share_record_poll_start(rd_kafka_t *rk) {
```
k-raina: Will these functions be executed when telemetry is turned off? i.e `rk_conf.enable_metrics_push=false`

src/rdkafka.c
```diff
@@ -3740,6 +3815,8 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
         if (error)
                 return error;
 
+        rd_kafka_share_record_poll_start(rk);
```
k-raina: Three error paths above return before reaching `record_poll_start`, so failed calls produce no sample at all. The next successful poll's `time_since_last_share_poll` then silently absorbs the entire gap, including all failed calls. Is this intentional?

src/rdkafka.c
```diff
@@ -3698,6 +3734,45 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
...
+}
+
+/**
+ * @brief Record the end of a share consume batch for telemetry.
+ *        Computes pollTime = now − pollStart, then idle ratio =
+ *        pollTime / (pollTime + timeSinceLastPoll), and records the sample.
+ */
+static void rd_kafka_share_record_poll_end(rd_kafka_t *rk) {
+        rd_ts_t end       = rd_clock();
+        rd_ts_t poll_time = end - rk->rk_telemetry.ts_share_poll_start;
```
k-raina: Regular consumer code: 
```  if (rkq->rkq_ts_last_poll_end && !RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
          int64_t poll_idle_ratio = 0;
          rd_ts_t poll_interval   = now - rkq->rkq_ts_last_poll_start;
          if (poll_interval) {
                  rd_ts_t idle_interval = rkq->rkq_ts_last_poll_end -
                                          rkq->rkq_ts_last_poll_start;
                  poll_idle_ratio =
                      idle_interval * 1000000 / poll_interval;
          }
```
Regular consumer is using last poll time as numerator, however current consumer is using current poll time as numerator. Is this intentional?

src/rdkafka_int.h
k-raina, line 1358: Are we setting poll end time to zero for share consumers also?

k-raina (commented):
Thanks for PR @Ankith-Confluent , added some comments.

Also, lets figure out if we can write integration tests also for this PR. We might need to explore , if we can modify CI to build AK with 
```
    metric.reporters=com.example.KafkaTopicTelemetryExporter
    telemetry.topic=__client_metrics
```
Where export() can be modified to produce metrics to temporary topic `shareconsumer_clientmetrics`. 

Then integration test can asset by consuming metrics from topic.

src/rdkafka.c
Ankith-Confluent: Sure I will add if/else block instead.

src/rdkafka.c
Ankith-Confluent: Yes will add.

src/rdkafka.c
Ankith-Confluent: no, it will still fill the storage fields with values, but the metric wont be pushed.
We are following the regular consumer, however I can add a flag to check and early return.

src/rdkafka.c
Ankith-Confluent: yes its expected and is following Java implementation. Will add a comment regarding this.

src/rdkafka.c
Ankith-Confluent: I have changed this conversion of μs -> ms to read time instead of write time.

src/rdkafka.c
Ankith-Confluent: Yes we are following the implementation of java.

src/rdkafka.c
Ankith-Confluent: `max.poll.interval.ms` would have removed the consumer from the group after the set time, mostly 5 minutes. So this wouldn't be required.

src/rdkafka.c
Ankith-Confluent: Yes we are following java implementation and the sampling location was discussed with @pranavrth and we came to a conclusion that it should be after the error handling, `rd_kafka_share_acknowledge_all_if_implicit` and `rd_kafka_share_ensure_all_acknowledged_if_explicit`.

src/rdkafka_int.h
Ankith-Confluent: ts_last_share_poll_start is a rolling previous poll start timestamp used to compute the interval, its not required for share consumer. If we set it to 0, we would lose the interval.

src/rdkafka_telemetry_decode.c
Ankith-Confluent: I didnt want to add a new parameter to the function which would have cause a lot of changes where the fucntion was called, so @pranavrth suggested this method.

src/rdkafka_telemetry_decode.c
Ankith-Confluent: The current unit test framework which is there for telemetry does not support this, it only supports storage→encode→decode pipeline. The regular consumer poll_start/end also does not have unit tests.

src/rdkafka_telemetry_decode.c
Ankith-Confluent: The idle-ratio is stored scaled by 1e6 (so values range [0, 1000000]), so both production and test should use (0, 1000 * 1000, 2), because in avg_t we store scaled by 1e6. The regular consumer should also follow the same, so its bug, but its harmless because the metric calculator reads ra_v.sum/cnt directly.

src/rdkafka_int.h
```diff
@@ -746,22 +746,39 @@ struct rd_kafka_s {
...
                                                          *   latency avg */
                         rd_avg_t
                             rk_avg_rebalance_latency; /**< Rolled over rebalance
                                                        *   latency avg */
                 } rd_avg_rollover;
 
+                /* Share consumer poll/batch tracking */
+                rd_ts_t ts_last_share_poll_start;
+                rd_ts_t ts_share_poll_start;
+                rd_ts_t time_since_last_share_poll;
```
ojasvajain: Can we have them inside a sub struct? 
I am thinking a sub struct named `rk_share_poll`. You can also drop the `share` from the variable names.

src/rdkafka_int.h
```diff
@@ -746,22 +746,39 @@ struct rd_kafka_s {
 
                 struct {
                         rd_avg_t rk_avg_poll_idle_ratio;
-                        rd_avg_t rk_avg_commit_latency; /**< Current commit
-                                                         *   latency avg */
+                        rd_avg_t rk_avg_share_poll_idle_ratio;
+                        rd_avg_t
+                            rk_avg_share_time_between_poll; /**< Current time
```
ojasvajain, line 825: Keep the share consumer related metrics below the regular consumer metrics. Currently they are in between.

src/rdkafka_int.h
```diff
@@ -746,22 +746,39 @@ struct rd_kafka_s {
...
                         rd_avg_t
                             rk_avg_rebalance_latency; /**< Current rebalance
                                                        *   latency avg */
                 } rd_avg_current;
 
                 struct {
                         rd_avg_t rk_avg_poll_idle_ratio;
+                        rd_avg_t rk_avg_share_poll_idle_ratio;
+                        rd_avg_t
+                            rk_avg_share_time_between_poll; /**< Rolled over
```
ojasvajain, line 840: [Same](https://github.com/confluentinc/librdkafka/pull/5435/changes#r3343347407)

src/rdkafka.c
```diff
@@ -2740,28 +2752,49 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
...
-                    rk->rk_conf.enable_metrics_push);
-                rd_avg_init(
-                    &rk->rk_telemetry.rd_avg_rollover.rk_avg_commit_latency,
-                    RD_AVG_GAUGE, 0, 500 * 1000, 2,
-                    rk->rk_conf.enable_metrics_push);
-                rd_avg_init(
-                    &rk->rk_telemetry.rd_avg_current.rk_avg_commit_latency,
-                    RD_AVG_GAUGE, 0, 500 * 1000, 2,
-                    rk->rk_conf.enable_metrics_push);
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
```
ojasvajain, line 2803: TODO on me: check exp_min and exp_max values

src/rdkafka.c
ojasvajain: Please add the comment.

src/rdkafka.c
```diff
@@ -3763,6 +3851,8 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
         rd_kafka_q_serve(rk->rk_rep, RD_POLL_NOWAIT, 0, RD_KAFKA_Q_CB_CALLBACK,
                          rd_kafka_poll_cb, NULL);
 
+        rd_kafka_share_record_poll_end(rk);
```
ojasvajain, line 4057: We have placed this after the serving the callback. Callback is user configured so if the callback fn is time-taking, it will inflate the poll duration as well. Did we check if Java accounts for it?

ojasvajain: Same for the start call, hope we are following Java's approach.

src/rdkafka_telemetry_encode.c
```diff
@@ -285,6 +285,38 @@ calculate_consumer_poll_idle_ratio_avg(rd_kafka_t *rk,
...
+static rd_kafka_telemetry_metric_value_t
+calculate_share_time_between_poll_max(rd_kafka_t *rk,
+                                      rd_kafka_broker_t *rkb_selected,
+                                      rd_ts_t now_ns) {
+        rd_kafka_telemetry_metric_value_t time_between_poll_max;
+        time_between_poll_max.int_value = calculate_max(
+            rk->rk_telemetry.rd_avg_rollover.rk_avg_share_time_between_poll,
+            THREE_ORDERS_MAGNITUDE);
+        return time_between_poll_max;
+}
```
ojasvajain, line 340: Let's move these functions below the regular consumer related functions. You can move them below `calculate_consumer_commit_latency_max`

src/rdkafka_telemetry_encode.h
```diff
@@ -296,6 +307,38 @@ static const rd_kafka_telemetry_metric_info_t RD_KAFKA_TELEMETRY_CONSUMER_METRIC
              .type          = RD_KAFKA_TELEMETRY_METRIC_TYPE_GAUGE},
 };
 
+static const rd_kafka_telemetry_metric_info_t
+    RD_KAFKA_TELEMETRY_SHARE_CONSUMER_METRICS_INFO
+        [RD_KAFKA_TELEMETRY_SHARE_CONSUMER_METRIC__CNT] = {
+            [RD_KAFKA_TELEMETRY_METRIC_SHARE_CONSUMER_POLL_IDLE_RATIO_AVG] =
+                {.name          = "consumer.share.poll.idle.ratio.avg",
+                 .description   = "The average fraction of time the consumer's "
+                                  "poll() is idle as opposed to waiting for the "
```
ojasvajain, line 316: Instead of poll(), should we use `consume_batch()`?

src/rdkafka_telemetry_encode.h
```diff
@@ -296,6 +307,38 @@ static const rd_kafka_telemetry_metric_info_t RD_KAFKA_TELEMETRY_CONSUMER_METRIC
...
+                                  "user code to process records.",
+                 .unit          = "1",
+                 .is_int        = rd_false,
+                 .is_per_broker = rd_false,
+                 .type          = RD_KAFKA_TELEMETRY_METRIC_TYPE_GAUGE},
+            [RD_KAFKA_TELEMETRY_METRIC_SHARE_CONSUMER_TIME_BETWEEN_POLL_AVG] =
+                {.name = "consumer.share.time.between.poll.avg",
+                 .description =
+                     "The average delay between invocations of poll() in "
```
ojasvajain, line 325: Same

src/rdkafka_telemetry_encode.h
```diff
@@ -296,6 +307,38 @@ static const rd_kafka_telemetry_metric_info_t RD_KAFKA_TELEMETRY_CONSUMER_METRIC
...
+                     "milliseconds.",
+                 .unit          = "ms",
+                 .is_int        = rd_false,
+                 .is_per_broker = rd_false,
+                 .type          = RD_KAFKA_TELEMETRY_METRIC_TYPE_GAUGE},
+            [RD_KAFKA_TELEMETRY_METRIC_SHARE_CONSUMER_TIME_BETWEEN_POLL_MAX] =
+                {.name = "consumer.share.time.between.poll.max",
+                 .description =
+                     "The max delay between invocations of poll() in "
```
ojasvajain, line 334: Same

src/rdkafka_telemetry_encode.h
ojasvajain: Can we check the existing metrics to see the convention we follow?

ojasvajain (commented):
I see some scope of using helpers to init, destroy and rollover. Will discuss it with you.

src/rdkafka.c
```diff
@@ -1056,18 +1056,30 @@ void rd_kafka_destroy_final(rd_kafka_t *rk) {
...
-                    &rk->rk_telemetry.rd_avg_current.rk_avg_commit_latency);
-                rd_avg_destroy(
-                    &rk->rk_telemetry.rd_avg_rollover.rk_avg_poll_idle_ratio);
-                rd_avg_destroy(
-                    &rk->rk_telemetry.rd_avg_rollover.rk_avg_rebalance_latency);
-                rd_avg_destroy(
-                    &rk->rk_telemetry.rd_avg_rollover.rk_avg_commit_latency);
+
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
+                        rd_avg_destroy(&rk->rk_telemetry.rd_avg_current
```
ojasvajain, line 1071: Please add a TODO to create a helper that destroys both current average and rollover average variables.

src/rdkafka.c
```diff
@@ -2740,28 +2752,49 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
...
-                rd_avg_init(
-                    &rk->rk_telemetry.rd_avg_current.rk_avg_commit_latency,
-                    RD_AVG_GAUGE, 0, 500 * 1000, 2,
-                    rk->rk_conf.enable_metrics_push);
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
+                        rd_avg_init(&rk->rk_telemetry.rd_avg_rollover
+                                         .rk_avg_share_poll_idle_ratio,
+                                    RD_AVG_GAUGE, 0, 1000 * 1000, 2,
+                                    rk->rk_conf.enable_metrics_push);
+                        rd_avg_init(&rk->rk_telemetry.rd_avg_current
```
ojasvajain, line 2808: Same as before. Please add a TODO to create a helper that inits both current avg and rollover vars.

src/rdkafka_telemetry_encode.c
```diff
@@ -796,23 +840,39 @@ rd_buf_t *rd_kafka_telemetry_encode_metrics(rd_kafka_t *rk) {
...
-
-                rd_avg_destroy(
-                    &rk->rk_telemetry.rd_avg_rollover.rk_avg_commit_latency);
-                rd_avg_rollover(
-                    &rk->rk_telemetry.rd_avg_rollover.rk_avg_commit_latency,
-                    &rk->rk_telemetry.rd_avg_current.rk_avg_commit_latency);
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
+                        rd_avg_destroy(&rk->rk_telemetry.rd_avg_rollover
+                                            .rk_avg_share_poll_idle_ratio);
+                        rd_avg_rollover(&rk->rk_telemetry.rd_avg_rollover
```
ojasvajain, line 849: Please add a TODO to create a helper that does both destroy and rollover

src/rdkafka.c
ojasvajain: Discussed with Ankith. The placement is discussed with Pranav. I will double check and discuss my doubt with Pranav.

src/rdkafka_telemetry_encode.h
ojasvajain: Discussed with Ankith, we are following the KIP so it should be fine to use `poll`

src/rdkafka_int.h
Ankith-Confluent: Sure, will add

src/rdkafka_int.h
Ankith-Confluent: Sure

src/rdkafka.c
Ankith-Confluent: Check with Java implementation, they include callback time too. Its a part of consume batch so I think it makes sense to include it.

src/rdkafka_telemetry_encode.c
Ankith-Confluent: Sure

src/rdkafka.c
Ankith-Confluent: Sure

src/rdkafka.c
Ankith-Confluent: Sure

src/rdkafka_telemetry_encode.c
Ankith-Confluent: Sure

src/rdkafka_int.h
```diff
@@ -1246,7 +1265,7 @@ static RD_INLINE RD_UNUSED void rd_kafka_app_poll_start(rd_kafka_t *rk,
                 now = rd_clock();
         if (is_blocking)
                 rd_atomic64_set(&rk->rk_ts_last_poll, INT64_MAX);
-        if (rkq->rkq_ts_last_poll_end) {
+        if (rkq->rkq_ts_last_poll_end && !RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
```
ojasvajain: Add this on line 1261. Basically we want to ensure 
`rd_kafka_app_poll_start` and rd_kafka_app_polled` is excluded for share consumers.

src/rdkafka_int.h
ojasvajain: Something like this:

```suggestion
        if (rk->rk_type != RD_KAFKA_CONSUMER || RD_KAFKA_IS_SHARE_CONSUMER(rk))
```

ojasvajain: Same exclusion here. This should not run for share consumers.

src/rdkafka.c
```diff
@@ -3698,6 +3737,61 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
...
+ */
+static void rd_kafka_share_record_poll_start(rd_kafka_t *rk) {
+        rd_ts_t now;
+
+        if (!rk->rk_conf.enable_metrics_push)
+                return;
+
+        now                                          = rd_clock();
+        rk->rk_telemetry.rk_share_poll.ts_poll_start = now;
+
```
ojasvajain, line 3915: We need to ensure that `max.poll.interval.ms` works for share consumer. If you see `rd_kafka_max_poll_exceeded` method, it only relies on `rk_ts_last_poll` value to determine if interval is exceeded or not.

`rk_ts_last_poll` was getting set in `rd_kafka_app_poll_start` and `rd_kafka_app_polled` but we don't want to run those functions for share consumers (as suggested in my comments above). 

So to ensure max.poll.interval.ms works as expected, add a line here to set `rk_ts_last_poll` to INT64_MAX (`rd_kafka_app_poll_start` also does it). No `if (is_blocking)` condition required here.

src/rdkafka.c
```diff
@@ -3698,6 +3737,61 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
...
+ *        pollTime / (pollTime + timeSinceLastPoll), and records the sample.
+ *        Mirrors Java's share consumer poll idle ratio formula.
+ */
+static void rd_kafka_share_record_poll_end(rd_kafka_t *rk) {
+        rd_ts_t end, poll_time, total;
+
+        if (!rk->rk_conf.enable_metrics_push)
+                return;
+
+        end       = rd_clock();
```
ojasvajain: From `rd_kafka_app_polled`, add this block here:

```
rd_atomic64_set(&rk->rk_ts_last_poll, end);
                if (unlikely(rk->rk_cgrp &&
                             rk->rk_cgrp->rkcg_group_protocol ==
                                 RD_KAFKA_GROUP_PROTOCOL_CONSUMER &&
                             rk->rk_cgrp->rkcg_flags &
                                 RD_KAFKA_CGRP_F_MAX_POLL_EXCEEDED)) {
                        rd_kafka_cgrp_consumer_expedite_next_heartbeat(
                            rk->rk_cgrp,
                            "app polled after poll interval exceeded");
                }
```
The first line sets the `rk_ts_last_poll` value which is used by `rd_kafka_max_poll_exceeded`.

The if block is used when max polling interval is already exceeded.

ojasvajain (commented):
Left some comments to ensure max poll interval check works with share consumer.

src/rdkafka_telemetry_decode.c
```diff
@@ -561,13 +568,25 @@ int unit_test_telemetry(rd_kafka_type_t rk_type,
 
         rd_avg_init(&rk->rk_telemetry.rd_avg_current.rk_avg_poll_idle_ratio,
                     RD_AVG_GAUGE, 0, 500 * 1000, 2, rd_true);
+        rd_avg_init(
```
k-raina, line 571: Are these getting destroyed in the teardown?

src/rdkafka_broker.c
k-raina, line 1953: Do we need check here 
"""
                          if (rkb->rkb_rk->rk_type == RD_KAFKA_CONSUMER &&
                              !RD_KAFKA_IS_SHARE_CONSUMER(rkb->rkb_rk))
"""

src/rdkafka_int.h
Ankith-Confluent: Sure

src/rdkafka_int.h
Ankith-Confluent: Sure

src/rdkafka.c
Ankith-Confluent: Sure

src/rdkafka.c
Ankith-Confluent: Done

src/rdkafka_telemetry_decode.c
Ankith-Confluent: I think I missed this in the unit test, I will add

src/rdkafka_broker.c
Ankith-Confluent: Share consumer cannot reach here I think, so it wouldn't be required.

src/rdkafka_int.h
```diff
@@ -1316,14 +1335,14 @@ static RD_INLINE RD_UNUSED void rd_kafka_app_poll_start(rd_kafka_t *rk,
...
-        if (rk->rk_type != RD_KAFKA_CONSUMER)
+        if (rk->rk_type != RD_KAFKA_CONSUMER || RD_KAFKA_IS_SHARE_CONSUMER(rk))
                 return;
 
         if (!now)
                 now = rd_clock();
         if (is_blocking)
                 rd_atomic64_set(&rk->rk_ts_last_poll, INT64_MAX);
-        if (rkq->rkq_ts_last_poll_end) {
+        if (rkq->rkq_ts_last_poll_end && !RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
```
ojasvajain: You can remove this now. It is not required.

src/rdkafka.c
```diff
@@ -3831,6 +3891,83 @@ rd_kafka_share_consumer_closed_err(rd_kafka_share_t *rkshare) {
...
+        rd_ts_t end, poll_time, total;
+
+        end = rd_clock();
+
+        /* max.poll.interval.ms enforcement: record the poll time and, if the
+         * interval had been exceeded, expedite the next heartbeat to rejoin. */
+        rd_atomic64_set(&rk->rk_ts_last_poll, end);
+        if (unlikely(rk->rk_cgrp &&
+                     rk->rk_cgrp->rkcg_group_protocol ==
+                         RD_KAFKA_GROUP_PROTOCOL_CONSUMER &&
```
ojasvajain: The condition `rk->rk_cgrp->rkcg_group_protocol ==
                      RD_KAFKA_GROUP_PROTOCOL_CONSUMER`

is not required as in case of share consumers it will always be consumer protocol.

tests/test.c
```diff
@@ -598,6 +600,8 @@ struct test tests[] = {
     _TEST(0181_share_consumer_topic_delete, 0, TEST_BRKVER(0, 4, 2, 0)),
     _TEST(0182_share_consumer_error_handling_mock, TEST_F_LOCAL),
     _TEST(0183_share_consumer_leader_change_mock, TEST_F_LOCAL),
+    _TEST(0185_share_consumer_max_poll_interval, 0, TEST_BRKVER(0, 4, 0, 0)),
```
```suggestion
    _TEST(0185_share_consumer_max_poll_interval, 0, TEST_BRKVER(0, 4, 2, 0)),
```

tests/0185-share_consumer_max_poll_interval.c
```diff
@@ -0,0 +1,730 @@
...
+                    test_mk_topic_name("0185_share_max_poll", 1));
+        test_str_id_generate(group, sizeof(group));
+
+        test_create_topic_wait_exists(NULL, topic, 1, 1, 5000);
+        if (msgcnt > 0)
+                test_produce_msgs_easy(topic, testid, 0, msgcnt);
+        test_share_set_auto_offset_reset(group, "earliest");
+
+        c = create_share_consumer(NULL, group, max_poll_interval_ms);
+        subscribe(c, topic);
```
ojasvajain: Possible to use `test_consumer_subscribe` in test.c?

tests/0185-share_consumer_max_poll_interval.c
```diff
@@ -0,0 +1,730 @@
...
+        TEST_SAY("Blocking in a single consume_batch() for %ds (> %dms)\n",
+                 (REAL_MAX_POLL_INTERVAL_MS + 2000) / 1000,
+                 REAL_MAX_POLL_INTERVAL_MS);
+        share_consume_once(c, REAL_MAX_POLL_INTERVAL_MS + 2000, &exceeded);
+
+        TEST_ASSERT(!exceeded,
+                    "A blocking consume_batch() must not trip "
+                    "max.poll.interval.ms");
+
+        /* And a subsequent poll should still be healthy. */
```
ojasvajain: Also check whether the consumer is able to consume msgs? For that, you can produce some msgs before the share_consume_once call.

tests/0185-share_consumer_max_poll_interval.c
```diff
@@ -0,0 +1,730 @@
...
+#define MOCK_MAX_POLL_INTERVAL_MS 3000
+/* Longer interval for the real-broker tests (matches 0089/0091), giving the
+ * group join/assignment enough headroom not to falsely trip. */
+#define REAL_MAX_POLL_INTERVAL_MS 10000
+
+
+/**
+ * @brief Enable the three share APIs on the mock cluster.
+ */
+static void enable_share_apis(rd_kafka_mock_cluster_t *mcluster) {
```
ojasvajain: This method is not required.

tests/0185-share_consumer_max_poll_interval.c
```diff
@@ -0,0 +1,730 @@
...
+                    errstr);
+
+        return rkshare;
+}
+
+
+/**
+ * @brief Subscribe a share consumer to a single topic.
+ */
+static void subscribe(rd_kafka_share_t *rkshare, const char *topic) {
```
ojasvajain: There is a helper in test.c for this. Check if you can use that.

tests/0185-share_consumer_max_poll_interval.c
```diff
@@ -0,0 +1,730 @@
...
+        TEST_ASSERT(!exceeded,
+                    "Consumer should remain in the group after a long "
+                    "blocking poll");
+
+        test_share_consumer_close(c);
+        test_share_destroy(c);
+        test_mock_cluster_destroy(mcluster);
+
+        SUB_TEST_PASS();
+}
```
ojasvajain: I don't see any mock behaviour injected like request delay or errors. Why do we need these test cases?

tests/0185-share_consumer_max_poll_interval.c
Ankith-Confluent: This is a regular consumer function right, it takes `rd_kafka_t`

tests/0185-share_consumer_max_poll_interval.c
Ankith-Confluent: Yes sure

tests/0185-share_consumer_max_poll_interval.c
Ankith-Confluent: Had a discussion about this, we dont need local versions of the tests, will be removing them

ojasvajain (approved):
Thanks @Ankith-Confluent 
Approved!

PR #5436, opened 2026-05-11
[KIP-932] Preserve pre-set INVALID_SHARE_SESSION_EPOCH on strip path 

pranavrth:


src/rdkafka_broker.c
```diff
@@ -3606,7 +3606,7 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                         rd_kafka_dbg(rkb->rkb_rk, BROKER, "SHAREFETCH",
                                      "Ignoring SHARE_FETCH op: "
                                      "instance or broker is terminating");
-                        rd_kafka_share_fetch_op_reply_with_err(
```
pranavrth, line 3609: Though I have changed this name as @PratRanj07 had asked in the previous PR but I think the older name was better. What do you think @PratRanj07?

tests/0182-share_consumer_error_handling_mock.c
```diff
@@ -861,20 +866,23 @@ static void test_consume_batch_at_epoch_zero_strips_piggyback_acks(void) {
...
+         * returns the full set. */
+        while (rcvd == 0 && attempts++ < 30) {
+                error = rd_kafka_share_consume_batch(rkshare, 3000, rkmessages,
+                                                     &rcvd);
+                if (error)
                         rd_kafka_error_destroy(error);
-                        continue;
-                }
-                total_consumed += (int)rcvd;
         }
```
PratRanj07, line 881: We should acknowledge records on the fly as the consumer is in explicit mode

tests/0182-share_consumer_error_handling_mock.c
```diff
@@ -984,15 +987,218 @@ static void test_consume_batch_at_epoch_zero_strips_piggyback_acks(void) {
...
+         * to be acknowledged before the next call, so we must not loop
+         * and accumulate without acking each batch first. msgcnt is far
+         * below max.poll.records (default 500), so a non-empty call
+         * returns the full set. */
+        while (rcvd == 0 && attempts++ < 30) {
+                error = rd_kafka_share_consume_batch(rkshare, 3000, rkmessages,
+                                                     &rcvd);
+                if (error)
+                        rd_kafka_error_destroy(error);
+        }
```
PratRanj07, line 1091: Same as [here](https://github.com/confluentinc/librdkafka/pull/5436/changes/e3383f93cdaaefd2d058d03a53632ac6ab2c753c#r3227254565)

src/rdkafka_broker.c
PratRanj07: The older name doesnt capture the essence that we are actually setting the err here in ack_details which shouldn't be expected by a normal op function, by name it meant like we are setting err on a share fetch reply op

PratRanj07 (commented):
Thanks @pranavrth for the PR! Some test improvem,ents!
Rest all looks good

tests/0182-share_consumer_error_handling_mock.c
pranavrth: I don't think this is needed. I am getting first batch of records and then breaking the loop. Check this condition

```
while (rcvd == 0 && attempts++ < 30) {
```

tests/0182-share_consumer_error_handling_mock.c
pranavrth: Same - https://github.com/confluentinc/librdkafka/pull/5436/changes#r3228011273

PratRanj07 (approved):

PR #5438, opened 2026-05-12
[KIP-932] Runtime acknowledgement callback for share consumer

PratRanj07:
Replace the config-based share acknowledgement callback with a runtime
API and add a single-thread access gate that all public share consumer
APIs route through.

Runtime acknowledgement callback
- Add rd_kafka_share_set_acknowledgement_cb() so the callback can be
  installed or swapped at any time. Lookup happens at invoke time, not
  at enqueue time, so an ack op enqueued before a callback swap will
  fire on the currently-registered callback (matches Java
  KafkaShareConsumer).
- Move the callback, opaque, in_callback flag, and registered flag onto
  rk_share_consumer in rd_kafka_s. The app-thread-owned fields (cb,
  opaque, in_callback) and the main-thread-owned field
  (acknowledgement_callback_registered) are each accessed from a single
  thread, so no locks are needed.
- Cross-thread registration is communicated via the new
  RD_KAFKA_OP_SHARE_ACK_COMMIT_CB_REGISTER op, only enqueued on
  registration state transitions (set <-> unset). The old
  RD_KAFKA_OP_SHARE_ACK_COMMIT_CB is split into _EXECUTE (main -> app)
  and _REGISTER (app -> main); the callback/opaque fields are removed
  from the execute op since they are now looked up from rk_share_consumer.
- Delete the conf entry, conf field, and
  rd_kafka_conf_set_share_acknowledgement_commit_cb() setter, along with
  its CONFIGURATION.md row. Mirror the Java behavior in which there is
  one and only one way to install the callback.

Share consumer access gate
- Add rd_kafka_share_acquire() / rd_kafka_share_release(). acquire()
  rejects:
    * NULL rkshare,
    * calls made from inside the acknowledgement callback
      (in_callback set), and
    * concurrent calls from another application thread (atomic
      test-and-set gate already held),
  all with RD_KAFKA_RESP_ERR__STATE. release() tolerates NULL so error
  paths can call it unconditionally.
- Wrap every public share API (consume_batch, commit_async, commit_sync,
  acknowledge, acknowledge_type, acknowledge_offset, subscribe,
  unsubscribe, subscription, consumer_close, consumer_close_queue,
  set_acknowledgement_cb) in acquire/release. Internal helpers expose
  acknowledge_offset0 so acknowledge_type and acknowledge_offset can
  share validation/inflight-map logic without re-entering the gate.
- The poll_cb handler for SHARE_ACK_COMMIT_CB_EXECUTE sets
  in_callback around the user callback so any share API called from
  within is rejected by acquire().

Tests are updated to install the callback through the new runtime API
and to exercise reentrancy rejection and dynamic callback swaps.


src/rdkafka.c
```diff
@@ -3062,6 +3062,12 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
         rkshare->rkshare_unacked_cnt      = 0;
         rkshare->rkshare_consumer_closing = rd_false;
 
+        rk->rk_share_consumer.acknowledgement_cb                  = NULL;
```
pranavrth: Add a TODO to change this to `rk_share` when the interface is finalized.

src/rdkafka.c
```diff
@@ -3075,6 +3081,88 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
 }
 
 
+rd_kafka_resp_err_t rd_kafka_share_acquire(rd_kafka_share_t *rkshare) {
```
pranavrth: 1. We should return error_t instead of err_t
2. Do exactly what Java does here.

src/rdkafka.c
```diff
@@ -5605,13 +5731,22 @@ rd_kafka_op_res_t rd_kafka_poll_cb(rd_kafka_t *rk,
...
                 break;
 
-        case RD_KAFKA_OP_SHARE_ACK_COMMIT_CB:
-                if (!rko->rko_u.share_ack_commit.cb)
+        case RD_KAFKA_OP_SHARE_ACK_COMMIT_CB_EXECUTE: {
+                rd_kafka_share_t *rkshare = rk->rk_rkshare;
+                /* Lookup callback at invoke time.
+                 * Locality: app thread */
+                if (!rkshare ||
+                    !rk->rk_share_consumer.acknowledgement_callback_registered)
```
pranavrth: This flag is main thread only. Why are we checking it here. Use direct callback pointer instead.

src/rdkafka.c
```diff
@@ -3019,6 +3037,88 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+    void *opaque) {
+        rd_kafka_t *rk;
+        rd_bool_t was_set, now_set;
+        rd_kafka_resp_err_t err;
+
+        if (unlikely((err = rd_kafka_share_acquire(rkshare))))
+                return err;
+
+        rk = rkshare->rkshare_rk;
+
```
pranavrth: We need close check here as well.

src/rdkafka.c
pranavrth: Please check both of these comments as well.

We need let a single thread pass multiple times. Python can call `rd_kafka_share_destroy` from the callback. In that case, java way will work but our way won't.

Let's discuss more if you have any question.

pranavrth (changes requested):

src/rdkafka.c
PratRanj07: Java throws an exception here. Concurrent Modification Exception
It is kept as rd_kafka_resp_err_t based on what our apis throw back to the user, 6 of our apis throw back rd_kafka_resp_err_t and 5 throw back error_t. 
So I decided to keep it as rd_kafka_resp_err_t and then we throw an error object from the other 5 apis

PratRanj07: I think we should return RD_KAFKA_RESP_ERR__CONFLICT in case of error and for each site where we are giving out rd_kafka_error_t we should return a function specific error string rather than a generic error string. WDYT?

src/rdkafka.c
```diff
@@ -3019,6 +3076,121 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+        /* Reject calls from inside the acknowledgement callback *before*
+         * claiming the gate. The callback runs on the owning thread, which the
+         * owner check below would otherwise let re-enter (bumping depth);
+         * rejecting after the claim would leak that depth count. */
+        if (unlikely(rk->rk_share_consumer.in_callback))
+                return RD_KAFKA_RESP_ERR__STATE;
+
+        mtx_lock(&rk->rk_share_consumer.guard_lock);
```
pranavrth: We don't want to introduce lock here. We just need a lightweight atomic variable.

pranavrth (changes requested):
I think we should divide this PR into 2 parts. 1 ack callback and another atomic lock. Let's discuss this over call.

src/rdkafka.c
pranavrth: __CONFLICT seems fine imo.

src/rdkafka.c
```diff
@@ -3019,6 +3076,126 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+    void *opaque) {
+        rd_kafka_t *rk;
+        rd_bool_t was_set, now_set;
+        rd_kafka_resp_err_t err;
+        rd_kafka_error_t *error;
+
+        if (unlikely((err = rd_kafka_share_acquire(rkshare))))
+                return err;
+
+        if (unlikely((error = rd_kafka_share_consumer_closed_error(rkshare)) !=
```
pranavrth, line 3173: error is leaking here.

src/rdkafka_int.h
```diff
@@ -345,6 +345,58 @@ struct rd_kafka_s {
...
+                 *  @c owner is a real per-thread identifier (from
+                 *  thrd_current_id()) rather than a @c thrd_t handle. */
+                mtx_t guard_lock;    /**< Protects owner/owned/depth. */
+                unsigned long owner; /**< Owning thread id while @c owned. */
+                rd_bool_t owned;     /**< Gate currently held. */
+                int depth;           /**< Re-entrancy count for the owner. */
+        } rk_share_consumer;
+        /**
+         * TODO KIP-932: Change to rk_share when interface is finalized.
+         */
```
pranavrth: TODO should be at the start instead of here.

src/rdkafka.c
```diff
@@ -1274,20 +1275,56 @@ void rd_kafka_destroy(rd_kafka_t *rk) {
...
+        if (unlikely((acq_err = rd_kafka_share_acquire(rkshare)) !=
+                     RD_KAFKA_RESP_ERR_NO_ERROR))
+                return rd_kafka_error_new(
+                    acq_err,
+                    "rd_kafka_share_destroy() called while another thread is "
+                    "using the share consumer; instance not destroyed");
+
         rd_kafka_destroy(rkshare->rkshare_rk);
+        return NULL;
 }
```
pranavrth, line 1300: Let's rethink if we need to gate this flow as well.

pranavrth (changes requested):

src/rdkafka.c
```diff
@@ -3019,6 +3065,142 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+                rd_kafka_dbg(
+                    rk, CGRP, "SHAREACK", "Share ack callback %sregistered",
+                    rko->rko_u.share_ack_commit_cb_register.registered ? ""
+                                                                       : "un");
+        }
+        return RD_KAFKA_OP_RES_HANDLED;
+}
+
+
+rd_kafka_resp_err_t rd_kafka_share_set_acknowledgement_cb(
```
pranavrth: We should return `error_t` itself.

src/rdkafka.h
```diff
@@ -2246,42 +2246,39 @@ RD_EXPORT size_t rd_kafka_share_partition_offsets_offsets_cnt(
...
- * @remark The \p partitions list and its contents are only valid for the
- *         duration of the callback.
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success, or:
+ *          - RD_KAFKA_RESP_ERR__STATE if called from within the callback.
  */
-RD_EXPORT
-void rd_kafka_conf_set_share_acknowledgement_commit_cb(
-    rd_kafka_conf_t *conf,
-    void (*share_acknowledgement_commit_cb)(
+RD_EXPORT rd_kafka_resp_err_t rd_kafka_share_set_acknowledgement_cb(
```
pranavrth: Use `error_t`.

src/rdkafka.h
```diff
@@ -2246,42 +2246,39 @@ RD_EXPORT size_t rd_kafka_share_partition_offsets_offsets_cnt(
...
+    rd_kafka_share_t *rkshare,
+    void (*acknowledgement_cb)(
```
pranavrth: `share_acknowledgement_commit_cb`

src/rdkafka.h
pranavrth: rd_kafka_share_set_acknowledgement_commit_cb

src/rdkafka.c
```diff
@@ -3006,6 +3045,13 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
         rkshare->rkshare_unacked_cnt      = 0;
         rkshare->rkshare_consumer_closing = rd_false;
 
+        rk->rk_share_consumer.acknowledgement_cb                  = NULL;
+        rk->rk_share_consumer.acknowledgement_opaque              = NULL;
```
pranavrth: acknowledgement_commit_cb and acknowledgement_commit_cb_opaque

src/rdkafka.c
```diff
@@ -3006,6 +3045,13 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+        rk->rk_share_consumer.in_callback                         = rd_false;
+        rk->rk_share_consumer.acknowledgement_callback_registered = rd_false;
```
pranavrth: `acknowledgement_commit_cb_registered`

src/rdkafka_int.h
```diff
@@ -1335,6 +1395,43 @@ rd_kafka_share_consumer_closed_error(rd_kafka_share_t *rkshare);
...
+
+/**
+ * @brief Sentinel value stored in rk_share_consumer.owner when the
+ *        gate is unowned. Chosen to never collide with a real
+ *        thread id on any supported platform. A zero-initialised struct
+ *        does NOT look unowned, which is intentional: a missing
+ *        init causes acquire to fail with __CONFLICT rather than
+ *        silently letting two threads pass.
+ */
+#define RD_KAFKA_SHARE_NO_OWNER ((int64_t) - 1)
```
pranavrth: - Remove space b/w `-` and `1`
- `RD_KAFKA_SHARE_NO_CURRENT_THREAD`

src/rdkafka.c
```diff
@@ -3006,6 +3045,13 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+        rd_atomic64_init(&rk->rk_share_consumer.owner, RD_KAFKA_SHARE_NO_OWNER);
```
pranavrth: `current_owner_thread`

src/rdkafka.c
```diff
@@ -3006,6 +3045,13 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+        rd_atomic32_init(&rk->rk_share_consumer.depth, 0);
```
pranavrth: `ref_count` instead of `depth`

src/rdkafka.c
```diff
@@ -3019,6 +3065,142 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+                rd_kafka_error_destroy(acq_err);
+                return err;
+        }
+
+        if (unlikely((error = rd_kafka_share_consumer_closed_error(rkshare)) !=
+                     NULL)) {
+                err = rd_kafka_error_code(error);
+                rd_kafka_error_destroy(error);
+                return err;
+        }
```
pranavrth: Should release the acquired lock above.
Use `goto done:`

src/rdkafka.c
```diff
@@ -3019,6 +3065,142 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+         * since the registered flag is already true on the main thread. */
+        if (was_set != now_set) {
+                rd_kafka_op_t *rko = rd_kafka_op_new_cb(
+                    rk, RD_KAFKA_OP_SHARE_ACK_COMMIT_CB_REGISTER,
+                    rd_kafka_share_ack_commit_cb_register_op);
+                rko->rko_u.share_ack_commit_cb_register.registered = now_set;
+                rd_kafka_q_enq(rk->rk_ops, rko);
+        }
+
+        rd_kafka_share_release(rkshare);
```
pranavrth, line 3200: goto done.

src/rdkafka.c
pranavrth: Initialize error to NULL for each of the share functions.

src/rdkafka.c
```diff
@@ -4913,21 +5113,25 @@ rd_kafka_error_t *rd_kafka_consumer_close_queue(rd_kafka_t *rk,
  */
 rd_kafka_error_t *rd_kafka_share_consumer_close_queue(rd_kafka_share_t *rkshare,
                                                       rd_kafka_queue_t *rkqu) {
-        rd_kafka_error_t *error = NULL;
+        rd_kafka_error_t *error;
```
pranavrth: initialize to null

src/rdkafka.c
```diff
@@ -4913,21 +5113,25 @@ rd_kafka_error_t *rd_kafka_consumer_close_queue(rd_kafka_t *rk,
...
         if (unlikely((error = rd_kafka_share_consumer_closed_error(rkshare)) !=
                      NULL))
-                return error;
+                goto done;
 
         rkshare->rkshare_consumer_closing = rd_true;
         error = rd_kafka_consumer_close_queue(rkshare->rkshare_rk, rkqu);
-        if (error) {
+        if (error)
                 rkshare->rkshare_consumer_closing = rd_false;
```
pranavrth, line 5131: goto done.

src/rdkafka_share_acknowledgement.c
```diff
@@ -348,33 +348,58 @@ rd_kafka_share_find_ack_batch_by_id(rd_list_t *ack_list,
...
+                         * misaligned read. Override every batch with
+                         * INVALID_RECORD_STATE — same treatment as a
+                         * partition omitted from a well-formed response —
+                         * so the consumer treats these as "outcome
+                         * unknown" and re-acquires. */
+                        RD_LIST_FOREACH(batch,
+                                        rko->rko_u.share_fetch.ack_details, i) {
+                                batch->rktpar->err =
+                                    RD_KAFKA_RESP_ERR_INVALID_RECORD_STATE;
+                        }
```
pranavrth: Keep same error as what is received.

src/rdkafka_share_acknowledgement.c
```diff
@@ -348,33 +348,58 @@ rd_kafka_share_find_ack_batch_by_id(rd_list_t *ack_list,
...
         if (err && rko->rko_u.share_fetch.ack_details) {
                 rd_kafka_share_ack_batches_t *batch;
                 int i;
-                RD_LIST_FOREACH(batch, rko->rko_u.share_fetch.ack_details, i) {
-                        if (batch->rktpar->err ==
-                            RD_KAFKA_RESP_ERR__IN_PROGRESS)
-                                batch->rktpar->err = err;
+
+                if (err == RD_KAFKA_RESP_ERR__BAD_MSG ||
+                    err == RD_KAFKA_RESP_ERR__UNDERFLOW) {
```
pranavrth: unlikely.

src/rdatomic.h
```diff
@@ -227,4 +227,47 @@ static RD_INLINE int64_t RD_UNUSED rd_atomic64_set(rd_atomic64_t *ra,
 #endif
 }
 
+/**
+ * @brief Atomic compare-and-swap on a 64-bit value.
+ *
+ *        If *ra == @p expected, atomically replace it with @p desired
+ *        and return 1. Otherwise leave *ra unchanged and return 0.
+ */
+static RD_INLINE int RD_UNUSED rd_atomic64_cas(rd_atomic64_t *ra,
```
pranavrth, line 236: Verify that type of thread id is correct across all the architecture.

src/rdkafka_int.h
```diff
@@ -1335,6 +1395,43 @@ rd_kafka_share_consumer_closed_error(rd_kafka_share_t *rkshare);
...
+ */
+rd_kafka_error_t *rd_kafka_share_acquire(rd_kafka_share_t *rkshare);
+
+/**
+ * @brief Release ownership acquired by rd_kafka_share_acquire().
+ *
+ * Tolerates NULL rkshare (no-op) so error paths can call it
+ * unconditionally.
+ */
+void rd_kafka_share_release(rd_kafka_share_t *rkshare);
```
pranavrth, line 1438: Add rd_dassert to make sure that ref count is >0.

src/rdkafka_int.h
```diff
@@ -1335,6 +1395,43 @@ rd_kafka_share_consumer_closed_error(rd_kafka_share_t *rkshare);
...
+ *
+ * @returns NULL on success, or a newly-allocated rd_kafka_error_t (owned
+ *          by the caller) on rejection:
+ *           - RD_KAFKA_RESP_ERR__STATE on NULL rkshare, or when called
+ *             from inside the user-defined acknowledgement commit
+ *             callback.
+ *           - RD_KAFKA_RESP_ERR__CONFLICT on concurrent entry from a
+ *             different application thread.
+ */
+rd_kafka_error_t *rd_kafka_share_acquire(rd_kafka_share_t *rkshare);
```
pranavrth, line 1430: rd_dassert for ref count to be >= 0

src/rdkafka_subscription.c
```diff
@@ -48,14 +48,23 @@ rd_kafka_resp_err_t rd_kafka_unsubscribe(rd_kafka_t *rk) {
...
-        if (unlikely((err = rd_kafka_share_consumer_closed_err(rkshare))))
+        if (unlikely((acq_err = rd_kafka_share_acquire(rkshare)) != NULL)) {
+                err = rd_kafka_error_code(acq_err);
+                rd_kafka_error_destroy(acq_err);
                 return err;
-        return rd_kafka_unsubscribe(rkshare->rkshare_rk);
+        }
+        if (unlikely((err = rd_kafka_share_consumer_closed_err(rkshare))))
+                goto done;
+        err = rd_kafka_unsubscribe(rkshare->rkshare_rk);
```
pranavrth, line 66: line break before.

src/rdkafka_subscription.c
```diff
@@ -129,36 +138,51 @@ rd_kafka_share_subscribe(rd_kafka_share_t *rkshare,
...
          * TODO KIP-932: Guard this with checks for rkshare and
          *               rkshare->rkshare_rk?
          */
-        if (unlikely((err = rd_kafka_share_consumer_closed_err(rkshare))))
+        if (unlikely((acq_err = rd_kafka_share_acquire(rkshare)) != NULL)) {
+                err = rd_kafka_error_code(acq_err);
+                rd_kafka_error_destroy(acq_err);
                 return err;
+        }
+        if (unlikely((err = rd_kafka_share_consumer_closed_err(rkshare))))
```
pranavrth, line 155: line break

src/rdkafka_subscription.c
```diff
@@ -326,14 +350,23 @@ rd_kafka_resp_err_t
...
          */
-        if (unlikely((err = rd_kafka_share_consumer_closed_err(rkshare))))
+        if (unlikely((acq_err = rd_kafka_share_acquire(rkshare)) != NULL)) {
+                err = rd_kafka_error_code(acq_err);
+                rd_kafka_error_destroy(acq_err);
                 return err;
-        return rd_kafka_subscription(rkshare->rkshare_rk, topics);
+        }
+        if (unlikely((err = rd_kafka_share_consumer_closed_err(rkshare))))
+                goto done;
```
pranavrth, line 369: Check this again once to make sure that we are not incorrectly acquiring and not releasing the lock.

pranavrth (changes requested):

src/rdatomic.h
PratRanj07: Yes the max size is 64bit so this should be fine

src/rdkafka.c
```diff
@@ -3006,6 +3046,14 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
         rkshare->rkshare_unacked_cnt      = 0;
         rkshare->rkshare_consumer_closing = rd_false;
 
+        rk->rk_share_consumer.share_acknowledgement_commit_cb      = NULL;
```
pranavrth, line 3049: "share_acknowledgement_commit_cb" has share but "acknowledgement_commit_cb_opaque" and "acknowledgement_commit_cb_registered" doesn't. We need to have same naming structure. Since this is internal, add a TODO for now.

src/rdkafka_subscription.c
```diff
@@ -48,14 +48,25 @@ rd_kafka_resp_err_t rd_kafka_unsubscribe(rd_kafka_t *rk) {
 
 rd_kafka_resp_err_t rd_kafka_share_unsubscribe(rd_kafka_share_t *rkshare) {
         rd_kafka_resp_err_t err;
+        rd_kafka_error_t *acq_err = NULL;
```
pranavrth, line 51: We should name `error_t` with `error` and not `err`

src/rdkafka_share_acknowledgement.c
```diff
@@ -1183,8 +1250,10 @@ void rd_kafka_share_dispatch_ack_callbacks(rd_kafka_t *rk,
         rd_kafka_share_ack_batches_t *ack_batch;
         int k;
 
-        if (!rk->rk_conf.share_acknowledgement_commit_cb || !ack_details ||
-            rd_list_cnt(ack_details) == 0)
+        /* Locality: main thread - checks runtime-set registration flag. */
```
pranavrth, line 1253: 1 point to note that Java sends the whole partition list in a single op and call onComplete for each partition but we send individual op.

src/rdkafka_share_acknowledgement.c
```diff
@@ -1114,17 +1179,19 @@ void rd_kafka_share_enqueue_ack_commit_cb_op(
         rd_kafka_op_t *cb_rko;
         rd_kafka_share_partition_offsets_list_t *partitions;
 
-        if (!rk->rk_conf.share_acknowledgement_commit_cb)
+        /* Check if a runtime callback is registered.
+         * Locality: main thread - reads flag owned by main thread.
+         * No race because the flag is only modified by the main thread
+         * via RD_KAFKA_OP_SHARE_ACK_COMMIT_CB_REGISTER op handler. */
+        if (!rk->rk_rkshare ||
```
pranavrth, line 1186: This was not needed but if fine for now.

src/rdkafka_share_acknowledgement.c
pranavrth, line 996: Add a TODO to make sure that this is the first check that we have.

pranavrth (commented):
Minor nits and docs.
LGTM!. Thanks @PratRanj07 !.

pranavrth (approved):
LGTM!.

PR #5449, opened 2026-05-18
[KIP-932]: Minor fixes in 0157, 0178, 0179 test cases

ojasvajain:
0178 and 0179
- Removed calls to `rd_kafka_mock_set_apiversion` as they are not required.
- Removed usage of `{..}` mid-function in 0178 and 0179 and moved variable declaration at top 

0157
- Reduced timeout of 0157 test case. (On CI the multiplier is 6 so the timeout gets multiplied by this value).


PratRanj07 (approved):
LGTM! Thanks @ojasvajain

PR #5453, opened 2026-05-19
[KIP-932] Windows Share Consumer Test Pipeline Optimization With Semaphore Cache 

Ankith-Confluent:
Wrap both downloads with `cache restore` / `Test-Path` guard /
`cache store`. The .tgz is cached rather than the extracted Kafka
directory so that the Add-Content append of share/txn broker
properties stays deterministic across cache hits.

Cache keys:
- jdk21-msi-${CACHE_TAG}
- kafka-windows-${KAFKA_VERSION}-${CACHE_TAG}


pranavrth (approved):
LGTM!. Thanks @Ankith-Confluent .

PR #5462, opened 2026-05-26
[KIP-932] : Add share consumer log queue forwarding. 

k-raina:
**Summary**

* Added the `rd_kafka_share_set_log_queue` function to the public API in `rdkafka.h`, enabling applications to forward a share consumer's internal log queue to another queue for use with poll calls. This is required for proper log handling when `log.queue=true` is set on a share consumer.
* Implemented the `rd_kafka_share_set_log_queue` function in `rdkafka_queue.c`, which validates the share consumer handle and delegates to the existing `rd_kafka_set_log_queue` function.

### DIFF
--- a/src/rdkafka.h
+++ b/src/rdkafka.h
@@ -3471,63 +3471,6 @@ rd_kafka_error_t *rd_kafka_share_sasl_set_credentials(rd_kafka_share_t *rkshare,
                                                       const char *username,
                                                       const char *password);
 
-
-/**
- * @brief Redirect the share consumer's main reply queue onto the underlying
- *        consumer-group queue so that main-queue events (errors, throttles,
- *        OAUTHBEARER authentication failures, etc.) are dispatched via the
- *        consumer-group queue returned by rd_kafka_share_queue_get_consumer().
- *
- * Share consumers wrap a regular RD_KAFKA_CONSUMER handle, so this is a
- * thin passthrough to rd_kafka_poll_set_consumer() on the underlying rk.
- *
- * @param rkshare Share consumer instance.
- *
- * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success;
- *          RD_KAFKA_RESP_ERR__INVALID_ARG if rkshare is NULL or uninitialised;
- *          RD_KAFKA_RESP_ERR__UNKNOWN_GROUP if no consumer group is configured.
- *
- * @sa rd_kafka_poll_set_consumer()
- * @sa rd_kafka_share_queue_get_consumer()
- */
-RD_EXPORT
-rd_kafka_resp_err_t
-rd_kafka_share_poll_set_consumer(rd_kafka_share_t *rkshare);
-
-
-/**
- * @brief Returns a reference to this share consumer's underlying consumer-
- *        group queue. Bindings can poll this queue independently of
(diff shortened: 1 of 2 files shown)


src/rdkafka.h
```diff
@@ -4006,6 +4015,41 @@ rd_kafka_resp_err_t rd_kafka_set_log_queue(rd_kafka_t *rk,
...
+ * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success;
+ *          RD_KAFKA_RESP_ERR__INVALID_ARG if rkshare is NULL or
+ *          uninitialized;
+ *          RD_KAFKA_RESP_ERR__NOT_CONFIGURED when log.queue is not set to
+ *          true.
+ *
+ * @sa rd_kafka_set_log_queue
+ */
+RD_EXPORT
+rd_kafka_resp_err_t
```
pranavrth: I think we should return `error_t` instead of `err_t` for the new APIs.

pranavrth (changes requested):

src/rdkafka_queue.c
```diff
@@ -1080,12 +1080,19 @@ rd_kafka_resp_err_t rd_kafka_set_log_queue(rd_kafka_t *rk,
...
         if (!rkshare || !rkshare->rkshare_rk)
-                return RD_KAFKA_RESP_ERR__INVALID_ARG;
-        return rd_kafka_set_log_queue(rkshare->rkshare_rk, rkqu);
+                return rd_kafka_error_new(
+                    RD_KAFKA_RESP_ERR__INVALID_ARG,
+                    "Share consumer handle is NULL or uninitialized");
+
+        err = rd_kafka_set_log_queue(rkshare->rkshare_rk, rkqu);
+
+        return err ? rd_kafka_error_new(err, NULL) : NULL;
```
pranavrth: Use `rd_kafka_err2str(err)` instead of NULL in `rd_kafka_error_new(err, NULL)`

```suggestion
        return err ? rd_kafka_error_new(err, rd_kafka_err2str(err)) : NULL;
```

pranavrth (changes requested):

src/rdkafka.h
```diff
@@ -4012,6 +4012,43 @@ rd_kafka_resp_err_t rd_kafka_set_log_queue(rd_kafka_t *rk,
...
 
+/**
+ * @brief Forward a share consumer's internal log queue onto another queue
+ *        for serving with one of the ..poll() calls.
+ *
+ *        Thin wrapper around rd_kafka_set_log_queue() for use with
+ *        rd_kafka_share_t handles. Required when `log.queue=true` is
+ *        configured on a share consumer — without this call the share
+ *        consumer's internal log queue is never drained and log_cb never
+ *        fires.
```
pranavrth:
```suggestion
 * Thin wrapper around rd_kafka_set_log_queue() for use with
 * rd_kafka_share_t handles. Required when `log.queue=true` is
 * configured on a share consumer — without this call the share
 * consumer's internal log queue is never drained and log_cb never
 * fires.
```

pranavrth (approved):
LGTM!. Thanks @k-raina !

pranavrth (approved):
LGTM!.

PR #5463, opened 2026-05-26
[KIP-932] Take snapshot of rktp_leader before using it in select_broker to avoid race

ojasvajain:
This addresses a race condition where rktp_leader can be freed by the broker thread (`rd_kafka_toppar_forget_leader`) while the main thread is selecting a broker for share fetch, causing a segfault. By keeping a reference to the leader, we are ensuring it is not null.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3114,22 +3114,21 @@ static rd_kafka_broker_t *rd_kafka_share_select_broker(rd_kafka_t *rk,
                  * 2. A share-fetch op must not already be enqueued on it.
                  * 3. The broker or instance must not be terminating. */
                 if (leader) {
-                        if (!rd_kafka_broker_or_instance_terminating(leader)) {
-                                rd_kafka_broker_lock(leader);
-                                if (!leader->rkb_share_fetch_enqueued) {
-                                        rd_kafka_broker_keep(leader);
-                                        selected_rkb = leader;
-                                        rd_kafka_dbg(
-                                            rk, CGRP, "SHARE",
-                                            "Selected broker %s (%p) for share "
-                                            "fetch "
-                                            "for partition %s [%" PRId32 "]",
-                                            rd_kafka_broker_name(selected_rkb),
-                                            selected_rkb, partition->topic,
-                                            partition->partition);
-                                }
-                                rd_kafka_broker_unlock(leader);
+                        if (!rd_kafka_broker_or_instance_terminating(leader) &&
+                            !leader->rkb_share_fetch_enqueued) {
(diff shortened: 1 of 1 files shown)


src/rdkafka.c
```diff
@@ -3103,32 +3103,35 @@ static rd_kafka_broker_t *rd_kafka_share_select_broker(rd_kafka_t *rk,
...
-                                rd_kafka_dbg(
-                                    rk, CGRP, "SHARE",
-                                    "Selected broker %s (%p) for share fetch "
-                                    "for partition %s [%" PRId32 "]",
-                                    rd_kafka_broker_name(selected_rkb),
-                                    selected_rkb, partition->topic,
-                                    partition->partition);
+                if (leader) {
+                        if (!rd_kafka_broker_or_instance_terminating(leader)) {
+                                rd_kafka_broker_lock(leader);
```
pranavrth: I think we can remove this lock.

src/rdkafka.c
```diff
@@ -3103,32 +3103,35 @@ static rd_kafka_broker_t *rd_kafka_share_select_broker(rd_kafka_t *rk,
...
+                                            "fetch "
+                                            "for partition %s [%" PRId32 "]",
+                                            rd_kafka_broker_name(selected_rkb),
+                                            selected_rkb, partition->topic,
+                                            partition->partition);
+                                }
+                                rd_kafka_broker_unlock(leader);
                         }
-                        rd_kafka_broker_unlock(rktp->rktp_leader);
+                        rd_kafka_broker_destroy(leader);
```
pranavrth, line 3132: Add comment to reference which keep are we destroying here.

src/rdkafka.c
```diff
@@ -3103,32 +3103,35 @@ static rd_kafka_broker_t *rd_kafka_share_select_broker(rd_kafka_t *rk,
...
-                                selected_rkb = rktp->rktp_leader;
-                                rd_kafka_dbg(
-                                    rk, CGRP, "SHARE",
-                                    "Selected broker %s (%p) for share fetch "
-                                    "for partition %s [%" PRId32 "]",
-                                    rd_kafka_broker_name(selected_rkb),
-                                    selected_rkb, partition->topic,
-                                    partition->partition);
+                if (leader) {
+                        if (!rd_kafka_broker_or_instance_terminating(leader)) {
```
pranavrth: Merge this and below if conditions.

pranavrth (changes requested):

pranavrth (approved):
LGTM!. Thanks @ojasvajain

PR #5466, opened 2026-05-28
[KIP-932]: Modify delay logic in 0178 subtest to only add delay for ShareAcknowledge request

ojasvajain:
This change should help in getting rid of flakiness observed in one of subtests in 0178.
We are now adding delay only to the ShareAcknowledge request instead of adding a general delay for all requests. Adding delay for all requests makes it difficult to predict the time taken by close() as sometimes there is a ShareFetch request enqueued on a broker which also gets delayed.


pranavrth (approved):
LGTM!.

PR #5468, opened 2026-05-29
[KIP-932]: Added more integration tests

PratRanj07:
Expands KIP-932 share consumer integration coverage across subscription,
consume, ack, commit (sync/async), concurrency, transactions, close,
destroy and config paths.

Also includes three small non-test touch-ups in `src/`:
- `rdkafka_fetcher.c`: comment fix (`REJECT` -> `RELEASE`) on the
  error-op ack_type branch, and the missing `%s err` added to the
  `BADMSG` debug log.
- `rdkafka_msgset_reader.c`: TODO comment in `rd_kafka_msgset_reader_v2`
  documenting CRC-corrupt `LastOffsetDelta` failure modes. No behavior
  change.

### 0170 - subscription
- `do_test_subscription_change_acks_pending` - subscription change
  drives an implicit ack of the pending records before the new
  subscription takes effect.
- `do_test_two_groups_earliest_vs_latest` - two share groups on the
  same topic with different `share.auto.offset.reset` (earliest vs
  latest) see disjoint record sets.
- `do_test_delete_records_advances_lso` - `DeleteRecords` advances the
  LSO and a new share group starts after the deleted offsets.

### 0171 - consume
- `test_headers_preserved` - record headers round-trip end-to-end
  through the share consumer.
- `test_fetch_max_bytes_small` - small `fetch.message.max.bytes` still
  delivers all records.
- `test_record_larger_than_fetch_max_bytes` - oversized records are
  delivered (broker grants them).
- `test_consumer_close_releases_to_group` - closing a consumer with
  unacked records redelivers them to the remaining group members.
- `test_partition_increase_during_consume` - `CreatePartitions` grows
  2->5 mid-consume; new partitions are picked up by the share group.
- `test_zero_byte_payload_and_key` - zero-length and NULL keys/values
  preserved.
- `test_message_at_max_bytes_boundary` - record sizes at the topic's
  `max.message.bytes` boundary are delivered.
- `test_all_compression_codecs` (driven by `do_test_compression_codec`)
  - gzip, snappy, lz4, zstd round-trip through the share consumer.
- `test_many_and_large_headers` - 100 headers per record plus one
  large header value preserved.

### 0172 - acknowledge
- `test_ack_message_from_earlier_batch` - re-acknowledging a record
  after the batch has been polled past returns `_STATE`.
- `test_ack_offset_before_consume` - `acknowledge_offset()` before any
  record has been consumed returns `_STATE`.
- `test_ack_offset_wrong_params` - `acknowledge_offset()` with wrong
  topic/partition/offset returns `_STATE`; correct values succeed.
- `test_mixed_ack_mode_same_group` - implicit and explicit ack-mode
  consumers coexist in the same share group and total delivery
  matches produced count.

### 0173 - commit_async
- `do_test_partial_batch_commit_async` - partial-batch commit_async
  semantics; un-acked records keep `consume_batch` returning `_STATE`.
- `do_test_implicit_callback_no_explicit_commit` - implicit mode fires
  the ack callback without explicit commits.
- `do_test_lock_expiry_callback_err` - lock-expiry surfaces as an
  error to the ack callback.
- `do_test_safe_api_from_callback` - API calls from inside the ack
  callback don't deadlock or misbehave.
- `do_test_callback_side_effects_dont_break` - heavy callback work
  doesn't break subsequent consume/commit flow.

### 0174 - concurrency
- `test_chaos_consumer_lifecycle` - consumer churn (start / close /
  detach wedged threads) during sustained production. Runs for both
  `implicit` and `explicit` ack modes.

### 0176 - commit_sync
- `do_test_commit_sync_after_topic_deletion` - commit_sync after the
  subscribed topic is deleted surfaces a per-partition error.
- Chaos suite stressing interleaved ack/commit patterns:
  - `do_test_chaos_111_ack_commit_interleave` - strict per-record
    ack/commit pattern: phase A (records 0-14) one commit per ack,
    alternating sync/async; phase B (records 15-29) one commit per
    every 2 acks, alternating sync/async.
  - `do_test_chaos_alternating_commits` - every ACCEPT immediately
    followed by alternating commit_sync / commit_async (40 records).
  - `do_test_chaos_random_ack_random_commit` - fixed seed (0xC0FFEE).
    Random ACCEPT/RELEASE/REJECT acks with ~1-in-3 commits, sync/async
    coin-flipped. Final commit_sync to flush.
  - `do_test_chaos_burst_commits` - after every ACCEPT, 5 back-to-back
    commit_async then 1 draining commit_sync (50 records).
  - `do_test_chaos_callback_receipt_match` - varying batch sizes
    between commits, alternating sync/async; asserts callback
    `total_offsets >= consumed`.

### 0177 - transactions
- `do_test_dynamic_committed_to_uncommitted_with_release` - dynamic
  isolation level change `READ_COMMITTED -> READ_UNCOMMITTED` with
  RELEASE ack type.
- `do_test_dynamic_committed_to_uncommitted_with_reject` - same with
  REJECT.
- `do_test_two_groups_different_isolation_levels` - two distinct share
  groups, one `read_committed` and one `read_uncommitted`, consuming
  the same topic concurrently. Verifies the committed group sees only
  committed txns; the uncommitted group sees all.

### 0178 - close
- `test_implicit_ack_callback_fires_on_close` - the implicit ack
  callback fires for un-committed records when the consumer is closed.

### 0179 - destroy
- `do_test_destroy_with_mixed_acks_and_commits` - destroy a consumer
  that drove a mixed-ack workload (round-robin ACCEPT/RELEASE/REJECT
  with interleaved commit_async / commit_sync). Runs destroy on a
  watchdog thread and fails the test with a clear diagnostic if it
  hangs past 15 s rather than wedging the binary.

### 0180 - config
- `test_max_poll_records_caps_batch_at_5` - producer with `linger.ms=0`
  emits 10 records with a 500 ms gap so each lands as its own broker
  batch. With `max.poll.records=5`, no consume_batch call returns more
  than 5 records and at least 2 batches are needed to drain.
- `test_max_poll_records_allows_full_drain_at_10` - same producer
  pattern. With `max.poll.records=10`, all batches stay <= 10 and the
  sum of batch sizes equals 10.

0180 is now split into:
- `main_0180_share_consumer_config` - broker-required (max.poll.records
  tests).
- `main_0180_share_consumer_config_local` - TEST_F_LOCAL (the existing
  construction-time conf-rejection tests).

Both are registered in [tests/test.c](tests/test.c), matching the same
base/`_local` split convention used by 0176, 0178, 0179.

### DIFF
--- a/tests/0170-share_consumer_subscription.c
+++ b/tests/0170-share_consumer_subscription.c
@@ -1676,7 +1676,6 @@ static void subchg_ack_cb(rd_kafka_share_t *rkshare,
         (void)rkshare;
 
         /* Mirror test_share_ack_cb's base accounting */
-        st->base.callback_cnt++;
         test_ack_cb_state_push_err(&st->base, err);
 
         entry = rd_kafka_share_partition_offsets_list_get(partitions, 0);
@@ -1803,11 +1802,11 @@ static void do_test_subscription_change_acks_pending(void) {
                     "Expected ack callback to fire after subscription change, "
                     "got %d callbacks",
                     state.base.callback_cnt);
-        TEST_ASSERT(test_ack_cb_state_first_err(&state.base) ==
-                        RD_KAFKA_RESP_ERR_NO_ERROR,
-                    "Expected no error in callback, got %s",
-                    rd_kafka_err2name(
-                        test_ack_cb_state_first_err(&state.base)));
+        TEST_ASSERT(
+            test_ack_cb_state_first_err(&state.base) ==
+                RD_KAFKA_RESP_ERR_NO_ERROR,
+            "Expected no error in callback, got %s",
+            rd_kafka_err2name(test_ack_cb_state_first_err(&state.base)));
         TEST_ASSERT(strcmp(state.first_callback_topic, t1) == 0,
                     "Expected first ack callback to be for t1 (%s), got %s", t1,
                     state.first_callback_topic);
--- a/tests/0170-share_consumer_subscription.c
+++ b/tests/0170-share_consumer_subscription.c
@@ -607,8 +607,13 @@ static void state_init(sub_test_state_t *state,
 
         memset(state, 0, sizeof(*state));
 
-        rd_snprintf(state->group_name, sizeof(state->group_name), "share-%s",
-                    scenario->name);
+        /* Append a per-invocation unique suffix so each run gets a
+         * fresh share-group on the broker (avoids collisions across
+         * re-runs or parallel runs against the same cluster). */
+        rd_snprintf(state->group_name, sizeof(state->group_name),
+                    "share-%s-rnd%" PRIx64, scenario->name, test_id_generate());
+        TEST_SAY("Scenario '%s' using group '%s'\n", scenario->name,
+                 state->group_name);
 
         state->consumer_cnt =
             scenario->consumer_cnt > 0 ? scenario->consumer_cnt : 1;
@@ -929,13 +934,10 @@ static void test_topic_deletion_does_not_surface_error(void) {
 
         SUB_TEST();
 
-        if (!strcmp(test_getenv("TEST_BROKER_OS", ""), "windows")) {
-                TEST_SAY(
-                    "Skipping topic-deletion-does-not-surface-error "
+        if (!strcmp(test_getenv("TEST_BROKER_OS", ""), "windows"))
+                SUB_TEST_SKIP(
+                    "topic-deletion-does-not-surface-error "
                     "(broker on Windows)\n");
-                SUB_TEST_PASS();
-                return;
-        }
 
         topic     = test_mk_topic_name("0170-delete-no-surface", 1);
         topic_dup = rd_strdup(topic);
@@ -945,12 +947,17 @@ static void test_topic_deletion_does_not_surface_error(void) {
 
         /* Custom consumer that disables the set_notexists defer window
          * (default 30s) so the log-and-drop path runs on the first
-         * post-delete metadata refresh instead of being deferred. */
+         * post-delete metadata refresh instead of being deferred.
+         * Also reduce the metadata refresh interval so a refresh actually
+         * happens during the post-delete settle window — the default is
+         * 5 minutes, which would never fire during the test. */
         test_conf_init(&conf, NULL, 60);
         rd_kafka_conf_set(conf, "group.id", group, errstr, sizeof(errstr));
         rd_kafka_conf_set(conf, "share.acknowledgement.mode", "explicit",
                           errstr, sizeof(errstr));
         rd_kafka_conf_set(conf, "topic.metadata.propagation.max.ms", "0",
+                          errstr, sizeof(errstr));
+        rd_kafka_conf_set(conf, "topic.metadata.refresh.interval.ms", "500",
                           errstr, sizeof(errstr));
         consumer = rd_kafka_share_consumer_new(conf, errstr, sizeof(errstr));
         TEST_ASSERT(consumer, "Failed to create share consumer: %s", errstr);
@@ -989,7 +996,7 @@ static void test_topic_deletion_does_not_surface_error(void) {
         TEST_ASSERT(!del_err, "DeleteTopics failed: %s",
                     rd_kafka_err2str(del_err));
 
-        /* Allow metadata propagation. */
+        /* Wait for the topic delete to propagate in the cluster. */
         rd_sleep(3);
 
         /* Drain consume_batch and verify no topic-level error reaches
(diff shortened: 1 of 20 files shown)


tests/0170-share_consumer_subscription.c
```diff
@@ -856,6 +861,145 @@ static void do_test_multi_consumer_overlap(void) {
...
+        int post_attempts;
+        rd_bool_t surfaced_topic_err = rd_false;
+
+        SUB_TEST();
+
+        if (!strcmp(test_getenv("TEST_BROKER_OS", ""), "windows")) {
+                TEST_SAY(
+                    "Skipping topic-deletion-does-not-surface-error "
+                    "(broker on Windows)\n");
+                SUB_TEST_PASS();
```
ojasvajain: Should we use SUB_TEST_SKIP()?

tests/0170-share_consumer_subscription.c
```diff
@@ -856,6 +861,145 @@ static void do_test_multi_consumer_overlap(void) {
...
+                    "record before deleting the topic");
+
+        TEST_SAY("Deleting topic %s\n", topic);
+        topics_for_delete[0] = topic_dup;
+        del_err              = test_DeleteTopics_simple(common_admin, NULL,
+                                                        topics_for_delete, 1, NULL);
+        TEST_ASSERT(!del_err, "DeleteTopics failed: %s",
+                    rd_kafka_err2str(del_err));
+
+        /* Allow metadata propagation. */
```
ojasvajain: I think we meant to say that we are waiting for delete to propagate in the cluster.

tests/0170-share_consumer_subscription.c
```diff
@@ -856,6 +861,145 @@ static void do_test_multi_consumer_overlap(void) {
...
+        rd_sleep(3);
+
+        /* Drain consume_batch and verify no topic-level error reaches
+         * the application. _STATE / generic timeouts are expected
+         * (no more records); only the metadata-derived topic codes
+         * must not surface. */
+        for (post_attempts = 0; post_attempts < 30; post_attempts++) {
```
ojasvajain, line 1006: Are we sure metadata refresh would happen? We should reduce the metadata refresh interval.

tests/0170-share_consumer_subscription.c
```diff
@@ -1343,6 +1487,316 @@ static void do_test_subscribe_caret_treated_as_literal_e2e(void) {
...
+                size_t batch_rcvd = 0;
+                err = rd_kafka_share_consume_batch(rkshare, 2000, batch + rcvd,
+                                                   &batch_rcvd);
+                if (err)
+                        rd_kafka_error_destroy(err);
+                rcvd += batch_rcvd;
+        }
+        TEST_ASSERT(rcvd == 1, "Expected 1 record from t1, got %zu", rcvd);
+        for (i = 0; i < rcvd; i++)
+                rd_kafka_message_destroy(batch[i]);
```
ojasvajain: just call destroy on batch[0]

tests/0170-share_consumer_subscription.c
```diff
@@ -1343,6 +1487,316 @@ static void do_test_subscribe_caret_treated_as_literal_e2e(void) {
...
+                rcvd += batch_rcvd;
+        }
+
+        TEST_ASSERT(state.callback_cnt >= 1,
+                    "Expected ack callback to fire after subscription change, "
+                    "got %d callbacks",
+                    state.callback_cnt);
+        TEST_ASSERT(state.last_err == RD_KAFKA_RESP_ERR_NO_ERROR,
+                    "Expected no error in callback, got %s",
+                    rd_kafka_err2name(state.last_err));
```
ojasvajain: Let's add assert for rcvd == 1

ojasvajain: Once done, you can remove the loop below and just destroy batch[0]

tests/0170-share_consumer_subscription.c
```diff
@@ -1343,6 +1487,316 @@ static void do_test_subscribe_caret_treated_as_literal_e2e(void) {
...
+        while ((rcvd < 1 || state.callback_cnt < 1) && attempts-- > 0) {
+                size_t batch_rcvd = 0;
+                err = rd_kafka_share_consume_batch(rkshare, 1000, batch + rcvd,
+                                                   &batch_rcvd);
+                if (err)
+                        rd_kafka_error_destroy(err);
+                rcvd += batch_rcvd;
+        }
+
+        TEST_ASSERT(state.callback_cnt >= 1,
```
ojasvajain: Hope this cb will not get triggered for a t2 record being acked

tests/0170-share_consumer_subscription.c
```diff
@@ -1343,6 +1487,316 @@ static void do_test_subscribe_caret_treated_as_literal_e2e(void) {
...
+        while (consumed < 5 && attempts-- > 0) {
+                batch_cnt = 0;
+                test_share_consume_batch(rkshare, 2000, NULL, 0, &batch_cnt);
+                consumed += batch_cnt;
+        }
+        TEST_ASSERT(consumed == 5,
+                    "Expected 5 records (offsets 5..9) after DeleteRecords, "
+                    "consumed %d",
+                    consumed);
+
```
ojasvajain, line 2030: It would be better if we can also add an assertion on the offset of the msg. You can check if the offset of first msg is 5.

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+
+                e = rd_kafka_share_consume_batch(consumer, 2000, batch + rcvd,
+                                                 &batch_rcvd);
+                if (e)
+                        rd_kafka_error_destroy(e);
+                rcvd += batch_rcvd;
+        }
+        TEST_ASSERT(rcvd == 1, "Expected 1 record, got %zu", rcvd);
+
+        for (i = 0; i < rcvd; i++) {
```
ojasvajain: No need for this loop. Just use 0 index.

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+
+        subs = rd_kafka_topic_partition_list_new(1);
+        rd_kafka_topic_partition_list_add(subs, topic, RD_KAFKA_PARTITION_UA);
+        rd_kafka_share_subscribe(consumer, subs);
+        rd_kafka_topic_partition_list_destroy(subs);
+
+        attempts = 200;
+        while (consumed < msgcnt && attempts-- > 0) {
+                int batch_cnt = 0;
+                test_share_consume_batch(consumer, 2000, NULL, 0, &batch_cnt);
```
ojasvajain, line 1178: We should also check if we are honoring the value of `fetch.message.max.bytes` or not. You can add a check to see no consume_batch call returns more than 1 msg.

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+        TEST_ASSERT(consumer, "Failed to create share consumer: %s", errstr);
+
+        test_share_set_auto_offset_reset(group, "earliest");
+
+        subs = rd_kafka_topic_partition_list_new(1);
+        rd_kafka_topic_partition_list_add(subs, topic, RD_KAFKA_PARTITION_UA);
+        rd_kafka_share_subscribe(consumer, subs);
+        rd_kafka_topic_partition_list_destroy(subs);
+
+        attempts = 200;
```
ojasvajain: this can be reduced.

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+                }
+
+                for (m = 0; m < rcvd; m++) {
+                        if (!batch[m]->err)
+                                consumed++;
+                        rd_kafka_message_destroy(batch[m]);
+                }
+        }
+        TEST_SAY("Pre-grow consumed: %d/%d\n", consumed, initial_expected);
+        TEST_ASSERT(consumed >= initial_expected,
```
ojasvajain: ==

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+                    rd_kafka_err2str(cp_err));
+
+        /* Force common_producer's metadata cache to refresh and observe the
+         * new partition count before producing to the new partitions —
+         * otherwise rd_kafka_produce() to partition >= cached count fails
+         * with UNKNOWN_PARTITION (producer's default metadata refresh
+         * interval is 5 minutes). */
+        rd_kafka_metadata_topic_t exp_md = {.topic         = (char *)topic,
+                                            .partition_cnt = grown_partitions};
+        test_wait_metadata_update(common_producer, &exp_md, 1, NULL, 0,
```
ojasvajain: Rather than waiting for 30s, you can configure the metadata refresh interval to a lower value like 500 ms or something.

ojasvajain: Same for consumer

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+                }
+
+                for (m = 0; m < rcvd; m++) {
+                        if (!batch[m]->err)
+                                consumed++;
+                        rd_kafka_message_destroy(batch[m]);
+                }
+        }
+
+        TEST_ASSERT(consumed >= total_expected,
```
ojasvajain: ==

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+        for (i = 0; i < rcvd; i++) {
+                rd_bool_t key_null  = batch[i]->key == NULL;
+                rd_bool_t key_empty = batch[i]->key_len == 0 && !key_null;
+                rd_bool_t val_null  = batch[i]->payload == NULL;
+                rd_bool_t val_empty = batch[i]->len == 0 && !val_null;
+
+                /* Brokers may normalize empty buffers to NULL; accept either.
+                 */
+                rd_bool_t key_is_empty_or_null = key_null || key_empty;
+                rd_bool_t val_is_empty_or_null = val_null || val_empty;
```
ojasvajain: Declarations at top

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+        test_conf_init(&conf, NULL, 60);
+        test_conf_set(conf, "message.max.bytes", "8192");
+        rd_kafka_conf_set_dr_msg_cb(conf, test_dr_msg_cb);
+        producer = test_create_handle(RD_KAFKA_PRODUCER, conf);
+
+        /* Produce one record per target size. */
+        for (i = 0; i < total; i++) {
+                char *payload = rd_malloc(sizes[i]);
+                rd_kafka_resp_err_t err;
+                size_t j;
```
ojasvajain: Declarations at the top

tests/0171-share_consumer_consume.c
```diff
@@ -844,8 +861,1008 @@ static void test_poll_callback_piggybacked_acks(void) {
...
+            hdr_cnt, big_hdr_size);
+
+        consumer = test_create_share_consumer(group, NULL);
+        topic    = test_mk_topic_name("0171-many-headers", 1);
+        test_create_topic_wait_exists(NULL, topic, 1, -1, 60 * 1000);
+
+        /* Build the producev arg list for a record with `hdr_cnt` headers.
+         * Each header name is "h-<i>" and each value is "v-<i>". */
+        {
+                rd_kafka_headers_t *hdrs = rd_kafka_headers_new(hdr_cnt);
```
ojasvajain: Declaratin at top

tests/0178-share_consumer_close.c
```diff
@@ -1054,7 +1062,7 @@ static void test_close_with_acknowledge(void) {
 
                 /* Close C2 */
                 TEST_SAY("C2: Closing consumer\n");
-                rd_kafka_share_consumer_close(c2);
+                test_share_consumer_close(c2);
```
ojasvajain: In this test case, we want to specifically print `C2` in the logs. So let's revert this.

tests/0178-share_consumer_close.c
```diff
@@ -1176,7 +1184,7 @@ static void test_close_without_acknowledge() {
                                         "no-ack-close");
 
                 TEST_SAY("C2: Closing consumer\n");
-                rd_kafka_share_consumer_close(c2);
+                test_share_consumer_close(c2);
```
ojasvajain: Same

tests/0179-share_consumer_destroy.c
```diff
@@ -2076,6 +2266,8 @@ int main_0179_share_consumer_destroy(int argc, char **argv) {
                                           rd_true);
         do_test_destroy_with_implicit_ack(RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE);
 
+        do_test_destroy_with_mixed_acks_and_commits();
```
ojasvajain: Also add the variant where we pass `RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE` as destroy flag. See above test cases for reference.

tests/0179-share_consumer_destroy.c
```diff
@@ -2048,6 +2048,196 @@ static void do_test_destroy_with_subscribe_unsubscribe(int do_subscribe,
...
+        rd_kafka_share_t *rkshare;
+        rd_atomic32_t done;
+} destroy_watchdog_arg_t;
+
+static int destroy_watchdog_thread(void *p) {
+        destroy_watchdog_arg_t *a = p;
+        rd_kafka_share_destroy(a->rkshare);
+        rd_atomic32_set(&a->done, 1);
+        return 0;
+}
```
ojasvajain: Helpers and related code is at the top of the file.  See existing structure.

tests/0172-share_consumer_acknowledge.c
```diff
@@ -1347,6 +1358,354 @@ static void test_scale_10_topics_3_partitions(void) {
...
+        /* Poll batch 2 - should be empty (we already consumed everything) */
+        size_t r = 0;
+        rd_kafka_error_t *e =
+            rd_kafka_share_consume_batch(rkshare, 2000, batch2, &r);
+        if (e)
+                rd_kafka_error_destroy(e);
+        TEST_SAY("Batch 2 received %zu records\n", r);
+        for (i = 0; i < r; i++)
+                rd_kafka_message_destroy(batch2[i]);
+
```
ojasvajain: Just assert for the msgs count to be zero.

tests/0172-share_consumer_acknowledge.c
```diff
@@ -1347,6 +1358,354 @@ static void test_scale_10_topics_3_partitions(void) {
...
+        size_t rcvd = 0;
+        size_t m;
+        rd_kafka_error_t *err =
+            rd_kafka_share_consume_batch(implicit_c, 1000, batch, &rcvd);
+        if (err)
+                rd_kafka_error_destroy(err);
+        for (m = 0; m < rcvd; m++)
+                rd_kafka_message_destroy(batch[m]);
+
+        TEST_ASSERT(implicit_cnt + explicit_cnt >= total_msgs,
```
ojasvajain: ==

tests/0173-share_consumer_commit_async.c
```diff
@@ -1643,8 +1643,513 @@ static void do_test_commit_async_callback(void) {
...
+                rcvd  = 0;
+                error = rd_kafka_share_consume_batch(rkshare, 3000,
+                                                     rkmessages + total, &rcvd);
+                if (error) {
+                        rd_kafka_error_destroy(error);
+                        continue;
+                }
+                total += rcvd;
+        }
+        TEST_ASSERT(total >= 3,
```
ojasvajain: Shouldn't it be total == 5?

ojasvajain: The while loop conditions would ensure total is 5. If it isn't, then we should fail the test case.

tests/0173-share_consumer_commit_async.c
```diff
@@ -1643,8 +1643,513 @@ static void do_test_commit_async_callback(void) {
...
+                        rd_kafka_error_destroy(error);
+                        continue;
+                }
+                for (j = 0; j < rcvd; j++) {
+                        if (!rkmessages[j]->err)
+                                consumed++;
+                        rd_kafka_message_destroy(rkmessages[j]);
+                }
+        }
+        TEST_ASSERT(consumed > 0, "Expected to consume some messages");
```
ojasvajain: consumed == 10

tests/0173-share_consumer_commit_async.c
```diff
@@ -1643,8 +1643,513 @@ static void do_test_commit_async_callback(void) {
...
+        test_wait_for_cb_with_poll(&state, rkshare, 1, 15000);
+
+        TEST_SAY("Callback count=%d, total_offsets=%zu, last_err=%s\n",
+                 state.callback_cnt, state.total_offsets,
+                 rd_kafka_err2name(state.last_err));
+
+        TEST_ASSERT(state.callback_cnt >= 1,
+                    "Expected callback to fire for stale ack, got %d",
+                    state.callback_cnt);
+        TEST_ASSERT(state.last_err != RD_KAFKA_RESP_ERR_NO_ERROR,
```
ojasvajain: can we assert for the exact error expected?

tests/0176-share_consumer_commit_sync.c
```diff
@@ -2049,12 +2049,752 @@ static void do_test_commit_sync_callback(void) {
...
+        /* Delete the subscribed topic */
+        TEST_SAY("Deleting topic %s\n", topic_name);
+        topics_for_delete[0] = topic_dup;
+        del_err              = test_DeleteTopics_simple(common_admin, NULL,
+                                                        topics_for_delete, 1, NULL);
+        TEST_ASSERT(del_err == RD_KAFKA_RESP_ERR_NO_ERROR,
+                    "DeleteTopics failed: %s", rd_kafka_err2str(del_err));
+
+        /* Give metadata propagation a moment */
+        rd_sleep(3);
```
ojasvajain, line 2155: This sleep is required for delete to be propagated in the cluster not for metadata refresh.

ojasvajain (commented):
Left comments. Thanks for this!

Since we have added so many test cases, there is a chance that some might be flaky. Once comments are addressed, let's ensure the builds pass consistently.

pranavrth:
Please upgrade to kafka version 4.3.0

tests/0170-share_consumer_subscription.c
PratRanj07: updated to check partition too

tests/0170-share_consumer_subscription.c
```diff
@@ -1482,6 +1489,405 @@ static void do_test_subscribe_caret_treated_as_literal_e2e(void) {
...
+        TEST_SAY("\n");
+        TEST_SAY(
+            "============================================================\n");
+        TEST_SAY("=== delete-records-advances-lso ===\n");
+        TEST_SAY(
+            "============================================================\n");
+
+        if (!strcmp(test_getenv("TEST_BROKER_OS", ""), "windows")) {
+                TEST_SAY(
+                    "Skipping delete-records scenario (broker on Windows)\n");
```
ojasvajain: SUB_TEST_SKIP

ojasvajain (commented):
3 more open comments, 2 from previous pass.
Looks good otherwise.

tests/0171-share_consumer_consume.c
PratRanj07: Done

tests/0171-share_consumer_consume.c
PratRanj07: Done

ojasvajain (approved):

PR #5469, opened 2026-05-30
[KIP-932] Share subscribe related improvements

pranavrth:
## Summary

Improvements and fixes around the share-consumer subscribe path.

- **Drop regex subscription path.** `rd_kafka_share_subscribe` forwards topic names verbatim to the broker as `ShareGroupHeartbeatRequest.SubscribedTopicNames`, matching the wire schema (no regex field) and Java's `ShareConsumer`. Removes regex compile at the API and `combine_regexes`/`remove_regexes` splitting in the share branch of `rd_kafka_cgrp_subscription_set`. Empty list and empty topic-name entries return `INVALID_ARG`; the dedup check is kept.
- **Revert URL-safe Base64 workaround for generated Uuids.** Reverts the `+`/`/` → `-`/`_` substitution in `rd_kafka_Uuid_base64str` (758c5cd64). Broker now accepts the standard Base64 form.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -5799,23 +5799,39 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
 
         if (rkcg->rkcg_subscription) {
                 rkcg->rkcg_flags |= RD_KAFKA_CGRP_F_SUBSCRIPTION;
-                if (rd_kafka_topic_partition_list_regex_cnt(
-                        rkcg->rkcg_subscription) > 0)
-                        rkcg->rkcg_flags |=
-                            RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION;
 
-                if (rkcg->rkcg_group_protocol ==
-                    RD_KAFKA_GROUP_PROTOCOL_CONSUMER) {
-                        rkcg->rkcg_subscription_regex =
-                            rd_kafka_topic_partition_list_combine_regexes(
-                                rkcg->rkcg_subscription);
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rkcg->rkcg_rk)) {
+                        /* Share consumer has no regex subscription: feed the
+                         * full subscription list to the heartbeat as
+                         * SubscribedTopicNames; let the broker reject any
+                         * invalid topic names. */
                         rkcg->rkcg_subscription_topics =
-                            rd_kafka_topic_partition_list_remove_regexes(
+                            rd_kafka_topic_partition_list_copy(
                                 rkcg->rkcg_subscription);
                         rkcg->rkcg_consumer_flags |=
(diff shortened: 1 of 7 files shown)


tests/0170-share_consumer_subscription.c
```diff
@@ -1162,6 +1136,194 @@ static void test_auto_offset_reset_default_latest(void) {
...
+ *   1. A topic name starting with '^' is treated as a literal — no
+ *      regex compilation, no rejection. The name is forwarded
+ *      verbatim to the broker via
+ *      ShareGroupHeartbeat.SubscribedTopicNames; the broker silently
+ *      drops names it cannot resolve.
+ *   2. Subscription list containing an empty topic name returns
+ *      INVALID_ARG.
+ *   3. Duplicate topic names return INVALID_ARG. (librdkafka enforces
+ *      dedup at the API; Java silently dedupes via Set — divergence
+ *      tracked as A6 in src/share-consumer-java-divergence.md.)
```
PratRanj07: Remove the reference of java and plan file

tests/0170-share_consumer_subscription.c
```diff
@@ -1162,6 +1136,194 @@ static void test_auto_offset_reset_default_latest(void) {
...
+ *
+ * Empty-list rejection is covered by the mock test
+ * do_test_empty_topic_list_subscription in
+ * 0155-share_group_heartbeat_mock.c.
```
PratRanj07: Do we need this?

src/rdkafka_cgrp.c
```diff
@@ -5799,23 +5799,39 @@ rd_kafka_cgrp_subscription_set(rd_kafka_cgrp_t *rkcg,
...
+                                rkcg->rkcg_flags |=
+                                    RD_KAFKA_CGRP_F_WILDCARD_SUBSCRIPTION;
+
+                        if (rkcg->rkcg_group_protocol ==
+                            RD_KAFKA_GROUP_PROTOCOL_CONSUMER) {
+                                rkcg->rkcg_subscription_regex =
+                                    rd_kafka_topic_partition_list_combine_regexes(
+                                        rkcg->rkcg_subscription);
+                                rkcg->rkcg_subscription_topics =
+                                    rd_kafka_topic_partition_list_remove_regexes(
```
PratRanj07, line 5827: Keep the normal consumer related things first

PratRanj07 (changes requested):
Hey @pranavrth !
Thanks for the PR!
Provided some comments

src/rdkafka_cgrp.c
pranavrth: I think this is a readability thing here as subscribe is not much called in consumer lifecycle. I would keep it like this for more readability. We did this change for fetch path as we wanted to make sure that there is least performance impact on the current normal consumer path.

PratRanj07 (approved):
LGTM! Thanks @pranavrth

PratRanj07 (approved):

PR #5470, opened 2026-05-30
[KIP-932]: Share consumers must not touch regular-consumer offset machinery

PratRanj07:
## Summary

Share consumers reuse a number of code paths with regular consumers
(metadata-refresh, leader change, broker decommission, cgrp
terminate, PARTITION_JOIN). Several of those paths still drove
regular-consumer offset machinery (`OffsetForLeaderEpoch`,
`ListOffsets`, `offset_query_tmr`, `offset_store_stop`, version
barriers) or enqueued ops onto `rktp_fetchq`, which the share path
doesn't drain. This PR adds the missing `RD_KAFKA_IS_SHARE_CONSUMER`
guards so those operations are no-ops for share rk.

### `src/rdkafka_topic.c` — `rd_kafka_toppar_leader_update`

Skip the `need_epoch_validation` block for share rk. The body
calls `rd_kafka_toppar_set_offset_validation_position` followed by
`rd_kafka_offset_validate`, which ultimately issues
`OffsetForLeaderEpoch` / `ListOffsets` and pipes the reply into
`rktp_fetchq`. Reachable on every metadata refresh.

### `src/rdkafka_partition.c` — `rd_kafka_toppar_enq_error`

Route `OP_ERR` to `rk_cgrp->rkcg_q` for share consumers instead of
`rktp_fetchq`. Topic-level errors surfaced through this helper
(`UNKNOWN_TOPIC_OR_PART` / `UNKNOWN_PARTITION` from the
metadata-refresh path) were previously stranded on a queue the share
path never drains.

### `src/rdkafka_partition.c` — `rd_kafka_toppar_broker_migrate`

Guard the `FETCH_OFFSET_WAIT -> offset_retry` transition with
`!IS_SHARE_CONSUMER`. Share rk never enters `OFFSET_WAIT` so this is
defensive today, but it keeps the offset machinery out of the
share leader-change path.

### `src/rdkafka_partition.c` — `rd_kafka_toppar_offset_retry`

Add an early-return at the top of the helper itself if
`IS_SHARE_CONSUMER`. Belt-and-suspenders: callers must guard, but if
a stray future call slipped through it would otherwise arm the
offset-query timer for a share-assigned rktp.

### `src/rdkafka_cgrp.c` — `cgrp_op_serve` `OP_PARTITION_JOIN`

Skip `rd_kafka_toppar_op_fetch_stop` for share rk in the
"PARTITION_JOIN while terminating" branch. `op_fetch_stop` bumps the
version barrier, touches `rktp_offset_query_tmr`, and calls
`offset_store_stop` — none of which apply to a share-assigned rktp
that never started a regular fetch.

### `src/rdkafka_broker.c` — `OP_PARTITION_JOIN` handler

Gate the `rktp_ts_fetch_backoff = 0` reset with
`!IS_SHARE_CONSUMER`. Surfaced by the broker-migrate call-graph
audit: share rk satisfies `rk_type == RD_KAFKA_CONSUMER`, so this
branch fired and wrote to a regular-fetch-only field. Dead-write
today (share doesn't read the field), but a forbidden-state
mutation worth closing.

### `src/rdkafka_msgset_reader.c` — `rd_kafka_msgset_reader_run`

Split the per-batch enqueue-debug log into share / non-share branches.
Three fields the regular log prints are misleading for share consumers:
`"fetch queue"` (share writes to the caller's `temp_fetchq`, not
`rktp_fetchq`), `v%d` (share always passes `tver.version=0`), and
`%d aborted msgsets` (share calls `share_msgset_parse` with
`aborted_txns=NULL`, so the counter is never bumped). The share
variant emits `"temp fetchq"`, drops the version field, and drops the
aborted-count field; `msetr_ctrl_cnt` stays as a "should be zero"
diagnostic.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -454,9 +454,6 @@ void rd_kafka_cgrp_destroy_final(rd_kafka_cgrp_t *rkcg) {
         rd_list_destroy(&rkcg->rkcg_toppars);
         rd_list_destroy(rkcg->rkcg_subscribed_topics);
         rd_kafka_topic_partition_list_destroy(rkcg->rkcg_errored_topics);
-        if (rkcg->rkcg_share_topic_errored)
-                rd_kafka_topic_partition_list_destroy(
-                    rkcg->rkcg_share_topic_errored);
         if (rkcg->rkcg_assignor && rkcg->rkcg_assignor->rkas_destroy_state_cb &&
             rkcg->rkcg_assignor_state)
                 rkcg->rkcg_assignor->rkas_destroy_state_cb(
@@ -542,9 +539,6 @@ rd_kafka_cgrp_t *rd_kafka_cgrp_new(rd_kafka_t *rk,
         rkcg->rkcg_next_target_assignment = NULL;
 
         rkcg->rkcg_errored_topics = rd_kafka_topic_partition_list_new(0);
-        if (RD_KAFKA_IS_SHARE_CONSUMER(rk))
-                rkcg->rkcg_share_topic_errored =
-                    rd_kafka_topic_partition_list_new(0);
 
         /* Create a logical group coordinator broker to provide
          * a dedicated connection for group coordination.
@@ -5718,86 +5712,43 @@ static void rd_kafka_propagate_consumer_topic_errors(
 
 
 /**
- * @brief Clear a share-consumer's errored-topic entry by topic_id.
- *
- *        Called from rd_kafka_topic_set_exists on recovery so a
- *        subsequent re-failure with the same code surfaces again,
- *        independent of any wholesale-replace happening in
- *        rd_kafka_propagate_consumer_topic_errors.
- *
- * @locality rdkafka main thread
- */
-void rd_kafka_share_clear_topic_err(rd_kafka_cgrp_t *rkcg,
-                                    rd_kafka_Uuid_t topic_id) {
-        int i;
-
-        if (!rkcg || !rkcg->rkcg_errored_topics)
-                return;
-
-        rd_dassert(thrd_is_current(rkcg->rkcg_rk->rk_thread));
-
-        for (i = 0; i < rkcg->rkcg_errored_topics->cnt; i++) {
-                if (rd_kafka_Uuid_cmp(rd_kafka_topic_partition_get_topic_id(
-                                          &rkcg->rkcg_errored_topics->elems[i]),
-                                      topic_id) == 0) {
-                        rd_kafka_topic_partition_list_del_by_idx(
-                            rkcg->rkcg_errored_topics, i);
-                        return;
-                }
-        }
-}
-
-
-/**
  * @brief Share-consumer entry point for end-of-metadata-cycle error
  *        propagation.
  *
  *        Drains the per-cycle pending list: app-facing codes
- *        (TOPIC_EXCEPTION, TOPIC_AUTHORIZATION_FAILED) are handed to
- *        rd_kafka_propagate_consumer_topic_errors for delivery and
- *        cross-cycle dedup; transient codes are debug-logged at most
- *        once per (topic, err) per cycle. Existing rkcg_errored_topics
- *        entries are always merged forward so propagate's wholesale-
- *        replace cannot drop entries for topics absent from this cycle.
- *
- *        After propagate, entries whose rkt has transitioned to
- *        RD_KAFKA_TOPIC_S_NOTEXISTS are pruned from the dedup memo.
- *        Recovery (back to S_EXISTS) is handled separately by
- *        rd_kafka_share_clear_topic_err on set_exists.
- *
- * TODO KIP-932: GA: Refactor this function and maybe use a new function
- *         instead of rd_kafka_propagate_consumer_topic_errors to avoid
- *         the complexity of merging the pending list with existing
- *         error topics.
+ *        (TOPIC_EXCEPTION, TOPIC_AUTHORIZATION_FAILED) are emitted
+ *        directly to the application on next poll; transient codes are
+ *        debug-logged.
  *
  * @locality rdkafka main thread
  */
 void rd_kafka_share_topic_err_propagate(rd_kafka_cgrp_t *rkcg) {
         rd_kafka_t *rk = rkcg->rkcg_rk;
-        rd_kafka_topic_partition_list_t *pending;
-        rd_kafka_topic_partition_list_t *to_surface;
-        int i, j;
+        int i;
 
-        if (!rkcg->rkcg_share_topic_errored)
+        if (!rkcg->rkcg_errored_topics)
                 return;
 
-        pending                        = rkcg->rkcg_share_topic_errored;
-        rkcg->rkcg_share_topic_errored = rd_kafka_topic_partition_list_new(0);
-
-        to_surface = rd_kafka_topic_partition_list_new(0);
-        for (i = 0; i < pending->cnt; i++) {
+        for (i = 0; i < rkcg->rkcg_errored_topics->cnt; i++) {
                 const rd_kafka_topic_partition_t *pending_errored_tp =
-                    &pending->elems[i];
+                    &rkcg->rkcg_errored_topics->elems[i];
 
                 switch (pending_errored_tp->err) {
                 case RD_KAFKA_RESP_ERR_TOPIC_EXCEPTION:
                 case RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED:
-                        rd_kafka_topic_partition_list_add_with_topic_name_and_id(
-                            to_surface,
-                            rd_kafka_topic_partition_get_topic_id(
-                                pending_errored_tp),
-                            pending_errored_tp->topic, RD_KAFKA_PARTITION_UA)
-                            ->err = pending_errored_tp->err;
+                        rd_kafka_dbg(rk, CONSUMER | RD_KAFKA_DBG_TOPIC,
+                                     "TOPICERR",
+                                     "Subscribed topic not available: %s: %s",
+                                     pending_errored_tp->topic,
+                                     rd_kafka_err2str(pending_errored_tp->err));
+                        rd_kafka_consumer_err(
+                            rkcg->rkcg_q, RD_KAFKA_NODEID_UA,
+                            pending_errored_tp->err, 0,
+                            pending_errored_tp->topic, NULL,
+                            RD_KAFKA_OFFSET_INVALID,
+                            "Subscribed topic not available: %s: %s",
+                            pending_errored_tp->topic,
+                            rd_kafka_err2str(pending_errored_tp->err));
                         break;
 
                 default:
@@ -5807,30 +5758,10 @@ void rd_kafka_share_topic_err_propagate(rd_kafka_cgrp_t *rkcg) {
                         break;
                 }
         }
-        rd_kafka_topic_partition_list_destroy(pending);
 
-        /* Carry previously-known errored entries forward so propagate's
-         * wholesale-replace doesn't drop them. */
-        if (rkcg->rkcg_errored_topics) {
-                for (j = 0; j < rkcg->rkcg_errored_topics->cnt; j++) {
-                        rd_kafka_topic_partition_t *existing =
-                            &rkcg->rkcg_errored_topics->elems[j];
-                        rd_kafka_Uuid_t existing_id =
-                            rd_kafka_topic_partition_get_topic_id(existing);
-
-                        if (rd_kafka_topic_partition_list_find_topic_by_id(
-                                to_surface, existing_id))
-                                continue; /* pending overrides existing */
-
-                        rd_kafka_topic_partition_list_add_with_topic_name_and_id(
-                            to_surface, existing_id, existing->topic,
-                            RD_KAFKA_PARTITION_UA)
-                            ->err = existing->err;
-                }
-        }
-
-        rd_kafka_propagate_consumer_topic_errors(
-            rkcg, to_surface, "Subscribed topic not available");
+        /* Clear the per-cycle accumulator; next cycle starts fresh. */
+        rd_kafka_topic_partition_list_destroy(rkcg->rkcg_errored_topics);
+        rkcg->rkcg_errored_topics = rd_kafka_topic_partition_list_new(0);
 }
 
 
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -5748,25 +5748,22 @@ void rd_kafka_cgrp_share_clear_topic_err(rd_kafka_cgrp_t *rkcg,
  *        (TOPIC_EXCEPTION, TOPIC_AUTHORIZATION_FAILED) are handed to
  *        rd_kafka_propagate_consumer_topic_errors for delivery and
  *        cross-cycle dedup; transient codes are debug-logged at most
- *        once per (topic, err) per cycle. propagate is called even on
- *        an empty surface list so recovered or unsubscribed topics
- *        are dropped from the dedup state.
+ *        once per (topic, err) per cycle. Existing rkcg_errored_topics
+ *        entries are always merged forward so propagate's wholesale-
+ *        replace cannot drop entries for topics absent from this cycle.
  *
- * @param should_merge_existing_errored true when this metadata response
- *        is a partial view of the subscription and previously-known
- *        errored topics must be carried forward to survive the
- *        wholesale-replace in propagate. False on a full-subscription
- *        view, where the response is authoritative.
+ *        After propagate, entries whose rkt has transitioned to
+ *        RD_KAFKA_TOPIC_S_NOTEXISTS are pruned from the dedup memo.
+ *        Recovery (back to S_EXISTS) is handled separately by
+ *        rd_kafka_cgrp_share_clear_topic_err on set_exists.
  *
  * @locality rdkafka main thread
  */
-void rd_kafka_cgrp_share_metadata_update_check(
-    rd_kafka_cgrp_t *rkcg,
-    rd_bool_t should_merge_existing_errored) {
+void rd_kafka_share_topic_err_propogate(rd_kafka_cgrp_t *rkcg) {
         rd_kafka_t *rk = rkcg->rkcg_rk;
         rd_kafka_topic_partition_list_t *pending;
         rd_kafka_topic_partition_list_t *to_surface;
-        int i;
+        int i, j;
 
         if (!rkcg->rkcg_share_topic_errored)
                 return;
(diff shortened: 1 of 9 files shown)


src/rdkafka_partition.c
```diff
@@ -2545,14 +2557,18 @@ void rd_kafka_toppar_enq_error(rd_kafka_toppar_t *rktp,
...
-         * UNKNOWN_PARTITION etc. from the metadata-refresh path in
-         * rdkafka_topic.c) may still be useful for share-consumer
-         * apps; verify and consider routing to rk_rep for share
-         * consumer instead. */
-        rd_kafka_q_enq(rktp->rktp_fetchq, rko);
+        /* Share consumers don't drain rktp_fetchq — route topic-level
+         * errors (UNKNOWN_TOPIC_OR_PART / UNKNOWN_PARTITION etc.
+         * surfaced from the metadata-refresh path in rdkafka_topic.c)
+         * to the cgrp queue so they reach the app via consume_batch. */
+        {
```
pranavrth: No inner braces.

src/rdkafka_partition.c
```diff
@@ -2545,14 +2557,18 @@ void rd_kafka_toppar_enq_error(rd_kafka_toppar_t *rktp,
...
+                rd_kafka_t *rk = rktp->rktp_rkt->rkt_rk;
+                rd_kafka_q_t *rkq =
+                    RD_KAFKA_IS_SHARE_CONSUMER(rk) && rk->rk_cgrp
+                        ? rk->rk_cgrp->rkcg_q
+                        : rktp->rktp_fetchq;
+                rd_kafka_q_enq(rkq, rko);
```
pranavrth: - No need for variable
- Check only IS_SHARE_CONSUMER and not rk->rk_cgrp. In future if someone changes anything then it might break.
- I would prefer we don't touch normal consumer path.

```
if(IS_SHARE_CONSUMER) {
    rd_kafka_q_enq(rk->rk_cgrp->rkcg_q, rko);
} else {
    rd_kafka_q_enq(rktp->rktp_fetchq, rko);
}
```

src/rdkafka_partition.c
pranavrth, line 2588: - It needs to be CONSUMER_ERR for our case. OP_ERR is not going to work.
- We just need to send TOPIC_EXCEPTION and TOPIC_AUTHORIZATION_ERROR to the user. We can log other errors or call this function from the correct places only. We log other errors only. Check java for parity on this path.
- Please add a deleted-topic share-consumer test covering the chosen behavior.

pranavrth, line 2577: With multiple changes, I think you should just create a new function altogether for share consumer.

src/rdkafka_partition.c
```diff
@@ -1138,8 +1138,13 @@ static void rd_kafka_toppar_broker_migrate(rd_kafka_toppar_t *rktp,
...
-        if (rktp->rktp_fetch_state == RD_KAFKA_TOPPAR_FETCH_OFFSET_WAIT)
+         * to time out..slowly).
+         *
+         * Share consumers never enter OFFSET_WAIT so this branch is
+         * unreachable for them today; guard regardless so the offset
+         * machinery stays out of the share leader-change path. */
+        if (rktp->rktp_fetch_state == RD_KAFKA_TOPPAR_FETCH_OFFSET_WAIT &&
+            !RD_KAFKA_IS_SHARE_CONSUMER(rktp->rktp_rkt->rkt_rk))
                 rd_kafka_toppar_offset_retry(rktp, 500,
                                              "migrating to new broker");
```
pranavrth: keep !RD_KAFKA_IS_SHARE_CONSUMER first condition. We don't want to touch the state part for share consumer. This is not hot path so it is fine.

pranavrth (changes requested):
Still checking if we didn't miss anything.

src/rdkafka_partition.c
```diff
@@ -2525,6 +2537,81 @@ rd_kafka_toppars_pause_resume(rd_kafka_t *rk,
...
+
+        rd_snprintf(buf, sizeof(buf), "%.*s [%" PRId32 "]: %s (%s)",
+                    RD_KAFKAP_STR_PR(rktp->rktp_rkt->rkt_topic),
+                    rktp->rktp_partition, reason, rd_kafka_err2str(err));
+
+        switch (err) {
+        case RD_KAFKA_RESP_ERR_TOPIC_EXCEPTION:
+        case RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED:
+                rd_kafka_share_toppar_enq_consumer_err(rktp, err, buf);
+                return;
```
pranavrth: We need to rate limit this. Check Java on how it is rate limited. Java sends once per topic (instead of per partition) in a certain time frame.

I will verify this for ShareAcknowledge and ShareFetch RPC path as well.

src/rdkafka_partition.c
```diff
@@ -2525,6 +2537,81 @@ rd_kafka_toppars_pause_resume(rd_kafka_t *rk,
...
+
+        case RD_KAFKA_RESP_ERR__UNKNOWN_TOPIC:
+        case RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION:
+        case RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART:
+                rd_kafka_log(rk, LOG_WARNING, "TOPICERR",
+                             "Received unknown topic or partition error "
+                             "for partition %.*s [%" PRId32 "]",
+                             RD_KAFKAP_STR_PR(rktp->rktp_rkt->rkt_topic),
+                             rktp->rktp_partition);
```
pranavrth: I think it should be debug level log or at max 1 warning log. Java also have debug logging for this case.

src/rdkafka_partition.c
```diff
@@ -2525,6 +2537,81 @@ rd_kafka_toppars_pause_resume(rd_kafka_t *rk,
...
+                return;
+
+        case RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_ID:
+                rd_kafka_log(rk, LOG_WARNING, "TOPICERR",
+                             "Received unknown topic ID error "
+                             "for partition %.*s [%" PRId32 "]",
+                             RD_KAFKAP_STR_PR(rktp->rktp_rkt->rkt_topic),
+                             rktp->rktp_partition);
```
pranavrth: Same.

src/rdkafka_partition.c
```diff
@@ -2525,6 +2537,81 @@ rd_kafka_toppars_pause_resume(rd_kafka_t *rk,
...
+        if (!rk->rk_cgrp)
+                return;
+
+        rko                   = rd_kafka_op_new(RD_KAFKA_OP_CONSUMER_ERR);
+        rko->rko_err          = err;
+        rko->rko_rktp         = rd_kafka_toppar_keep(rktp);
+        rko->rko_u.err.errstr = rd_strdup(errstr);
+
+        rd_kafka_q_enq(rk->rk_cgrp->rkcg_q, rko);
+}
```
pranavrth: I think we should use existing functions and shouldn't create a new function. There are somethings related to offset and rkm as well which might break existing APIs. Use the existing function.

pranavrth (changes requested):

pranavrth:
We should add tests to trigger these conditions atleast.

src/rdkafka_partition.c
```diff
@@ -2525,6 +2537,107 @@ rd_kafka_toppars_pause_resume(rd_kafka_t *rk,
...
+                                            const char *reason) {
+        rd_kafka_t *rk = rktp->rktp_rkt->rkt_rk;
+        char buf[512];
+
+        rd_snprintf(buf, sizeof(buf), "%.*s [%" PRId32 "]: %s (%s)",
+                    RD_KAFKAP_STR_PR(rktp->rktp_rkt->rkt_topic),
+                    rktp->rktp_partition, reason, rd_kafka_err2str(err));
+
+        if (rk->rk_cgrp &&
+            rd_kafka_share_topic_err_already_reported(
```
pranavrth: The dedup only ever adds to rkcg_errored_topics (or updates err on a code change) — it never removes an entry when the topic recovers or on unsubscribe. So once a (topic, err) is reported, the same error code is suppressed permanently:

- Topic auth fails → surfaces once → ACLs restored → revoked again → the second TOPIC_AUTHORIZATION_FAILED is never delivered. The app can't tell it re-failed.
- rkcg_errored_topics grows monotonically for a long-lived consumer cycling through topics.

The existing `rd_kafka_propagate_consumer_topic_errors()` (cgrp.c:5660) already handles this correctly — it wholesale-replaces `rkcg_errored_topics` each metadata cycle, so recovered topics drop out and a re-failure re-surfaces (it also does the `_UNKNOWN_TOPIC`→`UNKNOWN_TOPIC_OR_PART` normalization). Suggest reusing/factoring that instead of a parallel add-only dedup.

Note the (topic,err) tests pass because they never recover-then-refail with the same code — please add a test for revoke→restore→revoke (same err) that asserts it surfaces twice after the above fix.

pranavrth (changes requested):

src/rdkafka_partition.c
```diff
@@ -2525,6 +2537,107 @@ rd_kafka_toppars_pause_resume(rd_kafka_t *rk,
...
+ *          rd_false otherwise — records the (topic, err) for future
+ *          calls.
+ */
+static rd_bool_t
+rd_kafka_share_topic_err_already_reported(rd_kafka_cgrp_t *rkcg,
+                                          const char *topic,
+                                          rd_kafka_resp_err_t err) {
+        rd_kafka_topic_partition_t *prev;
+
+        prev = rd_kafka_topic_partition_list_find(rkcg->rkcg_errored_topics,
```
pranavrth: find by id.

pranavrth (changes requested):

src/rdkafka_partition.c
```diff
@@ -2525,6 +2537,37 @@ rd_kafka_toppars_pause_resume(rd_kafka_t *rk,
...
+                                            rd_kafka_resp_err_t err) {
+        rd_kafka_cgrp_t *rkcg = rktp->rktp_rkt->rkt_rk->rk_cgrp;
+        const char *topic     = rktp->rktp_rkt->rkt_topic->str;
+        rd_kafka_topic_partition_t *prev;
+
+        if (!rkcg || !rkcg->rkcg_share_pending_errored)
+                return;
+
+        prev = rd_kafka_topic_partition_list_find(
+            rkcg->rkcg_share_pending_errored, topic, RD_KAFKA_PARTITION_UA);
```
pranavrth:
```suggestion
            rkcg->rkcg_share_topic_errored, topic, RD_KAFKA_PARTITION_UA);
```

src/rdkafka_cgrp.c
```diff
@@ -5703,6 +5709,57 @@ static void rd_kafka_propagate_consumer_topic_errors(
...
+        rkcg->rkcg_share_pending_errored = rd_kafka_topic_partition_list_new(0);
+
+        to_surface = rd_kafka_topic_partition_list_new(0);
+        for (i = 0; i < pending->cnt; i++) {
+                const rd_kafka_topic_partition_t *t = &pending->elems[i];
+
+                switch (t->err) {
+                case RD_KAFKA_RESP_ERR_TOPIC_EXCEPTION:
+                case RD_KAFKA_RESP_ERR_TOPIC_AUTHORIZATION_FAILED:
+                        rd_kafka_topic_partition_list_add(to_surface, t->topic,
```
pranavrth: topic id.

Check that you are using topic id everywhere in the Share consumer flow.

src/rdkafka_cgrp.c
```diff
@@ -7439,6 +7503,14 @@ void rd_kafka_cgrp_metadata_update_check(rd_kafka_cgrp_t *rkcg,
...
+        /* Share consumers handle topic-level errors via the share
+         * variant; the classic rejoin path is not applicable since
+         * membership is driven by ShareGroupHeartbeat. */
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rkcg->rkcg_rk)) {
+                rd_kafka_share_metadata_update_check(rkcg);
+                return;
+        }
+
         if (rkcg->rkcg_group_protocol != RD_KAFKA_GROUP_PROTOCOL_CLASSIC)
                 return;
```
pranavrth, line 7504: Check how is this affecting the overall change.

pranavrth (changes requested):

src/rdkafka_cgrp.c
```diff
@@ -5703,6 +5709,113 @@ static void rd_kafka_propagate_consumer_topic_errors(
...
+
+        if (!rkcg->rkcg_share_topic_errored)
+                return;
+
+        pending                        = rkcg->rkcg_share_topic_errored;
+        rkcg->rkcg_share_topic_errored = rd_kafka_topic_partition_list_new(0);
+
+        to_surface = rd_kafka_topic_partition_list_new(0);
+        for (i = 0; i < pending->cnt; i++) {
+                const rd_kafka_topic_partition_t *t = &pending->elems[i];
```
pranavrth: Better name

src/rdkafka_cgrp.h
```diff
@@ -504,6 +508,10 @@ void rd_kafka_cgrp_coord_dead(rd_kafka_cgrp_t *rkcg,
                               const char *reason);
 void rd_kafka_cgrp_metadata_update_check(rd_kafka_cgrp_t *rkcg,
```
pranavrth, line 516: Function being used internally uses topic by name instead of id. Let's fix that.

`rd_kafka_propagate_consumer_topic_errors`

src/rdkafka_cgrp.c
```diff
@@ -7439,6 +7559,23 @@ void rd_kafka_cgrp_metadata_update_check(rd_kafka_cgrp_t *rkcg,
...
+         * For share, the share variant must additively merge previously-
+         * errored topics forward whenever this metadata response is a
+         * PARTIAL view of the subscription (i.e., the IF-branch
+         * gate in parse_Metadata0 was not taken and we entered this
+         * function via the share fallthrough). That gate's truth value
+         * is what `do_join` carries: do_join=true ↔ full-subscription
+         * view (no merge); do_join=false ↔ partial view (merge). */
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rkcg->rkcg_rk)) {
+                rd_kafka_share_metadata_update_check(
+                    rkcg, !do_join /*should_merge_existing_errored*/);
```
pranavrth: Can you please explain what are we doing here?

pranavrth (changes requested):

src/rdkafka_metadata.c
```diff
@@ -1101,15 +1101,28 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
               rd_kafka_cgrp_same_subscription_version(
-                  rkb->rkb_rk->rk_cgrp, cgrp_subscription_version))))
+                  rkb->rkb_rk->rk_cgrp, cgrp_subscription_version)));
+
+        if (cgrp_update_covers_subscription)
                 rd_kafka_cgrp_metadata_update_check(rkb->rkb_rk->rk_cgrp,
                                                     rd_true /*do join*/);
 
+        if (rk->rk_cgrp && RD_KAFKA_IS_SHARE_CONSUMER(rk))
+                rd_kafka_cgrp_share_metadata_update_check(
```
pranavrth: Let improve the name to make sure what is happening is reflected. Something similar to `rd_kafka_share_topic_err_propogate`

src/rdkafka_cgrp.h
```diff
@@ -228,6 +228,10 @@ typedef struct rd_kafka_cgrp_s {
         rd_list_t *rkcg_subscribed_topics; /**< (rd_kafka_topic_info_t *) */
         /** Subscribed topics that are errored/not available. */
         rd_kafka_topic_partition_list_t *rkcg_errored_topics;
+        /** Share-consumer per-metadata-cycle (topic, err) accumulator.
+         *  Populated by rd_kafka_share_toppar_enq_error and drained
+         *  by rd_kafka_share_topic_err_propogate at cycle end. */
+        rd_kafka_topic_partition_list_t *rkcg_share_topic_errored;
```
pranavrth: Add a TODO to move it to the correct place.

src/rdkafka_cgrp.h
```diff
@@ -504,6 +508,12 @@ void rd_kafka_cgrp_coord_dead(rd_kafka_cgrp_t *rkcg,
                               const char *reason);
 void rd_kafka_cgrp_metadata_update_check(rd_kafka_cgrp_t *rkcg,
                                          rd_bool_t do_join);
+
+void rd_kafka_share_topic_err_propogate(rd_kafka_cgrp_t *rkcg);
+
+void rd_kafka_cgrp_share_clear_topic_err(rd_kafka_cgrp_t *rkcg,
+                                         rd_kafka_Uuid_t topic_id);
```
pranavrth: Add TODO to move to correct place when resolving TODO for `rkcg_share_topic_errored`

src/rdkafka_cgrp.c
```diff
@@ -5703,6 +5709,146 @@ static void rd_kafka_propagate_consumer_topic_errors(
...
+        if (rkcg->rkcg_errored_topics) {
+                for (j = 0; j < rkcg->rkcg_errored_topics->cnt; j++) {
+                        rd_kafka_topic_partition_t *existing =
+                            &rkcg->rkcg_errored_topics->elems[j];
+                        rd_kafka_Uuid_t existing_id =
+                            rd_kafka_topic_partition_get_topic_id(existing);
+
+                        if (rd_kafka_topic_partition_list_find_topic_by_id(
+                                to_surface, existing_id))
+                                continue; /* pending overrides existing */
```
pranavrth: Check for existing and same error as previous.

src/rdkafka_cgrp.c
```diff
@@ -5703,6 +5709,146 @@ static void rd_kafka_propagate_consumer_topic_errors(
...
+        /* Post-propagate cleanup: drop memo entries whose rkt has moved
+         * to S_NOTEXISTS. For those topics the client stops re-asking,
+         * so pending won't refill */
+        if (rkcg->rkcg_errored_topics) {
+                j = 0;
+                while (j < rkcg->rkcg_errored_topics->cnt) {
+                        rd_kafka_topic_partition_t *entry =
+                            &rkcg->rkcg_errored_topics->elems[j];
+                        rd_kafka_topic_t *rkt =
+                            rd_kafka_topic_find(rk, entry->topic, 1 /*lock*/);
```
pranavrth: find by id.

src/rdkafka_cgrp.c
```diff
@@ -5703,6 +5709,146 @@ static void rd_kafka_propagate_consumer_topic_errors(
...
+                                rd_kafka_topic_destroy0(rkt);
+                        }
+
+                        if (drop)
+                                rd_kafka_topic_partition_list_del_by_idx(
+                                    rkcg->rkcg_errored_topics, j);
+                        else
+                                j++;
+                }
+        }
```
pranavrth: Remove this.

pranavrth (changes requested):

src/rdkafka_cgrp.c
```diff
@@ -5703,6 +5717,123 @@ static void rd_kafka_propagate_consumer_topic_errors(
...
+
+                        rd_kafka_topic_partition_list_add_with_topic_name_and_id(
+                            to_surface, existing_id, existing->topic,
+                            RD_KAFKA_PARTITION_UA)
+                            ->err = existing->err;
+                }
+        }
+
+        rd_kafka_propagate_consumer_topic_errors(
+            rkcg, to_surface, "Subscribed topic not available");
```
pranavrth: Change the reason.

pranavrth (approved):
Added a nit to add in next commit. Otherwise LGTM!. Thanks @PratRanj07 !.

src/rdkafka_cgrp.c
```diff
@@ -5703,6 +5703,60 @@ static void rd_kafka_propagate_consumer_topic_errors(
...
+                                     pending_errored_tp->topic,
+                                     rd_kafka_err2str(pending_errored_tp->err));
+                        rd_kafka_consumer_err(
+                            rkcg->rkcg_q, RD_KAFKA_NODEID_UA,
+                            pending_errored_tp->err, 0,
+                            pending_errored_tp->topic, NULL,
+                            RD_KAFKA_OFFSET_INVALID,
+                            "Subscribed topic not available: %s: %s",
+                            pending_errored_tp->topic,
+                            rd_kafka_err2str(pending_errored_tp->err));
```
pranavrth, line 5743: We need to throw it max twice (1 per error) per metadata response. Java sends the set of topic instead of sending the error topic number of times.

Maybe add a TODO for now.

We might not even do that as java exception can carry set of partitions but we can't have those. We might try to add but adding TODO is sufficient for now.

pranavrth (approved):
LGTM!. Thanks @PratRanj07 !.

PR #5471, opened 2026-05-31
[KIP-932] Improve share consumer ShareGroupHeartbeat error handling

pranavrth:
- Handle FENCED_MEMBER_EPOCH on ShareGroupHeartbeat by abandoning the assignment and rejoining the group, same as UNKNOWN_MEMBER_ID.
- Treat GROUP_ID_NOT_FOUND as fatal on a non-leave heartbeat (the leave path and the leaving-member case remain benign).
- Surface the specific fatal error code and message, with the fatal flag set, from rd_kafka_share_consume_batch instead of the generic local fatal sentinel; mark non-fatal share-consumer errors retriable.
- Re-enable the stable GROUP_ID_NOT_FOUND fatal subtest in 0155 and rework wait_fatal_error to poll consume_batch (returning the error object), asserting no records arrive and any error received is fatal.
- Add TODOs covering: unknown/unexpected heartbeat error codes, the regular consumer heartbeat GROUP_ID_NOT_FOUND path, retriable errors on the remaining consume_batch paths, session-close/leave behaviour on close after a fatal error, and whether all APIs should be gated once a fatal error is raised.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -939,6 +939,10 @@ int rd_kafka_set_fatal_error0(rd_kafka_t *rk,
          * consumer error so it is returned from consumer_poll(),
          * while for all other client types (the producer) we propagate to
          * the standard error handler (typically error_cb). */
+        /* TODO KIP-932: when a fatal error has been raised, check what the
+         * Java client does in the fatal-error case and decide whether we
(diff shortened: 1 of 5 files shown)


PratRanj07 (approved):
LGTM

PR #5472, opened 2026-06-01
[KIP-932] Improved fetch-broker selection and ShareFetch UNKNOWN_TOPIC_* handling on topic delete

pranavrth:
Hardens the share consumer when an assigned topic is deleted.

- **Fetch-broker selection**: `rd_kafka_share_select_broker()` now iterates the
  assigned toppar list (`rkcg_toppars`) instead of resolving each partition via
  `rd_kafka_toppar_get2()`. The old path marked deleted partitions *desired*
  (causing repeated forced metadata refreshes) and could fault on a missing
  cached toppar. The poll-loop re-trigger gate uses the same list.

- **ShareFetch UNKNOWN_TOPIC_***: per-partition `UNKNOWN_TOPIC_OR_PART` /
  `UNKNOWN_TOPIC_ID` / `INCONSISTENT_TOPIC_ID` are now debug-logged with no
  metadata refresh (left to the next refresh / heartbeat). Top-level errors do a
  non-forced refresh only on `UNKNOWN_TOPIC_ID`.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2385,19 +2385,14 @@ static int rd_kafka_thread_main(void *arg) {
                         rd_kafka_cgrp_serve(rk->rk_cgrp);
 
                 /* KIP-932: If share_fetch_more_records is set but no fetch
-                 * op is in-flight and assignments exist, re-trigger
-                 * the fetch flow by selecting a broker directly.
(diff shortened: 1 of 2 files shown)


ojasvajain (approved):
Looks good. Can see failures in 0174 and 0178 on CI. I am assuming they will be fixed in subsequent PRs merged by me and @PratRanj07 .

PR #5475, opened 2026-06-02
[KIP-932]: Prevent Share Session Lists to be mutated for internal broker thread

ojasvajain:
Internal broker thread does not make any ShareFetch and ShareAck network calls to the brokers so we shouldn't need to populate the `toppars_to_add` or `toppars_to_forget` lists when partitions migrate to/away from internal broker thread.
This PR adds check to update the lists only if the broker is not internal. It also removes the part where we enqueue SESSION_CLEAR op to the internal broker while destroying the client.

TODO - ~~Check if we should use `buf_enq2` instead of `buf_enq1` while enqueuing ShareAck and ShareFetch requests~~ Checked this - enq2 does NOT finalize the buffer. It assumes the caller already did. enq2 does NOT set the response callback or opaque. It assumes the rkbuf already has them. So it is probably not worth it to do it at the caller given that we ensure that we don't enqueue requests on internal thread.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3471,6 +3471,10 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                 rd_kafka_broker_lock(rkb);
                 TAILQ_INSERT_TAIL(&rkb->rkb_toppars, rktp, rktp_rkblink);
                 rkb->rkb_toppar_cnt++;
+                /* TODO KIP-932: Check if we can use
+                 * rkb->rkb_source == RD_KAFKA_LEARNED instead of !=
+                 * RD_KAFKA_INTERNAL.
+                 */
                 if (RD_KAFKA_IS_SHARE_CONSUMER(rkb->rkb_rk) &&
                     rd_kafka_toppar_is_on_cgrp(rktp, rd_false) &&
                     rkb->rkb_source != RD_KAFKA_INTERNAL) {
@@ -3583,6 +3587,10 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                 rd_kafka_broker_lock(rkb);
                 TAILQ_REMOVE(&rkb->rkb_toppars, rktp, rktp_rkblink);
                 rkb->rkb_toppar_cnt--;
+                /* TODO KIP-932: Check if we can use
+                 * rkb->rkb_source == RD_KAFKA_LEARNED instead of !=
+                 * RD_KAFKA_INTERNAL.
+                 */
                 if (RD_KAFKA_IS_SHARE_CONSUMER(rkb->rkb_rk) &&
                     rkb->rkb_source != RD_KAFKA_INTERNAL) {
                         rd_rkb_dbg(rkb, FETCH | RD_KAFKA_DBG_CGRP,
@@ -3594,7 +3602,6 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                                    rktp->rktp_partition);
                         rd_kafka_broker_share_session_toppar_remove(rkb, rktp);
                 }
-
 
                 rd_kafka_broker_unlock(rkb);
                 rd_kafka_broker_destroy(rktp->rktp_broker);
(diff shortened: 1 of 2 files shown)


src/rdkafka_broker.c
```diff
@@ -3581,15 +3583,19 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
...
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rkb->rkb_rk) &&
+                    rkb->rkb_source != RD_KAFKA_INTERNAL) {
                         rd_rkb_dbg(rkb, FETCH | RD_KAFKA_DBG_CGRP,
                                    "SHARESESSION",
                                    "%s [%" PRId32
                                    "]: share session LEAVE via "
                                    "PARTITION_LEAVE",
                                    rktp->rktp_rkt->rkt_topic->str,
                                    rktp->rktp_partition);
-                rd_kafka_broker_share_session_toppar_remove(rkb, rktp);
```
pranavrth, line 3592: IS_SHARE check is already present inside these functions. You can omit is share part. From here, just check if this is not internal thread.

src/rdkafka_broker.c
```diff
@@ -3817,7 +3825,9 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                            rko->rko_rktp->rktp_partition,
                            rkb->rkb_share_fetch_session.epoch);
 
-                rd_kafka_broker_share_session_toppar_remove(rkb, rko->rko_rktp);
+                if (rkb->rkb_source != RD_KAFKA_INTERNAL)
```
pranavrth, line 3855: Add a todo to skip from caller itself

pranavrth (changes requested):
minor

src/rdkafka_broker.c
ojasvajain: I kept it because of the log. If I keep only the internal thread check, then this log will be printed for regular consumer as well.

src/rdkafka_broker.c
ojasvajain: Done

src/rdkafka_broker.c
```diff
@@ -3581,15 +3583,19 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
...
                                    "SHARESESSION",
                                    "%s [%" PRId32
                                    "]: share session LEAVE via "
                                    "PARTITION_LEAVE",
                                    rktp->rktp_rkt->rkt_topic->str,
                                    rktp->rktp_partition);
-                rd_kafka_broker_share_session_toppar_remove(rkb, rktp);
+                        rd_kafka_broker_share_session_toppar_remove(rkb, rktp);
+                }
+
```
pranavrth, line 3605: Remove 1 extra line break.

src/rdkafka_broker.c
```diff
@@ -3806,7 +3812,10 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                            rko->rko_rktp->rktp_partition,
                            rkb->rkb_share_fetch_session.epoch);
 
-                rd_kafka_broker_share_session_toppar_add(rkb, rko->rko_rktp);
+                /* TODO KIP-932: Skip this from caller itself */
+                if (rkb->rkb_source != RD_KAFKA_INTERNAL)
```
pranavrth: "Received…" log still fires on internal then silently no-ops.
Emit a "skipping on internal" variant. Same for REMOVE.

src/rdkafka_broker.c
```diff
@@ -3581,15 +3583,19 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                 rd_kafka_broker_lock(rkb);
                 TAILQ_REMOVE(&rkb->rkb_toppars, rktp, rktp_rkblink);
                 rkb->rkb_toppar_cnt--;
-                if (RD_KAFKA_IS_SHARE_CONSUMER(rkb->rkb_rk))
+                if (RD_KAFKA_IS_SHARE_CONSUMER(rkb->rkb_rk) &&
+                    rkb->rkb_source != RD_KAFKA_INTERNAL) {
```
pranavrth, line 3595: Add a TODO to use `== RD_KAFKA_LEARNED` after validation. Reference same for other places from this TODO.

src/rdkafka_broker.c
pranavrth, line 3814: Add TODO to not even send this op from the caller itself. Caller can check if the broker is internal or learned.

pranavrth (changes requested):

src/rdkafka_broker.c
ojasvajain: Added at all 4 places

src/rdkafka_broker.c
ojasvajain: Added

pranavrth (approved):
LGTM!. Thanks @ojasvajain !

PR #5477, opened 2026-06-04
[KIP-932] Add ASAN/TSAN/Valgrind CI pipelines for share consumer tests

Ankith-Confluent:
Adds three Semaphore CI pipelines that run the share consumer test suite
(tests 0155+) under runtime sanitizers and memory checkers, to catch
leaks, data races, and memory errors introduced by the KIP-932 work.

- **ASAN + LSAN** — `.semaphore/run-asan-tests.yml`
- **TSAN** — `.semaphore/run-tsan-tests.yml`
- **Valgrind (memcheck)** — `.semaphore/run-valgrind-tests.yml`

### DIFF
--- a/.semaphore/run-asan-tests.yml
+++ b/.semaphore/run-asan-tests.yml
@@ -29,7 +29,7 @@ global_job_config:
       value: '8.0.0'
 
 blocks:
-  - name: 'Share consumer tests under ASAN+LSAN'
+  - name: 'Share consumer tests (ASAN + LSAN)'
     dependencies: []
     task:
       agent:
@@ -40,7 +40,7 @@ blocks:
           commands:
             - artifact push workflow artifacts/ --destination artifacts/asan/
       jobs:
-        - name: 'ASAN+LSAN: TESTS_SKIP_BEFORE=0155'
+        - name: 'Share consumer suite (ASAN + LSAN)'
           env_vars:
             - name: ASAN_OPTIONS
               value: 'detect_leaks=1:halt_on_error=0:print_stacktrace=1:abort_on_error=0:symbolize=1'
--- a/.semaphore/run-asan-tests.yml
+++ b/.semaphore/run-asan-tests.yml
@@ -1,0 +1,55 @@
+version: v1.0
+name: run-share-consumer-tests-asan
+
+agent:
+  machine:
+    type: s1-prod-ubuntu24-04-amd64-00
+
+execution_time_limit:
+  hours: 4
+
+global_job_config:
+  prologue:
+    commands:
+      - checkout
+      - '[[ -z "$GIT_REF" ]] || git checkout $GIT_REF'
+      - sudo apt remove -y needrestart
+      - sudo DEBIAN_FRONTEND=noninteractive apt install -y libcurl4-openssl-dev libssl-dev
+      - python3 -m pip install -U pip
+      - (cd tests && python3 -m pip install -r requirements.txt)
+      - ./dev-conf.sh asan
+      - sem-version java 17
+      - mkdir -p artifacts
+  env_vars:
+    - name: CACHE_TAG
+      value: '7'
+    - name: KAFKA_VERSION
+      value: '4.2.0'
+    - name: CP_VERSION
+      value: '8.0.0'
+
+blocks:
+  - name: 'Share consumer tests under ASAN+LSAN'
+    dependencies: []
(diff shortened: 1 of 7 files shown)


src/rdkafka_broker.c
```diff
@@ -739,6 +739,12 @@ void rd_kafka_broker_fail(rd_kafka_broker_t *rkb,
                 }
                 rd_kafka_toppar_unlock(rktp);
 
+                /* TODO: rktp_leader_id and rktp_broker_id are read
```
pranavrth: TODO KIP-932:

src/rdkafka_cgrp.h
```diff
@@ -126,6 +126,11 @@ typedef struct rd_kafka_cgrp_s {
         rd_kafka_q_t *rkcg_q;            /* Application poll queue */
         rd_kafka_q_t *rkcg_ops;          /* Manager ops queue */
         rd_kafka_q_t *rkcg_wait_coord_q; /* Ops awaiting coord */
+        /* TODO: rkcg_flags is written by the main thread (heartbeat
```
pranavrth: TODO KIP-932:

.semaphore/run-asan-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+          env_vars:
+            - name: ASAN_OPTIONS
+              value: 'halt_on_error=0:abort_on_error=0:exitcode=0:print_stacktrace=1:symbolize=1'
+          commands:
+            - cache restore trivup-kafka-${KAFKA_VERSION}-${CACHE_TAG}
+            - (cd tests && python3 -m trivup.clusters.KafkaCluster
+              --kraft
+              --version "${KAFKA_VERSION}"
+              --conf '["group.share.min.record.lock.duration.ms=1000"]'
+              --cmd "TESTS_SKIP_BEFORE=0155 make 2>&1 | tee asan-share-tests.log") || true
```
pranavrth: run-test.sh has asan and tsan mode as well. Have you checked?

.semaphore/run-valgrind-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+            - artifact push workflow artifacts/ --destination artifacts/valgrind/
+      jobs:
+        - name: 'Share consumer suite (Valgrind memcheck)'
+          commands:
+            - cache restore trivup-kafka-${KAFKA_VERSION}-${CACHE_TAG}
+            - (cd tests && python3 -m trivup.clusters.KafkaCluster
+              --kraft
+              --version "${KAFKA_VERSION}"
+              --conf '["group.share.min.record.lock.duration.ms=1000"]'
+              --cmd "TESTS_SKIP_BEFORE=0155 ./run-test.sh valgrind 2>&1 | tee valgrind-share-tests.log") || true
```
pranavrth: I think we need to change some of the properties in this script.

--------------- 



While developing `commit_sync`, I ran test `0176` under `./run-test.sh valgrind` and got a clean result — 0 definitely lost bytes. The leak existed but was invisible due to how two valgrind flags interact in `run-test.sh`:


`--errors-for-leak-kinds=all` treats all leak categories as errors, including "still reachable" blocks.
`--exit-on-first-error=yes` stops valgrind after the first error it encounters.


OpenSSL throws a "still reachable" block, and valgrind treated it as an error and exited immediately — hiding the actual "definitely lost" blocks that would have been reported later.

Updates you can do to `./run-test.sh` to fix this issue:
• Change `--errors-for-leak-kinds=all` to `--errors-for-leak-kinds=definite,possible` so "still reachable" blocks no longer trigger the early exit.
• Change `--exit-on-first-error=yes` to `--exit-on-first-error=no` so `valgrind` reports the full leak summary instead of stopping at the first one.

After these changes, valgrind correctly reports all real leaks.

pranavrth (changes requested):

src/rdkafka_broker.c
Ankith-Confluent: Sure, but this is not really part of kip 932 right?
These existed before itself.

.semaphore/run-asan-tests.yml
Ankith-Confluent: I have checked run-test.sh, it has valgrind mode but not ASAN and TSAN

pranavrth (approved):

PR #5478, opened 2026-06-04
[KIP-932]: Modify close() API to send leave even when fatal error is set

ojasvajain:
1. Currently if close() is called after a fatal error has been set, close() would return early to return the same fatal error. This behaviour is different from Java's Share consumer where session leave request is sent even if fatal error has been returned earlier. This PR modifies close() API to mimic Java's behaviour.

2. Destroy() API is untouched and a TODO is added to evalute it later.

3. Test cases have been added to verify fatal error handling in consume_batch and close.

### DIFF
--- a/tests/0155-share_group_heartbeat_mock.c
+++ b/tests/0155-share_group_heartbeat_mock.c
@@ -568,14 +568,11 @@ static void do_test_share_group_error_injection(void) {
                  rd_kafka_error_string(error));
         rd_kafka_error_destroy(error);
 
-        /* Cleanup. Consumer is in fatal state, so close is expected
-         * to return the stored fatal error. */
+        /* Cleanup. Consumer is in fatal state, but close() still flushes
+         * pending acks and leaves the share session, so it succeeds. */
         error = rd_kafka_share_consumer_close(share_c);
-        TEST_ASSERT(rd_kafka_error_code(error) ==
-                        RD_KAFKA_RESP_ERR_INVALID_REQUEST,
-                    "Expected close to return INVALID_REQUEST, got %s",
-                    rd_kafka_err2name(rd_kafka_error_code(error)));
-        rd_kafka_error_destroy(error);
+        TEST_ASSERT(!error, "Expected close to succeed, got %s",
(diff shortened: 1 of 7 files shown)


tests/0155-share_group_heartbeat_mock.c
```diff
@@ -2069,14 +2060,11 @@ static void do_test_group_id_not_found_while_stable_is_fatal(void) {
...
+        /* Cleanup. Consumer is in fatal state, but close() still flushes
+         * pending acks and leaves the share session, so it succeeds. */
         error = rd_kafka_share_consumer_close(share_c);
-        TEST_ASSERT(rd_kafka_error_code(error) ==
-                        RD_KAFKA_RESP_ERR_GROUP_ID_NOT_FOUND,
-                    "Expected close to return GROUP_ID_NOT_FOUND, got %s",
-                    rd_kafka_err2name(rd_kafka_error_code(error)));
-        rd_kafka_error_destroy(error);
+        TEST_ASSERT(!error, "Expected close to succeed, got %s",
+                    rd_kafka_error_name(error));
```
pranavrth, line 2067: Does Java send No error in this case as well? I think Java sends first error that it got in HB leave or session leave path from close.

pranavrth (changes requested):
LGTM!.

tests/0155-share_group_heartbeat_mock.c
ojasvajain: 1. If fatal error is already set before calling close() - In Java, we have an explicit check to see if member is in fatal state, if yes, HB leave request is not sent. Only Acks are flushed. Any error arising due to session leave request is ignored and only logged.

2. If fatal error is returned in HB leave request on calling close() - In Java, the future is resolved and member is still moved to fatal state. No errors are thrown to the user. Only logged. Acks are still flushed.

pranavrth (approved):
LGTM!. Thanks @ojasvajain !.

pranavrth (approved):
LGTM!. Thanks @ojasvajain !.

PR #5479, opened 2026-06-06
[KIP-932]: Add subscription gaurd while polling

PratRanj07:
Add subscription gaurd while polling
Treat empty list as unsubscribe instead of invalid arg (Java's parity)
Add tests to check the above behaviour

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3034,6 +3034,10 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
         rkshare->rkshare_unacked_cnt      = 0;
         rkshare->rkshare_consumer_closing = rd_false;
 
+        /**
+         * TODO KIP-932: Check for uniform naming structure for all
+         * share consumer variables.
+         */
         rk->rk_share_consumer.share_acknowledgement_commit_cb      = NULL;
         rk->rk_share_consumer.acknowledgement_commit_cb_opaque     = NULL;
         rk->rk_share_consumer.in_callback                          = rd_false;
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3882,10 +3882,10 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
                 goto done;
 
         /* Fail early if no subscription is active.
-         * RD_KAFKA_CGRP_F_SUBSCRIPTION is set by cgrp_subscription_set
-         * only when the subscription list is non-NULL, and cleared on
-         * unsubscribe or subscribe with an empty topic list. */
-        if (unlikely(!(rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION))) {
+         * rkshare_subscribed is app-thread-owned and set/cleared by
+         * rd_kafka_share_subscribe / rd_kafka_share_unsubscribe under
+         * the same rkshare_acquire/release lock that guards this call. */
+        if (unlikely(!rkshare->rkshare_subscribed)) {
                 error = rd_kafka_error_new(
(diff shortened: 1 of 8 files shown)


src/rdkafka.c
```diff
@@ -3881,6 +3881,17 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
         if (error)
                 goto done;
 
+        /* Fail early if no subscription is active.
+         * RD_KAFKA_CGRP_F_SUBSCRIPTION is set by cgrp_subscription_set
+         * only when the subscription list is non-NULL, and cleared on
+         * unsubscribe or subscribe with an empty topic list. */
+        if (unlikely(!(rkcg->rkcg_flags & RD_KAFKA_CGRP_F_SUBSCRIPTION))) {
```
pranavrth: This is main thread flag. Its not safe to use here.

pranavrth (changes requested):

tests/0182-share_consumer_error_handling_mock.c
```diff
@@ -2451,13 +2451,10 @@ static void test_share_consumer_resubscribe_re_emits_persistent_failure(void) {
...
                     "first TOPIC_EXCEPTION must surface");
 
-        /* Phase 2: unsubscribe. No surface. */
+        /* Phase 2: unsubscribe. */
         TEST_ASSERT(rd_kafka_share_unsubscribe(rkshare) ==
                         RD_KAFKA_RESP_ERR_NO_ERROR,
                     "unsubscribe");
-        share_topic_err_force_metadata(rkshare);
-        share_topic_err_assert_no_err(
-            rkshare, 5, "no error must surface while unsubscribed");
```
pranavrth, line 2460: Why this removed?

tests/0170-share_consumer_subscription.c
```diff
@@ -1482,6 +1522,143 @@ static void do_test_subscribe_caret_treated_as_literal_e2e(void) {
...
+                            rd_kafka_error_string(err));
+                rd_kafka_error_destroy(err);
+        }
+        for (i = 0; i < rcvd; i++)
+                rd_kafka_message_destroy(batch[i]);
+
+        /* Case 5: subscribe([]) is equivalent to unsubscribe — must
+         * return NO_ERROR and clear the F_SUBSCRIPTION flag, so the
+         * next consume_batch returns __STATE. */
+        {
```
pranavrth: No inner blocks. Check in other file as well. 182 as well has it.

src/rdkafka_subscription.c
```diff
@@ -183,6 +191,8 @@ rd_kafka_share_subscribe(rd_kafka_share_t *rkshare,
 
         err = rd_kafka_op_err_destroy(
             rd_kafka_op_req(rkcg->rkcg_ops, rko, RD_POLL_INFINITE));
+        if (!err)
+                rkshare->rkshare_subscribed = rd_true;
```
pranavrth, line 201: It can only return FATAL error from the main thread. In which case it unsubscribes. I am thinking if we should change the flag to false when FATAL error is received? Maybe we keep it as true which will let the user go through the normal consumer flow.

Let's add a TODO to check in future instead.

pranavrth (changes requested):
Minor nits.

tests/0182-share_consumer_error_handling_mock.c
PratRanj07: We were calling consume batch after unsubscribe, so it will give us an error, we can check that error but evidently there is no need for it, it doesnt meet with what the test is actually trying to do

pranavrth (approved):
LGTM!. Thanks @PratRanj07 !.

PR #5480, opened 2026-06-06
[KIP-932] Fix topic recreation issue in share consumer flow

pranavrth:
Fix topic recreation issue in share consumer flow. Though this is present because of non compliance with KIP-516 across all the client type, this PR tries to fix the issue for share consumer flow.


src/rdkafka_partition.h
```diff
@@ -612,6 +612,14 @@ rd_kafka_toppar_t *rd_kafka_toppar_get2(rd_kafka_t *rk,
                                         int32_t partition,
                                         int ua_on_miss,
                                         int create_on_miss);
+#define rd_kafka_toppar_get_by_id(rkt, partition)                              \
+        rd_kafka_toppar_get0(__FUNCTION__, __LINE__, (rkt), (partition),       \
+                             0 /*no ua-on-miss*/)
```
PratRanj07: Why have we added this macro? It is not being used

tests/0184-share_consumer_topic_recreate.c
```diff
@@ -0,0 +1,1061 @@
...
+
+        test_share_consumer_close(rkshare);
+        test_share_destroy(rkshare);
+
+        SUB_TEST_PASS();
+}
+
+
+/**
+ * @brief TC#6 — Recreate with a smaller partition count, HB-first regime.
```
PratRanj07: Inconsistent test case naming

tests/0184-share_consumer_topic_recreate.c
```diff
@@ -0,0 +1,1061 @@
...
+ */
+static rd_kafka_share_t *create_consumer_md_first(const char *group_id) {
+        rd_kafka_share_t *rkshare;
+        rd_kafka_conf_t *conf;
+        char errstr[512];
+
+        test_conf_init(&conf, NULL, 0);
+
+        test_conf_set(conf, "group.id", group_id);
+        // test_conf_set(conf, "enable.auto.commit", "false");
```
PratRanj07: remove this

tests/0184-share_consumer_topic_recreate.c
```diff
@@ -0,0 +1,1061 @@
...
+        /* TC#7: topic deletion within propagation window. */
+        // do_test_delete_within_propagation_window(
+        //     10000 /* propagation_max_ms */, 3000 /* post-window slack */,
+        //     "0184-delete-within-propagation-10s");
+
+        rd_kafka_destroy(common_admin);
+        rd_kafka_destroy(common_producer);
+
+        return 0;
+}
```
PratRanj07, line 959: add new line

src/rdkafka_metadata.c
```diff
@@ -1527,6 +1541,84 @@ rd_kafka_metadata_refresh_topics(rd_kafka_t *rk,
...
+        rd_list_t q_topic_ids;
+        int destroy_rkb = 0;
+
+        if (!rk) {
+                rd_assert(rkb);
+                rk = rkb->rkb_rk;
+        }
+
+        if (rd_list_cnt(topic_ids) == 0)
+                return RD_KAFKA_RESP_ERR__UNKNOWN_TOPIC;
```
PratRanj07, line 1586: Are you sure this is the correct error with respect to how this error is currently used in the library?

src/rdkafka_metadata.c
```diff
@@ -1527,6 +1527,84 @@ rd_kafka_metadata_refresh_topics(rd_kafka_t *rk,
...
+
+        if (!rkb) {
+                if (!(rkb = rd_kafka_broker_any_usable(
+                          rk, RD_POLL_NOWAIT, RD_DO_LOCK, 0, reason))) {
+                        rd_kafka_dbg(rk, METADATA, "METADATA",
+                                     "Skipping metadata refresh of %d topic "
+                                     "id(s): %s: no usable brokers",
```
PratRanj07: Instead of topic ids we should write topic(s) for more clarity? WDYT?

src/rdkafka_metadata.c
```diff
@@ -1613,6 +1691,24 @@ rd_kafka_metadata_refresh_consumer_topics(rd_kafka_t *rk,
...
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rk)) {
+                /* Share consumer identifies topics by id. Subscription
+                 * names are resolved server-side via
+                 * ShareGroupHeartbeat, so the only thing this
+                 * periodic path needs to do is status-check the
+                 * locally known topic ids. The broker's response
+                 * marks stale ids (e.g. those left behind by a
+                 * delete+recreate) as unknown, which drives cleanup
+                 * of the corresponding rkts. */
+                rd_list_t topic_ids;
```
PratRanj07: Declaration at top

PratRanj07 (changes requested):
Thanks @pranavrth for the PR
provided some nits and suggestions

src/rdkafka_partition.h
pranavrth: I think its doing the same thing as what `get()` does. Removing this.

tests/0184-share_consumer_topic_recreate.c
pranavrth: Removed.

src/rdkafka_metadata.c
pranavrth: Yes this is the error sent by other sibling functions - `metadata_refresh_known_topics` and `metadata_refresh_consumer_topics`

tests/0184-share_consumer_topic_recreate.c
pranavrth: I will keep it as it is for now as these are being updated by @ojasvajain in some other PR.

PratRanj07 (commented):
In rd_kafka_broker_share_fetch_reply in case of Unknown topic ids we are still refreshing with topics **rd_kafka_metadata_refresh_known_topics**. Please check that wont cause any problems

src/rdkafka_topic.c
```diff
@@ -1562,7 +1714,14 @@ int rd_kafka_topic_metadata_update2(
...
         rd_kafka_wrlock(rkb->rkb_rk);
 
-        if (likely(mdt->topic != NULL)) {
+        if (RD_KAFKA_IS_SHARE_CONSUMER(rkb->rkb_rk) &&
+            !RD_KAFKA_UUID_IS_ZERO(mdit->topic_id)) {
+                /* Share consumer identifies topics by id; looking up
+                 * by name risks matching a recreated topic's response
+                 * to a stale rkt that shares its name. */
+                rkt = rd_kafka_topic_find_by_topic_id(rkb->rkb_rk,
+                                                      mdit->topic_id);
```
PratRanj07, line 1723: Should we have a defensive check for if we dont have topic id from result here?

PratRanj07 (commented):
Hey @pranavrth 
I have posted just two concerns
Otherwise I am good with the PR
Lets rebase the PR and run all tests including normal consumers once since its a very big change and we get sure that it doesnt impact normal consumer functioning in any way

src/rdkafka_topic.c
pranavrth: This is the condition which checks this right now - `!RD_KAFKA_UUID_IS_ZERO(mdit->topic_id)`

pranavrth:
> In rd_kafka_broker_share_fetch_reply in case of Unknown topic ids we are still refreshing with topics rd_kafka_metadata_refresh_known_topics. Please check that wont cause any problems

Yes I will recheck this. Though it shouldn't cause any issue from the user's perspective, but we might keep on sending metadata request again and again. This is an optimization that I am planning to fix. Java has the current behavior what I have added.

src/rdkafka_topic.c
PratRanj07: If this happens in our case we would then search by name if its not NULL, but I think for this kip-932 brokers its very unrealistic, so we should be fine

PratRanj07 (approved):
LGTM! Thanks @pranavrth and @ojasvajain !

src/rdkafka_topic.h
```diff
@@ -193,6 +208,19 @@ struct rd_kafka_topic_s {
...
+ *        \p rkt->rkt_topic->str — each layer can be NULL on degenerate
+ *        flows that this helper is meant to absorb.
+ */
+static RD_INLINE RD_UNUSED const char *
+rd_kafka_topic_name_safe(const rd_kafka_topic_t *rkt) {
+        if (!rkt || !rkt->rkt_topic || !rkt->rkt_topic->str)
+                return "(null)";
+        return rkt->rkt_topic->str;
+}
+
```
PratRanj07, line 223: This function is not being used

pranavrth: Yes. I have created this for use in future. Earlier we never had topic name as null. Now we might have. So we should use this function wherever necessary. Right now there is no place where we have created rkt and don't have name.

pranavrth: But in general we should use this instead of using raw `topic->str`.

ojasvajain (approved):
LGTM

PratRanj07 (approved):
LGTM

PR #5481, opened 2026-06-07
[KIP-932]: Remove share fetch fanout with backoff

PratRanj07:
Removed share fetch fanout with backoff as it is no longer used


pranavrth (dismissed):
LGTM!. Thanks @PratRanj07 !

pranavrth (approved):
LGTM!. Thanks @PratRanj07 !

PR #5482, opened 2026-06-08
[KIP-932]: Remove response leader epoch handling from Share consumer implementation

PratRanj07:
Remove response leader epoch handling from Share consumer implementation


pranavrth (approved):
LGTM!. Thanks @PratRanj07 !

PR #5485, opened 2026-06-09
[KIP-848] Fix error handling for consumer group GroupIdNotFound error

izzyharker:
We identified a bug in the ConsumerGroupHeartbeat error handling for GroupIdNotFound 
which can cause consumers to fail silently and get stuck in an infinite loop. 

This fix updates the error handling to trigger a rejoin when GroupIdNotFound is received, matching the
UNKNOWN_MEMBER_ID and FENCED_MEMBER_EPOCH errors. 

Note: This behavior diverges from the Java client, which treats this error as fatal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+# librdkafka v2.14.3
+
+librdkafka v2.14.3 is a maintenance release:
+
+## Fixes
+
+### Consumer fixes
+
+* Fix `GROUP_ID_NOT_FOUND` not being handled in the KIP-848
+  `ConsumerGroupHeartbeat` response, which caused a consumer to silently
+  re-send heartbeats with a stale member epoch indefinitely (without
+  rejoining or surfacing an error) when the coordinator no longer had the
+  group. The error is now treated like `UNKNOWN_MEMBER_ID` /
+  `FENCED_MEMBER_EPOCH`: the consumer rejoins the group, re-creating it on
+  the coordinator.
+
+
 # librdkafka v2.14.2
 
 librdkafka v2.14.2 is a maintenance release:
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -3280,6 +3280,7 @@ err:
 
         case RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID:
         case RD_KAFKA_RESP_ERR_FENCED_MEMBER_EPOCH:
+        case RD_KAFKA_RESP_ERR_GROUP_ID_NOT_FOUND:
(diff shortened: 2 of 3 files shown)


lianetm:
Hi! Agree that we should ensure no loop, but in java we do so by treating this case as fatal (no rejoin), should we do the same here? Also in java we do consider the case where the error is received when the member is not in the group anymore (in that case we just "ignore" the error and leave the member as it is, out of the group already, not in a fatal state). Hope it helps!

src/rdkafka_cgrp.c
```diff
@@ -3280,6 +3280,7 @@ void rd_kafka_cgrp_handle_ConsumerGroupHeartbeat(rd_kafka_t *rk,
 
         case RD_KAFKA_RESP_ERR_UNKNOWN_MEMBER_ID:
         case RD_KAFKA_RESP_ERR_FENCED_MEMBER_EPOCH:
+        case RD_KAFKA_RESP_ERR_GROUP_ID_NOT_FOUND:
```
pranavrth, line 3283: As explained in the slack thread, we should add this error as FATAL.

src/rdkafka_cgrp.c
pranavrth, line 3300: We should add the error with a new case and mark actions as fatal if it is subscribed and skip if unsubscribed like Java client does.

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+## Fixes
+
+### Consumer fixes
+
+* Fix `GROUP_ID_NOT_FOUND` not being handled in the KIP-848
+  `ConsumerGroupHeartbeat` response, which caused a consumer to silently
+  re-send heartbeats with a stale member epoch indefinitely (without
+  rejoining or surfacing an error) when the coordinator no longer had the
+  group. The error is now treated like `UNKNOWN_MEMBER_ID` /
+  `FENCED_MEMBER_EPOCH`: the consumer rejoins the group, re-creating it on
```
pranavrth, line 14: This is incorrect.

pranavrth (changes requested):

pranavrth:
@lianetm - Thanks for the help.

>Hi! Agree that we should ensure no loop, but in java we do so by treating this case as fatal (no rejoin), should we do the same here? Also in java we do consider the case where the error is received when the member is not in the group anymore (in that case we just "ignore" the error and leave the member as it is, out of the group already, not in a fatal state). Hope it helps!

Yes this is what we should do here as well. We are treating this error as FATAL in `ShareGroupHeartbeat` flow and had planned to fix in CG heartbeat as well before the release.

izzyharker:
Closing - @pranavrth has a fix for this in a separate PR.

PR #5489, opened 2026-06-11
[KIP-932] Skip !UP leaders in share-consumer fetch path + related cleanup

pranavrth:
- Guard `rd_kafka_share_select_broker` with `rkb_state == STATE_UP`. Without this a DOWN leader caused an unbounded main↔broker thread bounce loop. The read is unlocked; stale values are bounded to one stray op per state transition.
- Drop the now-redundant `rkb_fetching` check from the share-consumer SHARE_FETCH op handler — `rkb_share_fetch_enqueued` on the main thread already enforces the at-most-one-in-flight invariant.
- New 0182 subtests:
  - Steady-state DOWN: no "broker not up" logs, internal retry refills the queue after recovery without consume_batch driving the fetch.
  - Broker chaos during active empty-poll: log count bounded by cycle count even under rapid up/down churn.
- 0176: add verbose tracing to `do_test_mock_uses_share_acknowledge` for easier diagnosis.
- Remove stale TODOs and commented-out debug / dead code across the share-consumer files.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3696,12 +3696,6 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                             rd_kafka_broker_state_names[rkb->rkb_state]);
                         rd_kafka_share_fetch_op_reply_and_update_ack_details_with_err(
                             rko, RD_KAFKA_RESP_ERR__STATE);
-                } else if (rkb->rkb_fetching) {
-                        rd_kafka_dbg(rkb->rkb_rk, BROKER, "SHAREFETCH",
-                                     "Ignoring SHARE_FETCH op: "
-                                     "already fetching");
-                        rd_kafka_share_fetch_op_reply_and_update_ack_details_with_err(
-                            rko, RD_KAFKA_RESP_ERR__PREV_IN_PROGRESS);
                 } else if (rko->rko_u.share_fetch.should_leave) {
                         rd_kafka_broker_share_fetch_session_leave(rkb, rko,
                                                                   rd_clock());
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3685,11 +3685,10 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                         rd_kafka_share_fetch_op_reply_and_update_ack_details_with_err(
                             rko, rd_kafka_broker_destroy_error(rkb->rkb_rk));
                 } else if (rkb->rkb_state != RD_KAFKA_BROKER_STATE_UP) {
-                        /* TODO KIP-932: The main thread should check
-                         * broker state before enqueuing SHARE_FETCH
-                         * ops, or back off after receiving ERR__STATE
-                         * replies, to avoid flooding the broker thread
-                         * with ops that are immediately rejected. */
+                        /* Backstop: the main thread filters on rkb_state
+                         * in rd_kafka_share_select_broker, so this only
+                         * fires when the broker transitioned to !UP
+                         * after that unlocked check. */
                         rd_kafka_dbg(
                             rkb->rkb_rk, BROKER, "SHAREFETCH",
                             "Ignoring SHARE_FETCH op: "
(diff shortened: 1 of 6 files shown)


tests/0176-share_consumer_commit_sync.c
```diff
@@ -1287,8 +1357,11 @@ static void do_test_mock_uses_share_acknowledge(void) {
                     "partitions count (%d)",
                     share_ack_cnt, commit_with_partitions);
 
+        TEST_SAY("Closing share consumer\n");
```
ojasvajain: We already log something similar in `test_share_consumer_close`. Can we remove this?

tests/0176-share_consumer_commit_sync.c
```diff
@@ -1287,8 +1357,11 @@ static void do_test_mock_uses_share_acknowledge(void) {
...
         test_share_consumer_close(rkshare);
+        TEST_SAY("Destroying share consumer\n");
```
ojasvajain: Same. We have logs in test_share_destroy.

tests/0182-share_consumer_error_handling_mock.c
```diff
@@ -2528,6 +2528,335 @@ static void test_share_consumer_does_not_surface_unknown_topic_or_part(void) {
...
+ *  Wire never sees a ShareFetch on the DOWN broker (broker thread
+ *  rejects the internal op before any RPC is built), so we assert on
+ *  the count of the broker-thread short-circuit debug log instead of
+ *  on mock_get_requests.
+ * =================================================================== */
+static void do_test_no_bounce_loop_on_down_broker(void) {
+        test_ctx_t ctx;
+        rd_kafka_share_t *rkshare;
+        rd_kafka_conf_t *conf;
+        rd_atomic32_t broker_not_up_cnt;
```
ojasvajain, line 2553: Do we need to make it atomic? All reads and updates will be on the same thread, right?

tests/0182-share_consumer_error_handling_mock.c
```diff
@@ -2528,6 +2528,335 @@ static void test_share_consumer_does_not_surface_unknown_topic_or_part(void) {
...
+        test_conf_set(conf, "group.id", group);
+        test_conf_set(conf, "share.acknowledgement.mode", "explicit");
+        test_conf_set(conf, "debug", "broker");
+        rd_kafka_conf_set_log_cb(conf, no_bounce_loop_log_cb);
+        rd_kafka_conf_set_opaque(conf, &broker_not_up_cnt);
+
+        rkshare = rd_kafka_share_consumer_new(conf, NULL, 0);
+        TEST_ASSERT(rkshare != NULL, "Failed to create share consumer");
+
+        subscribe_one(rkshare, topic);
```
ojasvajain: There is a `test_share_consumer_subscribe_multi` fn available in test.c. Can we remove `subscribe_one` from this file and use that?

src/rdkafka_broker.c
```diff
@@ -3685,49 +3685,24 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
...
                             "broker not up (state %s)",
                             rd_kafka_broker_state_names[rkb->rkb_state]);
                         rd_kafka_share_fetch_op_reply_and_update_ack_details_with_err(
                             rko, RD_KAFKA_RESP_ERR__STATE);
-                } else if (rkb->rkb_fetching) {
-                        rd_kafka_dbg(rkb->rkb_rk, BROKER, "SHAREFETCH",
-                                     "Ignoring SHARE_FETCH op: "
-                                     "already fetching");
-                        rd_kafka_share_fetch_op_reply_and_update_ack_details_with_err(
-                            rko, RD_KAFKA_RESP_ERR__PREV_IN_PROGRESS);
```
ojasvajain, line 3705: Just to confirm, we don't need a check on rkb_share_fetch_enqueued here, right? Because it only gets updated in main thread so the check in select broker should suffice (there will be no staleness).

pranavrth: Yes. That check is not needed here. We already make sure that we are asserting that while enqueuing the SHARE_FETCH op. Main thread controls that. This was done very long back when we didn't have full design ready.

tests/0182-share_consumer_error_handling_mock.c
pranavrth: If the `log.queue` is disabled, the client calls the `log_cb` from the caller thread.

ojasvajain (approved):
LGTM, thanks!

PR #5491, opened 2026-06-11
[KIP-932] Implement mock broker partition-level request error injection 

Ankith-Confluent:
Add rd_kafka_mock_partition_push_request_errors() to push a FIFO stack
of errors onto a single partition for a given ApiKey, complementing the
existing cluster- and broker-level injection which can only fail entire
requests. Errors are consumed one per request in push order until the
stack drains.

Consuming handlers: Produce, Fetch, ShareFetch, ShareAcknowledge.
The injected error suppresses the operation it claims failed, keeping
mock state consistent with the response (no append / no record
acquisition / no ack application), matching real broker behaviour.

Test 0182 adds three subtests: API mechanics, ShareFetch partition
scoping and recovery, ShareAcknowledge per-partition commit_sync
results.


PratRanj07 (approved):
LGTM! Thanks @Ankith-Confluent

PratRanj07 (approved):

PR #5494, opened 2026-06-12
[KIP-932]: Check todos in the code

PratRanj07:


src/rdkafka_share_acknowledgement.c
```diff
@@ -276,9 +276,6 @@ void rd_kafka_share_ack_all(rd_kafka_share_t *rkshare) {
  */
 static void rd_kafka_share_ack_details_batch_destroy(void *ptr) {
```
pranavrth, line 277: We have essentially the same function `rd_kafka_share_ack_batches_destroy_free` doing exactly the same thing. But I think we shouldn't remove anything right now as we are near release. Remember to remove this after the release.

pranavrth (approved):
LGTM!. Thanks @PratRanj07 !

PR #5495, opened 2026-06-12
[KIP-932]: Implement Interface change

PratRanj07:
Rename rd_kafka_share_consume_batch → rd_kafka_share_poll; return records via opaque rd_kafka_messages_t
Share consumer poll now allocates its own result, sized to the actual record count, instead of requiring callers to pre-allocate an rd_kafka_message_t ** array of size max.poll.records.

New API in rdkafka.h:

**rd_kafka_error_t *
rd_kafka_share_poll(rd_kafka_share_t *, int timeout_ms, rd_kafka_messages_t **);**

**size_t rd_kafka_messages_count(const rd_kafka_messages_t *);**
**rd_kafka_message_t *rd_kafka_messages_get(const rd_kafka_messages_t *, size_t i);**
**void rd_kafka_messages_destroy(rd_kafka_messages_t *);**
*rkmessages = NULL on error / empty timeout; _destroy(NULL) is a no-op.

### DIFF
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3885,12 +3885,7 @@ void rd_kafka_messages_destroy(rd_kafka_messages_t *messages) {
         if (!messages)
                 return;
         for (i = 0; i < messages->cnt; i++) {
-                /* Skip slots the caller has already destroyed individually
-                 * via rd_kafka_message_destroy() and NULLed; mixing
-                 * per-message and bulk destroy is permitted as long as
-                 * destroyed slots are NULLed out by the caller. */
-                if (messages->elems[i])
-                        rd_kafka_message_destroy(messages->elems[i]);
+                rd_kafka_message_destroy(messages->elems[i]);
         }
         rd_free(messages);
 }
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3884,8 +3884,14 @@ void rd_kafka_messages_destroy(rd_kafka_messages_t *messages) {
         size_t i;
         if (!messages)
                 return;
-        for (i = 0; i < messages->cnt; i++)
-                rd_kafka_message_destroy(messages->elems[i]);
+        for (i = 0; i < messages->cnt; i++) {
+                /* Skip slots the caller has already destroyed individually
+                 * via rd_kafka_message_destroy() and NULLed; mixing
+                 * per-message and bulk destroy is permitted as long as
+                 * destroyed slots are NULLed out by the caller. */
+                if (messages->elems[i])
+                        rd_kafka_message_destroy(messages->elems[i]);
+        }
         rd_free(messages);
 }
 
@@ -3907,7 +3913,6 @@ rd_kafka_error_t *rd_kafka_share_poll(rd_kafka_share_t *rkshare,
                                       rd_kafka_messages_t **rkmessages) {
         rd_kafka_t *rk = rkshare->rkshare_rk;
         rd_kafka_cgrp_t *rkcg;
-        size_t max_poll_records = (size_t)rk->rk_conf.share.max_poll_records;
         rd_bool_t has_records;
         rd_bool_t has_pending_acks;
         rd_kafka_error_t *error = NULL;
@@ -3979,9 +3984,13 @@ rd_kafka_error_t *rd_kafka_share_poll(rd_kafka_share_t *rkshare,
                 rd_kafka_share_fetch_fanout(rk, need_fetch_more_records,
                                             ack_batches);
 
-        scratch = rd_malloc(max_poll_records * sizeof(*scratch));
-        error   = rd_kafka_q_serve_share_rkmessages(
-            rkcg->rkcg_q, timeout_ms, scratch, max_poll_records, &cnt);
+        /* q_serve_share_rkmessages allocates the scratch buffer
+         * internally, sized to the actual op's message count, so it
+         * never overflows regardless of how many records a single
+         * SHARE_FETCH_RESPONSE aggregates (max.poll.records is enforced
+         * upstream during fetch scheduling, not here). */
+        error = rd_kafka_q_serve_share_rkmessages(rkcg->rkcg_q, timeout_ms,
+                                                  &scratch, &cnt);
 
         /* Drain rk_rep for callbacks again before returning */
         rd_kafka_q_serve(rk->rk_rep, RD_POLL_NOWAIT, 0, RD_KAFKA_Q_CB_CALLBACK,
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -6065,7 +6065,7 @@ rd_kafka_op_res_t rd_kafka_poll_cb(rd_kafka_t *rk,
                    be handled from this cb in Share Consumer except maybe in
                    case of closing as this op we don't forward rk_rep queue to
                    rkcg.q. This op is handled separately in
-                   rd_kafka_share_consume_batch() during normal working of Share
+                   rd_kafka_share_poll() during normal working of Share
                    Consumer. Check what is the correct way to handle this op
                    while closing. */
                 rd_kafka_assert(rk, rk->rk_rkshare->rkshare_consumer_closing);
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -3918,8 +3918,8 @@ rd_kafka_error_t *rd_kafka_share_poll(rd_kafka_share_t *rkshare,
         rd_kafka_error_t *error = NULL;
         rd_list_t *ack_batches  = NULL;
         rd_bool_t need_fetch_more_records;
-        rd_kafka_message_t **scratch = NULL;
-        size_t cnt                   = 0;
+        rd_kafka_message_t **rkshare_accumulated_msgs = NULL;
+        size_t cnt                                    = 0;
 
         /* Always NULL the out param so error/empty paths are well-defined. */
         *rkmessages = NULL;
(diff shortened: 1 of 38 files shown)


src/rdkafka.c
```diff
@@ -3948,16 +3979,37 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
...
-        error = rd_kafka_q_serve_share_rkmessages(rkcg->rkcg_q, timeout_ms,
-                                                  rkmessages, max_poll_records,
-                                                  rkmessages_size);
+        /* q_serve_share_rkmessages allocates the accumulated-msgs buffer
+         * internally, sized to the actual op's message count, so it
+         * never overflows regardless of how many records a single
+         * SHARE_FETCH_RESPONSE aggregates (max.poll.records is enforced
+         * upstream during fetch scheduling, not here). */
+        error = rd_kafka_q_serve_share_rkmessages(
+            rkcg->rkcg_q, timeout_ms, &rkshare_accumulated_msgs, &cnt);
```
pranavrth: Why do we still need `cnt` parameter?

src/rdkafka.h
```diff
@@ -3265,20 +3266,69 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
- * at least enough capacity for size max.poll.records.
- * @param rkmessages_size Output number of messages returned in rkmessages.
+ * @param rkmessages Output pointer to a newly-allocated message list (callee
+ *                   allocates). On return:
+ *                   - On success with at least one message: \c *rkmessages
+ *                     points to an \c rd_kafka_messages_t handle owned by the
+ *                     caller, who must release it with
+ *                     rd_kafka_messages_destroy().
+ *                   - On error, empty poll, or timeout with no messages:
+ *                     \c *rkmessages is set to NULL.
```
pranavrth: What is the meaning of empty poll?

src/rdkafka.h
```diff
@@ -3265,20 +3266,69 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
                                               size_t errstr_size);
 
 /**
- * @brief Consume a batch of messages from the share consumer instance.
+ * @brief Poll the share consumer for a batch of messages.
```
pranavrth: Add that it is different from normal consumer poll where a single record is returned. Here batch of records are returned.

src/rdkafka_queue.h
```diff
@@ -878,12 +878,23 @@ int rd_kafka_q_serve_rkmessages(rd_kafka_q_t *rkq,
...
+ * \c *rkmessages_size_out is 0. The returned rd_kafka_error_t (if non-NULL)
+ * is owned by the caller.
+ */
 rd_kafka_error_t *
 rd_kafka_q_serve_share_rkmessages(rd_kafka_q_t *rkq,
                                   int timeout_ms,
-                                  rd_kafka_message_t **rkmessages,
-                                  size_t rkmessages_size,
+                                  rd_kafka_message_t ***rkmessages_out,
                                   size_t *rkmessages_size_out);
```
pranavrth: Why don't we take the input here itself and correctly fill the details. It will also remove unnecessary allocation and destroy in the poll side.

src/rdkafka.c
```diff
@@ -3867,38 +3867,69 @@ static void rd_kafka_share_record_poll_end(rd_kafka_t *rk) {
...
         rd_kafka_t *rk = rkshare->rkshare_rk;
         rd_kafka_cgrp_t *rkcg;
-        size_t max_poll_records = (size_t)rk->rk_conf.share.max_poll_records;
         rd_bool_t has_records;
         rd_bool_t has_pending_acks;
         rd_kafka_error_t *error = NULL;
         rd_list_t *ack_batches  = NULL;
         rd_bool_t need_fetch_more_records;
+        rd_kafka_message_t **rkshare_accumulated_msgs = NULL;
+        size_t cnt                                    = 0;
```
pranavrth: I don't think we should have this here anymore. It should be part of `rd_kafka_q_serve_share_rkmessages` function.

src/rdkafka.c
```diff
@@ -3948,16 +3979,37 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
...
+                         * never both). Destroy messages to avoid leaks. */
+                        size_t i;
+                        for (i = 0; i < cnt; i++)
+                                rd_kafka_message_destroy(
+                                    rkshare_accumulated_msgs[i]);
+                } else {
+                        *rkmessages = rd_kafka_messages_new(
+                            rkshare_accumulated_msgs, cnt);
+                }
+        }
```
pranavrth: It should be part of `rd_kafka_q_serve_share_rkmessages` and this also lets us drop the unreachable if (error) defensive branch.

tests/0126-oauthbearer_oidc.c
```diff
@@ -591,15 +602,18 @@ static void do_test_produce_share_consumer_with_OIDC_should_fail(
...
                 rd_kafka_error_destroy(err);
+        rcvd = rd_kafka_messages_count(batch);
 
         TEST_ASSERT(rcvd == 0, "Expected no messages on auth failure, got %zu",
                     rcvd);
         TEST_ASSERT(error_seen,
                     "Expected authentication error for share consumer");
 
+        rd_kafka_messages_destroy(batch);
+        batch = NULL;
```
pranavrth, line 616: I was thinking if we should provide a function to reset it to NULL after destroy as well. Let me understand the impact of both of these and I will update here.

pranavrth (changes requested):
First pass comment - related to implementation on high level.

tests/test.h
```diff
@@ -1037,11 +1037,11 @@ rd_kafka_share_t *test_create_share_consumer(const char *group_id,
 
 #define TEST_SHARE_BATCH_SIZE 10001
```
pranavrth: This `BATCH_SIZE` (and the `TEST_SHARE_BATCH_SIZE` it aliases in `test.h`) is now unused — it sized the old pre-allocated batch array. Please remove both.

tests/0156-share_consumer_fetch_mock.c
```diff
@@ -860,12 +860,15 @@ static void do_test_sharefetch_fetch_error(rd_kafka_resp_err_t err) {
         consumer = new_share_consumer(ctx.bootstraps, "sg-fetch-error");
         subscribe_topics(consumer, &topic, 1);
 
-        rd_kafka_share_consume_batch(consumer, 2000, rkmessages, &rcvd);
+        rd_kafka_share_poll(consumer, 2000, &batch);
+        rcvd = rd_kafka_messages_count(batch);
         for (i = 0; i < rcvd; i++) {
-                if (!rkmessages[i]->err)
+                rd_kafka_message_t *rkm = rd_kafka_messages_get(batch, i);
+                if (rkm && !rkm->err)
```
pranavrth: `rkm` can't be `NULL` here — the loop bounds i by `rd_kafka_messages_count(batch)`, and _get only returns `NULL` for an out-of-range index or `NULL` list. Drop the `rkm` && to match the other migrated tests (`if (!msg->err)`). Same at line 965.

pranavrth (changes requested):

src/rdkafka.c
```diff
@@ -3867,38 +3867,55 @@ static void rd_kafka_share_record_poll_end(rd_kafka_t *rk) {
...
-    size_t *rkmessages_size /* out */) {
+size_t rd_kafka_messages_count(const rd_kafka_messages_t *messages) {
+        if (!messages)
+                return 0;
+        return messages->cnt;
+}
+
+rd_kafka_message_t *rd_kafka_messages_get(const rd_kafka_messages_t *messages,
+                                          size_t index) {
+        if (!messages || index >= messages->cnt)
```
pranavrth: No null or index check needed. Let user handle this as we do in other APIs.

src/rdkafka_int.h
```diff
@@ -939,6 +939,20 @@ struct rd_kafka_share_s {
...
+/**
+ * @brief Opaque message list returned by rd_kafka_share_poll().
+ *
+ * The struct is allocated as a flexible array sized to the exact message
+ * count returned by a single poll. The caller releases the list and all
+ * contained messages via rd_kafka_messages_destroy().
+ */
+struct rd_kafka_messages_s {
+        size_t cnt;                  /**< Number of messages. */
+        rd_kafka_message_t *elems[]; /**< Message pointer array. */
```
pranavrth, line 955: Ideally it should be FAM with message struct not with pointer to messages. I think it would take significant effort to do that right now as the internal of the struct itself hold many references and we use internal values as well. I think let's add a TODO to fix this and hopefully we fix this by GA. This will enable us to delete the whole messages struct with a single free and we don't need to worry about user destroying individual records.

pranavrth (changes requested):

src/rdkafka.c
```diff
@@ -3948,15 +3965,19 @@ rd_kafka_error_t *rd_kafka_share_consume_batch(
                 rd_kafka_share_fetch_fanout(rk, need_fetch_more_records,
                                             ack_batches);
 
+        /* q_serve_share_rkmessages allocates the caller-visible
+         * rd_kafka_messages_t directly, sized to the actual op's message
+         * count, so it never overflows regardless of how many records a
+         * single SHARE_FETCH_RESPONSE aggregates */
```
pranavrth: Remove this.

src/rdkafka_msg.c
```diff
@@ -1303,6 +1303,22 @@ rd_kafka_message_t *rd_kafka_message_new(void) {
...
+ * @brief Allocate an rd_kafka_messages_t whose flex \c elems[] array has
+ *        room for exactly \p cnt message pointers, with \c cnt populated.
+ *
+ * The caller is responsible for filling \c elems[0..cnt). Internal helper
+ * used by the share-consumer queue serve to allocate the caller-visible
+ * messages handle in a single step (no intermediate buffer).
+ */
+rd_kafka_messages_t *rd_kafka_messages_new(size_t cnt) {
+        rd_kafka_messages_t *messages =
+            rd_malloc(sizeof(*messages) + cnt * sizeof(*messages->elems));
```
pranavrth: `rd_calloc` instead.

src/rdkafka.h
```diff
@@ -3265,20 +3266,74 @@ rd_kafka_share_t *rd_kafka_share_consumer_new(rd_kafka_conf_t *conf,
...
+ *                     caller, who must release it with
+ *                     rd_kafka_messages_destroy().
+ *                   - On error or timeout with no messages:
+ *                     \c *rkmessages is set to NULL.
+ *                   \c rd_kafka_messages_destroy(NULL) is a safe no-op so
+ *                   callers may unconditionally destroy.
+ *
+ * Use rd_kafka_messages_count() and rd_kafka_messages_get() to access the
+ * returned messages.
+ *
```
pranavrth, line 3288: Add a note that this rkmessages should always be pointing to NULL otherwise it will be lost and there will be memory leak.

pranavrth (changes requested):
Some more minor changes.

pranavrth (approved):
LGTM!. Thanks for the changes @PratRanj07 !

PR #5498, opened 2026-06-14
[KIP-932] Let share ops through during SASL reauth; share-path isolation and cleanup

pranavrth:
Main change: during SASL reauthentication the connection and share session stay
live, so the broker-thread SHARE_FETCH op handler now queues the op (sent once
reauth completes) instead of bouncing it — avoiding failed acks for records
already acquired on the broker (which would otherwise be redelivered after the
lock lapses). Not-up bounce error code: `__STATE` → `__TRANSPORT` (retryable).

Also in this PR:
- Isolate share consumers from the rktp version/fetch machinery they don't use:
  `op_fetch_start/stop`, `op_seek`, `op_pause_resume` no-op for share handles;
  devel dasserts enforce `rktp_version == 0`.
- Devel-build fix: drop an `op_version_outdated` dassert that deref'd `rk_conf`
  on an incomplete `rd_kafka_t`.
- Comment/TODO cleanups on the share broker-serve loop and ShareAcknowledge
  handler (no behavior change).

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -3668,39 +3668,34 @@ rd_kafka_broker_op_serve(rd_kafka_broker_t *rkb, rd_kafka_op_t *rko) {
                            rd_kafka_broker_name(rkb),
                            rko->rko_u.share_fetch.should_fetch,
                            rko->rko_u.share_fetch.should_leave);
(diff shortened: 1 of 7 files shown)


PratRanj07 (approved):
LGTM! Thanks @pranavrth

PR #5499, opened 2026-06-15
[KIP-932]: Memory leak fixes and remove null processing from rd_kafka_messages_get

PratRanj07:


src/rdkafka_fetcher.c
```diff
@@ -919,8 +919,10 @@ void rd_kafka_share_filter_acquired_records_and_update_ack_type(
                 if (in_acquired_range) {
                         /* Set ack type based on op type */
                         rd_kafka_msg_t *rkm = NULL;
-                        if (unlikely(rd_kafka_op_is_ctrl_msg(rko)))
+                        if (unlikely(rd_kafka_op_is_ctrl_msg(rko))) {
+                                rd_kafka_op_destroy(rko);
                                 continue;
+                        }
```
pranavrth, line 925: Trying to understand why did our tests miss this memory leak? We are testing these (below one as well) flows in the tests as well right?

src/rdkafka_fetcher.c
```diff
@@ -1774,9 +1778,18 @@ rd_kafka_share_fetch_reply_handle(rd_kafka_broker_t *rkb,
...
+        /* Lists are NULLed on the success path after their inline destroy;
+         * here they're only non-NULL on the error path. filtered_msgs uses a
+         * NULL destructor, so any ops accumulated from earlier partitions
+         * before err_parse fired must be destroyed explicitly. */
+        RD_IF_FREE(inflight_acks, rd_list_destroy);
+        if (filtered_msgs) {
+                rd_kafka_op_t *leaked_rko;
+                int li;
+                RD_LIST_FOREACH(leaked_rko, filtered_msgs, li)
+                rd_kafka_op_destroy(leaked_rko);
```
pranavrth, line 1790: Don't call them `leaked_rko`. Only `rko` is fine. You are anyways explaining above.

src/rdkafka_fetcher.c
pranavrth, line 1750: I think we should check more places in the message handling for memory leak like in `rd_kafka_share_build_response_rko`, whole flow above etc.

tests/0174-share_consumer_concurrency.c
pranavrth, line 1: What was the issue related to memory leak in these files?

pranavrth (approved):
LGTM!. Please resolve the comments in another new PR.

PR #5500, opened 2026-06-15
[KIP-932] : Destroy Ack Batches in TERMINATE op if cgrp is already terminating

ojasvajain:


src/rdkafka_cgrp.c
```diff
@@ -6254,6 +6254,16 @@ void rd_kafka_cgrp_terminate0(rd_kafka_cgrp_t *rkcg, rd_kafka_op_t *rko) {
                             RD_KAFKA_OFFSET_INVALID, "Group is %s",
                             rkcg->rkcg_reply_rko ? "terminating"
                                                  : "terminated");
+
+                        /* For share consumers the TERMINATE op may carry the
+                         * final pending-ack batches. */
+                        if (RD_KAFKA_IS_SHARE_CONSUMER(rkcg->rkcg_rk) &&
+                            rko->rko_u.share_fetch_fanout.ack_batches) {
```
pranavrth: We are using share_fetch_fanout incorrectly here since starting I think. We should use `rko_u.terminate` for this otherwise we will have undefined behaviour. In that case op_destroy will take care of this destroy as well.

src/rdkafka_cgrp.c
```diff
@@ -6254,6 +6254,16 @@ void rd_kafka_cgrp_terminate0(rd_kafka_cgrp_t *rkcg, rd_kafka_op_t *rko) {
...
+                                rd_list_destroy(
+                                    rko->rko_u.share_fetch_fanout.ack_batches);
+                                rko->rko_u.share_fetch_fanout.ack_batches =
+                                    NULL;
+                        }
                         rd_kafka_q_destroy(rkq);
                         rd_kafka_op_destroy(rko);
```
pranavrth, line 6260: add destruction for the newly added field as well.

pranavrth (changes requested):

src/rdkafka_cgrp.c
ojasvajain: Addressed

src/rdkafka_cgrp.c
ojasvajain: Addressed

pranavrth (approved):
LGTM!. Thanks @ojasvajain !

PR #5501, opened 2026-06-15
  [KIP-932] Config changes to maintain parity with Java Client

ojasvajain:
Rule: for configs Java has, mirror Java (reject the ones Java rejects, accept+match
  min/max/default otherwise); for librdkafka-specific configs, reject if inapplicable
  or forced. Regular-consumer behavior is never changed.

| Config | What we did |
  | --- | --- |
  | `fetch.message.max.bytes` (alias `max.partition.fetch.bytes`) | **Accept** — Java accepts+ignores it. |
  | `message.max.bytes` | **Accept** — kept its value (functional for LZ4 decompress) and skipped the `fetch.max.bytes` floor for share consumers. |
  | `receive.message.max.bytes` | **Reject** user-set value and force to `INT_MAX` for share consumers ceiling check only (no extra memory), and Java has no client-side receive cap. |
  | `fetch.max.bytes` | **Accept** — floor gated for share so it can be set as low as `0` (Java parity); regular consumers unchanged. Note: `0` returns no records on both clients — intentionally allowed for parity. |
  | `fetch.min.bytes` | **Accept** — min lowered 1→0 and max raised to INT_MAX for share (full Java parity) ; non-share consumers re-restricted to `1..100000000`. |
  | `fetch.wait.max.ms` | **No change** — min (0) and default (500) match Java. Max differs: librdkafka caps at 300000 vs Java INT_MAX; left as-is (overflow-safety bound, added to `socket.timeout.ms` in an int deadline). |
  | `max.poll.records` | **No change** — already at parity (min 1, max INT_MAX, default 500; it is the share consumer's per-fetch record limit). |
  | `connections.max.idle.ms` | **Default 9 min for share consumers** (was `0`/disabled), matching Java; regular consumers unchanged. Azure heuristic still overrides to ~3m50s; user-set values honored. Note: disable sentinel differs — librdkafka uses `0`; Java `0` = expire-immediately, Java disables via `-1`. |
  | `coordinator.query.interval.ms` | **No change** — applicable; drives periodic coordinator re-query via the shared cgrp path. librdkafka-specific (Java re-queries only reactively on coordinator loss). |
  | `queued.min.messages` | **Reject** for share consumers — per-partition prefetch-queue tuning that share consumers (broker-driven via `max.poll.records`) don't use. No Java equivalent. |
  | `queued.max.messages.kbytes` | **Reject** for share consumers — same prefetch-queue tuning; also removed from the derived `fetch.max.bytes` clamp. No Java equivalent. |
  | `statistics.interval.ms` | **Accept** — removed earlier reject; applicable to share consumers, default `0` = disabled. |
  | `group.instance.id` | **Reject** for share consumers — static membership not supported for share groups (matches Java). |
  | `isolation.level` | **Reject** for share consumers — defined broker-side for share groups (matches Java). |
  | `group.remote.assignor` | **Reject** for share consumers — assignment is broker-driven (matches Java). |

### DIFF
--- a/src/rdkafka_conf.c
+++ b/src/rdkafka_conf.c
@@ -1482,7 +1482,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "Minimum number of bytes the broker responds with. "
      "If fetch.wait.max.ms expires the accumulated data will "
      "be sent to the client regardless of this setting.",
-     1, 100000000, 1},
+     0, 100000000, 1},
     {_RK_GLOBAL | _RK_CONSUMER | _RK_MED, "fetch.error.backoff.ms", _RK_C_INT,
      _RK(fetch_error_backoff_ms),
      "How long to postpone the next fetch request for a "
@@ -4273,7 +4273,12 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                  * end of the block, before the downstream
                  * group-protocol validation runs so its
                  * CONSUMER-branch checks also apply to share
-                 * consumers. */
+                 * consumers.
+                 *
+                 * TODO KIP-932: evaluate, per property, whether
+                 * rejecting is the right behavior or whether some
+                 * should instead be silently ignored (as the Java
+                 * client does for several of these). */
                 if (conf->share.is_share_consumer) {
                         if (conf->rebalance_cb)
                                 return "`rebalance_cb` is not applicable "
@@ -4312,6 +4317,36 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                                 return "`partition.assignment.strategy` is "
                                        "not applicable for share consumer";
 
+                        if (rd_kafka_conf_is_modified(
+                                conf, "fetch.message.max.bytes"))
+                                return "`fetch.message.max.bytes` "
+                                       "(`max.partition.fetch.bytes`) is not "
+                                       "applicable for share consumer";
+
+                        if (rd_kafka_conf_is_modified(conf,
+                                                      "message.max.bytes"))
+                                return "`message.max.bytes` is not applicable "
+                                       "for share consumer";
+
+                        if (rd_kafka_conf_is_modified(
+                                conf, "receive.message.max.bytes"))
+                                return "`receive.message.max.bytes` is not "
+                                       "applicable for share consumer";
+
+                        if (rd_kafka_conf_is_modified(conf,
+                                                      "group.instance.id"))
+                                return "`group.instance.id` is not applicable "
+                                       "for share consumer";
+
+                        if (rd_kafka_conf_is_modified(conf, "isolation.level"))
+                                return "`isolation.level` is not applicable "
+                                       "for share consumer";
+
+                        if (rd_kafka_conf_is_modified(conf,
+                                                      "group.remote.assignor"))
+                                return "`group.remote.assignor` is not "
+                                       "applicable for share consumer";
+
                         if (conf->topic_conf &&
                             rd_kafka_topic_conf_is_modified(
                                 conf->topic_conf, "auto.offset.reset"))
@@ -4321,6 +4356,17 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                         conf->enable_auto_commit = 0;
                         conf->group_protocol = RD_KAFKA_GROUP_PROTOCOL_CONSUMER;
                         conf->socket_max_fails = 1;
+                        /* This allows fetch.max.bytes to be set as low as 0 */
+                        conf->max_msg_size = 0;
+                        /* This is only a ceiling check, not a pre-allocation,
+                         * so it does not increase memory usage. */
+                        conf->recv_max_msg_size = INT_MAX;
+                } else if (conf->fetch_min_bytes < 1) {
+                        /* `fetch.min.bytes=0` (broker responds immediately,
+                         * no long-poll) is only supported for share
+                         * consumers. */
+                        return "`fetch.min.bytes` must be >= 1 for non-share "
+                               "consumers";
                 }
 
                 if (conf->group_protocol == RD_KAFKA_GROUP_PROTOCOL_CLASSIC) {
--- a/src/rdkafka_conf.c
+++ b/src/rdkafka_conf.c
@@ -1482,7 +1482,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "Minimum number of bytes the broker responds with. "
      "If fetch.wait.max.ms expires the accumulated data will "
      "be sent to the client regardless of this setting.",
-     0, 100000000, 1},
+     0, INT_MAX, 1},
     {_RK_GLOBAL | _RK_CONSUMER | _RK_MED, "fetch.error.backoff.ms", _RK_C_INT,
      _RK(fetch_error_backoff_ms),
      "How long to postpone the next fetch request for a "
@@ -4288,16 +4288,6 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                                 return "`RD_KAFKA_EVENT_REBALANCE` is not "
                                        "applicable for share consumer";
 
-                        /* TODO KIP-932: verify whether stats_cb should also
-                         * be rejected. Today we only reject the interval
-                         * because the test framework sets stats_cb
-                         * defensively; with interval=0 the callback is
-                         * never invoked, so the callback alone is inert. */
-                        if (rd_kafka_conf_is_modified(conf,
-                                                      "statistics.interval.ms"))
-                                return "`statistics.interval.ms` is not "
-                                       "applicable for share consumer";
-
                         if (rd_kafka_conf_is_modified(conf,
                                                       "enable.auto.commit"))
                                 return "`enable.auto.commit` is not "
@@ -4318,20 +4308,10 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                                        "not applicable for share consumer";
 
                         if (rd_kafka_conf_is_modified(
-                                conf, "fetch.message.max.bytes"))
-                                return "`fetch.message.max.bytes` "
-                                       "(`max.partition.fetch.bytes`) is not "
-                                       "applicable for share consumer";
-
-                        if (rd_kafka_conf_is_modified(conf,
-                                                      "message.max.bytes"))
-                                return "`message.max.bytes` is not applicable "
-                                       "for share consumer";
-
-                        if (rd_kafka_conf_is_modified(
                                 conf, "receive.message.max.bytes"))
-                                return "`receive.message.max.bytes` is not "
-                                       "applicable for share consumer";
+                                return "`receive.message.max.bytes` must be "
+                                       "INT_MAX for share consumer; changing "
+                                       "it is not allowed";
 
                         if (rd_kafka_conf_is_modified(conf,
                                                       "group.instance.id"))
@@ -4372,17 +4352,19 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
                         conf->enable_auto_commit = 0;
                         conf->group_protocol = RD_KAFKA_GROUP_PROTOCOL_CONSUMER;
                         conf->socket_max_fails = 1;
-                        /* This allows fetch.max.bytes to be set as low as 0 */
-                        conf->max_msg_size = 0;
                         /* This is only a ceiling check, not a pre-allocation,
-                         * so it does not increase memory usage. */
+                         * so it does not increase memory usage. Java client
+                         * has no client-side receive cap*/
                         conf->recv_max_msg_size = INT_MAX;
-                } else if (conf->fetch_min_bytes < 1) {
-                        /* `fetch.min.bytes=0` (broker responds immediately,
-                         * no long-poll) is only supported for share
-                         * consumers. */
-                        return "`fetch.min.bytes` must be >= 1 for non-share "
-                               "consumers";
+                } else if (conf->fetch_min_bytes < 1 ||
+                           conf->fetch_min_bytes > 100000000) {
+                        /* For non-share consumers preserve librdkafka's
+                         * historical `fetch.min.bytes` range of 1..100000000.
+                         * The property max was raised to INT_MAX so share
+                         * consumers can match the Java client; share consumers
+                         * are exempt from this check. */
+                        return "`fetch.min.bytes` must be in range "
+                               "1..100000000 for non-share consumers";
                 }
 
                 if (conf->group_protocol == RD_KAFKA_GROUP_PROTOCOL_CLASSIC) {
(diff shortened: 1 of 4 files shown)


src/rdkafka_msgset_reader.c
```diff
@@ -1591,6 +1591,11 @@ rd_kafka_msgset_reader_run(rd_kafka_msgset_reader_t *msetr) {
                                 err = RD_KAFKA_RESP_ERR_NO_ERROR;
 
                 } else if (rktp->rktp_fetch_msg_max_bytes < (1 << 30)) {
```
pranavrth, line 1593: I think we are reaching here and it will eventually be increased till 1<<30 and then this branch will be skipped.
Check in the test case, we might be reaching here.

src/rdkafka_conf.c
```diff
@@ -4312,15 +4317,72 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+
+                        if (rd_kafka_conf_is_modified(conf,
+                                                      "queued.min.messages"))
+                                return "`queued.min.messages` is not "
+                                       "applicable for share consumer";
+
+                        if (rd_kafka_conf_is_modified(
+                                conf, "queued.max.messages.kbytes"))
+                                return "`queued.max.messages.kbytes` is not "
+                                       "applicable for share consumer";
```
pranavrth, line 4334: Check for the equivalent config in Java for normal consumer.

src/rdkafka_conf.c
```diff
@@ -4312,15 +4317,72 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
                             rd_kafka_topic_conf_is_modified(
                                 conf->topic_conf, "auto.offset.reset"))
                                 return "`auto.offset.reset` is not "
                                        "applicable for share consumer";
 
+                        /* The Azure heuristic (which lowers it to <4 minutes)
+                         * still overrides this for Azure brokers. */
+                        if (!rd_kafka_conf_is_modified(
+                                conf, "connections.max.idle.ms"))
+                                conf->connections_max_idle_ms = 9 * 60 * 1000;
```
pranavrth, line 4354: Its fine to change the default for this property to match Java but we need to make sure that we are adhering to this and not reconnecting until explicitly required. Right now we are starting a new connection immediately if we have toppars to query from even if we don't receive any share fetch related request. Can you check that this is the behavior right now? Maybe with a test.

Add a TODO to check fix this before GA.

src/rdkafka_conf.c
```diff
@@ -4312,15 +4317,72 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+
                         conf->enable_auto_commit = 0;
                         conf->group_protocol = RD_KAFKA_GROUP_PROTOCOL_CONSUMER;
                         conf->socket_max_fails = 1;
+                        /* This allows fetch.max.bytes to be set as low as 0 */
+                        conf->max_msg_size = 0;
+                        /* This is only a ceiling check, not a pre-allocation,
+                         * so it does not increase memory usage. */
+                        conf->recv_max_msg_size = INT_MAX;
```
pranavrth, line 4365: I think we should write the test to make sure that default value is not affecting the overall flow first. Let's say we produce the record with 5KB size and set this to 1KB and try to see if we are getting the record or not.

We need to make sure that memory in soak test is not getting increased because of this change.

src/rdkafka_conf.c
```diff
@@ -4312,15 +4317,72 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                } else if (conf->fetch_min_bytes < 1) {
+                        /* `fetch.min.bytes=0` (broker responds immediately,
+                         * no long-poll) is only supported for share
+                         * consumers. */
+                        return "`fetch.min.bytes` must be >= 1 for non-share "
+                               "consumers";
                 }
```
pranavrth:
```suggestion
                } else {
                        if (conf->fetch_min_bytes < 1) {
                                /* `fetch.min.bytes=0` (broker responds immediately,
                                 * no long-poll) is only supported for share
                                 * consumers. */
                                 return "`fetch.min.bytes` must be >= 1 for non-share "
                                       "consumers";
                        }

                }
```

src/rdkafka_conf.c
```diff
@@ -4312,15 +4317,72 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+
+                        if (rd_kafka_conf_is_modified(conf,
+                                                      "message.max.bytes"))
+                                return "`message.max.bytes` is not applicable "
+                                       "for share consumer";
+
+                        if (rd_kafka_conf_is_modified(
+                                conf, "receive.message.max.bytes"))
+                                return "`receive.message.max.bytes` is not "
+                                       "applicable for share consumer";
```
pranavrth: I think we should relook at this. We might have to support this. Let's discuss in a call tomorrow.

pranavrth (changes requested):

src/rdkafka_conf.c
```diff
@@ -4312,15 +4317,72 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                         * still overrides this for Azure brokers. */
+                        if (!rd_kafka_conf_is_modified(
+                                conf, "connections.max.idle.ms"))
+                                conf->connections_max_idle_ms = 9 * 60 * 1000;
+
                         conf->enable_auto_commit = 0;
                         conf->group_protocol = RD_KAFKA_GROUP_PROTOCOL_CONSUMER;
                         conf->socket_max_fails = 1;
+                        /* This allows fetch.max.bytes to be set as low as 0 */
+                        conf->max_msg_size = 0;
```
pranavrth: I was going through this change and it seems to be incorrect. This property is used in LZ4 decompression path as well. The main problem is that it serves as floor for `fetch.max.bytes`. We should gate that instead of changing the default for this. Let's understand impact of that change. I would prefer a test as well for this.

src/rdkafka_conf.c
pranavrth, line 1: Please check the following:

- Add a TODO to match max value of `fetch.wait.max.ms` to Java before GA.
- We are verifying default for all these properties right?

pranavrth: add some guard here maybe.

pranavrth: Check this as well for `fetch_max_bytes`

src/rdkafka_conf.c
pranavrth: Also add a comment to make it equivalent to Java - unlimited.

pranavrth (changes requested):

src/rdkafka_conf.c
ojasvajain: I have added a test case to confirm that the connection is immediately re-established. Added a TODO to fix it.

src/rdkafka_conf.c
ojasvajain: Yes, I have enabled this field back so that user can set it. And have removed the logic to use max_msg_size as floor while calculating fetch.max.bytes (a test case is added too).

src/rdkafka_conf.c
ojasvajain: 1. Done
2. Yes

tests/0180-share_consumer_config.c
```diff
@@ -547,5 +1040,19 @@ int main_0180_share_consumer_config_local(int argc, char **argv) {
...
+        test_fetch_min_bytes_regular_consumer_range_rejected();
         return 0;
 }
+
+
+/*
+ * 1. receive.message.max.bytes - using normal consumer, we produce the record
+ * with 5KB size and set this to 1KB and try to see if we are getting the record
+ * or not
+ */
```
ojasvajain, line 1058: Self notes. Will remove before merge.

src/rdkafka_conf.c
ojasvajain: There are no prefetch queue related configs in Java. In normal consumer, on every poll() it send requests to all brokers who's fetch response isn't already buffered and then the response is buffered for the current or future poll to serve. And then we have fetch.max.bytes, max.poll.records and max.partition.fetch.bytes configs to control the size of the response.

src/rdkafka_msgset_reader.c
ojasvajain: I enabled the logs in test cases but I found none of the test cases are reaching here.

pranavrth: Check why this is not getting triggered in 0 message fetch case.

src/rdkafka_conf.c
pranavrth: Add as TODO.

src/rdkafka_conf.c
```diff
@@ -4312,15 +4303,78 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                        if (conf->fetch_min_bytes < 1 ||
+                            conf->fetch_min_bytes > 100000000) {
+                                /* For non-share consumers preserve librdkafka's
+                                 * historical `fetch.min.bytes` range
+                                 * of 1..100000000. The property max was raised
+                                 * to INT_MAX so share consumers can match the
+                                 * Java client; share consumers are exempt from
+                                 * this check. */
+                                return "`fetch.min.bytes` must be in range "
+                                       "1..100000000 for non-share consumers";
```
pranavrth, line 4376: Let's give the exact error what we are giving today. The behavior should be same for the normal consumer user.

src/rdkafka_conf.c
```diff
@@ -4370,16 +4424,24 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
-                            RD_MAX(RD_MIN(conf->fetch_max_bytes,
-                                          conf->queued_max_msg_kbytes * 1024),
-                                   conf->max_msg_size);
+                 * unless set by user.
+                 */
+                /* Share consumers are exempt: neither the message.max.bytes
+                 * floor nor the queued.max.messages.kbytes clamp applies
+                 * (both are rejected / not applicable), so fetch.max.bytes is
+                 * used as configured. */
+                if (!conf->share.is_share_consumer) {
```
pranavrth, line 4433: Check if we can use `IS_SHARE_CONSUMER` function.

tests/0171-share_consumer_consume.c
```diff
@@ -1203,12 +1203,11 @@ static void do_test_fetch_max_bytes_small(void) {
...
 
         /* Create share consumer with a small session-level fetch cap.
-         * fetch.max.bytes must be >= message.max.bytes per librdkafka
-         * config validation, so lower both to 1500. */
+         * For share consumers `message.max.bytes` does not floor
+         * `fetch.max.bytes` (and is rejected), so set `fetch.max.bytes`
+         * directly to 1500. */
         test_conf_init(&conf, NULL, 60);
         rd_kafka_conf_set(conf, "group.id", group, errstr, sizeof(errstr));
-        rd_kafka_conf_set(conf, "message.max.bytes", "1500", errstr,
```
pranavrth, line 1210: @PratRanj07 - is there any reason for using this?

tests/0180-share_consumer_config.c
```diff
@@ -192,6 +214,25 @@ static void test_socket_max_fails_rejected_at_construction(void) {
...
+ *        and values above 100000000 (e.g. INT_MAX) are rejected.
+ */
+static void test_fetch_min_bytes_regular_consumer_range_rejected(void) {
+        SUB_TEST_QUICK();
+
+        /* 0 is below the regular-consumer minimum of 1. */
+        verify_regular_consumer_conf_prop_rejected("fetch.min.bytes", "0");
+        /* INT_MAX exceeds the regular-consumer maximum of 100000000. */
+        verify_regular_consumer_conf_prop_rejected("fetch.min.bytes",
+                                                   "2147483647");
```
pranavrth, line 231:
```suggestion
        verify_regular_consumer_conf_prop_rejected("fetch.min.bytes",
                                                   "100000001");
```

tests/0171-share_consumer_consume.c
PratRanj07: Yes because fetch.max.bytes needed to be >= message.max.bytes

ojasvajain: @pranavrth - On second thoughts, we can revert this. This change was done when we were rejecting message.max.bytes. We are now keeping it as it is used in lz4 decompression.

src/rdkafka_conf.c
ojasvajain: After discussion, it was decided to not support this and go ahead with INT_MAX as max value.

src/rdkafka_conf.c
ojasvajain: Checked this, the handle is built after conf_finalize is completed so we don't have the handle here.

src/rdkafka_conf.c
```diff
@@ -4312,15 +4303,78 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                        /* TODO KIP-932: Fix max value for fetch.wait.max.ms to
+                         * keep it at par with Java client */
+
                         conf->enable_auto_commit = 0;
                         conf->group_protocol = RD_KAFKA_GROUP_PROTOCOL_CONSUMER;
                         conf->socket_max_fails = 1;
+                        /* This is only a ceiling check, not a pre-allocation,
+                         * so it does not increase memory usage. Java client
+                         * has no client-side receive cap*/
+                        conf->recv_max_msg_size = INT_MAX;
```
pranavrth, line 4365: Forcing `INT_MAX` is the right policy (a ShareFetch response can legitimately exceed the soft `fetch.max.bytes`; the client can't know the cluster's `max.message.bytes`; Java's consumer Selector is `UNLIMITED`). The issue is the failure mode: a large response passes the size check at `rdkafka_broker.c:2141` → `rd_buf_write_ensure_contig` → `rd_malloc(~2GB)`, and `rd_malloc` does an unconditional `rd_assert (NDEBUG unset)` → `SIGABRT` (or `segfault` under `-DNDEBUG`). The old 100 MB cap rejected oversized responses gracefully and never reached the allocation. Java degrades to a contained `OutOfMemoryError`. Scope is narrow (raised `max.message.bytes` + memory-constrained host), but it's a regression from graceful-reject to crash, with no operator recourse since `receive.message.max.bytes` is rejected.

Let's discuss and reason about all these options. We have 3 options:

- Reject and set INT_MAX. let it crash in the narrow window (current in the PR)
- Let user set it with default INT_MAX
- Let user set it with current default

src/rdkafka_conf.c
```diff
@@ -4312,15 +4303,78 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                         * assigned toppars (see
+                         * rd_kafka_broker_share_consumer_serve). Fix so the
+                         * connection stays closed until a fetch is actually
+                         * required. */
+                        if (!rd_kafka_conf_is_modified(
+                                conf, "connections.max.idle.ms"))
+                                conf->connections_max_idle_ms = 9 * 60 * 1000;
+
+                        /* TODO KIP-932: Fix max value for fetch.wait.max.ms to
+                         * keep it at par with Java client */
```
pranavrth, line 4357: Add - "Remember that raising the cap can overflow."

pranavrth (changes requested):

src/rdkafka_conf.c
ojasvajain: If I understand correctly, this crash would also occur if receive.message.max.bytes was set to a lesser value like 100 MB but the host does not have enough memory.

ojasvajain: As discussed, going ahead with 2nd option

ojasvajain:
As discussed, I am addressing the comments in a second PR.

pranavrth (approved):
LGTM!. Thanks @ojasvajain !

PR #5502, opened 2026-06-15
[KIP-932]: Error translation for commit sync and ack callback results

PratRanj07:


src/rdkafka_share_acknowledgement.c
pranavrth, line 1365: This logic should be part of `rd_kafka_share_commit_sync_send_response` function instead.

tests/0182-share_consumer_error_handling_mock.c
```diff
@@ -1852,22 +1854,25 @@ do_test_socket_timeout_partial_ack_then_remaining(int api_timeout_ms,
...
-                 * expected codes here become REQUEST_TIMED_OUT. */
-                expected_phase1_commit_err   = RD_KAFKA_RESP_ERR__TIMED_OUT;
-                expected_phase1_callback_err = RD_KAFKA_RESP_ERR__TIMED_OUT;
+                /* __TIMED_OUT from the broker-thread socket timer is
+                 * translated to REQUEST_TIMED_OUT at the app-facing
+                 * funnel. */
+                expected_phase1_commit_err =
+                    RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT;
+                expected_phase1_callback_err =
+                    RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT;
```
pranavrth, line 1863: Just stating that `__TIMED_OUT` and `REQUEST_TIMED_OUT` distinction was better as the user could have understood properly which was the real cause of the error.

src/rdkafka_share_acknowledgement.c
```diff
@@ -1235,6 +1235,29 @@ size_t rd_kafka_share_partition_offsets_offsets_cnt(
...
+ */
+static rd_kafka_resp_err_t
+rd_kafka_share_translate_app_err(rd_kafka_resp_err_t err) {
+        switch (err) {
+        case RD_KAFKA_RESP_ERR__TIMED_OUT:
+        case RD_KAFKA_RESP_ERR__TIMED_OUT_QUEUE:
+                return RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT;
+        case RD_KAFKA_RESP_ERR__DESTROY:
+        case RD_KAFKA_RESP_ERR__DESTROY_BROKER:
+                return RD_KAFKA_RESP_ERR__TRANSPORT;
```
pranavrth, line 1250: I think we should think about `__UNDERFLOW` and `__BAD_MSG` as well. Let's discuss.

pranavrth (changes requested):

src/rdkafka_share_acknowledgement.c
pranavrth: Its decided that `__UNDERFLOW` will be kept as it is as it is `KafkaException` in Java client and `__BAD_MSG` will be converted to `_INVALID_MSG` as it is `CorruptRecordException` in Java.

src/rdkafka_share_acknowledgement.c
```diff
@@ -1235,6 +1235,31 @@ size_t rd_kafka_share_partition_offsets_offsets_cnt(
...
+        case RD_KAFKA_RESP_ERR__BAD_MSG:
```
pranavrth, line 1251: 1 more case to add `RD_KAFKA_RESP_ERR_SASL_AUTHENTICATION_FAILED` for `__AUTHENTICATION` error.

src/rdkafka_share_acknowledgement.c
```diff
@@ -1235,6 +1235,31 @@ size_t rd_kafka_share_partition_offsets_offsets_cnt(
 }
 
 
+/**
+ * @brief Translate librdkafka-internal err sentinels into the broker-equivalent
+ *        codes the application is expected to handle.
+ *
+ *   __TIMED_OUT, __TIMED_OUT_QUEUE -> REQUEST_TIMED_OUT
+ *   __DESTROY, __DESTROY_BROKER    -> __TRANSPORT
+ *
```
pranavrth: Remove this mapping from here. Already easy to understand function.

pranavrth (changes requested):
Also check all the comments are correct or not.

src/rdkafka_share_acknowledgement.c
```diff
@@ -1235,6 +1235,28 @@ size_t rd_kafka_share_partition_offsets_offsets_cnt(
...
+        case RD_KAFKA_RESP_ERR__TIMED_OUT:
+        case RD_KAFKA_RESP_ERR__TIMED_OUT_QUEUE:
+                return RD_KAFKA_RESP_ERR_REQUEST_TIMED_OUT;
+        case RD_KAFKA_RESP_ERR__DESTROY:
+        case RD_KAFKA_RESP_ERR__DESTROY_BROKER:
+                return RD_KAFKA_RESP_ERR__TRANSPORT;
+        case RD_KAFKA_RESP_ERR__BAD_MSG:
+                return RD_KAFKA_RESP_ERR_INVALID_MSG;
+        case RD_KAFKA_RESP_ERR_SASL_AUTHENTICATION_FAILED:
+                return RD_KAFKA_RESP_ERR__AUTHENTICATION;
```
pranavrth: This is opposite of what I asked. We want `RD_KAFKA_RESP_ERR_SASL_AUTHENTICATION_FAILED` to be sent to the user in case of `__AUTHENTICATION` error. :)

pranavrth (changes requested):

pranavrth (approved):
Still check the comments in next PR.

LGTM!. Thanks @PratRanj07 & @Ankith-Confluent !

PR #5503, opened 2026-06-16
[KIP-932] Key share-consumer rebalance reconciler by topic id

pranavrth:
The cooperative reconciler diff and the group-assignment lookup keyed on (topic name, partition), so a same-name topic recreate collapsed the old and new topic ids into one bucket and never added the new generation to the share session. Key both by (topic id, partition) for share consumers, and add a deterministic regression test.

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -5105,6 +5105,40 @@ static map_toppar_member_info_t *rd_kafka_toppar_list_to_toppar_member_info_map(
 
 
 /**
+ * @brief Same as rd_kafka_toppar_list_to_toppar_member_info_map but keyed by
+ *        (topic_id, partition). Used by the share-consumer cooperative
+ *        reconciler where every entry carries a topic_id and a same-name
+ *        topic recreate can put OLD-id and NEW-id entries for the same
+ *        (name, partition) into the diff inputs; the by-id key keeps them
+ *        as distinct partitions so the diff correctly produces a revoke
+ *        for the OLD-id and an add for the NEW-id.
+ *
+ * @remark \p rktparlist may be NULL.
+ */
+static map_toppar_member_info_t *
+rd_kafka_toppar_list_to_toppar_member_info_map_by_id(
+    rd_kafka_topic_partition_list_t *rktparlist) {
+        map_toppar_member_info_t *map = rd_calloc(1, sizeof(*map));
+        const rd_kafka_topic_partition_t *rktpar;
+
+        RD_MAP_INIT(map, rktparlist ? rktparlist->cnt : 0,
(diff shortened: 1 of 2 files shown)


tests/0184-share_consumer_topic_recreate.c
```diff
@@ -976,9 +976,250 @@ int main_0184_share_consumer_topic_recreate(int argc, char **argv) {
...
+         * view: current rkcg_group_assignment still has the OLD-id
+         * partitions from PHASE 1; the next HB target carries the
+         * NEW-id partitions. That is the precise input that triggers
+         * the reconciler's (topic_name, partition) hash collision. */
+        new_assignment = rd_kafka_topic_partition_list_new(n_partitions);
+        for (p = 0; p < n_partitions; p++)
+                rd_kafka_topic_partition_list_add(new_assignment, topic, p);
+        member_id_arr[0]      = member_ids[0];
+        member_assignments[0] = new_assignment;
+        rd_kafka_mock_sharegroup_target_assignment(
```
ojasvajain, line 1156: As discussed yday, it would be better if we could have some logging or assertions to confirm if the assigmenment actually changed. 

Not a blocker though.

src/rdkafka_cgrp.c
```diff
@@ -5136,11 +5170,26 @@ static void rd_kafka_cgrp_handle_assignment_cooperative(
...
+                old_assignment_set =
+                    rd_kafka_toppar_list_to_toppar_member_info_map_by_id(
+                        rkcg->rkcg_group_assignment);
+        } else {
+                new_assignment_set =
+                    rd_kafka_toppar_list_to_toppar_member_info_map(assignment);
+                old_assignment_set =
+                    rd_kafka_toppar_list_to_toppar_member_info_map(
+                        rkcg->rkcg_group_assignment);
+        }
```
ojasvajain, line 5193: Came across `rd_kafka_share_assignment_subtract`, 
`rd_kafka_share_assignment_add` and 
`rd_kafka_share_assignment_serve_removals` which use topic name instead of id. Are they safe?

pranavrth: Yes those are safe today as removal and addition happens sequentially - removal first and then addition. So they won't collide as there cannot be same named different id partitions in single removal list or addition list.

ojasvajain (approved):
LGTM, thanks!

tests/0184-share_consumer_topic_recreate.c
pranavrth: We try to avoid log based assertions. We are explicitly setting the failing assignments as target.

I have verified that it is failing consistently without the fix. I have verified by looking at the logs as well.

PR #5507, opened 2026-06-17
[KIP-932]: Remove usage of rd_list from 0190 test

PratRanj07:


src/rdkafka_fetcher.c
```diff
@@ -1784,10 +1784,10 @@ rd_kafka_share_fetch_reply_handle(rd_kafka_broker_t *rkb,
          * before err_parse fired must be destroyed explicitly. */
         RD_IF_FREE(inflight_acks, rd_list_destroy);
         if (filtered_msgs) {
-                rd_kafka_op_t *leaked_rko;
+                rd_kafka_op_t *rko;
```
Ankith-Confluent, line 1787: Are these changes supposed to be there?

tests/0190-share_consumer_telemetry.c
```diff
@@ -69,6 +69,38 @@ static const char *telemetry_topic(void) {
...
+                if (strcmp(seen->names[i], name) == 0)
+                        return rd_true;
+        }
+        return rd_false;
+}
+
+static void seen_metrics_add(seen_metrics_t *seen, const char *name) {
+        if (seen_metrics_contains(seen, name))
+                return;
+        if (seen->cnt >= MAX_SEEN_METRICS) {
```
Ankith-Confluent, line 94: Can we make this TEST_FAIL instead of TEST_WARN? Right now if we ever go over 64 metrics it just drops the name and the test still passes, which we'd probably miss. Failing here would make it obvious and point to bump MAX_SEEN_METRICS.

Ankith-Confluent (commented):
Thanks for the PR @PratRanj07 
Just some minor comments

src/rdkafka_fetcher.c
PratRanj07: Yes its a comment from a previous PR

Ankith-Confluent (approved):
Looks like the build failures are not related to these changes.
LGTM! @PratRanj07

PR #5508, opened 2026-06-17
[KIP-932] Mock broker fix to scope share-session lock release to the current leader

Ankith-Confluent:
Fixes intermittent failures in 0183 (do_test_leader_change_consume_recovery),
e.g. "expected 102 records after leader change, got 137" — records delivered
twice. Failed only on slow CI, never locally.

Root cause: in the mock, share-group record state is cluster-global, and
rd_kafka_mock_sgrp_release_member_locks() released a member's acquired locks
across ALL partitions in the group on any single-broker session event
(connection close, epoch reset, timeout). So when one broker's session
closed, records still held through other brokers' intact sessions were
wrongly released and redelivered. A real broker can't do this — each broker
only holds lock state for partitions it leads.

Fix enforces the invariant "only a partition's current leader may release its
locks":
 - release is scoped to the closing session's partitions, and
 - guarded by a check that this broker is the partition's current leader
   (a session's partition list can be stale after a leader change).
The same leader guard is applied to the ShareFetch forgotten-partitions
release path.

### DIFF
--- a/src/rdkafka_mock_handlers.c
+++ b/src/rdkafka_mock_handlers.c
@@ -4289,7 +4289,22 @@ static int rd_kafka_mock_handle_ShareFetch(rd_kafka_mock_connection_t *mconn,
                                 }
 
                                 /* Release ACQUIRED records for this
-                                 * member on this partition. */
+                                 * member on this partition, but only
+                                 * if this broker currently leads it:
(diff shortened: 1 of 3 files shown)


PratRanj07 (approved):
LGTM! Thanks @Ankith-Confluent

PR #5509, opened 2026-06-18
[KIP-932] Share Consumer Config Evaluation Part 2

ojasvajain:
1. Addressed comments from [this](https://github.com/confluentinc/librdkafka/pull/5501) PR
2. Updated description of properties to indicate share consumer suppport.
3. The following additional configs have been evaluated as part of this PR

| Config | What we did |
  | --- | --- |
  | `fetch.queue.backoff.ms` | **Reject** for share consumers — backs off the per-partition prefetch-queue throttle (`queued.*`), which share consumers don't use. librdkafka-specific (no Java equivalent). |
  | `fetch.error.backoff.ms` | **Reject** for share consumers — only consulted on the regular (non-share) Fetch path; ShareFetch/ShareAcknowledge errors aren't retried with a fetch backoff (session reset / next poll / `reconnect.backoff.ms` instead). librdkafka-specific (Java uses generic `retry.backoff.ms`). |
  | `enable.partition.eof` | **Reject** for share consumers — emits PARTITION_EOF when the client's per-partition fetch cursor reaches the HWM; share consumers don't maintain that cursor, so it's never produced. librdkafka-specific (no Java equivalent). |
  | `retry.backoff.ms` | **Accept** — applicable (caps retry backoff for share HB/metadata/coordinator requests). Default (100) matches Java. Caveat - the min and max values are different from Java. We are intentionally keeping it this way. |
  | `retry.backoff.max.ms` | **Accept** — applicable; default (1000) matches Java. Caveat - the min and max values are different from Java. We are intentionally keeping it this way. |
  | `client.rack` | **Accept, no change** — applicable and functional; sent as the rack id in ShareGroupHeartbeat (rack-aware assignment / preferred-replica reads) in both clients. |
  | `reconnect.backoff.ms` | **Accept** — applicable (governs share-consumer reconnect). **Force default 50 for share consumers** (Java parity; librdkafka default is 100); user-set values honored, regular consumers/producers unchanged. |
  | `reconnect.backoff.max.ms` | **Accept** — applicable (reconnect backoff cap). **Force default 1000 for share consumers** (Java parity; librdkafka default is 10000); user-set values honored, regular consumers/producers unchanged. |
  | `reconnect.backoff.jitter.ms` | **No change** — deprecated and inert for all clients (jitter now applied automatically); no Java equivalent. Left as-is (not rejected). |


src/rdkafka_conf.c
```diff
@@ -4353,6 +4368,18 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                         * for share consumers (librdkafka's client-wide
+                         * defaults are 100 / 10000). Only when unset, so a
+                         * user-provided value still wins; regular consumers
+                         * and producers keep the librdkafka defaults. */
+                        if (!rd_kafka_conf_is_modified(conf,
+                                                       "reconnect.backoff.ms"))
+                                conf->reconnect_backoff_ms = 50;
+                        if (!rd_kafka_conf_is_modified(
+                                conf, "reconnect.backoff.max.ms"))
+                                conf->reconnect_backoff_max_ms = 1000;
```
pranavrth, line 4389: Generally we will immediately reject acknowledgements if we see that the broker is not up but this might affect the case when next share fetch request is being sent to this broker. If we keep our default of 10s, share fetch request will need to wait for ~10 seconds. It will be ~1s in Java. So let's understand changing the default doesn't have any major effect in our case.

src/rdkafka_conf.c
```diff
@@ -1634,15 +1634,15 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
+     0, 300 * 1000, 100},
 
     {_RK_GLOBAL | _RK_MED, "retry.backoff.max.ms", _RK_C_INT,
      _RK(retry_backoff_max_ms),
      "The max backoff time in milliseconds before retrying a protocol "
      "request, "
      "this is the atmost backoff allowed for exponentially backed off "
      "requests.",
-     1, 300 * 1000, 1000},
+     0, 300 * 1000, 1000},
```
pranavrth: Let's keep atleast 1 backoff. Check the effect of setting 0 wrt backoff and jitter.

src/rdkafka_conf.c
```diff
@@ -453,7 +453,8 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "case of protocol hickups. "
      "This value must be at least `fetch.max.bytes`  + 512 to allow "
      "for protocol overhead; the value is adjusted automatically "
-     "unless the configuration property is explicitly set.",
+     "unless the configuration property is explicitly set. For share "
+     "consumers, the default value is INT_MAX.",
```
pranavrth, line 457: We need to mention the allowed value range for normal consumer case as well now.

src/rdkafka_conf.c
```diff
@@ -1452,15 +1459,17 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
...
      0, 300 * 1000, 1000},
     {_RK_GLOBAL | _RK_CONSUMER | _RK_MED, "fetch.message.max.bytes", _RK_C_INT,
      _RK(fetch_msg_max_bytes),
      "Initial maximum number of bytes per topic+partition to request when "
      "fetching messages from the broker. "
      "If the client encounters a message larger than this value "
      "it will gradually try to increase it until the "
-     "entire message can be fetched.",
+     "entire message can be fetched. This property is not supported for share "
+     "consumers.",
```
pranavrth, line 1472: - Change the working to show that we are accepting it but ignoring it.
- We discussed that it is being used in the share path at the message set level. I think its better to change this property to `1 << 30` to always skip that branch. Check the feasibility of it.

src/rdkafka_conf.c
```diff
@@ -1474,7 +1483,8 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "The maximum message batch size accepted by the broker is defined "
      "via `message.max.bytes` (broker config) or "
      "`max.message.bytes` (broker topic config). "
-     "`fetch.max.bytes` is automatically adjusted upwards to be "
+     "For regular consumers, `fetch.max.bytes` is automatically adjusted "
+     "upwards to be "
```
pranavrth, line 1487:
```suggestion
     "For regular (not share) consumers, `fetch.max.bytes` is automatically adjusted "
     "upwards to be "
```

src/rdkafka_conf.c
```diff
@@ -1520,7 +1532,8 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
     {_RK_GLOBAL | _RK_CONSUMER, "enable.partition.eof", _RK_C_BOOL,
      _RK(enable_partition_eof),
      "Emit RD_KAFKA_RESP_ERR__PARTITION_EOF event whenever the "
-     "consumer reaches the end of a partition.",
+     "consumer reaches the end of a partition. This property is not supported "
```
pranavrth, line 1535: It is fine for now but let's update the `_RK_CONSUMER` part itself of all the properties.

src/rdkafka_conf.c
```diff
@@ -633,7 +634,7 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      _RK(connections_max_idle_ms),
      "Close broker connections after the specified time of "
      "inactivity. "
-     "Disable with 0. "
+     "Disable with 0. For share consumers, the default value is 32400 (9 mins)."
```
pranavrth, line 637:
```suggestion
     "Disable with 0. For share consumers, the default value is 540000 (9 mins). "
```

tests/0180-share_consumer_config.c
```diff
@@ -308,21 +310,6 @@ static void test_heartbeat_interval_ms_rejected_at_construction(void) {
         SUB_TEST_PASS();
 }
 
-/**
- * @brief `receive.message.max.bytes` is forced to INT_MAX for share
- *        consumers (a ShareFetch response can exceed fetch.max.bytes,
- *        which is a soft limit); any explicit set by the app is
- *        rejected.
- */
-static void test_receive_message_max_bytes_rejected_at_construction(void) {
```
pranavrth, line 317: Missing tests for the two new behaviors. The reject-additions are covered, but neither behavioral change is:

- `receive.message.max.bytes`: the old rejection test was removed and nothing replaced it. Add coverage that (a) when unset it resolves to INT_MAX, and (b) a value below `fetch.max.bytes` + 512 is rejected.
- `reconnect.backoff.ms` / `reconnect.backoff.max.ms`: the forced share-consumer defaults (50 / 1000) are a deliberate Java-parity choice, but nothing asserts them.

Non-blocking, but the reconnect.backoff defaults test in particular would protect an intentional decision from silent regression.

pranavrth (approved):
Some comments related to the description and tests. There is 1 comment for `fetch_msg_max_bytes` handling but I think we are good to merge this PR as we have already planned for improving the documentation. We should address the test comment also in that PR.

We can discuss `fetch_msg_max_bytes` part and decide if it makes sense to change that. Since that is in rare flow and will eventually correct itself, we can fix this later.

LGTM for now. Thanks @ojasvajain !

src/rdkafka_conf.c
ojasvajain: The default/min/max values for normal consumer are already mentioned in the other columns. We can decide on what can be the best way to document these values.

src/rdkafka_conf.c
ojasvajain: 1. 👍 
2. Can we just add a IS_SHARE_CONSUMER check in that branch? Would be simpler.

src/rdkafka_conf.c
ojasvajain: Are you suggesting defining a new enum like _RK_SHARE_CONSUMER?

src/rdkafka_conf.c
ojasvajain: Btw, the allowed range for `recv_max_msg_size` is still the same. We have only changed the default value for share consumers. For normal consumers, the default value is already mentioned in the table.

PR #5517, opened 2026-06-22
[KIP-932]  Share-consumer logging: levels, context/facility, and cleanup

pranavrth:
Improves observability of the KIP-932 share-consumer and heartbeat paths. Logging-only, except one preserved error code.

- Lower ShareFetch/ShareAcknowledge top-level errors from error to debug — they were double-logged and noisy for expected, recoverable errors; each is now reported once.
- Preserve the real broker error code on an unknown per-partition ShareFetch error instead of reporting a generic local state error.
- Log topic-authorization failures on the Share and Consumer group heartbeats at error level (throttled), alongside the existing error event.
- Route per-broker share logs through the broker logger so each line carries the broker id; group session-lifecycle logs under the SHARESESSION facility; move stray logs to the FETCH context; add topic/session-epoch references; drop a dead commented-out printf.
- Remove two duplicate share debug logs (per-partition ack build, session-update loop).

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -3771,6 +3771,11 @@ err:
                     (rd_clock() >
                      rkcg->rkcg_ts_last_err + min_error_interval)) {
                         rd_kafka_cgrp_set_last_err(rkcg, err);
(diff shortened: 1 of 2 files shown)


PratRanj07 (approved):
LGTM! Thanks @pranavrth

PR #5518, opened 2026-06-22
[KIP-932] Share Consumer Config Evaluation Part 3

ojasvajain:
1. Addressed comments received in Part 2 PR
2. The following configs were evaluate as part of this PR

 | # | Config | Action | Justification |
  |---|--------|--------|---------------|
  | 1 | `message.copy.max.bytes` | **Reject** | Producer-only copy/zero-copy threshold (used when building ProduceRequests). Flagged `_RK_GLOBAL` (not `_RK_PRODUCER`), so it's silently accepted by consumers; does nothing on the share path. No Java equivalent. |
  | 2 | `max.in.flight.requests.per.connection` | **No change** | Generic per-connection in-flight limit, primarily relevant to produce. The share fetch path self-limits to one outstanding request per broker, so it has no effect. Accepted silently (same as Java). |
  | 3 | `max.in.flight` | **No change** | Alias of `max.in.flight.requests.per.connection`; same reasoning. |
  | 4 | `topic.blacklist` | **Reject** | librdkafka-specific metadata filter that drops matching topics "as if they did not exist". No Java equivalent; rejected to keep the share-consumer config surface aligned with the Java client. |
  | 5 | `topic.metadata.refresh.interval.ms` | **No change** | Periodic metadata refresh to discover brokers/partitions/leaders. Functional and needed by share consumers (leader discovery for ShareFetch routing). Closest analog to Java's single `metadata.max.age.ms` (same 5 min default/intent). |
  | 6 | `metadata.max.age.ms` | **No change** | Metadata cache staleness threshold (librdkafka splits Java's single knob into refresh-interval + max-age). Functional for share consumers; not share-specific. (Default differs from Java by long-standing librdkafka design.) |
  | 7 | `topic.metadata.refresh.fast.interval.ms` | **No change** | Fast leader-recovery refresh after a leader loss, exponentially backed off up to `retry.backoff.max.ms`. Genuinely useful for share consumers (fast ShareFetch leader rediscovery). No Java config but mirrors Java's retry-backoff-gated fast re-query. |
  | 8 | `topic.metadata.refresh.sparse` | **No change** | Documented unimplemented no-op ("sparse not implemented, KIP-4"); the field is never read anywhere. Left as is. |
  | 9 | `topic.metadata.propagation.max.ms` | **No change** | Delays marking a topic NOTEXISTS to tolerate post-creation propagation. Producer-motivated but participates in the shared topic-state machine, so it has a real (if minor) effect for a share consumer subscribing to a new topic. Left functional. |
  | 10 | `rebalance_cb`, `offset_commit_cb`, `consume_cb` | **Reject** | None apply to the share path: share groups do server-side assignment (no rebalance), share consumers acknowledge rather than commit offsets (no offset-commit callback), and delivery is via `rd_kafka_share_poll()` not the `consume_cb` dispatch path. Java share consumer has no rebalance/offset-commit/consume callbacks (its commit analog is `AcknowledgementCommitCallback`). |
  | 11 | `offset.store.method` | **No change** | `_RK_DEPRECATED` legacy simple-consumer offset-store property; not used by the high-level/share consumer path. Left as the inert deprecated config it already is. |
  | 12 | `offset.store.sync.interval.ms` | **No change** | `_RK_DEPRECATED` legacy simple-consumer property; same reasoning as `offset.store.method`. |
  | 13 | `offset.store.path` | **No change** | `_RK_DEPRECATED` legacy simple-consumer property; same reasoning. |
  | 14 | `auto.commit.interval.ms` (global + legacy topic) | **Ignore** | Auto-commit cadence; meaningless when `enable.auto.commit` is rejected and share consumers don't commit offsets. Java accepts-but-ignores it for share consumers (not in its reject list), so we mirror that — accepted, no effect, description notes it's ignored. |
  | 15 | `auto.commit.enable` (topic, legacy) | **No change** | `_RK_DEPRECATED` legacy simple-consumer property (separate field from the global `enable.auto.commit`); inert on the share path. Not caught by the global `enable.auto.commit` reject; left as deprecated/inert. |
  | 16 | `enable.auto.commit` (topic alias) | **No change** | Topic-level alias resolving to the legacy `auto.commit.enable` field above (not the global `enable.auto.commit`). Inert on the share path; left as-is. The **global** `enable.auto.commit` is rejected (matches Java). |
  | 17 | `consume.callback.max.messages` | **Reject** | Caps the batch dispatched per `rd_kafka_consume_callback*()` call — a callback path share consumers (which use `rd_kafka_share_poll()`) never use. No Java equivalent. |
  | 18 | `plugin.library.paths` | **No change** | Native plugin/interceptor loader. Consume interceptors don't fire on the share path, but request-level and SASL/auth plugins still do, so it's only partially inert — left functional rather than blocking valid plugins. |
  | 19 | `internal.termination.signal` | **No change** | POSIX signal used to wake internal threads for fast `rd_kafka_destroy()`. Pure thread-lifecycle plumbing, identical for all client types; no consumer-protocol relevance. No Java equivalent. |
  | 20 | `allow.auto.create.topics` | **Reject** | Added a reject as it is not supported currently for share consumers in librdkafka. |


src/rdkafka_conf.c
pranavrth, line 1: - Check the interaction of `internal.termination.signal` with destroy.
- `rkcq` consumer queue is not exposed? check if user can't access it
- .topic level auto commit - check if not used anywhere
- `max.in.flight` related config should be no change instead of ignore
- Check `plugin.library.paths` again. Code path.
- Remove interceptor.

pranavrth (changes requested):

src/rdkafka_conf.c
ojasvajain: 1. Checking
2. Checked this. Using a rkshare handle, it is not possible to get `rkcg_q`. There is a `rd_kafka_queue_get_consumer` method but that only accepts the rk handle.
3. Checked. It is not used.
4. Fixed in PR description
5. Checking further.
6. Fixed. If users try to set any of these interceptors, they will get a error NOT_IMPLEMENTED in response. Interceptors:
on_destroy, on_consume, on_commit, on_send, on_acknowledgement, on_request_sent, on_response_received, on_thread_start, on_thread_exit, on_broker_state_change

ojasvajain: 1. Added a test case
5. `plugin.library.paths` allows users to provide paths to shared libraries. These libraries need to expose a conf_init API that librdkafka will call while loading them.  Inside conf_init, the plugin registers its interceptor hooks by calling `rd_kafka_(conf_)interceptor_add_on_*()` functions.

src/rdkafka_conf.c
```diff
@@ -818,7 +822,8 @@ static const struct rd_kafka_property rd_kafka_properties[] = {
      "Further, the consumer default value is different from the Java "
      "consumer (true), and this property is not supported by the Java "
      "producer. Requires broker version >= 0.11.0.0, for older broker "
-     "versions only the broker configuration applies.",
+     "versions only the broker configuration applies. This property "
```
pranavrth, line 825: You can expect it to be supported for GA. Also, add a TODO for us to remember this.

src/rdkafka_conf.c
```diff
@@ -4367,12 +4387,33 @@ const char *rd_kafka_conf_finalize(rd_kafka_type_t cltype,
...
+                                       "applicable for share consumer";
+
+                        if (rd_kafka_conf_is_modified(conf, "topic.blacklist"))
+                                return "`topic.blacklist` is not "
+                                       "applicable for share consumer";
+
+                        if (rd_kafka_conf_is_modified(
+                                conf, "allow.auto.create.topics"))
+                                return "`allow.auto.create.topics` is not "
+                                       "supported  for share consumer";
```
pranavrth:
```suggestion
                                return "`allow.auto.create.topics` is not "
                                       "supported  for share consumer in Share Consumer Preview release";
```

tests/0180-share_consumer_config.c
```diff
@@ -1055,17 +1458,183 @@ static void test_share_consumer_fetch_conn_idle_after_drain(void) {
...
-        test_share_consumer_fetch_conn_idle_after_drain();
+        // test_max_poll_records_caps_batch_at_5();
+        // test_max_poll_records_allows_full_drain_at_10();
+        // test_fetch_max_bytes_one_is_soft_limit();
+        // test_fetch_min_bytes_zero_drains_all();
+        // test_fetch_min_bytes_max_drains_all();
+        // test_fetch_max_bytes_zero_returns_no_records();
+        // test_receive_message_max_bytes_is_honored();
+        test_internal_termination_signal_destroy();
+        // test_share_consumer_fetch_conn_idle_after_drain();
```
pranavrth: uncomment.

pranavrth (changes requested):

src/rdkafka_conf.c
ojasvajain: Added a TODO.
Modified the above to "This property is currently not supported for share consumers and will be enabled in the General Availability (GA) release."

pranavrth (approved):
LGTM!. Thanks @ojasvajain !

PR #5533, opened 2026-06-25
[KIP-932] use int64 for offset count in share ack commit offsets list

sage-mode-hunter:
Offset count overflow in rd_kafka_share_build_partition_offsets_list:
- total_offsets is a 32-bit int but sums int64 ranges from broker AcquiredRecords; past INT_MAX it overflows
- the overflowed value sizes rd_kafka_share_partition_offsets_new while the fill loop writes one int64 per offset over the full range
- offset_idx is also int and can wrap to a negative index
- the sibling rd_kafka_share_build_response_rko already keeps this count in int64_t

Widened both to int64_t to match the int64 offsets array and that sibling.


sage-mode-hunter:
i have signed the CLA.

sage-mode-hunter:
Pushed an empty commit to re-run the CLA check. Signed it on my end already, so hopefully this clears the status.

sage-mode-hunter:
any update?

PR #5534, opened 2026-06-25
Fix flaky 0097_ssl_verify on SSL CI (accept TLS alert 40)

PratRanj07:
0097_ssl_verify's untrusted-client-cert subtest only accepted TLS alert 42
(bad_certificate), but some broker TLS stacks (e.g. the SSL CI runners) send
the RFC-correct alert 40 (handshake_failure) when the client withholds its
untrusted cert — so the assertion failed 

Fix: also accept alert 40, scoped to the client-auth-failure scenario so it
can't mask unrelated SSL errors. This resolves both the failure and the segfault.

### DIFF
--- a/tests/0097-ssl_verify.cpp
+++ b/tests/0097-ssl_verify.cpp
@@ -155,7 +155,8 @@ class TestEventCb : public RdKafka::EventCb {
         Test::Say("SSL error: " + event.str() + "\n");
         if (expect_client_auth_fail) {
           /* Broker rejects the untrusted client cert. Different broker TLS
-           * stacks send 42 (bad_certificate) or 40 (handshake_failure). */
+           * stacks send 42 (bad_certificate) or 40 (handshake_failure since
+           * Java 23). */
           if (event.str().find("alert number 42") != std::string::npos ||
               event.str().find("alert number 40") != std::string::npos)
             expected = true;
--- a/tests/0097-ssl_verify.cpp
+++ b/tests/0097-ssl_verify.cpp
@@ -126,8 +126,14 @@ class TestVerifyCb : public RdKafka::SslCertificateVerifyCb {
 class TestEventCb : public RdKafka::EventCb {
  public:
   bool should_succeed;
+  /* True only when the broker is expected to reject the client's cert
+   * (untrusted client key + mutual TLS). Used to accept the broker's TLS
+   * alert only in that case. */
+  bool expect_client_auth_fail;
 
-  TestEventCb(bool should_succeed) : should_succeed(should_succeed) {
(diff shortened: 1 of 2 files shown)


tests/0097-ssl_verify.cpp
```diff
@@ -147,14 +153,18 @@ class TestEventCb : public RdKafka::EventCb {
...
-        if (event.str().find("alert number 42") != std::string::npos)
-          /* Verify that certificate isn't sent if not trusted
-           * by the broker. We should receive 42 (bad_certificate)
-           * instead of 46 (certificate_unknown). */
-          expected = true;
-        else if (event.str().find("broker certificate could not be verified") !=
-                 std::string::npos)
+        if (expect_client_auth_fail) {
+          /* Broker rejects the untrusted client cert. Different broker TLS
+           * stacks send 42 (bad_certificate) or 40 (handshake_failure). */
```
emasab:
```suggestion
          /* Broker rejects the untrusted client cert. Different broker TLS
           * stacks send 42 (bad_certificate) or 40 (handshake_failure since Java 23). */
```

tests/0097-ssl_verify.cpp
emasab, line 1: add the JAVA_HOME to the run-all semaphore pipeline, after switching the java version
```
sem-version java 25
export JAVA_HOME="$(dirname "$(dirname "$(readlink -f "$(which java)")")")"
```

emasab (changes requested):

emasab (approved):
Thanks!

tests/0097-ssl_verify.cpp
PratRanj07: The Semaphore runner — KRaft setup crashed with:
kafka-run-class.sh: line 352: /home/semaphore/.jenv/bin/java: No such file or directory
Exception: KRaft setup failed: "...kafka-storage.sh format ..." returned exit code 32512

The cause is that sem-version is jenv-based here: **which java** returns ~/.jenv/shims/java, which is a wrapper script, not a symlink. So readlink -f returns the shim path unchanged, and the two dirnames resolve JAVA_HOME to ~/.jenv — Kafka then looks for $JAVA_HOME/bin/java (~/.jenv/bin/java), which doesn't exist.
I switched to deriving it from the JVM itself, which is jenv-agnostic:

- sem-version java 25
- export JAVA_HOME="$(java -XshowSettings:properties -version 2>&1 | sed -n 's/^ *java.home = //p')"

This asks the running JVM for its own java.home, so it always points at the real JDK regardless of how sem-version exposes java.

emasab (approved):

PR #5535, opened 2026-06-26
[KIP-932] Use broker-default replication factor for topic creation in 0185 testcase

Ankith-Confluent:
0185 created its topics with a hard-coded replication factor of 1, which
fails (or doesn't match cluster expectations) on real-broker runs where the
cluster has multiple brokers or a higher default.replication.factor.


pranavrth (approved):

PR #5539, opened 2026-06-29
[KIP-932] Share consumer Preview warning, example improvements, and docs

pranavrth:
Preview-readiness changes for the KIP-932 share consumer:

- **Preview warning:** log a warning on share consumer creation that it is in
  Preview and not recommended for production use.
- **Examples:** acknowledge by processing outcome (ACCEPT/RELEASE/REJECT),
  commit once per batch, demonstrate the async ack-commit callback, and stop
  on a fatal poll error. Drop the ad-hoc poll-timing code.
- **Docs:** note that RENEW is not yet available where the ack types are
  introduced; clarify that acknowledged-but-uncommitted records commit on close.


PratRanj07 (approved):
LGTM! Thanks @pranavrth

PratRanj07 (approved):

PratRanj07 (approved):

PR #5545, opened 2026-07-08
Fix cluster id over-read past the controller id in Metadata response

prashah-confluent:
### Summary
The Metadata response parser copied the `cluster_id` string using `strlen()`,
even though the source is not nul-terminated. This causes a read past the end
of the string into the following `controller_id` field, producing a corrupted
(over-long) cluster id and, in the worst case, a heap over-read past the receive
buffer.

### Root cause
In `rd_kafka_parse_Metadata0()` (`src/rdkafka_metadata.c`), `cluster_id` is read
with `rd_kafka_buf_read_str()`, which sets `cluster_id.str` to point **into the
receive buffer** with a known `cluster_id.len` but **no nul terminator**. The
copy into the internal metadata struct was done via
`rd_tmpabuf_write_str(&tbuf, cluster_id.str)`, which internally does
`strlen(cluster_id.str) + 1`.

On the wire, `cluster_id` is immediately followed by the `controller_id` int32.
So `strlen()` walks past the string into the `controller_id` bytes until it
happens to hit a `0x00`. It only stops correctly by luck - when the leading
(big-endian, i.e. network-order) byte of `controller_id` is `0x00`, which is the
case for typical small broker ids. It breaks when that byte is non-zero, e.g.:
- `controller_id = -1` (`0xFFFFFFFF`) — a valid "no controller" value, or
- any `controller_id >= 2^24`.

In those cases the copied cluster id is too long/corrupted, and if no `0x00`
appears before the end of the buffer the `strlen()` reads out of bounds.

### Impact
- The internal per-response cluster id (`mdi->cluster_id`) is corrupted. This is
  surfaced to applications via the Admin `DescribeCluster` API
  (`rd_kafka_DescribeCluster_result_cluster_id()`).
- Potential heap over-read (memory-safety issue) when the following bytes
  contain no `0x00`.
- Note: `rd_kafka_clusterid()` is **not** affected — the cached `rk_clusterid`
  is built with `RD_KAFKAP_STR_DUP` (length-based) and remains correct.

### Fix
Copy exactly `cluster_id.len` bytes and write our own nul terminator, instead of
relying on `strlen()`:

### Testing
- Added a mock-based regression test do_test_cluster_id_not_overread() in tests/0146-metadata_mock.c. It forces the mock cluster's controller_id to -1 and asserts the cluster id surfaced by DescribeCluster (the code path under test) exactly matches the authoritative value from rd_kafka_clusterid().
- Verified the test fails on the pre-fix code ("mockCluster…\xFF\xFF\xFF\xFF", len 25 vs expected len 20 — "cluster_id was over-read") and passes with the fix.
- Added a small mock helper, rd_kafka_mock_set_controller_id(), to let tests control the reported controller id (additive, RD_EXPORT).

### Compatibility
No public API/ABI changes to the client library; the fix is internal to metadata
parsing. The only added symbol is the additive mock helper
rd_kafka_mock_set_controller_id().

### DIFF
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -677,9 +677,14 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
         mdi->cluster_id = NULL;
         if (ApiVersion >= 2) {
                 rd_kafka_buf_read_str(rkbuf, &cluster_id);
-                if (cluster_id.str)
-                        mdi->cluster_id =
-                            rd_tmpabuf_write_str(&tbuf, cluster_id.str);
+                if (!RD_KAFKAP_STR_IS_NULL(&cluster_id)) {
+                        int clen        = RD_KAFKAP_STR_LEN(&cluster_id);
+                        mdi->cluster_id = rd_tmpabuf_alloc(&tbuf, clen + 1);
+                        if (mdi->cluster_id) {
+                                memcpy(mdi->cluster_id, cluster_id.str, clen);
+                                mdi->cluster_id[clen] = '\0';
+                        }
+                }
         }
 
         mdi->controller_id = -1;
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -678,12 +678,14 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
         if (ApiVersion >= 2) {
                 rd_kafka_buf_read_str(rkbuf, &cluster_id);
                 if (!RD_KAFKAP_STR_IS_NULL(&cluster_id)) {
-                        int clen        = RD_KAFKAP_STR_LEN(&cluster_id);
-                        mdi->cluster_id = rd_tmpabuf_alloc(&tbuf, clen + 1);
-                        if (mdi->cluster_id) {
-                                memcpy(mdi->cluster_id, cluster_id.str, clen);
-                                mdi->cluster_id[clen] = '\0';
-                        }
+                        int clen = RD_KAFKAP_STR_LEN(&cluster_id);
+                        if (!(mdi->cluster_id =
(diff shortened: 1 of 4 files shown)


src/rdkafka_metadata.c
```diff
@@ -677,9 +677,14 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
...
-                        mdi->cluster_id =
-                            rd_tmpabuf_write_str(&tbuf, cluster_id.str);
+                if (!RD_KAFKAP_STR_IS_NULL(&cluster_id)) {
+                        int clen        = RD_KAFKAP_STR_LEN(&cluster_id);
+                        mdi->cluster_id = rd_tmpabuf_alloc(&tbuf, clen + 1);
+                        if (mdi->cluster_id) {
+                                memcpy(mdi->cluster_id, cluster_id.str, clen);
+                                mdi->cluster_id[clen] = '\0';
+                        }
+                }
```
emasab, line 689:
```suggestion
                }  else {
                        rd_kafka_buf_parse_fail(
                            rkbuf,
                            "cluster_id: tmpabuf memory shortage");
                }
```

emasab: Just this comment to address memory shortage, thanks

emasab (changes requested):

src/rdkafka_metadata.c
prashah-confluent: Added the following change to maintain consistency with other places.

```
if (!(mdi->cluster_id =
          rd_tmpabuf_alloc(&tbuf, clen + 1)))
        rd_kafka_buf_parse_fail(
            rkbuf,
            "cluster_id: tmpabuf memory shortage");
memcpy(mdi->cluster_id, cluster_id.str, clen);
mdi->cluster_id[clen] = '\0';
```

emasab (approved):
Thanks @prashah-confluent ! Approving

PR #5547, opened 2026-07-08
Report ALL_BROKERS_DOWN only once per outage with metadata.recovery.strategy=rebootstrap

### LINKED ISSUE #5546: Repeated ALL_BROKERS_DOWN error on every re-bootstrap cycle during sustained outage (metadata.recovery.strategy=rebootstrap)
## Description

With `metadata.recovery.strategy=rebootstrap` (the default since v2.10.0 / KIP-899), a client
whose brokers are **all** unreachable for a sustained period receives
`RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN` (`Local_AllBrokersDown`, -187) on **every re-bootstrap
cycle** — roughly once per reconnect interval, for the whole duration of the outage — instead
of once per outage. First observed in production as a consumer that never got a single
successful bootstrap connection (the peer closed the connection during `APIVERSION_QUERY` on
every attempt) and logged the error every few seconds for minutes.

## Mechanism (traced against master / v2.15.0; identical since v2.11.1)

1. `rd_kafka_broker_set_state()` (`src/rdkafka_broker.c`) raises `ALL_BROKERS_DOWN` under the
   `/* Only 0 -> 1 */` `rkb_down_reported` guard when the last non-logical broker goes down,
   and calls `rd_kafka_rebootstrap()` in the same all-brokers-down block.
2. The re-bootstrap runs on an immediate one-shot timer whose callback (`src/rdkafka.c`)
   calls `rd_kafka_reset_any_broker_down_reported()`, clearing every broker's
   `rkb_down_reported`. This is intentional per #5126 (v2.11.1): all brokers must be re-tried
   before the all-down condition may be declared again.
3. During a sustained outage every broker keeps failing, so each broker's next `DOWN`
   transition passes the 0 → 1 guard again, the all-down threshold is crossed again, and the
   error is reported again — once per cycle, indefinitely.

Nothing rate-limits the loop: `metadata.recovery.rebootstrap.trigger.ms` only applies to the
separate metadata-timeout trigger, and `rk_rebootstrap_in_progress` is cleared at the end of
each timer callback. The #5126 commit message acknowledges the repeat ("given the down
reported state is reset, after retrying on re-bootstrap a new ALL_BROKERS_DOWN error is
issued"). With `metadata.recovery.strategy=none` the error fires once per outage, as the
guard intends.

## Expected behavior

`ALL_BROKERS_DOWN` should be reported once per distinct outage and re-armed when some broker
connection comes up again, independent of the recovery strategy.

## How to reproduce (deterministic, mock cluster)

Make every connection fail during the ApiVersion handshake — the mock broker closes the
connection when returning `RD_KAFKA_RESP_ERR__TRANSPORT`:

```c
rd_kafka_mock_push_request_errors_array(mcluster, RD_KAFKAP_ApiVersion,
                                        cnt, errs /* __TRANSPORT * cnt */);
```

An idle client connected to a 3-broker mock cluster set up this way receives one -187 error
event per re-bootstrap cycle: 17 events in 10 s with `reconnect.backoff.ms=100` /
`reconnect.backoff.max.ms=500`.

## Note on reconnect backoff

The reconnect backoff itself was initially suspected of not growing, but tracing and mock
reproduction ruled that out: during such an outage the backoff grows exponentially and caps
at `reconnect.backoff.max.ms`, and the steady-state cycle cadence equals the capped backoff.
Re-bootstrap does not reset any backoff state (re-added bootstrap brokers dedupe to no-ops
for existing `RD_KAFKA_CONFIGURED` brokers).

## Related issues

- #5088 — re-bootstrap loop from bootstrap-vs-learned broker connection preference; distinct,
  our case never reaches a learned-brokers state at all.
- #5301 — permanent disconnection after a `0 brokers, topic_cnt > 0` metadata response;
  opposite failure mode (stops retrying entirely).

## Environment

- librdkafka v2.14.0 (via confluent-kafka-dotnet / Confluent.Kafka 2.14.0); behavior verified
  unchanged on master (v2.15.0).
- Consumer client, `security.protocol=SASL_SSL`, `sasl.mechanism=PLAIN`, Confluent Cloud.
- `metadata.recovery.strategy` at its default (`rebootstrap`).

## Fix

Proposed in #5547: report the error only on the 0 → 1 transition of a per-instance
all-brokers-down latch, cleared when any non-logical broker enters the UP state.
Re-bootstrap behavior and the #5126 down-reported reset semantics are unchanged.

lahma:
Fixes #5546 (first part: repeated `ALL_BROKERS_DOWN` events).

## Problem

With `metadata.recovery.strategy=rebootstrap` (the default since v2.10.0 / KIP-899), a client
whose brokers are **all** unreachable for a sustained period reports
`RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN` to the application on **every re-bootstrap cycle**
(roughly every reconnect interval) instead of once per outage.

Mechanism:

- `rd_kafka_broker_set_state()` raises the error under the `/* Only 0 -> 1 */`
  `rkb_down_reported` guard when the last non-logical broker goes down, and starts a
  re-bootstrap in the same block (`src/rdkafka_broker.c`).
- The re-bootstrap timer callback calls `rd_kafka_reset_any_broker_down_reported()`, which
  resets every broker's `rkb_down_reported` so that all brokers are re-tried before the
  all-down condition can be reached again (intent of #5126).
- During a *sustained* outage every broker keeps failing, so the guard re-arms and the error
  re-fires on every cycle. Nothing rate-limits this
  (`metadata.recovery.rebootstrap.trigger.ms` only applies to the separate metadata-timeout
  trigger). The #5126 commit message acknowledges this side effect ("after retrying on
  re-bootstrap a new ALL_BROKERS_DOWN error is issued").

Reproduced with a mock cluster closing every connection during ApiVersion negotiation:
**17 error events in 10 seconds** on an idle consumer.

## Fix

Add an `rk_all_brokers_down_reported` latch:

- The error is reported only when the all-brokers-down condition is reached and the latch
  transitions 0 → 1; subsequent cycles of the same outage log a `BROKER` debug message
  instead.
- The latch is cleared when any non-logical broker enters the UP state, so a genuine
  recovery followed by a new outage reports the error again.
- Re-bootstrap behavior is unchanged: it still runs on every all-down condition, and the
  #5126 semantics (reset `rkb_down_reported` so all brokers are re-tried) are untouched —
  only the application-visible event is de-duplicated.

## Tests

Three new mock-cluster subtests in `tests/0152-rebootstrap.c` (local, no real cluster
needed), using ApiVersion + `RD_KAFKA_RESP_ERR__TRANSPORT` request-error injection to make
the mock broker close every connection during the handshake (the exact failure mode from
#5546):

1. Sustained outage (3 brokers, 10 s): exactly **1** `ALL_BROKERS_DOWN` error while
   re-bootstrap keeps cycling (≥ 2 sequences; observed ~19). Fails with **17** errors
   before the fix.
2. Second outage after recovery: exactly one more error (latch re-arms on broker UP).
3. Reconnect backoff regression guard: during a sustained outage the backoff keeps growing
   until capped by `reconnect.backoff.max.ms`, pacing connection attempts (3-6 attempts in
   10 s with 1000/5000 ms settings) — guards the investigated-and-disproved hypothesis from
   #5546 that re-bootstrap resets the reconnect backoff.

`0000` (unit tests), `0095`, `0151` and the local quick suite pass.

Note: the error de-duplication relies on `rd_atomic32_set()` returning the previous value,
which is broken in CMake builds before #5136 (shipped in v2.15.0) — this PR targets master
where that fix is present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,28 @@
+# librdkafka v2.15.1
+
+librdkafka v2.15.1 is a maintenance release:
+
+- Fix to report the `RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN` error only once per
+  outage instead of on every re-bootstrap cycle while all brokers stay
+  unreachable (#5546).
+
+
+## Fixes
+
+### General fixes
+
+- Issues: #5546.
+  The `RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN` error was reported on every
+  re-bootstrap cycle (roughly every reconnect interval) for the whole duration
+  of an outage in which all brokers are unreachable, instead of only once per
+  outage, because the re-bootstrap sequence resets the per-broker
+  down-reported state. The error is now reported once per outage and armed
+  again as soon as any broker connection comes up. Re-bootstrap sequences
+  continue on every cycle as before.
+  Happening since 2.11.1 (#5126).
+
+
+
 # librdkafka v2.15.0
 
 librdkafka v2.15.0 is a feature release:
--- a/src/rdkafka.c
+++ b/src/rdkafka.c
@@ -2610,6 +2610,7 @@ rd_kafka_t *rd_kafka_new(rd_kafka_type_t type,
         rd_atomic32_init(&rk->rk_logical_broker_cnt, 0);
         rd_atomic32_init(&rk->rk_broker_up_cnt, 0);
         rd_atomic32_init(&rk->rk_broker_down_cnt, 0);
+        rd_atomic32_init(&rk->rk_all_brokers_down_reported, 0);
         rd_atomic32_init(&rk->rk_rebootstrap_in_progress, 0);
 
         rk->rk_rep             = rd_kafka_q_new(rk);
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -372,20 +372,46 @@ void rd_kafka_broker_set_state(rd_kafka_broker_t *rkb, int state) {
                    rd_atomic32_set(&rkb->rkb_down_reported, 1) == 0) {
 
                 /* Propagate ALL_BROKERS_DOWN event if all brokers are
-                 * now down, unless we're terminating. */
+                 * now down, unless we're terminating.
+                 * The error is reported only once per outage: the
+                 * re-bootstrap sequence resets each broker's down reported
+                 * state to retry all brokers before reaching this condition
+                 * again, so the error would otherwise be reported on every
+                 * re-bootstrap cycle for the same continuing outage.
+                 * It's armed again when any broker connection comes up. */
                 if (rd_atomic32_add(&rkb->rkb_rk->rk_broker_down_cnt, 1) ==
                         rd_atomic32_get(&rkb->rkb_rk->rk_broker_cnt) -
                             rd_atomic32_get(
                                 &rkb->rkb_rk->rk_logical_broker_cnt) &&
                     !rd_kafka_terminating(rkb->rkb_rk)) {
                         rd_kafka_rebootstrap(rkb->rkb_rk);
-                        rd_kafka_op_err(
-                            rkb->rkb_rk, RD_KAFKA_RESP_ERR__ALL_BROKERS_DOWN,
-                            "%i/%i brokers are down",
-                            rd_atomic32_get(&rkb->rkb_rk->rk_broker_down_cnt),
-                            rd_atomic32_get(&rkb->rkb_rk->rk_broker_cnt) -
-                                rd_atomic32_get(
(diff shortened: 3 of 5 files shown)


emasab:
We were thinking about this during the development but we considered leaving it at every reconnect cycle, because that way we're able to see how long the downtime lasted on the client. Given the reconnect interval is 10s by default the amount of logging is low, even if you set it to 1s max it's around 4MB of logging for a 12h downtime.
There's also the possibility to set a logging callback where you can decide to filter the received logs if you want fewer or a single one.

lahma:
Thanks for the explanation, makes sense.  I think I can close this one then.

PR #4584, opened 2024-01-16
[KIP-951] Fetch RPC prework: upgrade to v15

mahajanadhitya:

PR #5550, opened 2026-07-15
Fix NULL key/value when consuming legacy MessageSets with flexible Fetch versions (v12+)

### LINKED ISSUE #5549: Consumer returns NULL key/value for legacy MessageSet format (message.format.version < 0.11) since v2.5.0
Description
===========
When consuming from a cluster or topic that stores messages in the legacy MessageSet format (`log.message.format.version` < 0.11, MsgVersion v0..v1), every message is delivered to the application with **NULL key and NULL value**: fetches succeed and offsets advance, but all payloads are NULL. Offsets and timestamps are correct. The consumer also raises errors such as:

```
Unsupported Message(Set) MagicByte 97 at offset 1525458452426752
```

Bisected to 6867b9f9 ("Fetch RPC prework: upgrade to v15", #4584, v2.5.0): since Fetch v12+ the response buffer is parsed as a flexible version, and the legacy MessageSet reader reads Message Key/Value with `rd_kafka_buf_read_kbytes()`, which switches to compact (varint) length parsing for flexible-version buffers. Legacy MessageSets always use fixed-width int32 length prefixes, so the first `0x00` length byte parses as a compact NULL and the reader drifts into the message bytes (the `MagicByte 97` above is a payload byte). On v2.5.x fetches fail/stall instead; from v2.6.0 onwards messages are delivered with NULL key/value as described. Still present on v2.15.0/master.

Found in production through node-rdkafka 3.6.1 (librdkafka 2.12.0), consuming from Apache Kafka 3.9.1 brokers that pin `log.message.format.version=0.10.2.2` for rolling-upgrade compatibility.

How to reproduce
================
Run any broker with `log.message.format.version=0.10.2.2` (e.g. Apache Kafka 3.9 in ZooKeeper mode with `inter.broker.protocol.version=0.10.2.2`), produce keyed messages, consume them: all messages arrive with NULL key/value. The fix PR adds a mock-broker regression test (0187) that reproduces this by capping the ProduceRequest version to <= v2 so the producer writes MsgVersion v1/v0 MessageSets.

Checklist
=========

 - [x] librdkafka version: v2.15.0 (also 2.6.0, 2.8.0, 2.11.1, 2.12.0, 2.14.2; regression since v2.5.0, bisected to 6867b9f9)
 - [x] Apache Kafka version: 3.9.1 with `inter.broker.protocol.version=0.10.2.2`, `log.message.format.version=0.10.2.2`
 - [x] librdkafka client configuration: default, `auto.offset.reset=earliest`
 - [x] Operating system: Linux (x64)
 - [x] Provide logs: `Unsupported Message(Set) MagicByte 97 at offset 1525458452426752` (`_NOT_IMPLEMENTED`) on every fetch
 - [ ] Provide broker log excerpts: no broker-side errors (messages stored intact, verified with `DumpLogSegments`)
 - [ ] Critical issue

delthas:
Fixes #5549

**Symptom**

When consuming from a cluster or topic using the legacy message format (`log.message.format.version` < 0.11, MsgVersion v0..v1), fetches succeed and offsets advance, but every message is delivered to the application with NULL key and NULL value (offsets and timestamps are correct), together with consumer errors such as:

```
Unsupported Message(Set) MagicByte 97 at offset 1525458452426752
```

Happening since v2.5.0. On v2.5.x fetches fail/stall instead; from v2.6.0 onwards messages are delivered with NULL key/value as described.

Found in production through node-rdkafka 3.6.1 (librdkafka 2.12.0), consuming from Apache Kafka 3.9.1 brokers that pin `log.message.format.version=0.10.2.2` for rolling-upgrade compatibility.

**Root cause**

Since the Fetch RPC was upgraded to flexible versions (#4584, v2.5.0), the FetchResponse buffer is flagged as a flexible-version response. The legacy MessageSet reader (`rd_kafka_msgset_reader_msg_v0_1()`) reads the Message Key and Value with `rd_kafka_buf_read_kbytes()`, which switches to compact (varint-prefixed) parsing when that flag is set — but legacy MessageSets always use fixed-width (int32-prefixed) encodings regardless of the Fetch version. The first `0x00` byte of the length prefix parses as a compact NULL, and the reader then drifts into the middle of the message bytes, misparsing the rest of the MessageSet.

**Fix**

Add `rd_kafka_buf_read_kbytes_fixed()`, which always reads the fixed-width encoding, and use it for the legacy Message Key and Value.

**Testing**

- The mock broker previously rejected any MessageSet with MagicByte != 2, so this path was untestable: a second commit adds support for legacy uncompressed v0..v1 MessageSets to the mock broker.
- New regression test 0187 caps the ProduceRequest version on the mock cluster so the producer writes MsgVersion v1 (or v0) MessageSets, and verifies a consumer using a flexible Fetch version receives them intact. Fails without the fix, passes with it.
- Also reproduced and verified against a real Apache Kafka 3.9 broker configured with `inter.broker.protocol.version=0.10.2.2` / `log.message.format.version=0.10.2.2`.

### DIFF
--- a/src/rdkafka_buf.h
+++ b/src/rdkafka_buf.h
@@ -770,6 +770,27 @@ struct rd_kafka_buf_s { /* rd_kafka_buf_t */
         } while (0)
 
 /**
+ * Read standard BYTES representation (4+N), regardless of the buffer's
+ * flexible-version flag: for structures that always use the non-compact
+ * encoding, such as legacy v0..v1 MessageSets in FetchResponses >= v12.
+ * The 'kbytes' will be updated to point to rkbuf data.
+ */
+#define rd_kafka_buf_read_kbytes_fixed(rkbuf, kbytes)                          \
+        do {                                                                   \
+                int32_t _klen;                                                 \
+                rd_kafka_buf_read_i32a(rkbuf, _klen);                          \
+                (kbytes)->len = _klen;                                         \
+                if (RD_KAFKAP_BYTES_IS_NULL(kbytes)) {                         \
+                        (kbytes)->data = NULL;                                 \
+                        (kbytes)->len  = 0;                                    \
+                } else if (RD_KAFKAP_BYTES_LEN(kbytes) == 0)                   \
+                        (kbytes)->data = "";                                   \
+                else if (!((kbytes)->data = rd_slice_ensure_contig(            \
+                               &(rkbuf)->rkbuf_reader, _klen)))                \
+                        rd_kafka_buf_check_len(rkbuf, _klen);                  \
+        } while (0)
+
+/**
  * @brief Read \p size bytes from buffer, setting \p *ptr to the start
  *        of the memory region.
  */
--- a/src/rdkafka_mock.c
+++ b/src/rdkafka_mock.c
@@ -96,11 +96,16 @@ static void rd_kafka_mock_msgset_destroy(rd_kafka_mock_partition_t *mpart,
 /**
  * @brief Create a new msgset object with a copy of \p bytes
  *        and appends it to the partition log.
+ *
+ * @param MagicByte the MsgVersion of \p bytes: v2 MessageSets get their
+ *        BaseOffset and PartitionLeaderEpoch updated, while legacy v0..v1
+ *        MessageSets get an absolute offset assigned to each message.
  */
 static rd_kafka_mock_msgset_t *
 rd_kafka_mock_msgset_new(rd_kafka_mock_partition_t *mpart,
                          const rd_kafkap_bytes_t *bytes,
-                         size_t msgcnt) {
+                         size_t msgcnt,
+                         int8_t MagicByte) {
         rd_kafka_mock_msgset_t *mset;
         size_t totsize = sizeof(*mset) + RD_KAFKAP_BYTES_LEN(bytes);
         int64_t BaseOffset;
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
...
+  The Message Key and Value fields were incorrectly parsed as compact
+  (varint-prefixed) bytes when the enclosing FetchResponse used a flexible
+  version (v12+), while legacy MessageSets always use the fixed-width
+  encoding. Messages were returned to the application with NULL key and
+  value, and the remainder of the MessageSet was misparsed, raising
+  `Unsupported Message(Set) MagicByte` errors.
+  Happening since v2.5.0 (#5549, @delthas).
+
+
+
```
Ankith-Confluent: Please drop the `CHANGELOG` changes from this PR. We add the changelog entry on our side during the next release.

Ankith-Confluent (changes requested):
Thanks for the PR @delthas! 
The PR looks good. Just one minor comment from my side.

CHANGELOG.md
delthas: Done — dropped the CHANGELOG commit. Thanks for the review!

Ankith-Confluent:
Hi @delthas 
Could you please rebase your PR with master.

delthas:
Hi @Ankith-Confluent,

Done.

Ankith-Confluent (approved):
I have approved the PR @delthas 
LGTM!

We will merge it after another round of review.

delthas:
@Ankith-Confluent Nice! Waiting on the next round of review.

k-raina (approved):
LGTM! Thanks for the fix

PR #5554, opened 2026-07-24
[FIX] Fall back to archive.apache.org for Kafka download in share telemetry pipeline

Ankith-Confluent:
The share-consumer telemetry CI job pins `KAFKA_VERSION=4.3.0`, but
`downloads.apache.org` only keeps the latest patch per minor — once 4.3.1
shipped, 4.3.0 was removed and the download 404'd on a cache miss (silently,
due to `wget -q`).

Fix: try `downloads.apache.org` first, then fall back to `archive.apache.org`
(which retains all releases), and use `wget -nv` so failures are visible.

### DIFF
--- a/packaging/tools/setup-share-telemetry-testing.sh
+++ b/packaging/tools/setup-share-telemetry-testing.sh
@@ -32,8 +32,13 @@ fi
 
 if [ ! -d "${KAFKA_DIR}" ]; then
     echo "==> Setting up Kafka broker ${KAFKA_VERSION}"
-    wget -q https://downloads.apache.org/kafka/${KAFKA_VERSION}/kafka_2.13-${KAFKA_VERSION}.tgz \
-        -O ${WORK_DIR}/kafka.tgz
+    KAFKA_TGZ=kafka_2.13-${KAFKA_VERSION}.tgz
+    if ! wget -nv "https://downloads.apache.org/kafka/${KAFKA_VERSION}/${KAFKA_TGZ}" \
+            -O ${WORK_DIR}/kafka.tgz; then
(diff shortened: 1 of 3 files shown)


packaging/tools/setup-share-telemetry-testing.sh
```diff
@@ -32,8 +32,13 @@ fi
...
     echo "==> Setting up Kafka broker ${KAFKA_VERSION}"
-    wget -q https://downloads.apache.org/kafka/${KAFKA_VERSION}/kafka_2.13-${KAFKA_VERSION}.tgz \
-        -O ${WORK_DIR}/kafka.tgz
+    KAFKA_TGZ=kafka_2.13-${KAFKA_VERSION}.tgz
+    if ! wget -nv "https://downloads.apache.org/kafka/${KAFKA_VERSION}/${KAFKA_TGZ}" \
+            -O ${WORK_DIR}/kafka.tgz; then
+        echo "==> ${KAFKA_VERSION} not on downloads.apache.org, trying archive.apache.org"
+        wget -nv "https://archive.apache.org/dist/kafka/${KAFKA_VERSION}/${KAFKA_TGZ}" \
+            -O ${WORK_DIR}/kafka.tgz
+    fi
```
pranavrth, line 41: Why don't we download directly from `archive.apache.org` instead of `downloads.apache.org`?

Ankith-Confluent: This is mainly for download speed, if the version is still available on `downloads.apache.org` it'll download from there at higher speeds, otherwise it falls back to `archive.apache.org`.

pranavrth (approved):

PR #5557, opened 2026-07-28
Fix Clang CFI icall type mismatch in rd_list_remove_cmp and string comparators

EricFXU:
### Problem
Under Clang Control Flow Integrity (`-fsanitize=cfi-icall`), function pointer signatures at indirect call sites must match the target function parameter types exactly.

Passing `(void *)strcmp` or comparators without `const` parameters to `rd_list_remove_cmp` and `rd_list_find` causes Clang CFI to flag due to type mismatches between `int (*)(void *, void *)` and `int (*)(const void *, const void *)`.

### Fix
1. Updated `rd_list_remove_cmp` and `rd_list_remove_multi_cmp` signatures in `src/rdlist.h` and `src/rdlist.c` to accept `int (*cmp)(const void *_a, const void *_b)`.
2. Replaced `(void *)strcmp` function pointer casts with `rd_list_cmp_str` in `src/rdkafka_metadata.c`, `src/rdkafka_partition.c`, `src/rdkafka_topic.c`, and `src/rdkafka_sasl_oauthbearer.c`.

### DIFF
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -951,7 +951,7 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
                         rd_list_free_cb(missing_topics,
                                         rd_list_remove_cmp(missing_topics,
                                                            md->topics[i].topic,
-                                                           (void *)strcmp));
+                                                           rd_list_cmp_str));
                 if (requested_topic_ids)
                         rd_list_free_cb(
                             missing_topic_ids,
--- a/src/rdkafka_metadata.c
+++ b/src/rdkafka_metadata.c
@@ -957,7 +957,7 @@ rd_kafka_parse_Metadata0(rd_kafka_broker_t *rkb,
                             missing_topic_ids,
                             rd_list_remove_cmp(missing_topic_ids,
                                                &mdi->topics[i].topic_id,
-                                               (void *)rd_kafka_Uuid_ptr_cmp));
+                                               rd_kafka_Uuid_ptr_cmp));
                 /* Only update cache when not asking
                  * for all topics or cache entry
                  * already exists. */
--- a/src/rdlist.c
+++ b/src/rdlist.c
@@ -194,7 +194,7 @@ void *rd_list_remove(rd_list_t *rl, void *match_elem) {
 
 void *rd_list_remove_cmp(rd_list_t *rl,
                          void *match_elem,
-                         int (*cmp)(void *_a, void *_b)) {
+                         int (*cmp)(const void *_a, const void *_b)) {
         void *elem;
         int i;
 
@@ -211,7 +211,7 @@ void *rd_list_remove_cmp(rd_list_t *rl,
 
 int rd_list_remove_multi_cmp(rd_list_t *rl,
                              void *match_elem,
-                             int (*cmp)(void *_a, void *_b)) {
+                             int (*cmp)(const void *_a, const void *_b)) {
 
         void *elem;
         int i;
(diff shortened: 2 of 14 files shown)


src/rdlist.h
```diff
@@ -161,7 +161,7 @@ void *rd_list_remove(rd_list_t *rl, void *match_elem);
  */
 void *rd_list_remove_cmp(rd_list_t *rl,
                          void *match_elem,
-                         int (*cmp)(void *_a, void *_b));
+                         int (*cmp)(const void *_a, const void *_b));
```
Ankith-Confluent, line 164: This `const` breaks the other caller of `rd_list_remove_cmp`, 4 lines below the one you fixed in `rdkafka_metadata.c`:

```c
/* rdkafka_metadata.c:960 */
rd_list_remove_cmp(missing_topic_ids, &mdi->topics[i].topic_id,
                   (void *)rd_kafka_Uuid_ptr_cmp);
```

`rd_kafka_Uuid_ptr_cmp` has no `const` (`rdkafka_proto.h:622`), so it matches today but won't after this change. The `(void *)` cast hides it from the compiler, so the build stays clean and CI won't catch it. 

Fix

```diff
-static RD_INLINE RD_UNUSED int rd_kafka_Uuid_ptr_cmp(void *a, void *b) {
-        rd_kafka_Uuid_t *a_uuid = a, *b_uuid = b;
+static RD_INLINE RD_UNUSED int rd_kafka_Uuid_ptr_cmp(const void *a,
+                                                     const void *b) {
+        const rd_kafka_Uuid_t *a_uuid = a, *b_uuid = b;
```

Then drop the `(void *)` at `rdkafka_metadata.c:960` and `rdkafka_cgrp.c:3049`. With the cast gone the compiler enforces this, so it can't silently regress again.

src/rdkafka_metadata.c
Ankith-Confluent: Line 960 passes `(void *)rd_kafka_Uuid_ptr_cmp` to this same function, and the `rdlist.h` change breaks it, see my comment on `rdlist.h`.

Ankith-Confluent (changes requested):
Thanks for the PR @EricFXU.
I've added a few comments. 
Once those are addressed we can run the CI and see if anything else comes up.

src/rdlist.h
EricFXU: Hi @Ankith-Confluent, thanks for catching this! Our usage is currently based on an older snapshot of the library (around v2.2.0) that didn't have this function, which was added later. I've made the fix.

Ankith-Confluent:
Hi @EricFXU 
There are few style fixes
Could you run `make style-fix` to resolve it.

EricFXU:
Hi @Ankith-Confluent, I had installed the wrong version of `cmake` (`pip install` installed `18.1.18`), and I realized `style-format.sh` only checks the major version number is set to 18. With the help of AI, checked and confirmed the dockerfile's image has `ubuntu:24.04` , which uses `18.1.3` version. After installing this version, I think the `make style-fix` runs the right format. Please LMK if this doesn't pass and I can try something else.

Also, could you please help answer when we can expect this fix to be released?

Ankith-Confluent:
Hey @EricFXU 
The dates on the release is not finalised yet, but it should be by the end of the month.

Ankith-Confluent (approved):
Thanks for the PR @EricFXU 
LGTM!

We will be merging it after another round of review.

src/rdlist.c
```diff
@@ -211,7 +211,7 @@ void *rd_list_remove_cmp(rd_list_t *rl,
 
 int rd_list_remove_multi_cmp(rd_list_t *rl,
```
k-raina, line 212: Nit: This function doesn't have callers.

k-raina (approved):
Thanks for PR!

PR #5561, opened 2026-08-07
[SSL] Report OpenSSL error when certificate chain can't be read (#5556)

### LINKED ISSUE #5556: Backward compatibility regression in ssl.certificate.pem parsing introduced by #4894
## Description

After #4894, `ssl.certificate.pem` parsing became stricter and is no longer backward compatible with librdkafka v2.6.1.

Before #4894, librdkafka read the first valid X509 certificate and ignored the remaining input. Starting with v2.8.0, the remaining input is parsed as a certificate chain. Any parsing error in the remaining data causes the already successfully parsed leaf certificate to be discarded.

This produces the following error:

```text
Failed to read certificate chain from PEM. Returning NULL certificate too.
ssl.certificate.pem failed: not in PEM format?
```

The behavior is still present in the current upstream master.

## Downstream observation

We observed this while upgrading a downstream application:

- With the older librdkafka version, an inline certificate passed through `ssl.certificate.pem` works.
- After upgrading librdkafka, the same configuration fails with the error above.
- Specifying the same certificates using file paths continues to work.

A similar regression was reported in a comment on #4894:

https://github.com/confluentinc/librdkafka/pull/4894#issuecomment-2752560095

## Minimal reproducer

The following can be added to `main_0097_ssl_verify_local()` in `tests/0097-ssl_verify.cpp`.

It starts with a valid certificate fixture and appends a malformed certificate block. This models input that older librdkafka versions accepted because they stopped parsing after the first valid certificate.

```cpp
const char *regression_pem_path =
    test_getenv("SSL_CERTIFICATE_PEM_REGRESSION_FILE",
                "fixtures/ssl/client2.certificate.pem");

std::vector<char> pem = read_file(regression_pem_path);
std::string certificate_pem(pem.begin(), pem.end());

certificate_pem += "\n-----BEGIN CERTIFICATE-----\n"
                   "AAAA\n"
                   "-----END CERTIFICATE-----\n";

RdKafka::Conf *conf =
    RdKafka::Conf::create(RdKafka::Conf::CONF_GLOBAL);
std::string errstr;

if (conf->set("security.protocol", "SSL", errstr) !=
    RdKafka::Conf::CONF_OK)
  Test::Fail(errstr);

if (conf->set("ssl.certificate.pem", certificate_pem, errstr) !=
    RdKafka::Conf::CONF_OK)
  Test::Fail(errstr);

RdKafka::Producer *producer = RdKafka::Producer::create(conf, errstr);
delete conf;

if (!producer)
  Test::Fail(
      "A PEM accepted by the compatibility baseline was rejected: " +
      errstr);

delete producer;
```

No Kafka broker is required because the failure happens during client construction.

Run with:

```bash
LD_LIBRARY_PATH=../src:../src-cpp \
./test-runner -p1 -l 0097_ssl_verify_local
```

## Results

| Version | Result |
|---|---|
| v2.6.1 | PASS — the first valid certificate is accepted |
| v2.13.0 | FAIL |
| master `4c3b017c3de8d9c10de12b9eec0ea688e8d5787d` | FAIL |

Failure on v2.13.0 and master:

```text
Failed to read certificate chain from PEM. Returning NULL certificate too.

A PEM accepted by the compatibility baseline was rejected:
ssl.certificate.pem failed: not in PEM format?:
kafka: No further error information available
```

## OpenSSL versions

The issue was reproduced with OpenSSL 3.0.13.

The same certificate-chain parsing algorithm was also tested with OpenSSL 3.2.1 and produced the expected ASN.1 parsing error for the malformed trailing block. Therefore, the compatibility change appears to be caused by librdkafka's new full-chain parsing behavior rather than by an OpenSSL regression.

A clean single certificate and the complete old v2.6.1 fixture are accepted by newer versions in our environment.

## Expected behavior

Please clarify whether rejecting the entire client certificate when a trailing certificate cannot be parsed is an intentional breaking change.

Possible compatibility options:

1. Keep the successfully parsed leaf certificate and ignore an invalid trailing block, preserving the pre-v2.8.0 behavior.
2. Introduce an explicit strict-chain parsing option.
3. If strict rejection is intended, document the breaking change and return the underlying OpenSSL error instead of:

   ```text
   kafka: No further error information available
   ```

File-based certificate loading and inline `ssl.certificate.pem` loading should ideally apply consistent parsing rules.

Ankith-Confluent:
Fixes https://github.com/confluentinc/librdkafka/issues/5556

A PEM in `ssl.certificate.pem` whose leaf certificate parses but whose
trailing block does not is rejected with an error that names neither the
reason nor the part that failed:

    ssl.certificate.pem failed: not in PEM format?: kafka: No further error information available

Both halves of that message are unhelpful. The leaf *was* in PEM format —
it parsed fine — and the OpenSSL reason is reported as unavailable even
though OpenSSL did produce one.

## Root cause

`rd_kafka_ssl_read_cert_chain_from_BIO()` called `ERR_clear_error()`
unconditionally, including on the failure path. By the time the caller
reached `rd_kafka_ssl_error()` the queue was empty, so the three ASN.1
errors describing the failure had already been discarded.

The `not in PEM format?` text was used for both a leaf failure and a chain
failure, so it pointed at the wrong part of the input.

The same applied to a PEM public key passed to
`rd_kafka_conf_set_ssl_cert()`, which reads the chain through the same
function and reported an empty reason.

## Changes

- Clear the OpenSSL error queue only on success, so a real parse error
  survives for the caller to report.
- Distinguish a leaf failure from a chain failure so the message names the
  part that actually failed.
- Clear the error queue at `rd_kafka_cert_new()`'s failure label. That path
  only *peeks* at the queue, so preserving errors would otherwise leave them
  behind to be reported again by a later, unrelated failure.

The error now reads:

    ssl.certificate.pem failed: error reading certificate chain: error:0488000D:PEM routines::ASN1 lib

## Not changed

Rejecting the certificate is intentional and stays. Since 2.8.0 the client
certificate chain is sent to the broker, so a chain that cannot be parsed
cannot be silently truncated — the leaf alone would produce an incomplete
chain and move the failure to the TLS handshake.

This also keeps the in-memory path consistent with the file-based one:
`SSL_CTX_use_certificate_chain_file()`, used by `ssl.certificate.location`,
rejects the same input with the same ASN.1 errors.

## Testing

New sub-test in `0097_ssl_verify_local` (no broker required) covering a
valid certificate, a malformed trailing block via `ssl.certificate.pem`,
and the same input via `set_ssl_cert()`. It fails on master with the exact
message from the issue and passes with this change.

### DIFF
--- a/src/rdkafka_cert.c
+++ b/src/rdkafka_cert.c
@@ -452,6 +452,10 @@ fail:
         if (p12)
                 PKCS12_free(p12);
 
+        /* The reason has been copied to errstr, don't leave it on the queue
+         * where a later unrelated failure would report it again. */
+        ERR_clear_error();
+
         return NULL;
 }
 #endif /* WITH_SSL */
--- a/src/rdkafka_cert.c
+++ b/src/rdkafka_cert.c
@@ -181,6 +181,9 @@ static rd_kafka_cert_t *rd_kafka_cert_new(const rd_kafka_conf_t *conf,
                 return NULL;
         }
 
+        /* Start clean so a failure below reports this call's error only. */
+        ERR_clear_error();
+
         action = "read memory";
         bio    = BIO_new_mem_buf((void *)buffer, (long)size);
         if (!bio)
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -776,8 +776,9 @@ static EVP_PKEY *rd_kafka_ssl_PKEY_from_string(rd_kafka_t *rk,
  *
  * @return 0 on success, -1 on error.
  *
- * @remark On error the OpenSSL error queue is left intact for the caller to
- *         report the underlying reason.
+ * @remark The error queue is cleared on entry. On success it is left
+ *         cleared; on error it holds this function's error for the caller
+ *         to report.
  */
 int rd_kafka_ssl_read_cert_chain_from_BIO(BIO *in,
                                           STACK_OF(X509) * chainp,
@@ -786,6 +787,9 @@ int rd_kafka_ssl_read_cert_chain_from_BIO(BIO *in,
         X509 *ca;
         int r, ret = 0;
         unsigned long err;
+
+        ERR_clear_error();
+
         while (1) {
                 ca = X509_new();
                 if (ca == NULL) {
@@ -840,11 +844,16 @@ static X509 *rd_kafka_ssl_X509_from_string(rd_kafka_t *rk,
         BIO *bio = BIO_new_mem_buf((void *)str, -1);
         X509 *x509;
 
+        /* Start clean so the reason reported below is from this parse only. */
+        ERR_clear_error();
+
         x509 =
             PEM_read_bio_X509(bio, NULL, rd_kafka_transport_ssl_passwd_cb, rk);
 
         if (!x509) {
-                *reasonp = "not in PEM format?";
+                /* Keep neutral: a well-formed block with an invalid payload
+                 * also fails here. The real reason is on the error queue. */
+                *reasonp = "error reading certificate";
                 BIO_free(bio);
                 return NULL;
         }
@@ -1329,6 +1338,10 @@ static int rd_kafka_ssl_set_certs(rd_kafka_t *rk,
                                 BIO_free(bio);
                                 return -1;
                         }
+
+                        /* The final read raised a benign "no start line";
+                         * clear it so it isn't mistaken for a real error. */
+                        ERR_clear_error();
 
                         BIO_free(bio);
 
--- a/src/rdkafka_ssl.c
+++ b/src/rdkafka_ssl.c
@@ -769,11 +769,15 @@ static EVP_PKEY *rd_kafka_ssl_PKEY_from_string(rd_kafka_t *rk,
 /**
  * Read a PEM formatted cert chain from BIO \p in into \p chainp .
  *
- * @param rk rdkafka instance.
  * @param in BIO to read from.
  * @param chainp Stack to push the certificates to.
+ * @param password_cb Password callback for encrypted certificates.
+ * @param password_cb_opaque Opaque passed to \p password_cb .
  *
  * @return 0 on success, -1 on error.
+ *
+ * @remark On error the OpenSSL error queue is left intact for the caller to
+ *         report the underlying reason.
  */
 int rd_kafka_ssl_read_cert_chain_from_BIO(BIO *in,
                                           STACK_OF(X509) * chainp,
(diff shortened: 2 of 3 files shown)


src/rdkafka_ssl.c
```diff
@@ -800,14 +804,16 @@ int rd_kafka_ssl_read_cert_chain_from_BIO(BIO *in,
...
-        /* When the while loop ends, it's usually just EOF. */
+        /* When the while loop ends, it's usually just EOF.
+         * Anything else is a real error and is left on the error queue. */
         err = ERR_peek_last_error();
         if (ERR_GET_LIB(err) == ERR_LIB_PEM &&
-            ERR_GET_REASON(err) == PEM_R_NO_START_LINE)
+            ERR_GET_REASON(err) == PEM_R_NO_START_LINE) {
                 ret = 0;
-        else
+                ERR_clear_error();
```
k-raina, line 817: Is there possibility that error set by other code in this thread , which was not cleared , be read by rd_kafka_ssl_read_cert_chain_from_BIO ? This could result in reporting unrelated errors

src/rdkafka_ssl.c
k-raina, line 1346: Do we need to clear error here? i.e `ERR_clear_error();`? 
Can we check whole workflow and make sure we are clearing errors which were set correctly?

k-raina (changes requested):

src/rdkafka_ssl.c
Ankith-Confluent: yes this is possible, I have added the fix by clearing the queue.

src/rdkafka_ssl.c
Ankith-Confluent: Yes I had missed this, I have added ERR_clear_error().

src/rdkafka_cert.c
```diff
@@ -181,6 +181,9 @@ static rd_kafka_cert_t *rd_kafka_cert_new(const rd_kafka_conf_t *conf,
                 return NULL;
         }
 
+        /* Start clean so a failure below reports this call's error only. */
+        ERR_clear_error();
```
k-raina, line 185: What happens when we paste two CA certificates together into ssl.ca.pem, but the second one is corrupted?

src/rdkafka_cert.c
k-raina, line 474: If we set a client cert via the rd_kafka_conf_set_ssl_cert API (public-key + PEM) where the leaf cert is fine but the CA chain appended below it is corrupt. Will message be correct ?

k-raina (changes requested):
Couple of more comments. Lets verify if they are pre-existing issues in repo.

src/rdkafka_cert.c
Ankith-Confluent: This PR only changes the error message on failure, it doesn't change which certs are accepted or rejected. 
However I tested this case, and the corrupt second cert is silently dropped.

src/rdkafka_cert.c
Ankith-Confluent: Yes, this case is covered in the test case.

src/rdkafka_cert.c
k-raina: Thanks let add it to backlog.

k-raina (approved):

PR #5563, opened 2026-08-08
Release the whole broker when thread creation fails

### LINKED ISSUE #4456: File Descriptor leak when broker thread creation fails
Description
===========
We had an issue where broker threads were failing to spawn because of we were hitting `Max processes` ulimit, but this also caused our process to eventually hit a `max open files` limit due to what I believe are file descriptors from rdkafka not being cleaned up when it fails to create a broker thread.

When adding a new broker, two pipes are set up [here](https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_broker.c#L4872), however if the [broker thread fails to be created](https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_broker.c#L4895-L4913) these don't appear to be cleaned up.

You can also see in an `strace` that the pipes are created, the `clone` fails, then the two file descriptors are never destroyed:
```
[pid 157895] pipe([597, 598])           = 0
[pid 157895] fcntl(597, F_GETFL)        = 0 (flags O_RDONLY)
[pid 157895] fcntl(597, F_SETFL, O_RDONLY|O_NONBLOCK) = 0
[pid 157895] fcntl(598, F_GETFL)        = 0x1 (flags O_WRONLY)
[pid 157895] fcntl(598, F_SETFL, O_WRONLY|O_NONBLOCK) = 0
[pid 157895] fcntl(597, F_SETPIPE_SZ, 100) = 4096
[pid 157895] fcntl(598, F_SETPIPE_SZ, 100) = 4096
[pid 157895] clone(child_stack=0x7f09c278c6f0, flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, parent_tid=0x7f09c278e9d0, tls=0x7f09c278e700, child_tidptr=0x7f09c278e9d0) = -1 EAGAIN (Resource temporarily unavailable)
```

Checklist
=========

 - [x] librdkafka version (release number or git tag): `v2.0.2`
 - [ ] Apache Kafka version: `<REPLACE with e.g., 0.10.2.3>`
 - [ ] librdkafka client configuration: `<REPLACE with e.g., message.timeout.ms=123, auto.reset.offset=earliest, ..>`
 - [x] Operating system: `Ubuntu 20.04`
 - [ ] Critical issue

BNCMK:
# The broker is abandoned, not only its descriptors, when thread creation fails

Follows #4456, which reports two pipe descriptors leaking when `thrd_create()`
fails in `rd_kafka_broker_add()`. That report is correct and its fix closes the
descriptors. The same line abandons everything else the broker owns, and this is
a separate defect on the same path.

Closes #4456. An earlier draft claimed it applied on top of #4456; that was wrong,
and @xhon-pelushi is right that no such fix is in `master`. Issue #4456 is still open
and the leak it describes is present today. This change removes it along with the
memory leaks, and nothing else needs to land first.

## What else is abandoned

`rd_kafka_broker_add()` allocates the broker, its wake-up pipe, its ops queue,
its averaging structures, its name, its locks and condition variables, and takes
two references on it, all before calling `thrd_create()`. On failure the function
reaches:

```c
rd_free(rkb);
```

`rd_free()` frees the outer struct and nothing else. Everything
`rd_kafka_broker_destroy_final()` would release is left behind:

| abandoned | consequence |
|---|---|
| `rkb_origname`, `rkb_ApiVersions` | heap |
| `rkb_ops` queue | heap, normally via `rd_kafka_q_destroy_owner` |
| six `rd_avg` structures | heap |
| locks, condition variables, refcount | never destroyed |
| the two references taken above | refcount bypassed entirely |

The descriptors #4456 reports are the visible part, because they accumulate
somewhere a person can look. The rest is invisible until a process that retries
broker creation runs long enough to notice its heap.

## Why the obvious fix aborts

Calling the destructor twice, once per reference, is the natural repair and it
fails:

```
Assertion `thrd_is_current(rkb->rkb_thread)' failed.
```

`rd_kafka_broker_destroy_final()` opens with that assertion because in every
other case it is reached from the broker's own thread. Here that thread was
never created, so the handle is uninitialised and the assertion fires on
whatever it contains.

## The fix

At this point the broker is not yet on `rk_brokers`, `rd_kafka_sasl_broker_init`
has not run, and nothing else holds a pointer to it, so the calling thread is the
only thread that owns it. Recording that before tearing down makes the assertion
true rather than bypassed:

```c
rkb->rkb_thread = thrd_current();
rd_kafka_broker_destroy(rkb); /* broker thread's refcnt */
rd_kafka_broker_destroy(rkb); /* rk_broker's refcnt */
```

Both references drop, the count reaches zero, and the destructor releases
everything including the descriptors, so the explicit closes added for #4456 are
no longer needed and this patch removes them.

The alternative is to keep closing resources inline as each is noticed. That is
how the path arrived here: the next resource added to a broker would have to be
remembered in two places, and the second place is the one nobody looks at.

If the assignment to `rkb_thread` is unwelcome, the same result comes from
relaxing the assertion for a broker whose thread was never started. That is a
maintainer's call about which invariant to state, and either way the teardown
belongs in one function.

## Measured

Same reproduction as #4456, which counts `/proc/self/fd` before and after.

| build | rounds | descriptors before | after | leaked |
|---|---|---|---|---|
| unpatched | 5 | 3 | 23 | **20** |
| with this patch | 8 | 3 | 3 | 0 |
| with this patch, `MALLOC_CHECK_=3` | 12 | 3 | 3 | 0, no abort |

Twelve rounds under `MALLOC_CHECK_=3` report no double free, and the normal path
with thread creation succeeding is unchanged.

## The diff

```diff
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -5438,8 +5438,7 @@
          * the broker thread until we've finalized the rkb. */
         rd_kafka_broker_lock(rkb);
         rd_kafka_broker_keep(rkb); /* broker thread's refcnt */
-        if (getenv("RDK_FORCE_THRD_FAIL") ||
-            thrd_create(&rkb->rkb_thread, rd_kafka_broker_thread_main, rkb) !=
+        if (thrd_create(&rkb->rkb_thread, rd_kafka_broker_thread_main, rkb) !=
             thrd_success) {
                 rd_kafka_broker_unlock(rkb);
 
@@ -5450,19 +5449,29 @@
                 rd_kafka_op_err(rk, RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE,
                                 "Unable to create broker thread");
 
-                /* The wake-up pipe was opened above and is closed by
-                 * rd_kafka_broker_destroy_final(), which this path does not
-                 * reach: rd_free() releases the struct holding the descriptor
-                 * numbers without closing them, so nothing can close them
-                 * afterwards. Close them here before the struct goes. */
-#ifndef _WIN32
-                if (rkb->rkb_wakeup_fd[0] != -1)
-                        rd_socket_close(rkb->rkb_wakeup_fd[0]);
-                if (rkb->rkb_wakeup_fd[1] != -1)
-                        rd_socket_close(rkb->rkb_wakeup_fd[1]);
-#endif
+                /* Unwind through the destructor rather than by hand.
+                 *
+                 * rd_free() releases the struct without running
+                 * rd_kafka_broker_destroy_final(), so the wake-up pipe opened
+                 * above is never closed and the two descriptors stay open for
+                 * the life of the process with nothing left holding their
+                 * numbers. It also bypasses the refcount, leaving the two
+                 * references taken here unaccounted for.
+                 *
+                 * At this point the broker is not yet on rk_brokers and nothing
+                 * else can reach it, so dropping both references takes the count
+                 * to zero and runs the destructor, which closes the pipe, tears
+                 * down the locks and queues, and frees the struct. */
+                /* The destructor asserts it runs on the broker's own thread,
+                 * which is how it is reached in every other case. That thread
+                 * was never created, and no other thread can reach this broker
+                 * because it is not yet on rk_brokers, so this thread is the one
+                 * that owns it. Recording that makes the assertion true rather
+                 * than bypassed. */
+                rkb->rkb_thread = thrd_current();
 
-                rd_free(rkb);
+                rd_kafka_broker_destroy(rkb); /* broker thread's refcnt */
+                rd_kafka_broker_destroy(rkb); /* rk_broker's refcnt */
 
 #ifndef _WIN32
                 /* Restore sigmask of caller */
```

## Tooling

The descriptor half was confirmed at the syscall boundary with
[ExecVis](https://github.com/BNCMK/ExecVis), which records what every process on
a machine did without the program being instrumented, recompiled or aware. The
heap half was found by reading what the destructor releases and comparing it with
what the failure path does, which is the kind of comparison a leak checker only
reaches once the path has actually run, and this one runs only when thread
creation fails.

### DIFF
--- a/src/rdkafka_broker.c
+++ b/src/rdkafka_broker.c
@@ -5449,7 +5449,29 @@ rd_kafka_broker_t *rd_kafka_broker_add(rd_kafka_t *rk,
                 rd_kafka_op_err(rk, RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE,
                                 "Unable to create broker thread");
 
-                rd_free(rkb);
+                /* Unwind through the destructor rather than by hand.
+                 *
+                 * rd_free() releases the struct without running
+                 * rd_kafka_broker_destroy_final(), so the wake-up pipe opened
+                 * above is never closed and the two descriptors stay open for
+                 * the life of the process with nothing left holding their
+                 * numbers. It also bypasses the refcount, leaving the two
+                 * references taken here unaccounted for.
+                 *
+                 * At this point the broker is not yet on rk_brokers and nothing
+                 * else can reach it, so dropping both references takes the count
+                 * to zero and runs the destructor, which closes the pipe, tears
+                 * down the locks and queues, and frees the struct. */
+                /* The destructor asserts it runs on the broker's own thread,
+                 * which is how it is reached in every other case. That thread
+                 * was never created, and no other thread can reach this broker
+                 * because it is not yet on rk_brokers, so this thread is the one
+                 * that owns it. Recording that makes the assertion true rather
+                 * than bypassed. */
+                rkb->rkb_thread = thrd_current();
+
+                rd_kafka_broker_destroy(rkb); /* broker thread's refcnt */
+                rd_kafka_broker_destroy(rkb); /* rk_broker's refcnt */
 
 #ifndef _WIN32
                 /* Restore sigmask of caller */


xhon-pelushi (commented):
Built this and exercised the failure path under ASan rather than reasoning about it. **The diagnosis is right, the fix works, and it also closes #4456 — which is still open and, contrary to the PR description, is not already fixed in `master`.**

## Reproduced both leaks, and the fix removes both

Built with `-fsanitize=address`, interposed `thrd_create` via `LD_PRELOAD` to fail the Nth call, and drove it with a two-broker producer. Failing call #2, #3 or #4 all land on `Unable to create broker thread`:

| | heap | file descriptors |
| --- | --- | --- |
| `master` | **111 789 bytes in 12 allocations** | **2** (the wake-up pipe) |
| this PR | **0** | **0** |

The leak stacks on `master` point exactly where you said:

```
Direct leak of 17520 byte(s) in 1 object(s) allocated from:
    #2 rd_hdr_histogram_new  src/rdhdrhistogram.c:134
    #3 rd_avg_init           src/rdavg.h:244
    #4 rd_kafka_broker_add   src/rdkafka_broker.c:5318
    #5 rd_kafka_new          src/rdkafka.c:2981
```

On the PR head the same run is clean — no leaks, no ASan errors, no assertion failures, `clean shutdown` — across failure indices 2/3/4 and repeats. Happy to share the ~30-line `LD_PRELOAD` shim if it's useful for a regression test.

## This supersedes #4456, it doesn't sit on top of it

The description says *"Applies on top of the #4456 fix"*, but there is no #4456 fix in `master` — the failure branch is still:

```c
                rd_kafka_op_err(rk, RD_KAFKA_RESP_ERR__CRIT_SYS_RESOURCE,
                                "Unable to create broker thread");

                rd_free(rkb);
```

and [#4456](https://github.com/confluentinc/librdkafka/issues/4456) is still open, filed 2023-09-29. Because `rd_kafka_broker_destroy_final()` does the `rd_socket_close(rkb->rkb_wakeup_fd[0/1])` itself, routing through the destructor fixes the descriptors as a side effect — which is what the 2 → 0 fd measurement above shows. So this PR stands alone and can close #4456. Worth fixing the description, since as written it suggests a dependency on an unmerged change.

## Audited the destructor's preconditions

`rd_kafka_broker_destroy_final()` opens with about fourteen `rd_assert`s and then unconditionally frees a lot. The PR asserts this is safe; here's the check:

- **Refcount is exactly 2.** `rd_refcnt_init(&rkb->rkb_refcnt, 0)`, then `rd_kafka_broker_keep()` for `rk_broker`'s count and again for the broker thread's, both before `thrd_create`. So the two `rd_kafka_broker_destroy()` calls take it to zero and run the destructor once. ✓
- **The broker really is unreachable.** `TAILQ_INSERT_HEAD(&rkb->rkb_rk->rk_brokers, ...)` is at line 5472, *after* the failure branch at 5452. Nothing else can hold a pointer. ✓
- **Every `rd_avg` is live.** 18 `rd_avg_init` calls before `thrd_create`, exactly 18 `rd_avg_destroy` in the destructor, and the conditional `rk_type == RD_KAFKA_CONSUMER && !SHARE_CONSUMER` / `PRODUCER` blocks are gated identically in both places. ✓
- **The unconditional frees are all initialized:** `rkb_origname` (`rd_strdup`), `rkb_ops` (`rd_kafka_q_new`), and `rkb_share_fetch_session.toppars_in_session` (the `rd_list_destroy` target) are all set before `thrd_create`. ✓
- **The guarded ones are safely NULL** — `rkb` comes from `rd_calloc`, so `rkb_recv_buf`, `rkb_rsal` and `rkb_ApiVersions` are zero and their `if` guards hold. ✓
- **No self-deadlock.** `rd_kafka_broker_add` documents *"Locks: rd_kafka_wrlock(rk) must be held"*, and neither `destroy_final` nor `rd_kafka_q_purge` / `rd_kafka_q_destroy_owner` / `rd_avg_destroy` takes an `rk` lock. ✓

Empirically none of the assertions fired in any run, which matches.

## One latent asymmetry worth a comment or a guard

`destroy_final()` calls `rd_kafka_sasl_broker_term(rkb)` when the protocol is `SASL_PLAINTEXT`/`SASL_SSL`. But `rd_kafka_sasl_broker_init(rkb)` is at line 5466 — on the **success** path, after the failure branch. So this new path can reach `broker_term` without a matching `broker_init`.

Today that is harmless, and I checked rather than assumed: `rd_kafka_sasl_broker_term()` is

```c
if (provider->broker_term)
        provider->broker_term(rkb);
```

and none of the five providers — Cyrus, SCRAM, Win32 SSPI, OAUTHBEARER, PLAIN — sets `.broker_init` or `.broker_term`, so both hooks are NULL everywhere. But the fix now quietly depends on that staying true. If a provider ever implements `broker_term`, this error path would call it on state that was never initialized. A one-line note next to the `rkb_thread` comment, or an explicit skip, would make the dependency visible.

Relatedly, `rkb->rkb_thread = thrd_current()` makes the assertion true rather than the invariant true, which the comment says plainly and I think is the right trade here. Nothing else in `destroy_final` is thread-affine today, so it holds — the comment is what stops that quietly changing.

## Two smaller things

**No test.** The path needs fault injection to reach, so it will silently regress. Given #4456 has been open nearly three years and the bare `rd_free(rkb)` survived that whole time, a test seems worth the effort here more than usual.

**CLA.** The bot still reports `BNCMK` as unsigned, which blocks merge regardless of review.

## What I did not test

x86-64 Linux, glibc C11 threads (`WITH_C11THREADS=y`), GCC 13. I did not exercise the `_WIN32` branch of the failure path. Most relevant to the point above: I configured with `--disable-sasl`, so the `rd_kafka_sasl_broker_term` call was never reached in any of my runs — that conclusion rests on reading the five providers, not on running them.

PR #5565, opened 2026-08-10
[DOCS-39462]-Add a Schema Registry producer example

jgillenwater:
## Summary

- Add `examples/producer_schema_registry.c`, a producer that registers an Avro schema in Schema Registry and produces a schema-framed message using libserdes and avro-c.
- Add an opt-in `producer_schema_registry` Makefile target. It is deliberately **not** in `$(EXAMPLES)`, so plain `make` still works for contributors without libserdes and avro-c installed — same precedent as `producer_cpp` and `rdkafka_consume_batch`.
- Point at the new example from `examples/README.md` and from the "Usage in code" section of the top-level `README.md`, noting that librdkafka has no built-in Schema Registry serializer.

Addresses DOCS-39462: the sibling clients all now lead with a Schema Registry producer. librdkafka's README carries no inline code example, so instead of swapping a first example this adds the missing Schema Registry example and links to it. This is one of five coordinated PRs across librdkafka, confluent-kafka-python, confluent-kafka-go, confluent-kafka-dotnet, and confluent-kafka-javascript.

## Verification

- Verified by an SME: compiled against librdkafka + avro-c + libserdes built from source in a container, and produced a real message to a local broker.
- `RD_KAFKA_PRODUCER` (not `RD_KAFKA_TYPE_PRODUCER`) confirmed against `src/rdkafka.h`.
- Not compiled on this machine — libserdes and avro-c are not installed locally.


rayokota (changes requested):
`libserdes` is deprecated.  We should have an example using `libschemaregistry` instead.  See https://github.com/confluentinc/libschemaregistry

noramullen1 (approved):

rayokota (approved):
LGTM

PR #5574, opened 2026-08-21
[FIX] Avoid duplicate FETCH_STOP for the same toppar during assignment removal

### LINKED ISSUE #5573: Duplicate FETCH_STOP can be enqueued for the same toppar during assignment removal
Description
===========

We found a race in the ordinary consumer assignment teardown path where the same `rd_kafka_toppar_t` can have more than one outstanding assignment FETCH_STOP reply.

In current master (`0623a303485f5c6c9278002b44f08e18979a26d1`) and latest release `v2.15.0`, `rd_kafka_assignment_serve_removals()` checks only `rktp->rktp_started` before calling `rd_kafka_toppar_op_fetch_stop()` and incrementing `assignment.wait_stop_cnt`:

https://github.com/confluentinc/librdkafka/blob/0623a303485f5c6c9278002b44f08e18979a26d1/src/rdkafka_assignment.c#L404-L411

`rktp_started` is cleared later, when `rd_kafka_assignment_partition_stopped()` consumes the FETCH_STOP reply:

https://github.com/confluentinc/librdkafka/blob/0623a303485f5c6c9278002b44f08e18979a26d1/src/rdkafka_assignment.c#L1177-L1183

There is also an asymmetry in `rd_kafka_assignment_serve()`: removals are served before the `wait_stop_cnt == 0` gate, while pending starts are gated on `wait_stop_cnt == 0`:

https://github.com/confluentinc/librdkafka/blob/0623a303485f5c6c9278002b44f08e18979a26d1/src/rdkafka_assignment.c#L714-L726

That section's own comment says pending entries wait for a previous assignment change, "an unassign in some form", to propagate. The pending-start side already waits for outstanding stops; the removal side can still enqueue another stop while the first stop reply is outstanding.

If another assignment/removal for the same toppar is processed after the first FETCH_STOP is enqueued but before that reply is consumed, the removal path can enqueue and count a second FETCH_STOP for the same started state. When the replies are later consumed, the first clears `rktp_started`; the second reaches `rd_assert(rktp->rktp_started)`.

`rd_kafka_assign()` returns after the cgrp ASSIGN op is handled without waiting for `assignment.wait_stop_cnt` to reach zero, so ordinary callers can issue another assign/remove while the first FETCH_STOP reply is outstanding. We observed the failure only under test teardown stress and have not measured production frequency.

A minimal fix is to track an assignment-owned pending stop state for the toppar: set it when enqueueing the first assignment FETCH_STOP, skip duplicate stop enqueue/count while it is pending, and clear it when the stop reply is consumed. The intended invariant is: one started toppar has at most one outstanding assignment FETCH_STOP.

A downstream bridge/proposed invariant test is here:

https://github.com/flarioncode/datafusion-accelerator/pull/4028

How to reproduce
================

A teardown-stress loop hit the assertion on iteration 7.

The captured reproducer used rust-rdkafka's `MockCluster::new(1)`, one topic (`range-confirmed-deny-terminal`) with one partition, four produced records, ordinary `rd_kafka_assign()`-based consumption through the Rust binding, and the trace tree normalized an empty user config to `group.id=comet-native`, `enable.auto.commit=false`, and `auto.offset.reset=error`. It also added Spark-parity defaults unrelated to the assignment state machine: `isolation.level=read_uncommitted`, `check.crcs=true`, `allow.auto.create.topics=true`, and `socket.receive.buffer.bytes=65536`.

Selected lines from a locally instrumented origin trace for the same `rktp=0x11f015000` are below; ellipses mark omitted neighboring trace events. The `offset-store-term` lines are stop completion/reply enqueue events, not the source of the second removal:

```text
FLARION_RDKAFKA_STOP_TRACE event=fetch-stop-request origin=assignment-serve-removals rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=1 fetch_state=active wait_stop_cnt=0 started_cnt=1 assignment_version=2
FLARION_RDKAFKA_STOP_TRACE event=wait-stop-incremented origin=assignment-serve-removals rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=1 fetch_state=active wait_stop_cnt=1 started_cnt=1 assignment_version=2
...
FLARION_RDKAFKA_STOP_TRACE event=fetch-stop-enqueue origin=offset-store-term rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=1 fetch_state=stopped wait_stop_cnt=1 started_cnt=1 assignment_version=3
FLARION_RDKAFKA_STOP_TRACE event=fetch-stop-request origin=assignment-serve-removals rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=1 fetch_state=stopped wait_stop_cnt=1 started_cnt=1 assignment_version=4
FLARION_RDKAFKA_STOP_TRACE event=wait-stop-incremented origin=assignment-serve-removals rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=1 fetch_state=stopped wait_stop_cnt=2 started_cnt=1 assignment_version=4
FLARION_RDKAFKA_STOP_TRACE event=fetch-stop-consume origin=consumer-op-serve rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=1 fetch_state=stopped wait_stop_cnt=2 started_cnt=1 assignment_version=4
FLARION_RDKAFKA_STOP_TRACE event=started-set-false origin=assignment-partition-stopped rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=1 fetch_state=stopped wait_stop_cnt=1 started_cnt=1 assignment_version=4
...
FLARION_RDKAFKA_STOP_TRACE event=fetch-stop-enqueue origin=offset-store-term rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=0 fetch_state=stopped wait_stop_cnt=1 started_cnt=0 assignment_version=4
FLARION_RDKAFKA_STOP_TRACE event=fetch-stop-consume origin=consumer-op-serve rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=0 fetch_state=stopped wait_stop_cnt=1 started_cnt=0 assignment_version=4
FLARION_RDKAFKA_STOP_TRACE event=assert-precheck origin=consume-fetch-stop rktp=0x11f015000 topic=range-confirmed-deny-terminal partition=0 started=0 fetch_state=stopped wait_stop_cnt=0 started_cnt=0 assignment_version=4
```

The final `assert-precheck` counter values are logged after `wait_stop_cnt--` in `rd_kafka_assignment_partition_stopped()` and before `rd_assert(rktp->rktp_started)`, so the failing assertion is `rd_assert(rktp->rktp_started)`.

We also have a deterministic state-machine-level test that pins the invariant "one started toppar has at most one outstanding assignment FETCH_STOP": removing the pending-stop predicate makes that test fail. That test is evidence for the invariant and the fix shape; it is not a timing-independent crash reproducer. We have not reduced the crash itself to a timing-independent reproducer.

Checklist
=========

 - [x] librdkafka version (release number or git tag): `v2.15.0` and master `0623a303485f5c6c9278002b44f08e18979a26d1`
 - [x] Apache Kafka version: No external Apache Kafka broker in the captured reproducer; it used rust-rdkafka/librdkafka `MockCluster::new(1)`.
 - [x] librdkafka client configuration: In the captured trace tree, the empty user config was normalized to `group.id=comet-native`, `enable.auto.commit=false`, `auto.offset.reset=error`, plus Spark-parity defaults `isolation.level=read_uncommitted`, `check.crcs=true`, `allow.auto.create.topics=true`, `socket.receive.buffer.bytes=65536`.
 - [x] Operating system: captured origin-trace reproducer on macOS 15.6.1 arm64 (`Darwin 24.6.0`).
 - [x] Provide logs (with `debug=..` as necessary) from librdkafka: included above as locally instrumented `FLARION_RDKAFKA_STOP_TRACE` output, not native `debug=..` output.
 - [x] Provide broker log excerpts: not applicable; the captured reproducer used librdkafka's mock cluster, and the failure is a client-side assertion.
 - [ ] Critical issue

Weijun-H:
Description
===========

Fixes #5573.

During assignment removal, `rktp_started` remains true until the FETCH_STOP reply is served. A later removal pass can therefore enqueue and count another FETCH_STOP for the same started toppar.

This change adds assignment-owned `rktp_wait_stop` state. A stop request now sets the flag, increments `assignment.wait_stop_cnt`, and enqueues FETCH_STOP in one place; later removal passes skip that toppar until `rd_kafka_assignment_partition_stopped()` clears the flag and the existing started/wait counters. Each actual request constructs a fresh `RD_KAFKA_REPLYQ(...)`, preserving one queue reference per op.

Correctness notes
=================

- Assignment removals send their stop replies to `rk->rk_ops`. The cgrp termination path uses `RD_KAFKA_NO_REPLYQ`, while the synchronous stop path waits on a local `tmpq`, so the new completion state applies only to assignment replies.
- `rktp_wait_stop` has the same owner and access points as `rktp_started`; no new lock domain is introduced.
- The native unit test runs the production removal and completion entry points. It checks version-barrier deltas of `+1` for the first request, `0` for a duplicate before completion, and `+1` for each of two partitions in one removal pass. It does not poll the asynchronous reply or reproduce the crash timing; per-request replyq ownership is preserved by code structure rather than a dynamic refcount assertion.

Tests
=====

- `make -j8`
- `cd tests && RD_UT_TEST=assignment TESTS=0000 ./run-test.sh -p1`
- `cd tests && TESTS=0000 ./run-test.sh -p1`
- `make style-check-changed` with clang-format 18

### DIFF
--- a/src/rdkafka_assignment.c
+++ b/src/rdkafka_assignment.c
@@ -387,11 +387,11 @@ static void rd_kafka_share_assignment_serve_removals(rd_kafka_t *rk) {
 static void rd_kafka_assignment_request_stop(rd_kafka_assignment_t *assignment,
                                              rd_kafka_toppar_t *rktp,
                                              rd_kafka_q_t *stop_replyq) {
-        if (!rktp->rktp_started || rktp->rktp_stop_pending)
+        if (!rktp->rktp_started || rktp->rktp_wait_stop)
                 return;
 
         rd_assert(assignment->started_cnt > 0);
-        rktp->rktp_stop_pending = rd_true;
+        rktp->rktp_wait_stop = rd_true;
         assignment->wait_stop_cnt++;
         rd_kafka_toppar_op_fetch_stop(rktp, RD_KAFKA_REPLYQ(stop_replyq, 0));
 }
@@ -405,8 +405,8 @@ static void rd_kafka_assignment_complete_stop(rd_kafka_assignment_t *assignment,
         rd_assert(assignment->wait_stop_cnt > 0);
         assignment->wait_stop_cnt--;
 
-        rd_assert(rktp->rktp_stop_pending);
-        rktp->rktp_stop_pending = rd_false;
+        rd_assert(rktp->rktp_wait_stop);
+        rktp->rktp_wait_stop = rd_false;
 
         rd_assert(rktp->rktp_started);
         rktp->rktp_started = rd_false;
@@ -486,10 +486,11 @@ static int rd_kafka_assignment_serve_removals0(rd_kafka_t *rk,
                 rd_kafka_dbg(rk, CGRP, "REMOVE",
                              "Removing %s [%" PRId32
                              "] from assignment "
-                             "(started=%s, pending=%s, queried=%s, "
-                             "stored offset=%s)",
+                             "(started=%s, wait_stop=%s, pending=%s, "
+                             "queried=%s, stored offset=%s)",
                              rktpar->topic, rktpar->partition,
                              RD_STR_ToF(rktp->rktp_started),
+                             RD_STR_ToF(rktp->rktp_wait_stop),
                              RD_STR_ToF(was_pending), RD_STR_ToF(was_queried),
                              rd_kafka_offset2str(rktpar->offset));
         }
@@ -585,7 +586,7 @@ static int rd_kafka_assignment_serve_pending(rd_kafka_t *rk) {
                     rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
 
                 rd_assert(!rktp->rktp_started);
-                rd_assert(!rktp->rktp_stop_pending);
+                rd_assert(!rktp->rktp_wait_stop);
 
                 if (!RD_KAFKA_OFFSET_IS_LOGICAL(rktpar->offset) ||
                     rktpar->offset == RD_KAFKA_OFFSET_BEGINNING ||
@@ -1254,7 +1255,9 @@ unittest_assignment_add_removed_partition(rd_kafka_t *rk,
                 rd_kafka_toppar_unlock(rktp);
         }
 
-        return rktp;
+        /* Return a test-owned reference that remains valid after
+         * serve_removals0() clears the removed assignment list. */
+        return rktp ? rd_kafka_toppar_keep(rktp) : NULL;
 }
 
 
@@ -1270,6 +1273,9 @@ static int unittest_assignment_stop_pending(void) {
         rd_kafka_conf_set(conf, "enable.auto.commit", "false", NULL, 0);
         rk = rd_kafka_new(RD_KAFKA_CONSUMER, conf, errstr, sizeof(errstr));
         RD_UT_ASSERT(rk, "failed to create consumer: %s", errstr);
+        /* Keep this fixture single-threaded and deterministic: no group.id
+         * means no consumer group, and FETCH_STOP replies go to a private
+         * test queue that is never polled. */
         test_ops = rd_kafka_q_new(rk);
 
         /* A removal for a partition that was never started must not request
@@ -1279,7 +1285,6 @@ static int unittest_assignment_stop_pending(void) {
         RD_UT_ASSERT(!RD_KAFKA_TOPPAR_IS_PAUSED(rktp),
                      "test requires an unpaused partition so only FETCH_STOP "
                      "bumps the version");
-        rk->rk_consumer.assignment.started_cnt = 1;
         version = rd_atomic32_get(&rktp->rktp_version);
         RD_UT_ASSERT(rd_kafka_assignment_serve_removals0(rk, test_ops) == 0,
                      "non-started removal should have no outstanding work");
@@ -1287,8 +1292,11 @@ static int unittest_assignment_stop_pending(void) {
                      "non-started removal must not request FETCH_STOP");
         RD_UT_ASSERT(rk->rk_consumer.assignment.wait_stop_cnt == 0,
                      "non-started removal must not increment wait_stop_cnt");
-        RD_UT_ASSERT(!rktp->rktp_stop_pending,
+        RD_UT_ASSERT(rk->rk_consumer.assignment.started_cnt == 0,
+                     "non-started removal must not alter started_cnt");
+        RD_UT_ASSERT(!rktp->rktp_wait_stop,
                      "non-started removal must not mark stop-pending");
+        rd_kafka_toppar_destroy(rktp);
 
         rktp = unittest_assignment_add_removed_partition(rk, "rdut", 0);
         RD_UT_ASSERT(rktp, "failed to re-add toppar to removed list");
@@ -1304,8 +1312,9 @@ static int unittest_assignment_stop_pending(void) {
                      "FETCH_STOP");
         RD_UT_ASSERT(rk->rk_consumer.assignment.wait_stop_cnt == 1,
                      "first stop should increment wait_stop_cnt once");
-        RD_UT_ASSERT(rktp->rktp_stop_pending,
+        RD_UT_ASSERT(rktp->rktp_wait_stop,
                      "first stop should mark the toppar stop-pending");
+        rd_kafka_toppar_destroy(rktp);
 
         rktp = unittest_assignment_add_removed_partition(rk, "rdut", 0);
         RD_UT_ASSERT(rktp, "failed to re-add toppar to removed list");
@@ -1316,7 +1325,7 @@ static int unittest_assignment_stop_pending(void) {
         rk->rk_consumer.assignment.wait_stop_cnt = 1;
         rk->rk_consumer.assignment.started_cnt   = 1;
         rktp->rktp_started                       = rd_true;
-        rktp->rktp_stop_pending                  = rd_true;
+        rktp->rktp_wait_stop                     = rd_true;
         version = rd_atomic32_get(&rktp->rktp_version);
         rd_kafka_assignment_serve_removals0(rk, test_ops);
         RD_UT_ASSERT(rd_atomic32_get(&rktp->rktp_version) == version,
@@ -1325,7 +1334,7 @@ static int unittest_assignment_stop_pending(void) {
         RD_UT_ASSERT(rk->rk_consumer.assignment.wait_stop_cnt == 1,
                      "duplicate removal must not increment wait_stop_cnt");
 
-        RD_UT_ASSERT(rktp->rktp_stop_pending,
+        RD_UT_ASSERT(rktp->rktp_wait_stop,
                      "completion fixture requires a pending stop");
         RD_UT_ASSERT(rk->rk_consumer.assignment.wait_stop_cnt == 1,
                      "completion fixture requires one awaited stop");
@@ -1336,11 +1345,12 @@ static int unittest_assignment_stop_pending(void) {
         rd_kafka_assignment_complete_stop(&rk->rk_consumer.assignment, rktp);
         RD_UT_ASSERT(rk->rk_consumer.assignment.wait_stop_cnt == 0,
                      "completion should clear wait_stop_cnt");
-        RD_UT_ASSERT(!rktp->rktp_stop_pending,
+        RD_UT_ASSERT(!rktp->rktp_wait_stop,
                      "completion should clear stop-pending");
         RD_UT_ASSERT(!rktp->rktp_started, "completion should clear started");
         RD_UT_ASSERT(rk->rk_consumer.assignment.started_cnt == 0,
                      "completion should decrement started_cnt");
+        rd_kafka_toppar_destroy(rktp);
 
         {
                 rd_kafka_toppar_t *rktp0, *rktp1;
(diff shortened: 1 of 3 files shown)


src/rdkafka_assignment.c
```diff
@@ -1196,6 +1237,192 @@ void rd_kafka_assignment_partition_stopped(rd_kafka_t *rk,
...
+                                          const char *topic,
+                                          int32_t partition) {
+        rd_kafka_topic_partition_t *rktpar;
+        rd_kafka_toppar_t *rktp;
+
+        rktpar = rd_kafka_topic_partition_list_add(
+            rk->rk_consumer.assignment.removed, topic, partition);
+        rktp = rd_kafka_topic_partition_ensure_toppar(rk, rktpar, rd_true);
+
+        if (rktp) {
```
prashah-confluent, line 1252: **Blocking: `rktp` is dangling after every `serve_removals0()` call — use-after-free.**

The toppar's only reference here is the one owned by the `removed`-list element
(`ensure_toppar()` caches it in the element's `_private`, which holds the ref).
`rd_kafka_assignment_serve_removals0()` ends with
`rd_kafka_topic_partition_list_clear(assignment.removed)`, and because this test runs with
no broker the partition is UNKNOWN — `rd_kafka_toppar_desired_del()` unlinks it, so clearing
the list drops the last reference and destroys the toppar. A `debug=topic` trace shows it
directly (note `DESTROY_FINAL` fires inside the serve call, before the test's asserts):

```
DESP:  Removing (un)desired topic rdut [0]
TOPPARDEL: rdut [0] not handled by any broker: not sending LEAVE for remove
TOPPARREMOVE: Removing toppar rdut [0] 0x...
DESTROY: rdut [0]: 0x... DESTROY_FINAL      ← during serve_removals0()
RDUT: FAIL: ... non-started removal must not request FETCH_STOP   ← reads freed memory
```

Every assertion after a `serve_removals0()` call, in all four scenarios, reads freed memory.
On macOS arm64 this fails 3/3 runs; where it passes, it passes by allocator luck (ASAN
catches it everywhere).

Fix that I verified locally (fails 3/3 before, passes 3/3 after): have the helper take a
test-owned reference, and release it once per acquisition (5 keeps / 5 destroys):

```c
        if (rktp) {
                rd_kafka_toppar_lock(rktp);
                rktp->rktp_flags |= RD_KAFKA_TOPPAR_F_ASSIGNED;
                rd_kafka_toppar_unlock(rktp);

                /* serve_removals0() clears the removed list, dropping the
                 * list element's toppar reference; with no broker the
                 * partition is unknown and gets destroyed with it. Hold a
                 * test-owned reference so the caller's asserts don't read
                 * freed memory. Released with rd_kafka_toppar_destroy(). */
                rd_kafka_toppar_keep(rktp);
        }
```

…plus `rd_kafka_toppar_destroy(rktp);` after the last assert of scenarios 1, 2, and 3
(before each re-acquisition), and `rd_kafka_toppar_destroy(rktp0); rd_kafka_toppar_destroy(rktp1);`
at the end of the two-partition block.

src/rdkafka_assignment.c
```diff
@@ -1196,6 +1237,192 @@ void rd_kafka_assignment_partition_stopped(rd_kafka_t *rk,
...
+                             "round should request one FETCH_STOP");
+                RD_UT_ASSERT(rk->rk_consumer.assignment.wait_stop_cnt == 2,
+                             "same removal round should await both stop "
+                             "replies");
+                RD_UT_ASSERT(rktp0->rktp_stop_pending &&
+                                 rktp1->rktp_stop_pending,
+                             "same removal round should mark both toppars "
+                             "stop-pending");
+
+                rd_kafka_assignment_partition_stopped(rk, rktp0);
```
prashah-confluent, line 1397: **Fragile ordering constraint worth making explicit (or removing).**

This consumer has no `group.id`, so `rk->rk_cgrp == NULL` (cgrp is only created when a group
id is configured, rdkafka.c ~2828). If `rd_kafka_assignment_partition_stopped()` ever drives
`wait_stop_cnt` to 0 here, it calls `rd_kafka_assignment_serve()`, which — with all counters
at zero — calls `rd_kafka_cgrp_assignment_done(rk->rk_cgrp)`, and that dereferences `rkcg`
unconditionally (rdkafka_cgrp.c:4995) → segfault.

The test only avoids this because `partition_stopped()` is called exactly once, while the
second stop is still pending (2→1), and the final completion goes through
`rd_kafka_assignment_complete_stop()` directly. That constraint is invisible to the next
person editing this test — could you add a comment stating it (e.g. "must not be the last
outstanding stop: with no cgrp, partition_stopped() → serve() → cgrp_assignment_done(NULL)
would crash"), or restructure so the last completion can't reach `serve()`?

src/rdkafka_assignment.c
```diff
@@ -1196,6 +1237,192 @@ void rd_kafka_assignment_partition_stopped(rd_kafka_t *rk,
...
+                rd_kafka_toppar_lock(rktp);
+                rktp->rktp_flags |= RD_KAFKA_TOPPAR_F_ASSIGNED;
+                rd_kafka_toppar_unlock(rktp);
+        }
+
+        return rktp;
+}
+
+
+static int unittest_assignment_stop_pending(void) {
```
prashah-confluent, line 1264: **Nit (robustness): please document the two facts that make this test thread-safe.**

The test mutates state owned by the main thread (`assignment.wait_stop_cnt`, `started_cnt`,
`rktp_started`, `rktp_stop_pending`) from the unit-test thread while the instance's threads
are running. That's safe today only because:

1. no `group.id` → no cgrp → nothing else drives the assignment state machine, and
2. stop replies are routed to the private `test_ops` queue that nothing polls, so
   `rd_kafka_assignment_partition_stopped()` can never be invoked concurrently by the
   main thread.

Both are load-bearing and neither is stated — a short comment here would prevent someone
from "simplifying" the test into a race later.

src/rdkafka_assignment.c
```diff
@@ -1196,6 +1237,192 @@ void rd_kafka_assignment_partition_stopped(rd_kafka_t *rk,
...
+        test_ops = rd_kafka_q_new(rk);
+
+        /* A removal for a partition that was never started must not request
+         * FETCH_STOP or alter the assignment stop counters. */
+        rktp = unittest_assignment_add_removed_partition(rk, "rdut", 0);
+        RD_UT_ASSERT(rktp, "failed to create toppar");
+        RD_UT_ASSERT(!RD_KAFKA_TOPPAR_IS_PAUSED(rktp),
+                     "test requires an unpaused partition so only FETCH_STOP "
+                     "bumps the version");
+        rk->rk_consumer.assignment.started_cnt = 1;
```
prashah-confluent: **Nit:** this `started_cnt = 1` is never read on this path — the partition isn't started, so
`rd_kafka_assignment_request_stop()` returns before its `started_cnt` assert. Dropping it
would make the fixture cleaner (or add a comment if it's intentional).

src/rdkafka_assignment.c
prashah-confluent, line 486: **Nit (debuggability):** now that `started` and `stop_pending` can legitimately differ — and
that difference is the whole point of this fix — it would be really useful to include
`stop_pending=%s` in this REMOVE debug line (and possibly in
`rd_kafka_assignment_dump()`), so the skipped-duplicate case is visible in `debug=cgrp`
output when diagnosing assignment issues in the field.

src/rdkafka_partition.h
```diff
@@ -398,11 +398,16 @@ struct rd_kafka_toppar_s {                           /* rd_kafka_toppar_t */
...
-                                 *   start.
-                                 *   This is used by cgrp to keep
-                                 *   track of whether the toppar has
-                                 *   been started or not. */
+        rd_bool_t rktp_started;      /**< Fetcher is instructured to
+                                      *   start.
+                                      *   This is used by cgrp to keep
+                                      *   track of whether the toppar has
+                                      *   been started or not. */
+        rd_bool_t rktp_stop_pending; /**< Assignment-owned FETCH_STOP reply is
```
prashah-confluent: **Naming nit:** suggest `rktp_wait_stop` instead of `rktp_stop_pending`, for two reasons:

1. The subsystem's existing idiom for "awaiting async completion" is `wait_*`
   (`wait_stop_cnt`, `wait_commit_cnt`), and this flag is exactly the per-partition
   shadow of `wait_stop_cnt` — with the rename, the invariant reads itself:
   `wait_stop_cnt` == number of toppars with `rktp_wait_stop` set. (There's currently
   no other `*_pending` boolean field in the codebase.)
2. "pending" already means the opposite direction in this same subsystem:
   `assignment.pending` is the list of partitions waiting to be *started*, whereas this
   flag means a *stop* is in flight.

Non-blocking, but since the field name is the API here, cheaper to settle now than after
merge. The helper names themselves read well (`request_stop`/`complete_stop` mirrors the
`rd_kafka_txn_complete()`-style action naming, and `serve_removals0` follows the standard
`foo0()` convention), so this is the only naming change I'd suggest.

prashah-confluent (changes requested):
Hi @Weijun-H , Gave some comments.

src/rdkafka_assignment.c
```diff
@@ -1196,6 +1238,207 @@ void rd_kafka_assignment_partition_stopped(rd_kafka_t *rk,
...
+        RD_UT_ASSERT(rd_kafka_assignment_serve_removals0(rk, test_ops) == 0,
+                     "non-started removal should have no outstanding work");
+        RD_UT_ASSERT(rd_atomic32_get(&rktp->rktp_version) == version,
+                     "non-started removal must not request FETCH_STOP");
+        RD_UT_ASSERT(rk->rk_consumer.assignment.wait_stop_cnt == 0,
+                     "non-started removal must not increment wait_stop_cnt");
+        RD_UT_ASSERT(rk->rk_consumer.assignment.started_cnt == 0,
+                     "non-started removal must not alter started_cnt");
+        RD_UT_ASSERT(!rktp->rktp_wait_stop,
+                     "non-started removal must not mark stop-pending");
```
prashah-confluent: nit: `wait_stop` instead of `stop-pending`

prashah-confluent (commented):
Have couple of `nit` comments. Otherwise LGTM!!

prashah-confluent (approved):
LGTM!! Approved.

We will merge the PR after one final review.

Thanks @Weijun-H !!

Ankith-Confluent (approved):
Thanks for the PR @Weijun-H 
LGTM!

PR #5580, opened 2026-08-25
fix(kafka): bound telemetry and Produce response arrays

shoemoney:
Malformed GetTelemetrySubscriptions array counts can reach allocation as negative or unbounded sizes. The Produce v8+ RecordErrors count also allowed any positive value, even when it exceeded the number of records sent.

This bounds accepted compression types at 1000, rejects nullable counts for both telemetry arrays, and bounds Produce record errors by the request batch size. Telemetry allocation ownership is established before reading array elements so a truncated response is cleaned up correctly.

Validation on macOS arm64:

- Produce v8 and v9 parser tests cover counts -1, 0, the request batch size, and batch size + 1. The over-limit case returned success before the upper bound and now returns BAD_MSG.
- All 25 built-in unit suites pass, including the telemetry count regressions.
- Build and clang-format 18 checks pass after rebasing onto current master.

Configured with SSL, SASL and curl disabled. No live Kafka cluster tests were run.


src/rdkafka_request.c
k-raina, line 6900: Do we need to check for negative error count here? 
Similiar to https://github.com/confluentinc/librdkafka/pull/5580/changes#diff-81d5315dc871bcfe090e0b28790b8e6ff1b929209cd1d9b3bf03d3d96d7e89faR6831-R6832

src/rdkafka_request.c
```diff
@@ -6827,25 +6827,40 @@ void rd_kafka_handle_GetTelemetrySubscriptions(rd_kafka_t *rk,
...
                 for (i = 0; i < (size_t)arraycnt; i++) {
                         int8_t AcceptedCompressionType;
                         rd_kafka_buf_read_i8(rkbuf, &AcceptedCompressionType);
-                        rk->rk_telemetry.accepted_compression_types[i] =
-                            AcceptedCompressionType;
+                        accepted_compression_types[i] = AcceptedCompressionType;
                 }
+
+                rk->rk_telemetry.accepted_compression_types_cnt = arraycnt;
+                rk->rk_telemetry.accepted_compression_types =
```
k-raina: Should this memory be freed in case of https://github.com/confluentinc/librdkafka/pull/5580/changes#diff-81d5315dc871bcfe090e0b28790b8e6ff1b929209cd1d9b3bf03d3d96d7e89faR6845 fails?

src/rdkafka_request.c
```diff
@@ -6827,25 +6827,40 @@ void rd_kafka_handle_GetTelemetrySubscriptions(rd_kafka_t *rk,
...
+        if (arraycnt < 0)
+                rd_kafka_buf_parse_fail(
+                    rkbuf, "ApiArrayCnt %" PRId32 " out of range", arraycnt);
 
         if (arraycnt) {
-                rk->rk_telemetry.accepted_compression_types_cnt = arraycnt;
-                rk->rk_telemetry.accepted_compression_types =
+                rd_kafka_compression_t *accepted_compression_types =
                     rd_calloc(arraycnt, sizeof(rd_kafka_compression_t));
+                if (!accepted_compression_types)
```
k-raina, line 6842: Will rd_calloc abort when failure to allocate accepted_compression_types?

src/rdkafka_request.c
```diff
@@ -6827,25 +6827,40 @@ void rd_kafka_handle_GetTelemetrySubscriptions(rd_kafka_t *rk,
                      "Parsing: subscription id %" PRId32,
                      rk->rk_telemetry.subscription_id);
 
-        rd_kafka_buf_read_arraycnt(rkbuf, &arraycnt, -1);
+        rd_kafka_buf_read_arraycnt(rkbuf, &arraycnt, 1000);
+        if (arraycnt < 0)
```
k-raina, line 6835: Should we make similiar change here https://github.com/shoemoney/librdkafka/blob/5214a830823795e9b4ae357f554f01024fa990a3/src/rdkafka_request.c#L3988-L3989

k-raina (commented):
Thanks for PR @shoemoney 
I have left few comments

src/rdkafka_request.c
```diff
@@ -3986,6 +3986,10 @@ rd_kafka_handle_Produce_parse(rd_kafka_broker_t *rkb,
                 int32_t RecordErrorsCnt;
                 rd_kafkap_str_t ErrorMessage;
                 rd_kafka_buf_read_arraycnt(rkbuf, &RecordErrorsCnt, -1);
```
k-raina: Should we update &RecordErrorsCnt also to upper limit?

k-raina (commented):
Thanks @shoemoney for addressing comments. Overall PR looks good, 
Provided one comment.

src/rdkafka_request.c
shoemoney: Yes. `4a0bd30c` caps `RecordErrorsCnt` at the number of records in the request batch, so the bound follows the actual request instead of imposing a fixed limit on valid batches. The negative-count guard remains.

Added Produce v8/v9 parser tests for counts -1, 0, the batch size and batch size + 1. The over-limit case failed before the fix and passes now. Rebased onto current master; the build, all 25 built-in unit suites and clang-format 18 checks pass locally (SSL/SASL/curl disabled; no live Kafka cluster tests).

shoemoney:
The CLA is signed and fully executed on Confluent's side as of 2026-09-19 02:02 UTC — `cla-status-check` is now green, alongside `ci/semaphoreci/pr: build-test-release` and `r2-agents/approval-gate`. Apologies for the delay on that; it was the only thing standing between your two `/sem-approve` runs and a merge. Nothing in the diff has changed since your review. Ready when you are.

src/rdkafka_request.c
k-raina: Are similar changes needed for sibling `rk->rk_telemetry.requested_metrics`?

k-raina: Should we have change for  NULL-check the alloc and Temp-pointer ownership-timing fix to requested_metrics & record_errors too, if applicable?

k-raina (commented):
@shoemoney Thanks for changes. 

Left one comment on checks for requested_metrics & record_errors.

shoemoney:
Good catch. Applied to both, though they needed slightly different treatment.

`requested_metrics` had the same bug `accepted_compression_types` had: the count was set before the allocation, so a failed alloc would leave a non-zero `requested_metrics_cnt` beside a NULL pointer. Switched it to the temp-pointer, NULL-check, assign-count-last pattern.

`record_errors` already assigned the count after the alloc, so that particular ordering failure was not present, but it had no NULL check at all. Added the same temp-pointer pattern for consistency. Worth noting that `rd_calloc` asserts internally on OOM rather than returning NULL, so these checks are defensive in the same way the existing `accepted_compression_types` check is.

Ran the full in-process unit suite after the change, with the `request` and `telemetry` suites specifically exercising both parsers. All green, and clang-format 18 is clean.

CHANGELOG.md
```diff
@@ -54,6 +54,15 @@ builds already got from 8.20.0).
 
 ### Producer fixes
 
+* Reject malformed Produce response record-error counts before allocating
+  memory (#5580, @shoemoney).
+
+* Guard the Produce response `record_errors` allocation, and the
+  GetTelemetrySubscriptions `requested_metrics` allocation, against
```
k-raina: nit: we should mention accepted_compression_types also.

shoemoney: Added `accepted_compression_types` to the changelog entry in `6921423f`. Squashed and rebased onto `c58a8dc4` as CONTRIBUTING requests.

Both clean-base and updated builds and all 25 built-in unit suites passed locally; clang-format 18 passed. SSL/SASL/curl were disabled; broker integration suites were not run. The new head has no code-test CI result yet.

k-raina (approved):
Thanks @shoemoney for PR and congratulations on your first contribution to this repo. 
We will wait for second reviewer to approve, before merging PR.

PR #5585, opened 2026-09-04
fix: Don't commit offsets of a lost assignment

k-raina:
A consumer with enable.auto.commit=true could send an OffsetCommit for an assignment it had already lost. On a client-side session timeout, the member id is reset to empty, the assignment is marked lost, and its partitions are revoked. 

The revoke-time auto-commit of those partitions was still sent because the assignment-lost flag was cleared inside rd_kafka_cgrp_unassign() and rd_kafka_cgrp_incremental_unassign(), before the removed partitions were served, defeating the guard that skips commits for a lost assignment.

The commit left with an empty member id and the previous generation. A broker rejects it with UNKNOWN_MEMBER_ID, or, for a static member (group.instance.id set), with FENCED_INSTANCE_ID, which librdkafka raises as a fatal error that stops the consumer.

Keep the assignment-lost flag set until the following rejoin applies a new assignment (assign()/incremental_assign() still clear it), so the offsets of a lost assignment are never committed. This restores the behavior documented in rd_kafka_cgrp_incremental_unassign()'s own comment.

Add test 0192 (static member, cooperative-sticky, mock) which asserts that no OffsetCommit is sent after a session timeout.

Happening since 1.6.0.

**Alternative Approach**
- Add a check at offset commit i.e "subscribed, but member id is blank → refuse locally with _ASSIGNMENT_LOST."
-  Only impact of alternative: A commit made after being kicked returns `_ASSIGNMENT_LOST` instead of broker `UNKNOWN_MEMBER_ID` (dynamic) or a fatal `FENCED_INSTANCE_ID` (static). Apps matching on UNKNOWN_MEMBER_ID should handle both. This needs to be mentioned in ChangeLog

### DIFF
--- a/src/rdkafka_cgrp.c
+++ b/src/rdkafka_cgrp.c
@@ -4883,8 +4883,12 @@ static rd_kafka_error_t *rd_kafka_cgrp_incremental_unassign(
                     RD_KAFKA_CGRP_JOIN_STATE_WAIT_INCR_UNASSIGN_TO_COMPLETE);
         }
 
-        rd_kafka_cgrp_assignment_clear_lost(rkcg,
-                                            "incremental_unassign() called");
+        /* Note: the assignment-lost flag is intentionally NOT cleared here.
+         * A lost assignment must stay marked lost until the following rejoin
+         * applies a new assignment (assign()/incremental_assign()), so that the
+         * revoke-time auto-commit of the removed partitions is correctly
+         * skipped. Clearing it here previously let an OffsetCommit go out with
+         * an empty member id, fatally fencing a static member. */
 
         return NULL;
 }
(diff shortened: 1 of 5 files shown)


src/rdkafka_cgrp.c
```diff
@@ -4883,8 +4883,12 @@ static rd_kafka_error_t *rd_kafka_cgrp_incremental_unassign(
                     RD_KAFKA_CGRP_JOIN_STATE_WAIT_INCR_UNASSIGN_TO_COMPLETE);
         }
 
-        rd_kafka_cgrp_assignment_clear_lost(rkcg,
-                                            "incremental_unassign() called");
```
k-raina: rd_kafka_cgrp_assignment_clear_lost 
- Need to clear before assignement CB rebalabce & 
or wait assignment call 
- Or before changing to steady state. 

Overall check when to update.

emasab (approved):
LGTM, Thanks @k-raina !

emasab (approved):

PR #5600, opened 2026-09-17
[FIX] Decommission learned brokers when a KIP-1102 re-bootstrap sequence starts

emasab:
Description
===========

A KIP-1102 re-bootstrap sequence never actually reached a bootstrap broker while
the learned brokers were still connected. `rd_kafka_rebootstrap_tmr_cb()` re-added
the bootstrap servers, but `rd_kafka_broker_weight_usable()` weighs learned brokers
above bootstrap ones, so `rd_kafka_broker_any_usable()` always found a learned
broker and the client kept asking the very brokers that had reported its metadata
as stale. The sequence started, logged, and was inert.

Learned brokers are now decommissioned when a re-bootstrap sequence starts
(`rd_kafka_brokers_decommission_learned()`), so only the bootstrap servers are used
until a Metadata response rebuilds the broker list. Queued messages on a
decommissioned broker are handed back to their partitions and re-sent once new
leaders are known.

Three follow-on races surfaced while testing this and are fixed here:

- **Spurious second sequence.** `rd_kafka_broker_decommission()` forced its broker
  down without marking the failure as planned, so decommissioning several still-up
  learned brokers at once could itself cross the "all brokers down" threshold and
  start another re-bootstrap. It now uses `rd_kafka_broker_planned_fail()`, so a
  decommission is never counted toward `ALL_BROKERS_DOWN`. This applies to every
  decommission, not only the re-bootstrap path.

- **Learned brokers re-created from a decommissioning broker's response.** An
  already-received FindCoordinator or Metadata response from a broker that is being
  taken down carries information from a source the client just stopped trusting.
  Applying it re-created the coordinator as a learned broker, which then won broker
  selection over the bootstrap brokers and defeated the re-bootstrap. A successful
  response whose source broker has `termination_in_progress` is now failed with
  `RD_KAFKA_RESP_ERR__DESTROY_BROKER` in the three handlers that can create brokers
  (`rd_kafka_cgrp_handle_FindCoordinator`, `rd_kafka_coord_req_handle_FindCoordinator`,
  `rd_kafka_handle_Metadata`), which existing callers already treat as a retry.
  This was the cause of the `0151` `bootstrap_only_metadata_cnt > 0` failure seen on
  Windows.

- **Mock cgrp assertion.** Decommissioning the current group coordinator marks it
  dead immediately and triggers a JoinGroup/SyncGroup retry on a new connection,
  which can reach the mock broker before it has processed the old connection's
  closure. The mock's per-member response tracking only tolerated one live request
  per member and asserted on the retry; the stale, now-unreachable response is
  discarded instead.

Upgrade considerations
=====================

Admin requests in flight on a decommissioned broker now fail with
`RD_KAFKA_RESP_ERR__TRANSPORT` instead of `RD_KAFKA_RESP_ERR__DESTROY_BROKER`, so
callers retry them instead of treating them as a hard failure.

Tests
=====

`tests/0151-purge-brokers.c` gains coverage for the KIP-899 and KIP-1102
re-bootstrap cases. The pre-existing KIP-1102 cases only counted
`Starting re-bootstrap sequence` log lines, which are emitted before any broker is
added or connected and so cannot tell a working sequence from an inert one; and
they could not be told apart by address, since the harness puts every mock broker
in `bootstrap.servers` and all of them are in Metadata. The new cases remove one
broker from the mock Metadata response and use it as the sole bootstrap server, then
assert via request tracking that a Metadata request actually reaches it.

`do_test_down_then_up_no_rebootstrap_loop` no longer asserts a measured range of
re-bootstrap sequences (7-12), which failed branch-wide on the arm64 `local_quick`
jobs with 20-21. The attempts are not paced by `reconnect.backoff.ms`:
`RD_KAFKA_OP_CONNECT` zeroes `rkb_ts_reconnect` to expedite the reconnect, so the
backoff never ramps. The rate follows how many subsystems ask for a cluster
connection while every broker is down, including ProducerID acquisition, which
`run_local_quick` enables via `-P` and which takes the rate from ~1/s to ~3/s.
It now asserts the one bound that is guaranteed: `sparse_connect_intvl` gates
`rd_kafka_connect_any()` at 50 ms by default, allowing at most 120 sequences in the
6 s window. A loop not gated by connection attempts still exceeds that.

`tests/0118-commit_rebalance.c` was flaky in the `consumer` protocol. It expects the
commit issued from the rebalance callback to fail, but that failure comes only from
the cgrp being fully terminated (its ops queue disabled, giving `_DESTROY`) — a
commit after an ordinary mid-life revocation legitimately succeeds, and in the
`consumer` protocol every assignor is cooperative, so live members get partial
revocations that reach the same callback. The sleep and commit are now gated on a
flag set immediately before `c2` is closed.

- `make -j` and `cd tests && make build`
- `cd tests && LD_LIBRARY_PATH=../src:../src-cpp TESTS=0000,0105,0117,0138,0151,0152 ./test-runner` — all passed
- `cd tests && TESTS=0118 ./run-test.sh` against a trivup cluster with
  `group.protocol=consumer`, 4/4 runs passed (2 of them exercising the exact
  `_DESTROY` path the test asserts); the `classic` protocol was verified before the
  change but not re-run after it
- `packaging/tools/style-format.sh` on the changed sources — clean


k-raina (approved):
Thanks @emasab for PR! Looks great. 
I have triggered integration tests. Once they pass, we can merge

PR #5606, opened 2026-09-18
[KIP-1102] Re-bootstrap reconnects to bootstrap servers on REBOOTSTRAP_REQUIRED

### LINKED ISSUE #5605: KIP-1102: `REBOOTSTRAP_REQUIRED` does not make the client reconnect to its bootstrap servers
librdkafka receives the top-level Metadata error `REBOOTSTRAP_REQUIRED` (129), logs `Starting re-bootstrap sequence`, and keeps using the same brokers. The re-bootstrap sequence only re-adds the `bootstrap.servers` entries to the broker list. They are already there, so nothing changes. If `bootstrap.servers` is a DNS name or load balancer that was moved to a new cluster, the client never finds the new cluster.

The Metadata response that carries the error is also discarded in full, so the client keeps stale partition leaders. A producer stops delivering without reporting an error.

### Expected behavior

[KIP-1102](https://cwiki.apache.org/confluence/spaces/KAFKA/pages/327977769/KIP-1102+Enable+clients+to+rebootstrap+based+on+timeout+or+error+code#KIP1102%3AEnableclientstorebootstrapbasedontimeoutorerrorcode-ProposedChanges) defines `REBOOTSTRAP_REQUIRED` as a signal for the client to go back to its bootstrap servers and fetch metadata again. The client should drop its learned brokers, reconnect to `bootstrap.servers` and resolve their addresses again.

Java `kafka-clients` does this (tested on 4.1.1). It closes every connection, including the bootstrap one (`Disconnecting from node -1 due to client rebootstrap.`), and dials the bootstrap servers again.

### How to reproduce

Branch in my fork with a failing test: [Link](https://github.com/confluentinc/librdkafka/compare/master...r-vasquez:librdkafka:kip1102-rebootstrap-noop-repro?expand=1). It includes a small mock cluster change so the mock can send the error: Metadata v13 with the pushed request error written as top-level `ErrorCode`.

1. Mock cluster with 3 brokers. Only broker 1 is in `bootstrap.servers`.
2. The client learns all brokers and its startup connections settle.
3. Every following Metadata response carries top-level `REBOOTSTRAP_REQUIRED`.
4. Assert that a new connection reaches broker 1 within 5 seconds.

```
$ cd tests && TESTS=0152 SUBTESTS=reconnects ./run-test.sh -l
[0152_rebootstrap_local / 9.109s] re-bootstrap sequences: 1, new connections to bootstrap server: 0
Test assertion failed: "connect_cnt >= 1": client did not reconnect to its bootstrap server
after REBOOTSTRAP_REQUIRED
```

Same result for producer and consumer. With `TEST_DEBUG=broker,metadata` the bootstrap broker
is `DOWN` before the error and has no state change after it:

```
127.0.0.1:49255/bootstrap: Broker changed state UP -> DOWN
127.0.0.1:49256/2: Metadata response: received top level error code 129: Broker: Client metadata is stale, client should rebootstrap to obtain new metadata
REBOOTSTRAP [thrd:main]: Starting re-bootstrap sequence
```

r-vasquez:
Fixes #5605

When a Metadata response carries `REBOOTSTRAP_REQUIRED` the re-bootstrap sequence only re-added the bootstrap servers, which were already in the broker list, so nothing changed and the client stayed on the old brokers. The sequence now decommissions learned and configured brokers and adds the bootstrap servers again as new brokers, so they connect and resolve their address from scratch. The all-brokers-down trigger keeps the previous behavior.

The second commit fixes a related problem found while writing the test: a successful Metadata response arriving right after the error cancelled the pending re-bootstrap and left the in-progress flag set, so later errors were ignored. Test 0152 reproduces the first bug and passes with the fix. The mock cluster now supports Metadata v13 so it can return the top-level error code.


emasab:
Duplicate of #5600

